PR #82, opened 2018-05-11
aws: add assume-role support

otterley:
Add two new functions:

* `CreateAwsSessionFromRole` returns a new AWS session after assuming the role
   whose ARN is provided in `roleARN`.

* `AssumeRole` mutates the provided session by obtaining new credentials by
   assuming the role provided in `roleARN`.

### DIFF
--- a/modules/aws/auth.go
+++ b/modules/aws/auth.go
@@ -27,6 +27,22 @@ func NewAuthenticatedSession(region string) (*session.Session, error) {
 	return sess, nil
 }
 
+// NewAuthenticatedSessionFromRole returns a new AWS Session after assuming the
+// role whose ARN is provided in roleARN. If the credentials are not properly
+// configured in the underlying environment, an error is returned.
+func NewAuthenticatedSessionFromRole(region string, roleARN string) (*session.Session, error) {
+	sess, err := CreateAwsSessionFromRole(region, roleARN)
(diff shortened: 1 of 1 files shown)


modules/aws/auth.go
```diff
@@ -25,6 +27,24 @@ func NewAuthenticatedSession(region string) (*session.Session, error) {
 	return sess, nil
 }
 
+// CreateAwsSessionFromRole returns a new AWS session after assuming the role
+// whose ARN is provided in roleARN.
+func CreateAwsSessionFromRole(region string, roleARN string) (*session.Session, error) {
```
brikis98: The `NewAuthenticatedSession` method above calls `sess.Config.Credentials.Get()` to check that the credentials are valid and returns an error right away if they are not. This ensures the test fails as early as possible. 

Any reason not to implement a similar mechanic in this method? If so, it would be worth naming it in a similar manner (e.g., `NewAuthenticatedSessionFromRole`).

otterley:
Added `NewAuthenticatedSessionFromRole` per your request.

brikis98:
I'll let the tests run and if they pass, I'll create a new release and post the link here.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.9.3

PR #85, opened 2018-05-14
Update acronyms to have consistent case

chrsblck:
- https://github.com/golang/go/wiki/CodeReviewComments#initialisms
- Ran `goimports`

I read CONTRIBUTING.md, assuming steps 1-3 don't apply for this example test. Let me know if I should have.

### DIFF
--- a/test/terraform_http_example_test.go
+++ b/test/terraform_http_example_test.go
@@ -1,13 +1,14 @@
 package test
 
 import (
-	"testing"
 	"fmt"
+	"testing"
 	"time"
-	"github.com/gruntwork-io/terratest/modules/terraform"
+
+	"github.com/gruntwork-io/terratest/modules/aws"
 	"github.com/gruntwork-io/terratest/modules/http-helper"
 	"github.com/gruntwork-io/terratest/modules/random"
-	"github.com/gruntwork-io/terratest/modules/aws"
+	"github.com/gruntwork-io/terratest/modules/terraform"
 )
 
 // An example of how to test the Terraform module in examples/terraform-http-example using Terratest.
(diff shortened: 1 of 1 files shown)


brikis98:
Thanks for the PR! It seems like running `gofmt` would be a good thing to do in this repo in general, perhaps as a [pre-commit hook](https://golang.org/misc/git/pre-commit)? We could fail the CI on it too. Let me know if you'd be up for that sort of change instead of doing it on just one file.

chrsblck:
@brikis98 I use [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), which also formats the code like `gofmt`. 

I use these:
- [vim-go](https://github.com/fatih/vim-go)
- [VS Code](https://github.com/Microsoft/vscode-go#options)

Yeah, I can send a separate PR (today or tomorrow) that runs `goimports` on `*.go` files.

Would you want me to revert the `goimports` from this PR, first?

brikis98:
> I use goimports, which also formats the code like gofmt.

`goimports` sounds great!

> Yeah, I can send a separate PR (today or tomorrow) that runs goimports on *.go files.

Wonderful, thanks. The approach we recommend is to use [pre-commit](https://pre-commit.com/) to add a pre-commit hook that runs `goimports`. You can also add it to the CircleCI config to fail the build if someone didn't run the pre-commit hook and committed unformatted code.

> Would you want me to revert the goimports from this PR, first?

Nah, instead, just run `goimports` on all the code in this codebase so everything is formatted once the pre-commit hook is added. Thank you!

chrsblck:
OK, cool. Should I create a new PR for this?

brikis98:
Sure, that would be great! I'll close this one for now.

Issue #94, opened 2018-05-30
Add GOCACHE=off to best practises


mastertinner:
Since Go 1.10, test results get cached. This leads to the tests not being run sometimes because Go has cached the results and just assumes it will get the same results because the Go code hasn't changed. We should add a point to the documentation "Testing best practises" that asks users to set the `GOCACHE` environment variable to `off` for terratests:

```shell
$ GOCACHE=off go test -v -timeout 45m ./...

```

brikis98:
Oh, interesting, I had not heard about this. Where can I find more info on `GOCACHE`?

mastertinner:
I learned about it from here: https://github.com/golang/go/issues/22758

brikis98:
Heh, that's quite hidden. Good find though.

Would you be up for a quick PR to add this as a note to the README?

PR #95, opened 2018-05-31
Fix linting issues and add best practise about Go cache

mastertinner:
Fixes https://github.com/gruntwork-io/terratest/issues/94

I also fixed a lot of linting issues and implemented many of the Go best practises without changing the API of any of the packages.

@brikis98, let me know what you think. I would also advise to change the exported variables to follow the Go best pracises (e.g. change any variable that contains `Ssh` to `SSH` as described in https://github.com/golang/go/wiki/CodeReviewComments#initialisms) but I didn't want to break the API with this PR.


Gopkg.toml
```diff
@@ -17,18 +16,27 @@
...
-  version = "1.12.74"
+  version = "1.13.59"
 
 [[constraint]]
   name = "github.com/google/uuid"
   version = "0.2.0"
 
+[[constraint]]
+  name = "github.com/gruntwork-io/terratest"
+  version = "0.9.4"
```
brikis98: Wait, why does this code depend on itself?

README.md
```diff
@@ -1,37 +1,36 @@
 # Terratest
 
+[![Go Report Card](https://goreportcard.com/badge/github.com/gruntwork-io/terratest)](https://goreportcard.com/report/github.com/gruntwork-io/terratest)
```
brikis98: Cool project, I didn't know about this!

README.md
```diff
@@ -1,37 +1,36 @@
...
 
 Terratest was developed at [Gruntwork](https://gruntwork.io/) to help maintain the [Infrastructure as Code
 Library](https://gruntwork.io/infrastructure-as-code-library/), which contains over 250,000 lines of code written
 in Terraform, Go, Python, and Bash, and is used in production by hundreds of companies. See [Open sourcing Terratest: a
 swiss army knife for testing infrastructure code](https://blog.gruntwork.io/open-sourcing-terratest-a-swiss-army-knife-for-testing-infrastructure-code-5d883336fcd5)
 for more info.
 
-
-
-
```
brikis98: What tool is reformatting the Markdown files? I actually prefer having more spacing between sections, as it make sit easier to visually see the structure of the file when looking at the raw Markdown text rather than the HTML output.

modules/aws/account.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/iam"
 )
 
-// Get the Account ID for the currently logged in IAM User.
+// GetAccountId gets the Account ID for the currently logged in IAM User.
```
brikis98: What's the point of repeating the method name here? Seems 99% likely to break when the name is changed and the comment accidentally left the old way... I noticed this PR does this on most (but not all?) of the methods and it seems like it doesn't add much, but makes maintenance harder.

modules/aws/ami.go
```diff
@@ -63,28 +66,34 @@ type imageSort []*ec2.Image
...
-	itime, _ := time.Parse(time.RFC3339, *a[i].CreationDate)
-	jtime, _ := time.Parse(time.RFC3339, *a[j].CreationDate)
-	return itime.Unix() < jtime.Unix()
+	iTime, err := time.Parse(time.RFC3339, *a[i].CreationDate)
+	if err != nil {
+		panic(fmt.Errorf("error parsing creation date of EC2 image: %s", err))
+	}
+	jTime, err := time.Parse(time.RFC3339, *a[j].CreationDate)
+	if err != nil {
+		panic(fmt.Errorf("error parsing creation date of EC2 image: %s", err))
```
brikis98: We do not want to `panic` in Terratest, ever, if we can avoid it. It will prevent cleanup code from running, such as the call to `terratest.Destroy`, and therefore, leave lots of resources running in your AWS account (or wherever you're testing). 

If we're going to make this change, then anything that uses `imageSort` needs to `recover` from possible panics, and we should probably have an automated test to verify it actually works.

modules/aws/aws.go
```diff
@@ -0,0 +1,2 @@
+// Package aws allows to interact with resources on Amazon Web Services.
+package aws
```
brikis98: What does this empty file do? I noticed there's a bunch of these and it's not obvious to me what purpose they serve.

modules/http-helper/dummy_server.go
```diff
@@ -35,11 +35,17 @@ func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
 	logger.Logf(t, "Starting dummy HTTP server in port %d that will return the text '%s'", port, text)
 
 	listener, err := net.Listen("tcp", ":"+strconv.Itoa(port))
-
-	if err == nil {
-		go http.Serve(listener, nil)
+	if err != nil {
```
brikis98: Oh, good catch!

modules/http-helper/dummy_server.go
```diff
@@ -35,11 +35,17 @@ func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
...
+		return nil, 0, fmt.Errorf("error listening: %s", err)
 	}
 
+	go func() {
+		err = http.Serve(listener, nil)
+		if err != nil {
+			panic(err)
```
brikis98: Again, we should generally not `panic` anywhere in Terratest... Not sure what effect a `panic` would have in a `goroutine` though... Would it result in the tests all exiting? Or just be ignored?

modules/shell/command.go
```diff
@@ -66,7 +66,8 @@ func RunCommandAndGetOutputE(t *testing.T, command Command) (string, error) {
 		return "", err
 	}
 
-	if err := cmd.Start(); err != nil {
+	err = cmd.Start()
+	if err != nil {
```
brikis98: I though the one-line `if err := xxx; err != nil { ... }` pattern was idiomatic Go?

test/terraform_redeploy_example_test.go
```diff
@@ -152,7 +152,7 @@ func fetchSyslogForAsg(t *testing.T, awsRegion string, workingDir string) {
 
 	logger.Logf(t, "===== Syslog for instances in ASG %s =====\n\n", asgName)
 
-	for instanceId, logs := range asgLogs {
-		logger.Logf(t, "Most recent syslog for Instance %s:\n\n%s\n", instanceId, logs)
+	for id, logs := range asgLogs {
```
brikis98: Seems like `instanceID` would be a clearer name here than `id`

README.md
mastertinner: I'm using https://github.com/prettier/prettier to automatically format the Markdown. If you don't like it, I can revert it.

modules/aws/account.go
mastertinner: https://golang.org/doc/effective_go.html#commentary
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences
https://blog.golang.org/godoc-documenting-go-code

modules/aws/aws.go
mastertinner: They are simply there to document the package. I think it's advantageous if the file containing the comment always has the same name as the package itself (and the directory it's contained in) so you don't have to search which file contains the package comment.

modules/shell/command.go
mastertinner: It is but it lead to `err` shadowing `err` from its parent scope. We can change it to

```go
if err = cmd.Start(); err != nil {
```

if you like that better. I think the version I changed it to is easier to read but that's a personal preference and I'm happy to change it back if that's what you prefer.

test/terraform_redeploy_example_test.go
mastertinner: You're right. I changed it back.

Gopkg.toml
mastertinner: I was wondering that too... The file was auto generated by `dep` so it wasn't a conscious decision but it might be worth investigating.

modules/aws/account.go
brikis98: I assume you're mainly referring to this:

> If every doc comment begins with the name of the item it describes, the output of godoc can usefully be run through grep. Imagine you couldn't remember the name "Compile" but were looking for the parsing function for regular expressions, so you ran the command,
>
> $ godoc regexp | grep -i parse
>
> If all the doc comments in the package began, "This function...", grep wouldn't help you remember the name. But because the package starts each doc comment with the name, you'd see something like this, which recalls the word you're looking for.
> 
> $ godoc regexp | grep parse
>    Compile parses a regular expression and returns, if successful, a Regexp
>   parsed. It simplifies safe initialization of global variables holding
>    cannot be parsed. It simplifies safe initialization of global variables

Honest question, but, have you ever actually done that sort of search? I'm sure it happens time to time, but it seems like an exceptionally rare occurrence. On the other hand, renaming methods happens all the time, and forgetting to update the comment will be confusing for everyone, including anyone who tries to search for method names using `godoc`.

modules/aws/ami.go
mastertinner: Good point. So I guess, in this case, it's preferable just not to handle the error...

modules/shell/command.go
brikis98: It seems like in these cases, shadowing, rather than reusing the same variable name, is a good thing, no? It makes each error check standalone and you can never accidentally reuse an error from some previous call...

README.md
brikis98: I'm not religious about these sorts of decisions, but which of these makes the structure of the Markdown easier to see?

https://raw.githubusercontent.com/mastertinner/terratest/813222c9ce7fbdf4320e5b596f006602c98078c4/README.md
https://raw.githubusercontent.com/gruntwork-io/terratest/master/README.md

modules/aws/ami.go
brikis98: It's crappy either way, but yea, probably.

mastertinner: Ok. I removed it.

modules/http-helper/dummy_server.go
mastertinner: I removed it (see comments above).

Gopkg.toml
brikis98: Does something import the root `terratest` folder in the code?

mastertinner: No

brikis98: Hm, I have no clue. Perhaps it's confused by the folder structure, where there are no `.go` files in the root, and only in `modules/xxx` subfolders? Not sure how to fix that, other than manually removing this, as it will likely cause issues.

mastertinner: Sorry, it was my bad. I was working on my fork in `$GOPATH/src/github.com/mastertinner/terratest`. That lead to this import statement. It's now fixed.

mastertinner: Sorry, it was my bad. I was working on my fork in `$GOPATH/src/github.com/mastertinner/terratest`. That lead to this import statement. It's now fixed.

modules/aws/account.go
mastertinner: I actually think it's quite useful. Another advantage is that it gets people to write full sentences when they're commenting exported entities. That helps a lot to improve the quality o the documentation.

As for the renaming, I think most people use a tool like [gorename] to rename variables which automatically changes the comment as well so I don't see any big risks in that.

README.md
mastertinner: I usually try to avoid having to maintain the formatting of my files manually because you'll almost always forget about doing it in one or more places and there's no way of enforcing/checking that it's actually done consistently. That's why I love tools like prettier because they automatically take care of it for me upon saving a file.

I'll leave the choice up to you. I agree that the file is visually easier to scan with the spacing you had before I auto-formatted it so if that's what you prefer, I'll gladly change it back.

Gopkg.toml
brikis98: Ah, OK, great 👍

README.md
brikis98: Totally agree on using automatic code formatting where possible, but it's important that the resulting formatting is easy to read. I've found prettier often favors consistency over readability, which is not always a good trade-off. If there's a way to configure prettier to enforce better spacing in Markdown, I'm all for it. If not, I'd prefer to keep the file as it was before. Thx!

modules/aws/account.go
brikis98: Hmm, OK, I'm willing to try it. I've never used `gorename`. I use IntelliJ to do refactoring. Not sure if it handles comments.

README.md
mastertinner: I put the spacing back.

modules/aws/account.go
mastertinner: I personally use [vim-go](https://github.com/fatih/vim-go) but I'm sure that [Go Land](https://www.jetbrains.com/go/) and the IntelliJ Go plugin also use `gorename` under the hood.

mastertinner:
I did try to run it but haven't gotten a run yet when all tests passed... I then tried to run all tests on your `master` branch and also couldn't get a fully green run... Could you try as well?

brikis98:
Tests passed in CI just a few minutes ago.

What errors are you seeing?

mastertinner:
I'm getting `aws_alb_target_group.web_servers: aws_alb_target_group.web_servers: Error retrieving Target Group: ValidationError: 'arn:aws:elasticloadbalancing:ap-northeast-2:814024550722:targetgroup/redeploy-test-ZcD2Nr/f27f78da2c89599f' is not a valid target group ARN` when running `go test -timeout 30m -p 1 ./test/... -run TerraformRedeployExample`.

That's the only failure I get (on your `master` branch as well as on my own branch).

brikis98:
Oh, that might be an intermittent Terraform bug that the test is catching!

Could you try updating that example with these workarounds (some `depends_on`):

https://github.com/hashicorp/terraform/issues/12634#issuecomment-321633155
https://github.com/hashicorp/terraform/issues/12634#issuecomment-371555338

mastertinner:
That fixed it (https://github.com/gruntwork-io/terratest/pull/95/commits/7bb45b35001b127dc594faadfc786d1c1351e74b). Now all the tests pass!

brikis98:
Doesn't look like the public API changed, so no need for a new release. Thank you again for the PR!

PR #97, opened 2018-06-01
Temp folder subtest fix

eak12913:
Fixes the issue where if tests would fail if you were copying files to a temporary folder from a subtest (from inside of a `t.Run(...) in your test case)


eak12913:
https://github.com/gruntwork-io/terratest/releases/tag/v0.9.5

PR #100, opened 2018-06-09
Initial work on GCP module.

Dahs81:
Hey,

Before jumping in too far, I just wanted to do this simple PR. Currently only has create and delete for Google Storage. I also stubbed out accounts and vpc files.

The one thing that I'm not sure how to handle yet is dealing with passing in the projectId. Currently, it is being passed in via an env var, but I was thinking that it should probably pull that out of the `~.google/credentials.json` path.  Then if that is not there, use the env var.  Otherwise call fatal as your code is currently doing.

Any thoughts on how you want that done?


examples/terraform-gcp-example/main.tf
```diff
@@ -0,0 +1,9 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# CREATE A GOOGLE STORAGE BUCKET
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "google_storage_bucket" "test_bucket" {
+    name     = "${var.bucket_name}"
+    location = "${var.location}"
+    project  = "${var.project}"
+}
```
brikis98: Please run `terraform fmt` on the code. I think the default is 2 spaces, not 4.

examples/terraform-gcp-example/README.md
```diff
@@ -0,0 +1,41 @@
+# Terraform GCP Example
+
+This folder contains a simple Terraform module that deploys resources in [GCP](https://cloud.google.com/) to demonstrate
+how you can use Terratest to write automated tests for your GCP Terraform code. This module deploys an [Compute
+Instance](https://cloud.google.com/compute/) and gives that Instance a `Name` tag with the value specified in the
+`instance_name` variable.
```
brikis98: Make sure to update the description of what this example deploys once it's done.

modules/gcp/storage.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	// Sets the name for the new bucket.
+	bucketName := name
+
+	// Creates a Bucket instance.
+	bucket := client.Bucket(bucketName)
+
+	// Creates the new bucket.
+	if err := bucket.Create(ctx, projectId, attr); err != nil {
```
brikis98: You can probably just `return bucket.Create(...)` here.

modules/gcp/storage.go
```diff
@@ -0,0 +1,71 @@
...
+	ctx := context.Background()
+
+	// Creates a client.
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
+	}
+
+	// Sets the name for the new bucket.
+	bucketName := name
```
brikis98: No need for an extra variable here, just use `name` directly in the line below.

modules/gcp/storage.go
```diff
@@ -0,0 +1,71 @@
...
+// CreateStorageBucketE creates a Google Cloud bucket with the given BucketAttrs. Note that Google Storage bucket names must be globally unique.
+func CreateStorageBucketE(t *testing.T, projectId string, name string, attr *storage.BucketAttrs) error {
+	logger.Logf(t, "Creating bucket %s", name)
+
+	ctx := context.Background()
+
+	// Creates a client.
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
```
brikis98: `log.Fatalf`, believe it or not, calls `os.Exit`. Instead, just return the `err` here so the user can decide what to do with it. Same with all usages below.

modules/gcp/storage.go
```diff
@@ -0,0 +1,71 @@
...
+
+// DeleteS3BucketE destroys the S3 bucket in the given region with the given name.
+func DeleteStorageBucketE(t *testing.T, name string) error {
+	logger.Logf(t, "Deleting bucket %s", name)
+
+	ctx := context.Background()
+
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
```
brikis98: `return` the `err` here, don't use `log.Fatal`

modules/gcp/storage.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if err := client.Bucket(name).Delete(ctx); err != nil {
```
brikis98: Return `client.Bucket(...)` directly.

modules/gcp/storage_test.go
```diff
@@ -0,0 +1,26 @@
...
+func TestCreateAndDestroyStorageBucket(t *testing.T) {
+	t.Parallel()
+
+	id := random.UniqueId()
+	logger.Logf(t, "Random values selected Id = %s\n", id)
+
+	gsBucketName := "gruntwork-terratest-" + strings.ToLower(id)
+
+	CreateStorageBucket(t, projectId, gsBucketName, nil)
+	DeleteStorageBucket(t, gsBucketName)
```
brikis98: Might be worth adding a `FindStorageBucket` method and making sure it returns the expected value after the Create and no longer finds it after the Delete.

test/terraform_gcp_example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+var projectId = os.Getenv("GOOGLE_CLOUD_PROJECT_ID")
+
+func TestTerraformGcpExample(t *testing.T) {
+	t.Parallel()
+
+	// Give this EC2 Instance a unique ID for a name tag so we can distinguish it from any other EC2 Instance running
+	// in your AWS account
```
brikis98: Update comment

test/terraform_gcp_example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+var projectId = os.Getenv("GOOGLE_CLOUD_PROJECT_ID")
```
brikis98: Should probably display a clear error message if this env var is not set

test/terraform_gcp_example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	// This will run `terraform init` and `terraform apply` and fail the test if there are any errors
+	terraform.InitAndApply(t, terraformOptions)
+
+	// Run `terraform output` to get the value of an output variable
+	bucketURL := terraform.Output(t, terraformOptions, "url")
+
+	expectedURL := fmt.Sprintf("gs://%s", expectedName)
+
+	// Verify that our expected name tag is one of the tags
+	assert.Equal(t, expectedURL, bucketURL)
```
brikis98: Test looks great 👍

PR #108, opened 2018-06-13
Add support for SCPing content over SSH

### LINKED ISSUE #102: Add support for SCPing content to instances

Adding support for SCPing content onto the instances is something that I feel would be handy to include as an easier way to get files onto instances.

Support for SCPing content _off_ of instances while may be nice isn't as critical, due to the fact you can always just `cat filename`.

The use case I have in mind is creating an AMI using Packer, including baking in relevant binaries, and then testing the AMI by spinning up a simple test instance and copying configuration over to the test instance.

wjam:
Fixes #102

### DIFF
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -9,8 +9,9 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"strconv"
+
+	"path/filepath"
 
 	"golang.org/x/crypto/ssh"
 )
@@ -36,8 +37,7 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents 
 	if err != nil {
 		return err
 	}
-	file := path.Base(remotePath)
-	dir := path.Dir(remotePath)
+	dir, file := filepath.Split(remotePath)
 
 	hostOptions := SshConnectionOptions{
 		Username:    host.SshUserName,
@@ -47,14 +47,7 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents 
 		AuthMethods: authMethods,
 	}
 
-	scp := func(w io.WriteCloser) {
-		// https://web.archive.org/web/20170215184048/https://blogs.oracle.com/janp/entry/how_the_scp_protocol_works
-		defer w.Close()
-		octalMode := "0" + strconv.FormatInt(int64(os.FileMode(mode)), 8)
-		fmt.Fprintln(w, "C"+octalMode, len(contents), file)
-		fmt.Fprint(w, contents)
-		fmt.Fprint(w, "\x00")
-	}
+	scp := sendScpCommandsToCopyFile(mode, file, contents)
 
 	sshSession := &SshSession{
 		Options:  &hostOptions,
@@ -181,7 +174,10 @@ func runSSHCommand(sshSession *SshSession) (string, error) {
 		if err != nil {
 			return "", err
 		}
-		go (*sshSession.Input)(w)
+		go func() {
+			defer w.Close()
+			(*sshSession.Input)(w)
+		}()
 	}
 
 	bytes, err := sshSession.Session.CombinedOutput(sshSession.Options.Command)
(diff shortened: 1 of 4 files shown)


wjam:
Gist of test run https://gist.github.com/wjam/52875b51f6baa926500d948c871c5577

modules/ssh/ssh.go
```diff
@@ -16,6 +22,52 @@ type Host struct {
...
+}
+
+// ScpFileToE uploads the contents using SCP to the given host and return an error if the process fails.
+func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents string) error {
+	authMethods, err := createAuthMethodsForHost(host)
+	if err != nil {
+		return err
+	}
+	file := path.Base(remotePath)
+	dir := path.Dir(remotePath)
```
brikis98: Shouldn't this be `filepath` instead of `path`?

modules/ssh/ssh.go
```diff
@@ -16,6 +22,52 @@ type Host struct {
...
+
+	hostOptions := SshConnectionOptions{
+		Username:    host.SshUserName,
+		Address:     host.Hostname,
+		Port:        22,
+		Command:     "/usr/bin/scp -t " + dir,
```
brikis98: Are you relying on `scp` being installed instead of implementing the file copying functionality in Go?

modules/ssh/ssh.go
```diff
@@ -16,6 +22,52 @@ type Host struct {
...
+	}
+
+	scp := func(w io.WriteCloser) {
+		// https://web.archive.org/web/20170215184048/https://blogs.oracle.com/janp/entry/how_the_scp_protocol_works
+		defer w.Close()
+		octalMode := "0" + strconv.FormatInt(int64(os.FileMode(mode)), 8)
+		fmt.Fprintln(w, "C"+octalMode, len(contents), file)
+		fmt.Fprint(w, contents)
+		fmt.Fprint(w, "\x00")
+	}
```
brikis98: Could you extract this into a named function with an explanation of what this is doing?

modules/ssh/ssh.go
```diff
@@ -124,6 +176,14 @@ func runSSHCommand(sshSession *SshSession) (string, error) {
 		return "", err
 	}
 
+	if sshSession.Input != nil {
+		w, err := sshSession.Session.StdinPipe()
+		if err != nil {
+			return "", err
+		}
+		go (*sshSession.Input)(w)
```
brikis98: Does anything need to be closed (in a `defer`) after this?

test/terraform_ssh_example_test.go
```diff
@@ -167,3 +170,44 @@ func testSSHToPrivateHost(t *testing.T, terraformOptions *terraform.Options, key
...
+
+		actualText, err := ssh.CheckSshCommandE(t, publicHost, "cat /tmp/test.txt")
+
+		if err != nil {
+			return "", err
+		}
+
+		if strings.TrimSpace(actualText) != expectedText {
+			return "", fmt.Errorf("Expected SSH command to return '%s' but got '%s'", expectedText, actualText)
+		}
```
brikis98: Nice tests!

modules/ssh/ssh.go
wjam: You can't implement the functionality in Go as scp must be installed to handle the SCP protocol. The only other alternative would be to chain a large number of commands to `echo 'foo' >> file` and hope to not go over the maximum command limit.

modules/ssh/ssh.go
wjam: Refactored the code to make it more obvious where the closing was being handled

modules/ssh/ssh.go
wjam: Done

modules/ssh/ssh.go
wjam: Changed

modules/ssh/ssh.go
brikis98: Oh, I gotcha. Despite having used it many times, I didn't actually realize that's how SCP worked. TIL! 😄

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.9.7

PR #109, opened 2018-06-15
Add log message when running SSH commands

### LINKED ISSUE #104: Add `logger.Logf` message to `ssh.runSSHCommand`

Adding a log message to the `runSSHCommand` method would mean that when `CheckSshCommand` fails, you aren't just left with a message of how the exit status for some command wasn't 0. An example can be found in `RunCommandAndGetOutputE` which makes it easier to track what went wrong when a test fails.

Use case: It's difficult to track down what failed when executing numerous `CheckSshCommand` when testing an AMI

wjam:
Fixes #104

### DIFF
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -13,6 +13,7 @@ import (
 
 	"path/filepath"
 
+	"github.com/gruntwork-io/terratest/modules/logger"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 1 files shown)


wjam:
Rebased to fix conflict

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.9.8

PR #110, opened 2018-06-17
Add support for deleting EBS snapshots that back an AMI

### LINKED ISSUE #105: Add support for deleting an AMI along with its backing snapshot

As part of building a pipeline for some Packer/Terraform infrastructure, it's common that AMIs will be created that are instantly destined for the bin - either because a branch/PR is being tested or because the AMI failed testing. It'd be useful to add a method that is capable of both deregistering the AMI as well as deleting the snapshot behind the AMI.

wjam:
Fixes #105

### DIFF
--- a/modules/aws/ami.go
+++ b/modules/aws/ami.go
@@ -18,6 +18,36 @@ const (
 	AmazonAccountId    = "amazon"
 )
 
+// DeleteAmiAndAllSnapshots will delete the given AMI along with all EBS snapshots that backed that AMI
+func DeleteAmiAndAllSnapshots(t *testing.T, region string, ami string) {
+	err := DeleteAmiAndAllSnapshotsE(t, region, ami)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+// DeleteAmiAndAllSnapshots will delete the given AMI along with all EBS snapshots that backed that AMI
+func DeleteAmiAndAllSnapshotsE(t *testing.T, region string, ami string) error {
+	snapshots, err := GetEbsSnapshotsForAmiE(t, region, ami)
+	if err != nil {
+		return err
+	}
+
+	err = DeleteAmiE(t, region, ami)
+	if err != nil {
+		return err
+	}
+
+	for _, snapshot := range snapshots {
+		err = DeleteEbsSnapshotE(t, region, snapshot)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // GetEbsSnapshotsForAmi retrieves the EBS snapshots which back the given AMI
 func GetEbsSnapshotsForAmi(t *testing.T, region string, ami string) []string {
 	snapshots, err := GetEbsSnapshotsForAmiE(t, region, ami)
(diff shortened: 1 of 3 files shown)


wjam:
Output from test: https://gist.github.com/wjam/ef7339ba190d950310350f5b59ebff88

modules/aws/ami.go
```diff
@@ -17,6 +18,44 @@ const (
...
+	}
+	return snapshots
+}
+
+// GetEbsSnapshotsForAmi retrieves the EBS snapshots which back the given AMI
+func GetEbsSnapshotsForAmiE(t *testing.T, region string, ami string) ([]string, error) {
+	logger.Logf(t, "Retrieving EBS snapshots backing AMI %s", ami)
+	ec2Client, err := NewEc2ClientE(t, region)
+	if err != nil {
+		return []string{}, err
```
brikis98: Can simplify to `return nil, err`. Applies to the code below too.

test/packer_basic_example_test.go
```diff
@@ -12,7 +12,7 @@ func TestPackerBasicExample(t *testing.T) {
 	t.Parallel()
 
 	// Pick a random AWS region to test in. This helps ensure your code works in all regions.
-	awsRegion := aws.GetRandomRegion(t, nil, nil)
+	awsRegion := "eu-central-1" //aws.GetRandomRegion(t, nil, nil)
```
brikis98: I'm guessing this was just for testing?

If so, you can override the behavior of `aws.GetRandomRegion` without changing any code by using the `TERRATEST_REGION` env var.

test/packer_basic_example_test.go
```diff
@@ -28,5 +28,13 @@ func TestPackerBasicExample(t *testing.T) {
 	amiID := packer.BuildAmi(t, packerOptions)
 
 	// Clean up the AMI after we're done
-	defer aws.DeleteAmi(t, awsRegion, amiID)
+	defer deleteAmi(t, awsRegion, amiID)
+}
+
+func deleteAmi(t *testing.T, awsRegion string, amiID string) {
```
brikis98: Perhaps we should add a `DeleteAmiAndAllSnapshots` method to the Terratest API that does exactly this?

test/packer_basic_example_test.go
wjam: Yes it was. Only a few regions in the account I'm using has a default VPC

modules/aws/ami.go
```diff
@@ -17,6 +18,74 @@ const (
...
+
+// GetEbsSnapshotsForAmi retrieves the EBS snapshots which back the given AMI
+func GetEbsSnapshotsForAmiE(t *testing.T, region string, ami string) ([]string, error) {
+	logger.Logf(t, "Retrieving EBS snapshots backing AMI %s", ami)
+	ec2Client, err := NewEc2ClientE(t, region)
+	if err != nil {
+		return nil, err
+	}
+
+	images, err := ec2Client.DescribeImages(&ec2.DescribeImagesInput{
```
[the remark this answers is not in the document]
wjam: It doesn't look like the DescribeImages is paginated (CLI documentation says it is but the Go SDK doesn't have it as a paginated version of the method), but we're only expecting one image to be returned as we are passing in a specific AMI id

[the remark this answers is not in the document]
brikis98: Understood, thx!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.9.9

PR #116, opened 2018-06-20
Add support for the Terraform `-no-color` flag

### LINKED ISSUE #115: Allow setting of the terraform `-no-color` option
It should be possible to use the `-no-color` option when running terraform. This is because in certain situations, such as when running on a CI server, the colour escape codes get displayed rather than changing the colour of the output text.

wjam:
Fixes #115

### DIFF
--- a/modules/terraform/apply_test.go
+++ b/modules/terraform/apply_test.go
@@ -17,11 +17,15 @@ func TestApplyNoError(t *testing.T) {
 
 	options := &Options{
 		TerraformDir: testFolder,
+		NoColor: true,
 	}
 
 	out := InitAndApply(t, options)
 
 	assert.Contains(t, out, "Hello, World")
+
+	// Check that NoColor correctly doesn't output the colour escape codes which look like [0m,[1m or [32m
+	assert.NotRegexp(t, `\[\d*m`, out, "Output should not contain color escape codes")
 }
 
 func TestApplyWithErrorNoRetry(t *testing.T) {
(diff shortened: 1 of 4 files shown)


wjam:
Test output, although obviously doesn't show the lack of color:
https://gist.github.com/wjam/8bca85887db769777204bff442fcf55b

test/terraform_basic_example_test.go
```diff
@@ -21,6 +21,8 @@ func TestTerraformBasicExample(t *testing.T) {
 		Vars: map[string]interface{}{
 			"example": expectedText,
 		},
+
+		NoColor: true,
```
brikis98: +1 on adding this to something that goes through automated tests. However, instead of adding it here, could you add it to [apply_test.go](https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/apply_test.go) and actually update one of those tests to look at the stdout/stderr returned by the `Apply` method and ensure it doesn't contain any of the terminal color codes?

wjam:
Test output: https://gist.github.com/wjam/546b6415356948f48bf2e4616cee3fef

modules/terraform/apply_test.go
```diff
@@ -17,11 +17,15 @@ func TestApplyNoError(t *testing.T) {
...
 		TerraformDir: testFolder,
+		NoColor: true,
 	}
 
 	out := InitAndApply(t, options)
 
 	assert.Contains(t, out, "Hello, World")
+
+	// Check that NoColor correctly doesn't output the colour escape codes which look like [0m,[1m or [32m
+	assert.NotRegexp(t, `\[\d*m`, out, "Output should not contain color escape codes")
```
brikis98: Nice, thanks 👍

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.9.12

Issue #112, opened 2018-06-19
Attempting to use different AWS auth methods


Briansbum:
I'm just starting to have a go at using this package and I've come across an issue with how AWS authentication is handled.

In the AWS module there are a lot of auth methods such as `NewAuthenticatedSession`, `CreateAwsSessionWithMfa` and `NewAuthenticatedSessionFromRole`. The issue is that none of these apart from `NewAuthenticatedSession` are implemented anywhere which makes them unusable.

Am I missing something or should I have a look in to how to get these properly implemented?

brikis98:
What do you mean those methods are not implemented? They are all implemented here: https://github.com/gruntwork-io/terratest/blob/9348b5bd2fbfbe5fcf3c40b59117c16670c1f3b9/modules/aws/auth.go

Briansbum:
Sorry I probably misspoke there.

I mean they're not available to be used in any other sections of the AWS module.

For example the S3 functions terratest/modules/aws/s3.go only make use of `NewAuthenticatedSession` with no way to use another method of authentication.

brikis98:
Ah, I see what you mean. I can definitely see cases where you may want to pass in a custom `Session` of some sort. 

I suspect this would require a fairly large change. One way to do it that is mostly backwards compatible is to update every method that uses the AWS APIs from:

```go
func SomeMethodThatUsesAWS(t *testing.T, someParam string) {
  sess := NewAuthenticateSession()
  ec2 := ec2.New(sess)
}
```

To:

```go
func SomeMethodThatUsesAWS(t *testing.T, someParam string, sess ... *session.Session) {
  sess := NewAuthenticateSession(sess)
  ec2 := ec2.New(sess)
}
```

And to update `NewAuthenticatedSession` to use the passed in `Session`, if one is specified:

```go
func NewAuthenticatedSession(sess ... *session.Session) {
  if len(sess) > 0 {
    return sess[0]
  }

  // Otherwise, use the original logic
}
```

Briansbum:
This looks like it would work nicely, I'll have a look at making it work.

Briansbum:
Resolved in #119

PR #119, opened 2018-06-22
Add an environment variable to use aws role auth instead of access keys

Briansbum:
[Issue](https://github.com/gruntwork-io/terratest/issues/112)
[Tests](https://gist.github.com/Briansbum/450553b11f33bf567c39596ce6d095fe)

Currently there's no way to use AWS role based authentication to get a session and then use that session in other functions in terratest. This allows you to set the variable "AWS_ROLE_ARN" in your environment and this will override any session made with access keys.

### DIFF
--- a/examples/terraform-aws-example/README.md
+++ b/examples/terraform-aws-example/README.md
@@ -25,7 +25,7 @@ it should be free, but you are completely responsible for all AWS charges.
 1. Sign up for [AWS](https://aws.amazon.com/).
 1. Configure your AWS credentials using one of the [supported methods for AWS CLI
    tools](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html), such as setting the
-   `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables.
+   `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables. If you're using the `~/.aws/config` file for profiles then export `AWS_SDK_LOAD_CONFIG` as "True".
 1. Set the AWS region you want to use as the environment variable `AWS_DEFAULT_REGION`.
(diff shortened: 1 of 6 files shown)


brikis98:
Thanks for the PR! 

One sanity check: you can assume an IAM role and expose env vars for it that automatically get picked up by all AWS SDK tools, including Terratest. To do that, you run the [assume-rule](https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html) command with the AWS CLI and export the parameters you get back as the env vars `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`.

Would using that (standardized) approach work better for you than this custom env var?

Briansbum:
I was looking deeper in to what caused me to need to use the env var.

Turns out that with the GO SDK for aws you don't need to give it a profile or a role so long as they exist in your `~/.aws/config` file. All that you need do is export `AWS_SDK_LOAD_CONFIG` and set it to true.

I guess this means that this PR should either be to set that to true all of the time or to put this information in docs somewhere as it's specific to the SDK.

Which would you prefer @brikis98?

brikis98:
It's not just the GO SDK; e.g., the [Node SDK does it too](https://stackoverflow.com/questions/44433527/how-to-load-config-from-aws-config). At any rate, I think a documentation update would do the trick, thanks!

brikis98:
It looks like the changes to `examples/terraform-aws-example/main.tf` and `Gopkg.lock` are still in the PR. Also, did you mean to add all the IAM/S3 policy methods? Those look like they belong in a separate PR...

Briansbum:
I went and messed it up yesterday. I'll get it sorted and drop a comment here once it's done.

Briansbum:
@brikis98 It's reverted, sorry for the mix up there.

Issue #122, opened 2018-06-26
Testing Terragrunt?


kamsz:
Is it possible to swap executable that is used from `terraform` to `terragrunt`?

brikis98:
Not currently, but that's a great idea. A `TerraformBinary` field, which if empty defaults to `terraform`, would be a great addition to [terraform.Options](https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/options.go).

kamsz:
@brikis98 I'd stick with a separate module to include additional functionalities of Terragrunt like apply-all.

brikis98:
I think it would be cleaner to avoid duplicating all those code paths. To handle `apply-all`, we could expose a `RunTerraformCommand(command string, args[]string, options *terraform.Options)` method.

kamsz:
I'll try to work on this then.

so87:
I'm confused here: doesn't gruntwork exclusively use terragrunt to work on infrastructure? And shouldn't all "dev" "staging" and "production" folders have terratests written for them (because that is what you seem to suggest in both of your books).  So how does gruntwork actually test terragrunt - do you just write terratests for modules, and use another tool to do smoke tests on terragrunt deployments? @brikis98

brikis98:
> I'm confused here: doesn't gruntwork exclusively use terragrunt to work on infrastructure?

Not exclusively. All of the modules in our [IaC Library](https://gruntwork.io/infrastructure-as-code-library/), where we do the majority of our testing, are written in pure Terraform and do NOT require Terragrunt at all. Also, many of the modules are written in Go, Python, Bash, etc too, as the IaC Library is designed to solve problems across the _entire_ infrastructure, and not just those problems that can be solved by Terraform.

>  So how does gruntwork actually test terragrunt

Terragrunt itself is tested in this repo. See the `_test.go` files. 

> do smoke tests on terragrunt deployments

Correct

so87:
Thank you for the fast response and time, you seem like an extremely busy person.  
After further analyzing the terragrunt and terratest repos let me see if you think how I'm going to test my future configurations make sense.  I tried looking for people using writing go tests for a terragrunt project layout, but couldn't find any - so maybe tons of people are using terragrunt but not actually writing any tests for their infrastructure?.

I want to take a docker image I have and get it running in ec2.  So first I 
1. break out the necessary infrastructure pieces into a modules repository
2. write failing tests for that terraform code using terratest.  Keep working until tests pass
[...]

PR #123, opened 2018-06-27
Terragrunt module

kamsz:
This PR adds Terragrunt module (which is basically a copy of Terraform module).

It also adds an ability to disable stderr from shell commands. This is because Terragrunt prints alot of debug information to stderr, which completely mess up retrieving outputs.

Related to #122.


brikis98:
Thanks for the PR! Instead of copying all these methods, see https://github.com/gruntwork-io/terratest/issues/122#issuecomment-400623844 for a simpler approach.

PR #134, opened 2018-07-12
add empty bucket method

mcalhoun:


modules/aws/s3.go
```diff
@@ -139,6 +139,72 @@ func DeleteS3BucketE(t *testing.T, region string, name string) error {
...
+		}
+
+		//Running the Bulk delete job (limit 1000)
+		_, err = s3Client.DeleteObjects(deleteParams)
+		if err != nil {
+			return err
+		}
+
+		if *(*bucketObjects).IsTruncated { //if there are more objects in the bucket, IsTruncated = true
+			params.Marker = (*deleteParams).Delete.Objects[len((*deleteParams).Delete.Objects)-1].Key
```
brikis98: `bucketObjects.Marker` should do this in a much simpler manner :)

modules/aws/s3.go
```diff
@@ -139,6 +139,72 @@ func DeleteS3BucketE(t *testing.T, region string, name string) error {
...
+			logger.Logf(t, "Requesting next batch | %s", *(params.Marker))
```
brikis98: Do you need to sleep in between to avoid being throttled?

brikis98:
Also, does this work if the bucket has versioning enabled?

modules/aws/s3.go
mcalhoun: It's actually a bulk delete so only 1 operation, so I don't think it will be throttled.

mcalhoun:
Ok, this has been updated to include tests which upload a random number of files to the bucket with both versioned and non-versioned buckets.

PR #138, opened 2018-08-05
Remove build updates

robmorgan:
Have you essentially already implemented these?


robmorgan:
Ignore, I realised its a changelog.

PR #141, opened 2018-08-09
modules/gcp - GCP support

robmorgan:
Hi Guys,

I'm contributing a module with initial support for GCP in Terratest.

Cheers,

Rob

### Packer Module
- [x] Extend Packer template `packer-basic-example` to build both AWS & GCP resources.
- [x] Add `BuildArtifact` & `BuildArtifactE`
- [x] Modify Regex to work with GCE Packer logs: `extractAMIID -> extractArtifactID`.

The change to the Packer module is potentially controversial. I have deprecated the `BuildAmi` and `BuildAmiE` methods and added `BuildArtifact` and `BuildArtifactE`. I thought this makes more sense when working with multiple Packer builders. Thoughts?

### Test Structure Module
- [x] `SaveArtifactID`: A possible replacement for `SaveAmiId`.
- [x] `LoadArtifactID`: A possible replacement for `LoadAmiId`.

#### Misc
- [ ] Find a better way to work with the ENV vars/credentials.
- [ ] GCP refers to Public IPs as External IPs. Should we change the methods accordingly?

#### compute.go
- [x] `GetPublicIPOfInstance`
- [x] `GetPublicIPOfInstanceE`
- [x] `GetLabelsForComputeInstance`
- [x] `GetLabelsForComputeInstanceE`
- [x] `AddLabelsToInstance`
- [x] `AddLabelsToInstanceE`
- [x] `GetInstanceIdsForInstanceGroup`
- [x] `GetInstanceIdsForInstanceGroupE`
- [x] `DeleteImage`
- [x] `DeleteImageE`
- [x] `NewComputeService`
- [x] `NewComputeServiceE`

#### provider.go
- [x] `GetGoogleCredentialsFromEnvVar`
- [x] `GetGoogleProjectIDFromEnvVar`
- [x] `GetGoogleRegionFromEnvVar`
- [x] `multiEnvSearch`

#### region.go
- [x] `GetRandomRegion`
- [x] `GetRandomRegionE`
- [x] `GetRandomZone`
- [x] `GetRandomZoneE`
- [x] `GetAllGcpRegions`
- [x] `GetAllGcpRegionsE`
- [x] `GetAllGcpZones`
- [x] `GetAllGcpZonesE`

#### region_test.go
- [x] `TestGetRandomRegion`
- [x] `TestGetRandomZone`
- [x] `TestGetRandomRegionExcludesForbiddenRegions`
- [x] `TestGetRandomZoneExcludesForbiddenZones`
- [x] `TestGetAllGcpRegions`
- [x] `TestGetAllGcpZones`
- [x] `assertLooksLikeRegionName`
- [x] `assertLooksLikeZoneName`

#### storage.go
- [x] `CreateStorageBucket`
- [x] `CreateStorageBucketE`
- [x] `DeleteStorageBucket`
- [x] `DeleteStorageBucketE`
- [x] `ReadBucketObject`
- [x] `ReadBucketObjectE`
- [x] `WriteBucketObject`
- [x] `WriteBucketObjectE`
- [x] `EmptyStorageBucket`
- [x] `EmptyStorageBucketE`
- [x] `AssertStorageBucketExists`
- [x] `AssertStorageBucketExistsE`

#### storage_test.go
- [x] Add a test that reads and writes an objects to a storage bucket
- [x] Look at removing env vars: `GOOGLE_STORAGE_PROJECT_ID`
- [x] `TestCreateAndDestroyStorageBucket`
- [x] `TestAssertStorageBucketExistsNoFalseNegative`
- [x] `TestAssertStorageBucketExistsNoFalsePositive`

#### terraform-gcp-example
- [x] Format TF code better and add comment blocks
- [x] Create compute resources

#### packer-basic-example
- [x] Add an additional builder for GCP

### Tests
- [x] `test/packer_basic_example_test.go` should only build the AWS AMI.
- [x] `test/packer_gcp_basic_example_test.go` should only build the GCP Image.
- [x] `test/packer_gcp_basic_example_test.go` should bake using a random zone.
- [x] `test/terraform_gcp_example_test.go` should create and validate Compute resources.


modules/gcp/storage.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if _, err := bucket.Attrs(ctx); err != nil {
+		// ErrBucketNotExist
+		return err
+	}
+
+	it := bucket.Objects(ctx, nil)
+	if _, err := it.Next(); err == storage.ErrBucketNotExist {
+		return err
+	}
```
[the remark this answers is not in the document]
robmorgan: Hi @brikis98, I spent a while digging through their Go & Java SDK, but it didn't turn up much luck. It appears some calls e.g: `client.Bucket(name)` create an object but don't actually do any activity across the network.

I stumbled on their own integration tests and adapted a bit of the code from there.

You might want to take a look: https://github.com/GoogleCloudPlatform/google-cloud-go/blob/de879f7be552d57556875b8aaa383bce9396cc8c/storage/integration_test.go#L1231

I thought it might be a tad easier and theres still the possibilty I'm on the wrong track.

examples/terraform-gcp-example/main.tf
```diff
@@ -0,0 +1,9 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# CREATE A GOOGLE STORAGE BUCKET
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "google_storage_bucket" "test_bucket" {
+  name     = "${var.bucket_name}"
+  location = "${var.location}"
+  project  = "${var.project}"
+}
```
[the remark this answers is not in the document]
robmorgan: Yeah, that's a good point. I think this example was originally added by @Dahs81. I'll either cleanup the README or even better should we actually add compute resources like `terraform-aws-example`, then add tests for those?

And yes we should definitely add a provider block just for clarity.

Right now I'm only setting the following ENV vars locally to test the GCP module:

* `GOOGLE_APPLICATION_CREDENTIALS`
* `GOOGLE_CLOUD_PROJECT_ID`
* `GOOGLE_STORAGE_PROJECT_ID`

We might be able to consolidate that a bit better.

modules/gcp/storage_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/random"
+)
+
+var (
+	projectId = os.Getenv("GOOGLE_STORAGE_PROJECT_ID")
```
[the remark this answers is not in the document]
robmorgan: Not that I'm aware of.

I have a feeling @Dahs81 may have adapted it from the Google examples, see: https://cloud.google.com/storage/docs/reference/libraries#client-libraries-usage-go.

What would be the best way for us to handle `PROJECT_ID` for GCP in Terratest? I guess the AWS GO SDK just automatically reads ENV vars right?

modules/gcp/storage_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func TestCreateAndDestroyStorageBucket(t *testing.T) {
+	t.Parallel()
+
+	id := random.UniqueId()
+	logger.Logf(t, "Random values selected Id = %s\n", id)
+
+	gsBucketName := "gruntwork-terratest-" + strings.ToLower(id)
+
+	CreateStorageBucket(t, projectId, gsBucketName, nil)
```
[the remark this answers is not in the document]
robmorgan: ok good point!

Dahs81:
Thanks, @robmorgan for taking over. I have been busy doing a bunch of k8s stuff and haven't had time to work on this. I appreciate your willingness to take over. Sorry for not getting too far with this.

modules/gcp/storage_test.go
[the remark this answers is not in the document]
Dahs81: One option is that you could create something similar to the ~/.aws directory, but for terratest. Maybe have a ~/.terratest/config.toml that looks something like this:

```toml
[google]
project_id=somegoogleprojectidhere
```

Then maybe have a `terratest config` command that configures it. Then check if that file exists. I'd probably use env variable as a failback though.

I'm not sure if that is too much for this feature. Just an idea.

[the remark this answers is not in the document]
brikis98: @josh-padnick Could you chime on with thoughts on how to handle GCP env vars?

robmorgan:
@Dahs81 yeah sure, not a problem!

modules/gcp/compute.go
```diff
@@ -0,0 +1,196 @@
...
+	ctx := context.Background()
+
+	service, err := NewComputeServiceE(t)
+	if err != nil {
+		return nil, err
+	}
+
+	rb := &compute.InstanceGroupsListInstancesRequest{
+		InstanceState: "ALL",
+	}
```
josh-padnick: noob question, but what does `rb` stand for?

modules/gcp/compute.go
```diff
@@ -0,0 +1,196 @@
...
+		return nil, fmt.Errorf("Failed to get default client: %v", err)
+	}
+
+	service, err := compute.New(client)
+	if err != nil {
+		return nil, err
+	}
+
+	return service, nil
+}
```
josh-padnick: This code is clear and well-documented. Nicely done. 👍

modules/gcp/region.go
```diff
@@ -0,0 +1,188 @@
...
+	return out
+}
+
+// GetAllGcpRegionsE gets the list of GCP regions available in this account.
+func GetAllGcpRegionsE(t *testing.T) ([]string, error) {
+	logger.Log(t, "Looking up all GCP regions available in this account")
+
+	// TODO - NewComputeServiceE creates a context, should we get that somehow
+	// or use a new one here?
+	ctx := context.Background()
```
josh-padnick: This seems cleaner for now. I still haven't seen the `context` actually used for anything these GCP functions so far, so my vote is to keep it simple and consider alternatives if/when necessary.

modules/gcp/region_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// The typical account had access to 52 zones as of August, 2018: https://cloud.google.com/compute/docs/regions-zones/
+	assert.True(t, len(zones) >= 52, "Number of zones: %d", len(zones))
+	for _, zone := range zones {
+		assertLooksLikeZoneName(t, zone)
+	}
+}
+
+func assertLooksLikeRegionName(t *testing.T, regionName string) {
+	assert.Regexp(t, "[a-z]+-[a-z]+?[[:digit:]]+", regionName)
```
josh-padnick: Should this be `[a-z]+-[a-z]+[[:digit:]]+` without the `?`?

modules/gcp/storage.go
```diff
@@ -0,0 +1,214 @@
...
+
+	if _, err := io.Copy(w, body); err != nil {
+		return "", err
+	}
+	if err := w.Close(); err != nil {
+		return "", err
+	}
+
+	const publicURL = "https://storage.googleapis.com/%s/%s"
+	return fmt.Sprintf(publicURL, bucketName, filePath), nil
```
josh-padnick: Does it make more sense to return the bucket object itself and have a separate `GetBucketUrl(bucketObject)` function?

modules/gcp/storage.go
```diff
@@ -0,0 +1,214 @@
...
+
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	// List all objects in the bucket
+	//
+	// TODO - we should really do a bulk delete call here, but I couldn't find
+	// anything in the SDK.
```
josh-padnick: AFAIK, the AWS API has the same limitations!

modules/gcp/storage.go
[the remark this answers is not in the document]
josh-padnick: I had the same reaction as @brikis98, though I agree the API here is confusing. Let's add a comment clarifying the uncertainty and linking back to the URL you posted above. If this passes tests, I think it's good enough for now.

modules/gcp/storage_test.go
```diff
@@ -0,0 +1,75 @@
...
+	fileReader := ReadBucketObject(t, gsBucketName, testFilePath)
+
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(fileReader)
+	result := buf.String()
+
+	require.Equal(t, testFileBody, result)
+
+	// Empty the storage bucket so we can delete it
+	EmptyStorageBucket(t, gsBucketName)
```
josh-padnick: If this test fails, the `defer DeleteStorageBucket(t, gsBucketName)` will run, but it looks like that will only succeed if `EmptyStorageBucket(t, gsBucketName)` runs, so does this need `defer` also?

modules/gcp/storage_test.go
```diff
@@ -0,0 +1,75 @@
...
+	gsBucketName := "gruntwork-terratest-" + strings.ToLower(id)
+	logger.Logf(t, "Random values selected Id = %s\n", id)
+
+	// Don't create a new storage bucket so we can confirm that our function works as expected.
+
+	err := AssertStorageBucketExistsE(t, gsBucketName)
+	if err == nil {
+		t.Fatalf("Function claimed that the Storage Bucket '%s' exists, but in fact it does not.", gsBucketName)
+	}
+}
```
josh-padnick: These tests are great.

modules/test-structure/save_test_data.go
```diff
@@ -103,14 +103,30 @@ func LoadInt(t *testing.T, testFolder string, name string) int {
...
+// LoadArtifactID loads and unserializes an Artifact ID from the given folder. This allows you to reuse an Artifact that was created during an
+// earlier setup step in later validation and teardown steps.
+func LoadArtifactID(t *testing.T, testFolder string) string {
+	return LoadString(t, testFolder, "Artifact")
+}
+
 // SaveAmiId serializes and saves an AMI ID into the given folder. This allows you to build an AMI during setup and to reuse that
 // AMI later during validation and teardown.
+//
+// deprecated.
```
josh-padnick: It's a nit, but it looks there is actually a [golang convention for marking a function as deprecated](https://rakyll.org/deprecated/). Can update accordingly?

modules/test-structure/save_test_data.go
```diff
@@ -103,14 +103,30 @@ func LoadInt(t *testing.T, testFolder string, name string) int {
...
 func SaveAmiId(t *testing.T, testFolder string, amiId string) {
 	SaveString(t, testFolder, "AMI", amiId)
 }
 
 // LoadAmiId loads and unserializes an AMI ID from the given folder. This allows you to reuse an AMI  that was created during an
 // earlier setup step in later validation and teardown steps.
+//
+// deprecated.
```
josh-padnick: See above comment.

test/packer_gcp_basic_example_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Template: "../examples/packer-basic-example/build.json",
+
+		// Variables to pass to our Packer build using -var options
+		Vars: map[string]string{
+			"gcp_project_id": projectID,
+			"gcp_zone":       zone,
+		},
+
+		// Only build the AWS AMI
+		Only: "googlecompute",
```
josh-padnick: s/Only build the AWS AMI/Only build the Google Compute Image/

modules/gcp/storage.go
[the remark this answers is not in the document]
robmorgan: @josh-padnick ok sure

Gopkg.toml
```diff
@@ -52,3 +52,11 @@
...
   go-tests = true
   unused-packages = true
+
+[[constraint]]
+  name = "cloud.google.com/go"
+  version = "0.26.0"
+
+[[constraint]]
+  branch = "master"
+  name = "google.golang.org/api"
```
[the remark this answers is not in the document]
robmorgan: Hi @josh-padnick, I agree that the new Google Cloud Platform Go SDK looks like the right way to go, but for reasons, I'm not entirely sure of, it seems they are still yet to implement all of the necessary API calls. For example, there are essentially no available methods for working with Compute Instances. I noticed in their own integration tests they also use the autogenerated library (see: https://github.com/GoogleCloudPlatform/google-cloud-go/blob/8007d2620d6fbc9f7eeb0a0ee571fe802b866353/storage/go110.go).

In addition I found other tools like Terraform also depend on this SDK (see: https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_storage_bucket_object.go#L14).

I had to use `branch = "master"` because it appears they don't push tags or use the GitHub release system which is required for Dep. Maybe I'm wrong?

examples/terraform-gcp-example/main.tf
```diff
@@ -0,0 +1,47 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# CONFIGURE THE GOOGLE CLOUD PROVIDER
+# ---------------------------------------------------------------------------------------------------------------------
+
+provider "google" {
+  project = "${var.project}"
+  version = "~> 1.16"
```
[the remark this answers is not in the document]
robmorgan: Yep I agree, let me change.

examples/terraform-gcp-example/main.tf
```diff
@@ -0,0 +1,47 @@
...
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "google_compute_instance" "example" {
+  name         = "${var.instance_name}"
+  machine_type = "${var.machine_type}"
+  zone         = "${var.zone}"
+
+  boot_disk {
+    initialize_params {
+      image = "${data.google_compute_image.ubuntu.self_link}"
```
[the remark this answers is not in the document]
robmorgan: Yes, its possible and I'll change. For some reason, I deliberately did this, because I thought the verbosity might be clearer for people reading the tests.

examples/terraform-gcp-example/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "instance_id" {
+  value = "${google_compute_instance.example.id}"
+}
+
+output "bucket_url" {
+  value = "${google_storage_bucket.example_bucket.url}"
+}
```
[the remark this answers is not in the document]
robmorgan: I found some methods in the Google Go SDKs return URLs whereas some return IDs etc. Therefore I assume it depends whether or not you need a specific property for making another API request. In terms of storage, the URL is fairly handly if you want to construct the publicly accessible URL, but yeah beyond that I found it pretty random.

Here's another example where a method returns a URL, when really I need an ID for the next call: https://github.com/gruntwork-io/terratest/pull/141/files#diff-aef32c063b72fb294a6da41ed62b52baR161

modules/gcp/storage_test.go
robmorgan: Yes, good point!

modules/gcp/storage_test.go
robmorgan: thanks!

modules/test-structure/save_test_data.go
robmorgan: Ok, I'll update accordingly.

test/packer_gcp_basic_example_test.go
robmorgan: thanks

examples/terraform-gcp-example/main.tf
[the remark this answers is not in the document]
robmorgan: Turns out this will only work as long as the user has one of the Environment variables set: https://www.terraform.io/docs/providers/google/index.html. I might need to clarify somewhere in the docs.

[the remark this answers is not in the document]
robmorgan: I was using `GOOGLE_CLOUD_PROJECT_ID`. I've now changed the GCP code to prefer: `GOOGLE_CLOUD_PROJECT` as it plays better with other tools like Terraform.

examples/terraform-gcp-example/variables.tf
```diff
@@ -0,0 +1,47 @@
...
+# GOOGLE_APPLICATION_CREDENTIALS
+
+# ---------------------------------------------------------------------------------------------------------------------
+# REQUIRED PARAMETERS
+# You must provide a value for each of these parameters.
+# ---------------------------------------------------------------------------------------------------------------------
+
+variable "project" {
+  description = "The ID of the project to apply any resources to."
+}
```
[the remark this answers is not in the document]
robmorgan: I've removed it altogether as it was only used in the `provider` block.

modules/gcp/accounts.go
```diff
@@ -0,0 +1 @@
+package gcp
```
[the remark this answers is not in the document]
robmorgan: I might remove them altogether to make it more obvious.

modules/gcp/compute.go
robmorgan: request body ;), can change if you like?

robmorgan: technically its probably in the URL

modules/gcp/storage.go
robmorgan: I would argue in that case it should return only the error object as you pass in the `io.Reader` as a parameter. Let me know your thoughts and I'll adjust accordingly.

modules/gcp/region_test.go
robmorgan: Yes, good pickup!

modules/gcp/region.go
robmorgan: Okay so leave as is?

modules/gcp/storage.go
robmorgan: Dang!

robmorgan:
@josh-padnick Hi Josh,

Yes definitely, Jim mentioned this to me last week. I was going to try and break everything down into smaller PRs, but it looks like you've reviewed it all now.. so thanks!

No, I've replied to all of your feedback. Just work through the remaining issues and then merge it as it stands. I'm happy to implement new functionality in subsequent PRs.

examples/terraform-gcp-example/variables.tf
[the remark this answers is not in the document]
josh-padnick: Oh, interesting. Are you saying that you set the project ID in the GCP client code anyway, so it's not actually necessary for the `provider` block?

One sanity check we like to use for AWS is making sure that the AWS Account ID you're authenticated to matches the AWS Account ID passed to your Terraform provider. The GCP equivalent of that would probably be `project_id`. For that reason, my vote is to keep it in.

modules/gcp/accounts.go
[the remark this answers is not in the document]
josh-padnick: Sounds good.

modules/gcp/compute.go
josh-padnick: Oh! Haha, wouldn't have guessed request body. Changing to `requestBody` or `req` sounds good to me.

modules/gcp/region.go
josh-padnick: Yeah, let's leave as is. I was just affirming your decision.

modules/gcp/region_test.go
josh-padnick: Actually, there's one more RegEx that has the `?` as well. I believe it was for Zone.

modules/gcp/storage.go
josh-padnick: Oh, interesting. We can just leave as is.

josh-padnick:
Alright, I think we're ready to merge! There were one or two open comments, so I'll wait for your ok to officially merge. Thanks again for this awesome work.

modules/gcp/compute.go
robmorgan: ok going with `requestBody`

modules/gcp/region.go
robmorgan: cool

modules/gcp/storage.go
robmorgan: ok

robmorgan:
Hi @josh-padnick, I've made another two micro changes based on your feedback, but I'm happy with how it stands now. Ok to merge! 😎 👍

josh-padnick:
'cc @infosecgithub on this significant update to Terratest which will be directly leveraged to add tests for the Consul and Vault modules for GCP.

brikis98:
@robmorgan The original merge of this PR failed because we hadn't yet set up our GCP creds in our CircleCI account. I've now done that and all the GCP tests pass except one:

```
--- FAIL: TestTerraformGcpExample (204.21s)
	terraform_gcp_example_test.go:68: 
			Error Trace:	terraform_gcp_example_test.go:68
			Error:      	Should be true
			Test:       	TestTerraformGcpExample
	terraform_gcp_example_test.go:69: 
			Error Trace:	terraform_gcp_example_test.go:69
			Error:      	Not equal: 
			            	expected: "testing-tag-value2"
			            	actual  : ""
			            	
			            	Diff:
			            	--- Expected
			            	+++ Actual
			            	@@ -1 +1 @@
			            	-testing-tag-value2
			            	+
			Test:       	TestTerraformGcpExample
```

Any idea why?

robmorgan:
@brikis98 Hi Jim,

Just scanning through some of your commits - I believe you may need to create a new project under your Gruntwork GCP account and then add `GOOGLE_CLOUD_PROJECT=foo-123` to the CircleCI build environment.

Here's another test that failed:

```
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region us-east4
--- FAIL: TestGetAllGcpRegions (0.17s)
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region asia-east1
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region europe-west3
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region europe-north1
	region.go:112: googleapi: got HTTP response code 404 with body: Not Found
--- FAIL: TestGetRandomZone (0.23s)
	region.go:76: googleapi: got HTTP response code 404 with body: Not Found
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region southamerica-east1
--- FAIL: TestGetAllGcpZones (0.24s)
	region.go:153: googleapi: got HTTP response code 404 with body: Not Found
--- PASS: TestAssertStorageBucketExistsNoFalsePositive (0.21s)
--- FAIL: TestAssertStorageBucketExistsNoFalseNegative (0.23s)
	storage.go:18: googleapi: Error 400: Unknown project id: , invalid
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region europe-west3
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region us-east1
--- FAIL: TestCreateAndDestroyStorageBucket (0.23s)
	storage.go:18: googleapi: Error 400: Unknown project id: , invalid
TestGetRandomRegionExcludesForbiddenRegions 2018-08-27T19:13:37Z region.go:66: Using region europe-north1
--- FAIL: TestGetRandomRegion (0.24s)
	region.go:38: googleapi: got HTTP response code 404 with body: Not Found
```

I can see it complaining about a project id. We could extend `provider.go` to throw a fatal error?

Your commit:
- https://github.com/gruntwork-io/terratest/commit/5d7c405c19b2ec079b8c4fd2567b7e70b49193bc

brikis98:
Could it be an eventual consistency thing? The tags are applied after a few seconds (sometimes), so it's timing dependent? If so, you may just need to use `retry.DoWithRetry`.

robmorgan:
@brikis98 yeah I believe it probably is. From memory I found during development there was a lag for this operation, I'm going to go down the `retry` path and see if it makes a different. Should I essentially move my asserts to the `retry func`?

brikis98:
Yea. In the example automated tests, anything that is eventually consistent should be done in a retry loop.

PR #142, opened 2018-08-12
Add support for running terraform plan and parsing the results using go-terraform-plan.

justinbarrick:


Gopkg.toml
```diff
@@ -49,6 +49,10 @@
   branch = "master"
   name = "golang.org/x/net"
 
+[[constraint]]
+  name = "github.com/justinbarrick/go-terraform-plan"
```
brikis98: Hm, one worry I have is that currently, we use whatever version of Terraform the user has installed for `terraform apply`, `terraform plan`, etc. This library seems to be built around a specific version of Terraform, so if it doesn't match with what the user has installed, will this fail to parse the `plan` output?

test/fixtures/terraform-plan-no-error/main.tf
```diff
@@ -0,0 +1,12 @@
+provider "digitalocean" {
+  token = "hello"
+}
```
brikis98: We don't have any DigitalOcean creds set up in our CI system. Would it make more sense to use `null_resource` here? After all, we just want to check that the `plan` output is correct, not that it works with any particular provider.

justinbarrick: Changed

Gopkg.toml
justinbarrick: Hmm, this is a good point. It seems like it would be unable to parse the plan output from before they bumped the planFormatVersion:

* https://github.com/hashicorp/terraform/blame/bb37637a139151be0ff9bf00c85c448538e5b3d2/terraform/plan.go#L170
* https://github.com/hashicorp/terraform/commit/1b673746fd0d7a0f97127bf59d1e85b9e99afd62

So it looks like it would be compatible with terraform versions back to v0.10.0, which is only about a year.

It might be better to attempt to parse the human readable output, but that doesn't seem right either.

brikis98: Here's an idea for how to handle this:

1. Add a new set of methods to the `terraform` package that can be used to check the version of Terraform installed. E.g. `GetTerraformVersion()` and `AssertTerraformVersion(constraint)`. 

1. Update the new `Plan` methods you've added to assert the version is greater than v0.10.0... We could also consider logging a warning if the version is greater than 0.11.x to say "this method has not yet been tested with this version..." Might help users debug issues if 0.12.x or newer makes backwards incompatible changes in the plan file format.

nandac:
@here Is anyone still looking into adding this feature to terratest?
I would love to see this feature merged into master and be available in a new release.

Are there any other go packages that would give the ability to parse a plan in test cases. Kindly let me know.

ellisonc:
@justinbarrick, thanks for your contribution. We are removing old PRs. If this change is still required, we encourage you to re-open.

PR #144, opened 2018-08-19
Add methods to fetch files from servers over SSH

brikis98:
This PR adds helper methods that can be used to fetch the contents of files from severs over SSH, as well as some wrapper methods designed specifically for use with EC2 Instances and ASGs. The main use case for this is to fetch the contents of log files (or any other files you need!) from EC2 Instances at the end of the test to make debugging a test failure 10x easier. 

I'm hitting a confusing, intermittent test failure in our Vault code, and I can't repro it locally, so having the test show me the logs of the Vault servers is going to be a huge help.


brikis98:
Merging so I can try this out with the Vault codebase. Feedback very welcome!

PR #146, opened 2018-08-27
Add helper methods for creating SshAgent with KeyPair

joshpurvis:
I was writing a few tests which utilized `remote-exec` for newly generated keys not in my local SSH agent, and I found myself copying the same bits of code found in the `test/terraform_ssh_example_test.go` example.

Here I propose adding two new methods to the ssh package, `SshAgentWithKeyPair` and `SshAgentWithKeyPairs`. With these users will only need to add something such as:

```
sshAgent := ssh.SshAgentWithKeyPair(t, keyPair)
defer sshAgent.Stop()
```


modules/ssh/agent.go
```diff
@@ -0,0 +1,112 @@
...
+type SshAgent struct {
+	stop       chan bool
+	stopped    chan bool
+	socketDir  string
+	socketFile string
+	agent      agent.Agent
+	ln         net.Listener
+}
+
+// Create SSH agent, start it in background and returns control back to the main thread
```
brikis98: You should tell the user that it's their responsibility to stop the agent, and how to do it.

modules/ssh/agent.go
```diff
@@ -0,0 +1,112 @@
...
+// Stop and clean up SSH agent
+func (s *SshAgent) Stop() {
+	close(s.stop)
+	s.ln.Close()
+	<-s.stopped
+	os.RemoveAll(s.socketDir)
+}
+
+// Instantiates and returns an in-memory ssh agent with the given KeyPair already added
+func SshAgentWithKeyPair(t *testing.T, keyPair *KeyPair) *SshAgent {
```
brikis98: Please add `SshAgentWithKeyPairE` and `SshAgentWithKeyPairsE` methods that return the `error` rather than calling `t.Fatal`.

modules/ssh/agent.go
```diff
@@ -0,0 +1,112 @@
...
+func SshAgentWithKeyPairs(t *testing.T, keyPairs []*KeyPair) *SshAgent {
+	t.Log("Generating SSH Agent with given KeyPair(s)")
+
+	// Instantiate a temporary SSH agent
+	socketDir, err := ioutil.TempDir("", "ssh-agent-")
+	if err != nil {
+		t.Fatal(err)
+	}
+	socketFile := filepath.Join(socketDir, "ssh_auth.sock")
+	os.Setenv("SSH_AUTH_SOCK", socketFile)
```
brikis98: Hm, this won't work with multiple tests running in parallel. Is it strictly necessary to set this env var?

modules/ssh/agent_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// ensure that socket directory is set in environment, and it exists
+	sockFile := filepath.Join(sshAgent.socketDir, "ssh_auth.sock")
+	sockFileEnv, found := os.LookupEnv("SSH_AUTH_SOCK")
+	assert.FileExists(t, sockFile)
+	assert.True(t, found)
+	assert.Equal(t, sockFileEnv, sockFile)
+
+	// assert that there's 1 key in the agent
+	keys, _ := sshAgent.agent.List()
```
brikis98: Don't ignore errors. `require.NoError` on it instead.

modules/ssh/agent_test.go
```diff
@@ -0,0 +1,42 @@
...
+	assert.Len(t, keys, 1)
+
+
+	// is socketDir removed as expected?
+	if _, err := os.Stat(sockFile); os.IsExist(err) {
```
brikis98: When does it get removed?

test/terraform_ssh_example_test.go
```diff
@@ -243,50 +236,9 @@ func testSSHAgentToPublicHost(t *testing.T, terraformOptions *terraform.Options,
...
-	if err != nil {
-		t.Fatal(err)
-	}
-	socketFile := filepath.Join(socketDir, "ssh_auth.sock")
-	os.Setenv("SSH_AUTH_SOCK", socketFile)
-	sshAgent, err := NewSSHAgent(socketDir, socketFile)
-	if err != nil {
-		t.Fatal(err)
-	}
+	sshAgent := ssh.SshAgentWithKeyPair(t, keyPair.KeyPair)
```
brikis98: Much nicer API 👍

modules/ssh/agent.go
joshpurvis: Added

modules/ssh/agent.go
joshpurvis: Fixed

modules/ssh/agent.go
joshpurvis: I just copied this from the existing example, but I just tested with that line commented out and it fails to connect. Will look into this further to see if there's away around it.

modules/ssh/agent_test.go
joshpurvis: Fixed

modules/ssh/agent.go
joshpurvis: Here is what requires the env variable: https://github.com/gruntwork-io/terratest/blob/master/modules/ssh/ssh.go#L328

Question -- does it ever make sense to read from the local ssh-agent, when `Host.SshAgent == true`? In my opinion they shouldn't do that, because it makes the tests brittle and not very portable. They'll only work if the key(s) are present in the agent of the machine running the tests. 

Either way I think the solution is to store a reference to the instance of `SshAgent` on the Host struct:

	privateHost := ssh.Host{
		Hostname:    privateInstanceIP,
		SshUserName: "ubuntu",
		OverrideSshAgent: sshAgent,  // or replace the existing SshAgent bool?
	}

Then you could add another if statement (or replace the SshAgent bool) here: https://github.com/gruntwork-io/terratest/blob/master/modules/ssh/ssh.go#L327

(Replacing the bool would probably be a backwards incompatible change.)

brikis98: Ah, good find. Adding an `OverrideSshAgent` to `ssh.Host` sounds like a clean solution 👍

joshpurvis: Ok.. updated branch with the `OverrideSshAgent` change. I also just added another unit test which confirms that it can run multiple agents at once without mixing up keys. One thing to note is that I refactored the conditionals in `createAuthMethodsForHost` (they were getting unwieldy and confusing). As far as I can tell, the behavior is the same, but a second set of eyes couldn't hurt.

modules/ssh/agent.go
```diff
@@ -0,0 +1,136 @@
...
+func SshAgentWithKeyPairsE(t *testing.T, keyPairs []*KeyPair) (*SshAgent, error) {
+	t.Log("Generating SSH Agent with given KeyPair(s)")
+
+	// Instantiate a temporary SSH agent
+	socketDir, err := ioutil.TempDir("", "ssh-agent-")
+	if err != nil {
+		return nil, err
+	}
+	socketFile := filepath.Join(socketDir, "ssh_auth.sock")
+	//os.Setenv("SSH_AUTH_SOCK", socketFile)
```
brikis98: Remove?

modules/ssh/agent_test.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSshAgentWithKeyPair(t *testing.T) {
```
brikis98: `t.Parallel()` at the top of each test?

modules/ssh/ssh.go
```diff
@@ -23,9 +23,11 @@ import (
 type Host struct {
 	Hostname    string // host name or ip address
 	SshUserName string // user name
-	// set one or both authentication methods
+	// set one or more authentication methods,
+	// the first valid method will be used
 	SshKeyPair *KeyPair // ssh key pair to use as authentication method (disabled by default)
 	SshAgent   bool     // enable authentication using your existing local SSH agent (disabled by default)
+	OverrideSshAgent *SshAgent
```
brikis98: Add comment explaining that this is an in-process SSH Agent that will be used for the test, where as `SsgAgent` (`bool`) uses the default SSH Agent running on the OS.

brikis98:
I'll let the tests run, and once they pass, I'll create a new release and paste the link here.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.10.1

joshpurvis:
Awesome! ... However, I just thought of a possible problem, one that was implicitly working before because of the environment variable swap magic.

Should `terraform.InitAndApply()` (namely the apply step) also be driven by the in-process ssh agent?

The use case I'm considering is if the terraform plan uses a remote-exec that relies on the SSH agent for connection. 

Is this a use case you want to cover with terratest?

Here's a snippet that might illustrate it better:

```
keyPair := aws.CreateAndImportEC2KeyPair(t, awsRegion, uniqueName)

	TerraformDir: terraformDirectory,
	Vars: map[string]interface{}{
		"aws_region":        awsRegion,
		"instance_name":     uniqueName,
		"aws_key_pair_name": keyPair.Name, // this is attached to the instance, 
                                                   // so that remote-exec works
	},
}

sshAgent := ssh.SshAgentWithKeyPair(t, ec2KeyPair.KeyPair)
defer sshAgent.Stop()

// In the previous iteration `SSH_AUTH_SOCK` env var was changed to the
//  in-process agent's socket file at this point, which then was then 
// passed along to terraform:
terraform.InitAndApply(t, terraformOptions)
```

brikis98:
How are you thinking of forcing Terraform to use the in-process SSH agent?

joshpurvis:
Forcing terraform isn't a problem, all you need to do is pass the ssh agent into the `terraform.Options` struct, and then set the environment before launching terraform:
```
os.Setenv("SSH_AUTH_SOCK", SshAgent.socketFile)
exec.Command("terraform", ... )
```
.. I've tested that and it seems to work.

But then you run into the problem with `t.parallel()` again. 

The solution that comes to mind is to use a mutex to only allow one test to change that environment variable at a time. I'm not sure if that's considered hacky or not, I'm still kind of new to go. I also don't know off hand if that env var needs to be in place for the life of the terraform process, or just on initiation. The latter would obviously be more ideal, for concurrency's sake.

If all that seems too convoluted, I understand. I don't need this feature for anything important right now, and I've already hacked around it. I just figured I should point out that terraform modules with `remote-exec` (or any terraform ssh connection) are potentially untestable with terratest, unless the keys are already present in the local SSH Agent. (making the tests less portable)

brikis98:
Oh, you know what, using env vars, I think we can make this work even with `t.parallel()`. To execute `terraform`, [Terratest calls shell. RunCommandAndGetOutputE](https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/cmd.go#L31-L38). This method allows you to [configure the environment variables](https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/cmd.go#L35) to use when shelling out to `terraform`. We could set `SSH_AUTH_SOCK` there.

joshpurvis:
Nice find! I guess I had assumed go used whatever environment was available to it when launching subshells. I'll take a look at this in the morning. It should be pretty easily to implement, if that does what it says it does.

Should I amend this PR or open a new one?

brikis98:
New PR please and thank you 🍺

PR #150, opened 2018-08-31
Add ability to override local SSH Agent when testing modules

joshpurvis:
This PR gives Terratest the ability to override the local SSH agent when testing modules which use `remote-exec`, `files`, and other ssh-based [provisioners](https://www.terraform.io/docs/provisioners/index.html).

This can be achieved by creating an in-process ssh agent with the keyPair, and then passing it to terraform via `terraform.Options`:
```
sshAgent := ssh.SshAgentFromKeyPair(t, keyPair.KeyPair)
defer sshAgent.Stop()

	TerraformDir: terraformDirectory,
	Vars: map[string]interface{}{
		"aws_region":    awsRegion,
		"instance_name": instanceName,
		"key_pair_name": keyPair.Name,
	},
	SshAgent: sshAgent, // Overrides local SSH agent with our new agent
}

terraform.InitAndApply(t, terraformOptions)
```

I've also added a new integration test/example which is linked from the README file. This test could probably be combined with another integration test, if testing time or testing costs are a constraint. I chose to make it a separate test for readability.


README.md
```diff
@@ -91,6 +91,7 @@ such as:
 1.  [Terraform Packer Example](/examples/terraform-packer-example): A more complicated example that shows how to use
     Packer to build an AMI with a web server installed and deploy that AMI in AWS using Terraform.
 1.  [Terraform GCP Example](/examples/terraform-gcp-example): A simple Terraform configuration that creates a GCP Compute Instance and Storage Bucket.
+1.  [Terraform remote-exec Example](/examples/terraform-remote-exec-example): A terraform configuration that creates and AWS instance and then uses `remote-exec`to provision it.
```
brikis98: Minor typo: missing space after `remote-exec`.

examples/terraform-remote-exec-example/files/get-public-ip.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# For example purposes, print the public IP address of this instance
+echo $(wget -q -O- http://169.254.169.254/latest/meta-data/public-ipv4)
```
brikis98: Is the `echo` and subshell necessary? Won't calling `wget` directly produce the exact same output?

test/terraform_remote_exec_example_test.go
```diff
@@ -0,0 +1,109 @@
...
+		// Give this EC2 Instance and other resources in the Terraform code a name with a unique ID so it doesn't clash
+		// with anything else in the AWS account.
+		instanceName := fmt.Sprintf("terratest-remote-exec-example-%s", uniqueID)
+
+		// Pick a random AWS region to test in. This helps ensure your code works in all regions.
+		awsRegion := aws.GetRandomRegion(t, nil, nil)
+
+		// Create an EC2 KeyPair that we can use for SSH access
+		keyPairName := fmt.Sprintf("terratest-remote-exec-example-%s", uniqueID)
+		keyPair := aws.CreateAndImportEC2KeyPair(t, awsRegion, keyPairName)
```
brikis98: `defer` deleting the KeyPair.

test/terraform_remote_exec_example_test.go
```diff
@@ -0,0 +1,109 @@
...
+		// Check that public IP from output matches public IP generated by script on the server
+		b, err := ioutil.ReadFile(testFile)
+		if err != nil {
+			fmt.Print(err)
+		}
+		assert.Equal(t, strings.TrimSpace(publicIP), strings.TrimSpace(string(b)))
+	})
+
+	// At the end of the test, run `terraform destroy` to clean up any resources that were created
+	defer test_structure.RunTestStage(t, "teardown", func() {
```
brikis98: This should be the very first thing called in this test. Otherwise, if an earlier stage fails, it'll never even get to this point to `defer` the cleanup code.

test/terraform_remote_exec_example_test.go
joshpurvis: This was already in the tearDown step that I just moved from the bottom. Let me know if you meant something else.

PR #152, opened 2018-09-03
Prevent possible resource leakage in packer example

joshpurvis:
As it was previously, if one were to use `SKIP_*` env variables, it could potentially generate a different `awsRegion` each iteration and fail and/or leak AWS resources. To correct, I've saved the random `awsRegion` in the first step using `test_structure.SaveString` and loaded it as needed in the subsequent steps using `test_structure.LoadString`.

(This particular example caused me some confusion when I was first learning the `test_structure` module last week. It made me wrongly assume `test_structure` was doing something more magical. )

### DIFF
--- a/test/terraform_packer_example_test.go
+++ b/test/terraform_packer_example_test.go
@@ -22,19 +22,20 @@ import (
 func TestTerraformPackerExample(t *testing.T) {
 	t.Parallel()
 
-	// Pick a random AWS region to test in. This helps ensure your code works in all regions.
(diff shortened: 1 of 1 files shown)


brikis98:
Ah, good catch. 👍 These are definitely bugs!

> Also just noticed this as pushing this PR.. should the defers be moved to the top like you suggested for my remote-exec example?

Yes please!

joshpurvis:
@brikis98 Should be ready to go now!

PR #155, opened 2018-09-17
Terraform plan and assert for changes present and absent

mikhailadvani:
Methods Added:
- `AssertPlanHasNoChanges`
- `AssertPlanHasChanges`
- `GetExitCodeForTerraformCommand`
- `InitAndPlan`
- `PlanExitCode`

### DIFF
--- a/modules/terraform/cmd.go
+++ b/modules/terraform/cmd.go
@@ -59,3 +59,46 @@ func RunTerraformCommandE(t *testing.T, options *Options, args ...string) (strin
 		return out, retry.FatalError{Underlying: err}
 	})
 }
+
+// GetExitCodeForTerraformCommand runs terraform with the given arguments and options and return stdout/stderr.
+func GetExitCodeForTerraformCommand(t *testing.T, options *Options, args ...string) int {
+	exitCode, err := GetExitCodeForTerraformCommandE(t, options, args...)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return exitCode
+}
+
+// GetExitCodeForTerraformCommandE runs terraform with the given arguments and options and return stdout/stderr.
+func GetExitCodeForTerraformCommandE(t *testing.T, options *Options, args ...string) (int, error) {
+	if options.NoColor && !collections.ListContains(args, "-no-color") {
+		args = append(args, "-no-color")
+	}
+
+	// if SshAgent is provided, override the local SSH agent with the socket of our in-process agent
+	if options.SshAgent != nil {
+		// Initialize EnvVars, if it hasn't been set yet
+		if options.EnvVars == nil {
+			options.EnvVars = map[string]string{}
+		}
+		options.EnvVars["SSH_AUTH_SOCK"] = options.SshAgent.SocketFile()
+	}
+
+	fmt.Printf("Running terraform %v", args)
+	cmd := shell.Command{
+		Command:    "terraform",
+		Args:       args,
+		WorkingDir: options.TerraformDir,
+		Env:        options.EnvVars,
+	}
+
+	_, err := shell.RunCommandAndGetOutputE(t, cmd)
+	if err == nil {
+		return 0, nil
+	}
+	exitCode, getExitCodeErr := shell.GetExitCodeForRunCommandError(err)
+	if getExitCodeErr == nil {
+		return exitCode, nil
+	}
+	return -1, getExitCodeErr
+}
--- a/modules/terraform/plan.go
+++ b/modules/terraform/plan.go
@@ -1,0 +1,63 @@
+package terraform
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// InitAndPlan runs terraform init and plan with the given options and return stdout/stderr from the apply command.
+func InitAndPlan(t *testing.T, options *Options) int {
+	exitCode, err := InitAndPlanE(t, options)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return exitCode
+}
+
+// InitAndPlanE runs terraform init and plan with the given options and return stdout/stderr from the apply command.
+func InitAndPlanE(t *testing.T, options *Options) (int, error) {
+	if _, err := InitE(t, options); err != nil {
+		return -1, err
+	}
+
+	if _, err := GetE(t, options); err != nil {
+		return -1, err
+	}
+
+	return PlanExitCodeE(t, options)
+}
+
+// PlanExitCode runs terraform apply with the given options and returns the detailed exitcode.
+func PlanExitCode(t *testing.T, options *Options) int {
+	exitCode, err := PlanExitCodeE(t, options)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return exitCode
+}
+
+// PlanExitCodeE runs terraform apply with the given options and returns the detailed exitcode.
+func PlanExitCodeE(t *testing.T, options *Options) (int, error) {
+	exitCode, err := GetExitCodeForTerraformCommandE(t, options, FormatArgs(options.Vars, "plan", "-input=false", "-lock=false", "-detailed-exitcode")...)
+	if err != nil {
+		return -1, err
+	}
+	return exitCode, nil
+}
+
+// AssertPlanHasNoChanges run terraform plan with detailed-exitcode and asserts for no changes
+func AssertPlanHasNoChanges(t *testing.T, options *Options) {
+	t.Parallel()
+
+	exitCode := InitAndPlan(t, options)
+	assert.Equal(t, 0, exitCode)
+}
+
+// AssertPlanHasChanges run terraform plan with detailed-exitcode and asserts for changes
+func AssertPlanHasChanges(t *testing.T, options *Options) {
+	t.Parallel()
+
+	exitCode := InitAndPlan(t, options)
+	assert.Equal(t, 2, exitCode)
+}
--- a/modules/terraform/plan_test.go
+++ b/modules/terraform/plan_test.go
@@ -1,0 +1,57 @@
+package terraform
+
+import (
+	"terratest/modules/aws"
+	"terratest/modules/files"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPlanWithNoChanges(t *testing.T) {
+	testFolder, err := files.CopyTerraformFolderToTemp("../../test/fixtures/terraform-no-error", t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	awsRegion := aws.GetRandomRegion(t, nil, nil)
+	options := &Options{
+		TerraformDir: testFolder,
+
+		EnvVars: map[string]string{
+			"AWS_DEFAULT_REGION": awsRegion,
+		},
+	}
+	AssertPlanHasNoChanges(t, options)
+}
+
+func TestPlanWithChanges(t *testing.T) {
+	testFolder, err := files.CopyTerraformFolderToTemp("../../examples/terraform-aws-example", t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	awsRegion := aws.GetRandomRegion(t, nil, nil)
+	options := &Options{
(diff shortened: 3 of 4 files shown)


modules/terraform/cmd.go
```diff
@@ -59,3 +59,46 @@ func RunTerraformCommandE(t *testing.T, options *Options, args ...string) (strin
 		return out, retry.FatalError{Underlying: err}
 	})
 }
+
+// GetExitCodeForTerraformCommand runs terraform with the given arguments and options and return stdout/stderr.
```
brikis98: Comment doesn't seem accurate? It doesn't return stdout/stderr, it returns the exit code, right?

modules/terraform/cmd.go
```diff
@@ -59,3 +59,46 @@ func RunTerraformCommandE(t *testing.T, options *Options, args ...string) (strin
...
+	}
+
+	// if SshAgent is provided, override the local SSH agent with the socket of our in-process agent
+	if options.SshAgent != nil {
+		// Initialize EnvVars, if it hasn't been set yet
+		if options.EnvVars == nil {
+			options.EnvVars = map[string]string{}
+		}
+		options.EnvVars["SSH_AUTH_SOCK"] = options.SshAgent.SocketFile()
+	}
```
brikis98: Don't we have the `NoColor` and `SshAgent` logic elsewhere already? If so, we should extract it into a common place where we can reuse it rather than copy/pasting.

modules/terraform/cmd.go
```diff
@@ -59,3 +59,46 @@ func RunTerraformCommandE(t *testing.T, options *Options, args ...string) (strin
...
+
+	fmt.Printf("Running terraform %v", args)
```
brikis98: Please use the Terratest logger rather than `fmt.Printf`.

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
+package terraform
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// InitAndPlan runs terraform init and plan with the given options and return stdout/stderr from the apply command.
```
brikis98: Again, the comment doesn't seem right

modules/terraform/cmd.go
```diff
@@ -59,3 +59,46 @@ func RunTerraformCommandE(t *testing.T, options *Options, args ...string) (strin
...
+
+	_, err := shell.RunCommandAndGetOutputE(t, cmd)
+	if err == nil {
+		return 0, nil
+	}
+	exitCode, getExitCodeErr := shell.GetExitCodeForRunCommandError(err)
+	if getExitCodeErr == nil {
+		return exitCode, nil
+	}
+	return -1, getExitCodeErr
```
brikis98: We need to document this `-1` default somewhere and put it into a constant

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
...
+	return exitCode
+}
+
+// InitAndPlanE runs terraform init and plan with the given options and return stdout/stderr from the apply command.
+func InitAndPlanE(t *testing.T, options *Options) (int, error) {
+	if _, err := InitE(t, options); err != nil {
+		return -1, err
+	}
+
+	if _, err := GetE(t, options); err != nil {
```
brikis98: I don't think we need a separate call to `Get` if we're already calling `Init`?

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
...
+	exitCode, err := PlanExitCodeE(t, options)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return exitCode
+}
+
+// PlanExitCodeE runs terraform apply with the given options and returns the detailed exitcode.
+func PlanExitCodeE(t *testing.T, options *Options) (int, error) {
+	exitCode, err := GetExitCodeForTerraformCommandE(t, options, FormatArgs(options.Vars, "plan", "-input=false", "-lock=false", "-detailed-exitcode")...)
```
brikis98: Why `-lock=false`?

brikis98: Also, you can just do `return GetExitCodeForTerraformCommandE(...)` here.

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return -1, err
+	}
+	return exitCode, nil
+}
+
+// AssertPlanHasNoChanges run terraform plan with detailed-exitcode and asserts for no changes
+func AssertPlanHasNoChanges(t *testing.T, options *Options) {
+	t.Parallel()
```
brikis98: Remove

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
...
+
+	exitCode := InitAndPlan(t, options)
+	assert.Equal(t, 0, exitCode)
```
brikis98: Put 0 into a constant with a name. Add a link to the terraform plan docs.

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// AssertPlanHasChanges run terraform plan with detailed-exitcode and asserts for changes
+func AssertPlanHasChanges(t *testing.T, options *Options) {
+	t.Parallel()
```
brikis98: Remove

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
...
+
+	exitCode := InitAndPlan(t, options)
+	assert.Equal(t, 2, exitCode)
```
brikis98: Put 2 into a constant with a name. Add a link to the terraform plan docs.

modules/terraform/plan.go
```diff
@@ -0,0 +1,63 @@
...
+func PlanExitCodeE(t *testing.T, options *Options) (int, error) {
+	exitCode, err := GetExitCodeForTerraformCommandE(t, options, FormatArgs(options.Vars, "plan", "-input=false", "-lock=false", "-detailed-exitcode")...)
+	if err != nil {
+		return -1, err
+	}
+	return exitCode, nil
+}
+
+// AssertPlanHasNoChanges run terraform plan with detailed-exitcode and asserts for no changes
+func AssertPlanHasNoChanges(t *testing.T, options *Options) {
```
brikis98: There should probably be `AssertPlanHasNoChangesE` method that returns an error if the plan has changes... And this method merely calls that other one and calls `t.Fatal` if an error is returned.

modules/terraform/cmd.go
mikhailadvani: updated

modules/terraform/cmd.go
mikhailadvani: fixed

modules/terraform/plan.go
mikhailadvani: updated

modules/terraform/plan.go
mikhailadvani: done

modules/terraform/plan.go
mikhailadvani: done

modules/terraform/plan.go
mikhailadvani: as suggested have deleted the function `AssertPlanHasNoChanges` altogether.

modules/terraform/plan.go
mikhailadvani: done

modules/terraform/plan.go
mikhailadvani: done

modules/terraform/cmd.go
mikhailadvani: done

modules/terraform/cmd.go
```diff
@@ -59,3 +65,35 @@ func RunTerraformCommandE(t *testing.T, options *Options, args ...string) (strin
...
+	cmd := shell.Command{
+		Command:    "terraform",
+		Args:       args,
+		WorkingDir: options.TerraformDir,
+		Env:        options.EnvVars,
+	}
+
+	_, err := shell.RunCommandAndGetOutputE(t, cmd)
+	if err == nil {
+		return TerraformPlanNoChangeExitCode, nil
```
brikis98: Seems like this isn't the best variable to use here, since this is a generic `GetExitCodeForTerraformCommandE` method, and has nothing to do with the `plan` command. Perhaps add one more constant for use here called `DefaultSuccessExitCode`? It's also set to 0, but it'll communicate the intent a bit more clearly.

modules/terraform/cmd.go
```diff
@@ -59,3 +65,35 @@ func RunTerraformCommandE(t *testing.T, options *Options, args ...string) (strin
...
+	}
+	exitCode, getExitCodeErr := shell.GetExitCodeForRunCommandError(err)
+	if getExitCodeErr == nil {
+		return exitCode, nil
+	}
+	return TerraformPlanExceptionExitCode, getExitCodeErr
```
brikis98: Similarly here, we may want to use `DefaultErrorExitCode`, which is set to 1.

modules/terraform/terraform.go
```diff
@@ -1,2 +1,13 @@
...
+// https://www.terraform.io/docs/commands/plan.html#detailed-exitcode
+
+// TerraformPlanChangesPresentExitCode is the exit code returned by terraform plan detailed exitcode when changes are present
+const TerraformPlanChangesPresentExitCode = 2
+
+// TerraformPlanNoChangeExitCode is the exit code returned by terraform plan detailed exitcode when no changes are present
+const TerraformPlanNoChangeExitCode = 0
+
+// TerraformPlanExceptionExitCode is the exit code returned by terraform plan detailed exitcode when some exception occurs
+const TerraformPlanExceptionExitCode = -1
```
brikis98: I think exit codes have to be between 0 and 255 in *nix. So `1` is probably a better, more canonical value here.

modules/terraform/plan.go
```diff
@@ -0,0 +1,37 @@
...
+	exitCode, err := PlanExitCodeE(t, options)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return exitCode
+}
+
+// PlanExitCodeE runs terraform apply with the given options and returns the detailed exitcode.
+func PlanExitCodeE(t *testing.T, options *Options) (int, error) {
+	return GetExitCodeForTerraformCommandE(t, options, FormatArgs(options.Vars, "plan", "-input=false", "-lock=false", "-detailed-exitcode")...)
```
brikis98: It's still not obvious to me why we would want `-lock=false` here... If people are making changes concurrently to the same state, we want the lock to kick in!

modules/terraform/cmd.go
mikhailadvani: done

modules/terraform/cmd.go
mikhailadvani: done

modules/terraform/terraform.go
mikhailadvani: switched to `DefaultErrorExitCode` with value 1

modules/terraform/plan.go
mikhailadvani: If developer A is running a long running update and developer B has pushed commit (in a different module) to the repo which triggered a build on CI, B's commit might fail the build waiting for a lock(exit 1) whereas if there is deviation the plan should only detect change(exit 2). do not want to put in an arbitrary value for the timeout and hence want to go ahead with non locking. thoughts?

brikis98: Isn't this code just used in automated tests? What other devs are operating on this state concurrently?

mikhailadvani: my bad. got confused in my thoughts. fixed.

brikis98:
The tests failed. Looks like some of the imports were broken. Let's see if this fixes it: https://github.com/gruntwork-io/terratest/commit/3abebb895728c2ab5c74ef833cdafb3466f92a9b

brikis98:
OK, that fixed it. New release is here: https://github.com/gruntwork-io/terratest/releases/tag/v0.10.4

PR #156, opened 2018-09-21
Check AMI permissions

mikhailadvani:
Check permissions of AWS AMIs:
- `GetAmiPubliclyAccessible` returns `true` or `false` based on whether the AMI is publicly shared or not
- `GetAccountsWithLaunchPermissions` returns a list of accounts the AMI is shared with

Usage of the methods can be seen here: https://gist.github.com/mikhailadvani/11813b9a789a45965f04b5e11eb5b90c


modules/aws/ec2.go
```diff
@@ -221,6 +221,68 @@ func TerminateInstanceE(t *testing.T, region string, instanceID string) error {
...
+	for _, launchPermission := range launchPermissions {
+		if aws.StringValue(launchPermission.UserId) != "" {
+			accountIDs = append(accountIDs, aws.StringValue(launchPermission.UserId))
+		}
+	}
+	return accountIDs, nil
+}
+
+// GetLaunchPermissions returns launchPermissions as configured in AWS
+func GetLaunchPermissions(t *testing.T, awsRegion string, amiID string) ([]*ec2.LaunchPermission, error) {
```
brikis98: By convention, should this method be called `GetLaunchPermissionsE` since it returns an error?

mikhailadvani: fixed

mikhailadvani:
Adding the tests was the aim initially at least for the check for public images, but the problem is that you cannot query the image attributes of AMIs not owned by you. hence I took the approach of putting info in the gist. I tried looking up documentation which I could quote but could not find it.

You can test it out using any of the publicly available AMIs returned by https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/finding-an-ami.html#finding-quick-start-ami. I tried with the Amazon Linux(owner: amazon) and Ubuntu(owner: 099720109477) but didn't have any luck.

mikhailadvani:
I have renamed the `GetAccountsWithLaunchPermissions `. to `GetAccountsWithLaunchPermissionsForAmi`. Just for better naming consistency. Gist has alos been updated.

brikis98:
> Adding the tests was the aim initially at least for the check for public images, but the problem is that you cannot query the image attributes of AMIs not owned by you.

Ah, I see. In that case, you can add an example that builds a custom AMI with a Packer template and the helpers in the `packer` module, and then uses the new APIs in this PR to fetch info about that AMI. And yes, you should definitely mention this limitation in the comments for this method!

mikhailadvani:
@brikis98 tests added

modules/aws/ec2_test.go
```diff
@@ -0,0 +1,114 @@
...
+	ShareAmi(t, amiID, requestingAccount, ec2Client)
+	accountsWithLaunchPermissions := GetAccountsWithLaunchPermissionsForAmi(t, awsRegion, amiID)
+	assert.Contains(t, accountsWithLaunchPermissions, requestingAccount)
+}
+
+func BuildAmi(t *testing.T, packerVars map[string]string) string {
+
+	packerOptions := &packer.Options{
+		// The path to where the Packer template is located
+		Template: "../../examples/packer-basic-example/build.json",
```
brikis98: In the `test` folder, we already have example automated tests that build this AMI. It makes sense to just add your checks to those existing tests, as that'll also make it easier for users to use this code as an example.

modules/aws/ec2_test.go
```diff
@@ -0,0 +1,114 @@
...
+	awsRegion := "eu-west-1"
+	packerVars := map[string]string{
+		"aws_region": awsRegion,
+	}
+
+	amiID := BuildAmi(t, packerVars)
+	defer DeleteAmi(t, awsRegion, amiID)
+
+	ec2Client := NewEc2Client(t, awsRegion)
+	ShareAmi(t, amiID, "099720109477", ec2Client)
```
brikis98: What is `099720109477`?

modules/aws/ec2_test.go
mikhailadvani: have implemented the changes

modules/aws/ec2_test.go
mikhailadvani: have switched to using the aws.CanonicalAccountId constant. it was the same thing

PR #157, opened 2018-09-21
Update GCP Module to Support Regional Instance Groups, Cleaner Usage Model

josh-padnick:
This PR was originally motivated by my need to finish setting up automated tests for the https://github.com/hashicorp/terraform-google-consul repo. In that repo, I added support for Regional Instance Groups to Consul. To add support for that in Terratst led to a cascade of changes, plus a number of insights on creating a cleaner API for Terratest. Included in this PR:

- **Update the Terratest GCP API to "fetch" new resources.** GCP's Golang SDK often returns just an ID for a resource, requiring a follow-up query task. I switched to instantiating new GCP resources by fetching their data from GCP to create a cleaner usage model that more closely aligns with the GCP SDK.

- **Add More Helper Functions.** For example, `GetRandomInstance()` to return a randomly selected, fully populated Instance object from a Zonal or Regional Instance Group.

- **Add Better Test Coverage.** There are now readily available helper functions for launching and destroying a Compute Instance with Go instead of just with Terraform. This made writing tests easier. I also added more test coverage to some functions.


modules/gcp/compute.go
```diff
@@ -7,146 +7,259 @@ import (
...
+}
+
+// Corresponds to a GCP Regional Instance Group (https://cloud.google.com/compute/docs/instance-groups/)
+type RegionalInstanceGroup struct {
+	projectID string
+	*compute.InstanceGroup
+}
+
+// FetchInstance queries GCP to return an instance of the (GCP Compute) Instance type
+func FetchInstance(t *testing.T, projectID string, name string) *Instance {
```
brikis98: Please see the error handling docs: https://github.com/gruntwork-io/terratest#error-handling. Each test method should have a `fooE` flavor that returns all errors and a `foo` version that calls `t.Fatal` if any error is returned. This applies to all new methods in this PR.

modules/gcp/compute.go
```diff
@@ -7,146 +7,259 @@ import (
...
+	if err != nil {
+		t.Fatal(err)
 	}
 
-	instance, err := service.Instances.Get(projectID, zone, instanceID).Context(ctx).Do()
+	return &ZonalInstanceGroup{projectID, instanceGroup}
+}
+
+// GetPublicIP gets the public IP address of the given Compute Instance.
+func (c *Instance) GetPublicIp(t *testing.T) string {
```
brikis98: Please don't use short, non-descriptive names like `c`. 2 lines of code later and I have no idea what `c` is.

modules/gcp/compute.go
```diff
@@ -7,146 +7,259 @@ import (
...
-	instance, err := service.Instances.Get(projectID, zone, instanceID).Context(ctx).Do()
-	if err != nil {
-		return nil, fmt.Errorf("Instances.Get(%s) got error: %v", instanceID, err)
+	req := compute.InstancesSetLabelsRequest{Labels: labels, LabelFingerprint: c.LabelFingerprint}
+	if _, err := service.Instances.SetLabels(c.projectID, c.GetZone(t), c.Name, &req).Context(ctx).Do(); err != nil {
+		return fmt.Errorf("Instances.SetLabels(%s) got error: %v", c.Name, err)
 	}
 
-	return instance.Labels, nil
+	return err
```
brikis98: `return nil`?

modules/gcp/compute.go
```diff
@@ -7,146 +7,259 @@ import (
...
+	if err := req.Pages(ctx, func(page *compute.InstanceGroupsListInstances) error {
+		for _, instance := range page.Items {
+			// For some reason service.InstanceGroups.ListInstances returns us a collection
+			// with Instance URLs and we need only the Instance ID for the next call. Use
+			// the path functions to chop the Instance ID off the end of the URL.
+			instanceID := path.Base(instance.Instance)
+			instanceIDs = append(instanceIDs, instanceID)
+		}
+		return nil
+	}); err != nil {
```
brikis98: The `if err := xxx; err != nil` idiom works well for short amounts of code, but not for code with a multi-line function defined in `xxx`. Please break into multiple lines.

modules/gcp/compute.go
```diff
@@ -171,13 +286,66 @@ func GetInstanceIdsForInstanceGroupE(t *testing.T, projectID string, zone string
...
+	}
+
+	return instance
+}
+
+// GetRandomInstanceE returns a randomly selected Instance from the Regional Instance Group
+func (ig *RegionalInstanceGroup) GetRandomInstanceE(t *testing.T) (*Instance, error) {
+	instanceIDs := ig.GetInstanceIds(t)
+	clusterSize := int(ig.Size)
+	randIndex := random.Random(1, clusterSize)
```
brikis98: Should this be `(0, clusterSize)`?

modules/gcp/compute.go
```diff
@@ -171,13 +286,66 @@ func GetInstanceIdsForInstanceGroupE(t *testing.T, projectID string, zone string
...
+
+	if randIndex > len(instanceIDs) {
```
brikis98: When would this happen?

modules/gcp/compute.go
```diff
@@ -171,13 +286,66 @@ func GetInstanceIdsForInstanceGroupE(t *testing.T, projectID string, zone string
...
+		return nil, fmt.Errorf("Could not find any instances in Regional Instance Group %s in Region %s", ig.Name, ig.Region)
+	}
+
+	instanceID := instanceIDs[randIndex-1]
+	instance := FetchInstance(t, ig.projectID, instanceID)
+
+	return instance, nil
+}
+
+// TODO: Is there some way to avoid the total duplication of this function and the Regional Instance Group equivalent in Go?
```
brikis98: Define a common interface for `ZonalInstanceGroup` and `RegionalInstanceGroup` and create a standalone method that takes in that interface as a parameter and returns a random `Instance`.

modules/gcp/region.go
```diff
@@ -186,3 +213,17 @@ func GetAllGcpZonesE(t *testing.T) ([]string, error) {
 
 	return zones, nil
 }
+
+// Given a GCP Zone URL formatted like https://www.googleapis.com/compute/v1/projects/project-123456/zones/asia-east1-b,
+// return "asia-east1-b".
+func ZoneUrlToZone(zoneUrl string) string {
+	tokens := strings.Split(zoneUrl, "/")
+	return tokens[len(tokens)-1]
```
brikis98: To avoid really confusing errors—e.g., I could call this method with "foo" or "foo/bar" and still get back data—should this be a regex to do a bit more sanity checking?

modules/gcp/region.go
```diff
@@ -186,3 +213,17 @@ func GetAllGcpZonesE(t *testing.T) ([]string, error) {
...
+}
+
+// Given a GCP Zone URL formatted like https://www.googleapis.com/compute/v1/projects/project-123456/regions/southamerica-east1,
+// return "southamerica-east1".
+func RegionUrlToRegion(zoneUrl string) string {
+	tokens := strings.Split(zoneUrl, "/")
+	return tokens[len(tokens)-1]
```
brikis98: Same here.

robmorgan:
@josh-padnick nice enhancements and thanks for improving the test coverage 👍

modules/gcp/compute.go
josh-padnick: Ah, thanks for clarifying. I thought there was more discretion around this but it makes sense that these are recommended for all methods. Updated in 78e3ad59fd8561c0fafcf0ad1aa4e15682b7799e.

modules/gcp/compute.go
josh-padnick: Hm, in the all examples I read online (like [GoByExample](https://tour.golang.org/methods/3)), the idiomatic way to name the receiver variable is `i *Instance`. I even assume when I see a single-letter var when reading Go methods that it's the receiver var. 

That being said, I'll update the receiver vars to use the first letter of the object name (i.e. `i` instead of `c` for `*Instance`).

modules/gcp/compute.go
josh-padnick: Good catch. I had to fix this in other func as well.

josh-padnick:
@robmorgan I'm heading to bed now, but please use this branch with your current work. I'll plan to merge this on my Monday.

robmorgan:
@josh-padnick ok sure

modules/gcp/compute.go
josh-padnick: Thanks, updated in 485ca9ad89b866e972ef841a2d915e4f9d2b9ce3.

modules/gcp/region.go
josh-padnick: Good call. I'm unfortunately out of time, so I've left a ToDo for now.

josh-padnick:
Assuming tests pass, this is now ready to merge and release.

Issue #163, opened 2018-10-03
Is there a way to work with Terraform workspaces via Terratest?


sw-samuraj:
**Question:**
In our _Terraform_ scripts, we are using a concept of [workspaces](//www.terraform.io/docs/state/workspaces.html). For example, it's part of naming convention of resources:

```hcl
resource "oci_core_instance" "CPServer" {
  display_name = "cpServer${count.index}-${terraform.workspace}"
  # other configuration
}
```

Is there a way to work with _workspaces_ in the _Terratest_ test? E.g. switch to the existing, or create the new one.

Is it even a valid use case?

sw-samuraj:
I think that `RunTerraformCommand` should be fine for now. :+1: 

Would be OK if I write some proposal/PR for those functions? Something like:

* `WorkspaceNew(t *testing.T, name string) string`,
* `WorkspaceSelect(t *testing.T, name string) string`,

or maybe just:

* `WorkspaceSelectOrNew(t *testing.T, name string) string`.

I guess, they could be modeled like [Destroy](//github.com/gruntwork-io/terratest/blob/be2c7d366654f28812e037c3e8455a408ad2483c/modules/terraform/destroy.go#L17).

brikis98:
Sure, that would be great 👍

sw-samuraj:
New function `WorkspaceSelectOrNew` satisfies the initial question.

PR #164, opened 2018-10-03
Implement a function for workspace selection, or creation

sw-samuraj:
Related to #163 (Terraform workspace support)

A new function `WorkspaceSelectOrNew` has been implemented and covered via three unit tests. The function takes a workspace name as an argument and:

* either select the given workspace, if it already exists,
* or creates a new workspace with the given name.

### DIFF
--- a/modules/terraform/workspace.go
+++ b/modules/terraform/workspace.go
@@ -1,12 +1,15 @@
 package terraform
 
 import (
+	"fmt"
+	"regexp"
 	"strings"
 	"testing"
 )
 
-// WorkspaceSelectOrNew runs terraform workspace with the given options and returns workspace name.
-// It tries to select a workspace with the given name, or it creates a new one if it doesn't exist.
+// WorkspaceSelectOrNew runs terraform workspace with the given options and the workspace name
+// and returns a name of the current workspace. It tries to select a workspace with the given
+// name, or it creates a new one if it doesn't exist.
 func WorkspaceSelectOrNew(t *testing.T, options *Options, name string) string {
 	out, err := WorkspaceSelectOrNewE(t, options, name)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


modules/terraform/workspace.go
```diff
@@ -0,0 +1,36 @@
...
+
+// WorkspaceSelectOrNewE runs terraform workspace with the given options and returns workspace name.
+// It tries to select a workspace with the given name, or it creates a new one if it doesn't exist.
+func WorkspaceSelectOrNewE(t *testing.T, options *Options, name string) (string, error) {
+	out, err := RunTerraformCommandE(t, options, "workspace", "list")
+	if err != nil {
+		return out, nil
+	}
+
+	if strings.Contains(out, name) {
```
brikis98: This seems error prone. If you have a workspace called `foobar` and you set `name` to `foo`, this will return true even though `foo` doesn't actually exist. I think you need to split the output into individual lines and loop over each line looking for an exact match to `name`. Note that the currently selected workspace may have an asterisk next to it, so it might require some extra cleaning.

modules/terraform/workspace.go
```diff
@@ -0,0 +1,36 @@
+package terraform
+
+import (
+	"strings"
+	"testing"
+)
+
+// WorkspaceSelectOrNew runs terraform workspace with the given options and returns workspace name.
+// It tries to select a workspace with the given name, or it creates a new one if it doesn't exist.
+func WorkspaceSelectOrNew(t *testing.T, options *Options, name string) string {
```
brikis98: What is the return value?

modules/terraform/workspace.go
```diff
@@ -0,0 +1,36 @@
...
+		return out, nil
+	}
+
+	if strings.Contains(out, name) {
+		_, err = RunTerraformCommandE(t, options, "workspace", "select", name)
+	} else {
+		_, err = RunTerraformCommandE(t, options, "workspace", "new", name)
+	}
+	if err != nil {
+		return out, nil
```
brikis98: `return "", nil`

modules/terraform/workspace.go
sw-samuraj: The return value is the name of the current workspace.

modules/terraform/workspace.go
sw-samuraj: That's correct, I'll modify it.

modules/terraform/workspace.go
brikis98: Add that to the docs!

modules/terraform/workspace.go
sw-samuraj: :+1:

sw-samuraj:
@brikis98 Improved.

modules/terraform/workspace.go
```diff
@@ -0,0 +1,55 @@
...
+	for _, ws := range workspaces {
+		if nameMatchesWorkspace(name, ws) {
+			return true
+		}
+	}
+	return false
+}
+
+func nameMatchesWorkspace(name string, workspace string) bool {
+	match, _ := regexp.MatchString(fmt.Sprintf("^\\*?\\s*%s$", name), workspace)
```
brikis98: Could you add a comment to explain what format you expect `workspace` to have? It'll make it easier to understand the regex.

sw-samuraj: Regex documented.

PR #165, opened 2018-10-03
Add ScpDirFrom and ScpFileFrom

eak12913:
- Added ScpDirFrom - a helper method that will transfer many/all (can be controlled via ScpDownloadOptions) files from a remote dir to a local dir
    - ScpDownloadOptions allows user to specify a filename filter to grab only files matching the filter
    - ScpDownloadOptions allows user to specify a max size (MB) filter to only grab files smaller than XMB
- Added ScpFileFrom - a helper method that will transfer a remote file to your local machine
- Added new test cases to validate functionality

### DIFF
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -334,14 +334,6 @@ func listFileInRemoteDir(t *testing.T, sshSession *SshSession, options ScpDownlo
 
 	var result []string
 
-	if err := setUpSSHClient(sshSession); err != nil {
-		return result, err
-	}
-
-	if err := setUpSSHSession(sshSession); err != nil {
-		return result, err
-	}
-
 	findCommand := fmt.Sprintf("find %s -type f", options.RemoteDir)
 
 	if options.FileNameFilter != "" {
@@ -352,14 +344,7 @@ func listFileInRemoteDir(t *testing.T, sshSession *SshSession, options ScpDownlo
 		findCommand = fmt.Sprintf("%s -size -%dM", findCommand, options.MaxFileSizeMB)
 	}
 
-	r, err := sshSession.Session.Output(findCommand)
-
-	if err != nil {
-		return result, err
-	}
-	defer sshSession.Session.Close()
-
-	resultString := string(r)
+	resultString := CheckSshCommand(t, options.RemoteHost, findCommand)
 	// The last character returned is `\n` this results in an extra "" array
 	// member when we do the split below. Cut off the last character to avoid
 	// having to remove the blank entry in the array.
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -14,10 +14,11 @@ import (
 
 	"path/filepath"
 
+	"strings"
+
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
-	"strings"
 )
 
 // Host is a host on AWS.
@@ -32,11 +33,11 @@ type Host struct {
 }
 
 type ScpDownloadOptions struct {
-	FileNameFilter	string //bash style match string for files
-	MaxFileSizeMB		int //Don't grab any files > MaxFileSizeMB
-	RemoteDir 		string //Copy this directory on the remote machine
-	LocalDir   		string //Copy RemoteDir to this directory on the local machine
-	RemoteHost 		Host //Connection information for the remote machine
+	FileNameFilter string //bash style match string for files
+	MaxFileSizeMB  int    //Don't grab any files > MaxFileSizeMB
+	RemoteDir      string //Copy this directory on the remote machine
+	LocalDir       string //Copy RemoteDir to this directory on the local machine
+	RemoteHost     Host   //Connection information for the remote machine
 }
 
 // ScpFileToE uploads the contents using SCP to the given host and fails the test if the connection fails.
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -78,6 +78,15 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents 
 	return err
 }
 
+// ScpFileFromE downloads the file from remotePath on the given host using SCP.
+func ScpFileFrom(t *testing.T, host Host, remotePath string, localDestination *os.File) {
+	err := ScpFileFromE(t, host, remotePath, localDestination)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
 // ScpFileFromE downloads the file from remotePath on the given host using SCP and returns an error if the process fails.
 func ScpFileFromE(t *testing.T, host Host, remotePath string, localDestination *os.File) error {
 	authMethods, err := createAuthMethodsForHost(host)
@@ -103,6 +112,15 @@ func ScpFileFromE(t *testing.T, host Host, remotePath string, localDestination *
 
 	err = copyFileFromRemote(t, sshSession, localDestination, remotePath)
 	return err
+}
+
+// ScpFileFromE downloads all the files from remotePath on the given host using SCP.
+func ScpDirFrom(t *testing.T, options ScpDownloadOptions) {
+	err := ScpDirFromE(t, options)
+
+	if err != nil {
+		t.Fatal(err)
+	}
 }
 
 // ScpFileFromE downloads all the files from remotePath on the given host using SCP and returns an error if the process fails.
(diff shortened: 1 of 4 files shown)


modules/ssh/ssh.go
```diff
@@ -30,6 +33,14 @@ type Host struct {
 	OverrideSshAgent *SshAgent // enable an in process `SshAgent` for connections to this host (disabled by default)
 }
 
+type ScpDownloadOptions struct {
+	FileNameFilter string //bash style match string for files
+	MaxFileSizeMB  int    //Don't grab any files > MaxFileSizeMB
```
brikis98: `MaxNumFiles` too?

modules/ssh/ssh.go
```diff
@@ -68,6 +79,104 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents
...
+	}
+}
+
+// ScpFileFromE downloads the file from remotePath on the given host using SCP and returns an error if the process fails.
+func ScpFileFromE(t *testing.T, host Host, remotePath string, localDestination *os.File) error {
+	authMethods, err := createAuthMethodsForHost(host)
+	if err != nil {
+		return err
+	}
+	dir, _ := filepath.Split(remotePath)
```
brikis98: https://golang.org/pkg/path/filepath/#Dir perhaps?

modules/ssh/ssh.go
```diff
@@ -68,6 +79,104 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents
...
+	}
+
+	sshSession := &SshSession{
+		Options:  &hostOptions,
+		JumpHost: &JumpHostSession{},
+	}
+
+	defer sshSession.Cleanup(t)
+
+	err = copyFileFromRemote(t, sshSession, localDestination, remotePath)
```
brikis98: You can just `return` here

modules/ssh/ssh.go
```diff
@@ -68,6 +79,104 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents
...
+
+	defer sshSession.Cleanup(t)
+
+	filesInDir, err := listFileInRemoteDir(t, sshSession, options)
+
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(options.LocalDir); os.IsNotExist(err) {
```
brikis98: https://github.com/gruntwork-io/terratest/blob/master/modules/files/files.go#L12

modules/ssh/ssh.go
```diff
@@ -68,6 +79,104 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents
...
+		os.Mkdir(options.LocalDir, 0755)
+	}
+
+	errorsOccurred := []error{}
+
+	for _, fullRemoteFilePath := range filesInDir {
+		logger.Logf(t, "Copying remote file file: %s", fullRemoteFilePath)
```
brikis98: copying remote file x to local path y

modules/ssh/ssh.go
```diff
@@ -68,6 +79,104 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents
...
+		_, fileName := filepath.Split(fullRemoteFilePath)
```
brikis98: https://golang.org/pkg/path/filepath/#Base perhaps?

modules/ssh/ssh.go
```diff
@@ -68,6 +79,104 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents
...
+
+		localFilePath := fmt.Sprintf("%s/%s", options.LocalDir, fileName)
```
brikis98: `filepath.Join`

modules/ssh/ssh.go
```diff
@@ -217,6 +326,65 @@ func FetchContentsOfFileE(t *testing.T, host Host, useSudo bool, filePath string
...
+func listFileInRemoteDir(t *testing.T, sshSession *SshSession, options ScpDownloadOptions) ([]string, error) {
+	logger.Logf(t, "Running command %s on %s@%s", sshSession.Options.Command, sshSession.Options.Username, sshSession.Options.Address)
+
+	var result []string
+
+	if err := setUpSSHClient(sshSession); err != nil {
+		return result, err
+	}
+
+	if err := setUpSSHSession(sshSession); err != nil {
```
brikis98: Is there a reason you're running all these steps manually instead of [CheckSshCommand](https://github.com/gruntwork-io/terratest/blob/master/modules/ssh/ssh.go#L95)?

modules/ssh/ssh.go
```diff
@@ -217,6 +326,65 @@ func FetchContentsOfFileE(t *testing.T, host Host, useSudo bool, filePath string
...
+
+	r, err := sshSession.Session.Output(findCommand)
+
+	if err != nil {
+		return result, err
+	}
+	defer sshSession.Session.Close()
+
+	resultString := string(r)
+	resultString = resultString[:len(resultString)-1]
```
brikis98: Why?

modules/ssh/ssh.go
```diff
@@ -217,6 +326,65 @@ func FetchContentsOfFileE(t *testing.T, host Host, useSudo bool, filePath string
...
+
+	result = append(result, strings.Split(resultString, "\n")...)
+	return result, nil
```
brikis98: Should the return type be a map from filename to string?

test/terraform_ssh_example_test.go
```diff
@@ -91,6 +137,164 @@ func configureTerraformOptions(t *testing.T, exampleFolder string) (*terraform.O
...
+		SshUserName: "ubuntu",
+	}
+
+	// It can take a minute or so for the Instance to boot up, so retry a few times
+	maxRetries := 30
+	timeBetweenRetries := 5 * time.Second
+	description := fmt.Sprintf("SSH to public host %s", publicInstanceIP)
+	remoteTempFolder := "/tmp/testFolder"
+	remoteTempFilePath := fmt.Sprintf("%s/test.foo", remoteTempFolder)
+	remoteTempFilePath2 := fmt.Sprintf("%s/test.baz", remoteTempFolder)
```
brikis98: Always use `filepath.Join`. Building paths through string concatenation almost always leads to a bug.

modules/ssh/ssh.go
eak12913: Done.

modules/ssh/ssh.go
eak12913: Done

modules/ssh/ssh.go
eak12913: done

modules/ssh/ssh.go
brikis98: My main thought is to avoid downloading everything on the entire computer if you accidentally specify `/` as a path. If `MaxNumFiles` is exceeded, you stop downloading further files, and add that as a message in the MultiError. 

BTW, does this code only download files or does it recursively download folders too? Either way, it should be documented.

modules/ssh/ssh.go
eak12913: Done.

modules/ssh/ssh.go
eak12913: done

modules/ssh/ssh.go
eak12913: done

modules/ssh/ssh.go
eak12913: Unfortunately after doing the `find` command on the remote server, the resulting string of files always has a `\n` last character - so I'm cutting that off.

Added comment to code to explain:
```	
        // The last character returned is `\n` this results in an extra "" array
	// member when we do the split below. Cut off the last character to avoid
	// having to remove the blank entry in the array.
```

modules/ssh/ssh.go
eak12913: What will the `string` be? Right now, I'm just returning the full path of the matching files on the remote machine.

test/terraform_ssh_example_test.go
eak12913: fixed

modules/ssh/ssh.go
brikis98: Oh, my bad, I misread this as though these were the contents of the files rather than the paths.

modules/ssh/ssh.go
eak12913: Nope, no real reason when getting the list of remote files. Just did the same thing I was doing when getting binary because I was on auto-pilot :-)

Switched to using CheckSshCommand.

modules/ssh/ssh.go
eak12913: This code will only download files within the folder you specified. It won't recursively download subfolders or follow symlinks.

So If you accidentally specify `/` it won't just download your whole computer :-)

Updated comments on the method to say the following:
```
// ScpFileFromE downloads all the files from remotePath on the given host using SCP 
// and returns an error if the process fails. NOTE: only files within remotePath will
// be downloaded. This function will not recursively download subdirectories or follow
// symlinks.
```

Issue #167, opened 2018-10-08
Practice recommendation for dealing with an output list


sw-samuraj:
**Question:**
The _Terraform_ output can be defined as a list, e.g.

```tf
output "CPPublicIPs" {
  value = ["${oci_core_instance.CPServer.*.public_ip}"]
}
```

will provide a following output:

```
CPPublicIPs = [
    129.146.109.157,
    129.146.111.161,
    129.146.101.69
]
```

However, _Terratest_ function `Output` provides just a `string`. Is there a recommendation how to deal with _output list_? Does it have to be parsed manually? E.g.

```go
out := terraform.Output(t, terraformOptions, "CPPublicIPs")
ips := strings.Split(out, "\n")
```

Wouldn't be convenient to have a function like `OutputList(t *testing.T, options *Options, key string) []string`?

brikis98:
> Wouldn't be convenient to have a function like OutputList(t *testing.T, options *Options, key string) []string?

Yes. We actually have functions like this in some of our individual tests, but have not yet extracted them into Terratest itself. A PR for that would be very welcome! Using the JSON output and a JSON parser is probably the way to go.

yorinasub17:
`OutputList` and `OutputMap` were introduced since [v0.13.6](https://github.com/gruntwork-io/terratest/releases/tag/v0.13.6)

PR #169, opened 2018-10-09
Implement a function for output list

sw-samuraj:
Related to #167 (output list)

A new function `OutputList` has been implemented and covered via unit tests. The function takes an output key as an argument and:

* either returns a multiple-items `[]string` if the value is a list,
* or returns a single item `[]string` if the value is not a list (i.e. a single value).

The output values are parsed from _Terraform_ JSON output obtained via `-json` switch.

### DIFF
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -78,14 +78,13 @@ func OutputListE(t *testing.T, options *Options, key string) ([]string, error) {
 		return nil, fmt.Errorf("Output doesn't contain a value for the key %q", key)
 	}
 
-	var list []string
+	list := []string{}
 	switch t := value.(type) {
 	case []interface{}:
-		list = make([]string, len(t))
 		for i, item := range t {
 			list[i] = fmt.Sprintf("%v", item)
 		}
-	case interface{}:
+	default:
 		return nil, fmt.Errorf("Output value %q is not a list", value)
 	}
 
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -1,6 +1,7 @@
 package terraform
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
@@ -47,6 +48,45 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
 	}
 
 	return out, nil
+}
+
+// OutputList calls terraform output for the given variable and return its value as a list.
+// If the output is a single value, then it returns a list with just one item.
+func OutputList(t *testing.T, options *Options, key string) []string {
+	out, err := OutputListE(t, options, key)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return out
+}
+
+// OutputListE calls terraform output for the given variable and return its value as a list.
+// If the output is a single value, then it returns a list with just one item.
+func OutputListE(t *testing.T, options *Options, key string) ([]string, error) {
+	out, err := RunTerraformCommandE(t, options, "output", "-no-color", "-json", key)
+	if err != nil {
+		return nil, err
+	}
+
+	outputMap := map[string]interface{}{}
+	if err := json.Unmarshal([]byte(out), &outputMap); err != nil {
+		t.Fatalf("Failed to parse JSON for value %s: %v", out, err)
+	}
+
+	value := outputMap["value"]
+	var list []string
+	switch t := value.(type) {
+	case []interface{}:
+		list = make([]string, len(t))
+		for i, item := range t {
+			list[i] = item.(string)
+		}
+	case interface{}:
+		list = make([]string, 1)
+		list[0] = t.(string)
+	}
+
+	return list, nil
 }
 
 // EmptyOutput is an error that occurs when an output is empty.
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -50,8 +50,8 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
 	return out, nil
 }
 
-// OutputList calls terraform output for the given variable and return its value as a list.
-// If the output is a single value, then it returns a list with just one item.
+// OutputList calls terraform output for the given variable and returns its value as a list.
+// If the output value is not a list type, then it fails the test.
 func OutputList(t *testing.T, options *Options, key string) []string {
 	out, err := OutputListE(t, options, key)
 	if err != nil {
@@ -60,8 +60,8 @@ func OutputList(t *testing.T, options *Options, key string) []string {
 	return out
 }
 
-// OutputListE calls terraform output for the given variable and return its value as a list.
-// If the output is a single value, then it returns a list with just one item.
+// OutputListE calls terraform output for the given variable and returns its value as a list.
+// If the output value is not a list type, then it returns an error.
 func OutputListE(t *testing.T, options *Options, key string) ([]string, error) {
 	out, err := RunTerraformCommandE(t, options, "output", "-no-color", "-json", key)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


modules/terraform/output.go
```diff
@@ -49,6 +50,45 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
...
+func OutputList(t *testing.T, options *Options, key string) []string {
+	out, err := OutputListE(t, options, key)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return out
+}
+
+// OutputListE calls terraform output for the given variable and return its value as a list.
+// If the output is a single value, then it returns a list with just one item.
```
brikis98: If it's a non-list value, shouldn't this return an error instead? It seems like that might indicate a bug with types or a typo in the variable name...

modules/terraform/output.go
```diff
@@ -49,6 +50,45 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
...
+	if err != nil {
+		return nil, err
+	}
+
+	outputMap := map[string]interface{}{}
+	if err := json.Unmarshal([]byte(out), &outputMap); err != nil {
+		t.Fatalf("Failed to parse JSON for value %s: %v", out, err)
+	}
+
+	value := outputMap["value"]
```
brikis98: Should you check if `value` is actually in the map? 

```go
if value, containsValue := outputMap["value"]; !containsValue {
  return nil, fmt.Errorf(...)
}
```

modules/terraform/output.go
```diff
@@ -49,6 +50,45 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
...
+// If the output is a single value, then it returns a list with just one item.
+func OutputListE(t *testing.T, options *Options, key string) ([]string, error) {
+	out, err := RunTerraformCommandE(t, options, "output", "-no-color", "-json", key)
+	if err != nil {
+		return nil, err
+	}
+
+	outputMap := map[string]interface{}{}
+	if err := json.Unmarshal([]byte(out), &outputMap); err != nil {
+		t.Fatalf("Failed to parse JSON for value %s: %v", out, err)
```
brikis98: Return an error, don't call `t.Fatal`

modules/terraform/output.go
```diff
@@ -49,6 +50,45 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
...
+	}
+
+	value := outputMap["value"]
+	var list []string
+	switch t := value.(type) {
+	case []interface{}:
+		list = make([]string, len(t))
+		for i, item := range t {
+			list[i] = item.(string)
```
brikis98: Might be safer to do `fmt.Sprintf("%v", item)`.

modules/terraform/output.go
```diff
@@ -49,6 +50,45 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
...
+	out, err := OutputListE(t, options, key)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return out
+}
+
+// OutputListE calls terraform output for the given variable and return its value as a list.
+// If the output is a single value, then it returns a list with just one item.
+func OutputListE(t *testing.T, options *Options, key string) ([]string, error) {
```
brikis98: Do we need to handle lists of ints? Or bools? Or nested structures, such as lists of maps or lists of lists?

modules/terraform/output_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	InitAndApply(t, options)
+	out := OutputList(t, options, "giant_steps")
+
+	expectedLen := 4
+	expectedItem := "John Coltrane"
+
+	assert.Len(t, out, expectedLen, "Output should contain %d items", expectedLen)
+	assert.Contains(t, out, expectedItem, "Output should contain %q item", expectedItem)
+	assert.Equal(t, out[0], expectedItem, "First item should be %q, got %q", expectedItem, out[0])
```
brikis98: Shouldn't you compare `out` to `[]string{"John Coltrane", "Tommy Flanagan", "Paul Chambers", "Art Taylor"}`?

modules/terraform/output.go
sw-samuraj: I was thinking about throwing an error too. It's a design decision and I don't have strong opinion here. I'll change it to throwing the error then.

modules/terraform/output.go
sw-samuraj: I'm not sure - formerly, I was just counting with strings. I can investigate _Terraform_ source code to catch some inspiration.

sw-samuraj: According the [documentation](//www.terraform.io/docs/configuration/outputs.html#value), output type can be a `string`, a `list`, or a `map`. However, it doesn't state a precise value type for the later two. Checking the _Terraform_ [source code](//github.com/hashicorp/terraform/blob/5e511e1ea1ccbc6cfc1646d3dd3e88e8d81f3864/terraform/state.go#L917), the type is defined as:

* `string`
* `[]interface{}`
* `map[string]interface{}`
* `[]map[string]interface{}`

So theoretically, the list can be of other types than just `string`.

Maybe, let's keep it as `[]string` for the start and make it more generic on (later) demand? It can be quite complex to have it generic for complex structures.

For a `map`, I would say that a new function `OutputMap` would be good match.

brikis98: > Maybe, let's keep it as []string for the start and make it more generic on (later) demand? It can be quite complex to have it generic for complex structures.

Sounds good!

modules/terraform/output.go
```diff
@@ -49,6 +50,48 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
...
+	}
+
+	var list []string
+	switch t := value.(type) {
+	case []interface{}:
+		list = make([]string, len(t))
+		for i, item := range t {
+			list[i] = fmt.Sprintf("%v", item)
+		}
+	case interface{}:
```
brikis98: Change to `default` here.

modules/terraform/output.go
```diff
@@ -49,6 +50,48 @@ func OutputRequiredE(t *testing.T, options *Options, key string) (string, error)
...
+
+	value, containsValue := outputMap["value"]
+	if !containsValue {
+		return nil, fmt.Errorf("Output doesn't contain a value for the key %q", key)
+	}
+
+	var list []string
+	switch t := value.(type) {
+	case []interface{}:
+		list = make([]string, len(t))
```
brikis98: Since these lists are short, you may as well declare `list` above via `list := []string{}`.

sw-samuraj:
Done.

brikis98:
Hm, CI build failed:

```
 --- FAIL: TestOutputList (0.79s)
 panic: runtime error: index out of range [recovered]
   panic: runtime error: index out of range
 
 goroutine 18 [running]:
 testing.tRunner.func1(0xc000184e00)
   /usr/local/go/src/testing/testing.go:792 +0x387
 panic(0x7f7d40, 0xb99280)
   /usr/local/go/src/runtime/panic.go:513 +0x1b9
 github.com/gruntwork-io/terratest/modules/terraform.OutputListE(0xc000184e00, 0xc000301f48, 0x86ce9f, 0xb, 0xc000184e00, 0xc000301f48, 0xc0001a42d0, 0xe5, 0x0)
   /home/circleci/.go_workspace/src/github.com/gruntwork-io/terratest/modules/terraform/output.go:85 +0x52f
 github.com/gruntwork-io/terratest/modules/terraform.OutputList(0xc000184e00, 0xc000301f48, 0x86ce9f, 0xb, 0xc0002f60c0, 0x32, 0x0)
   /home/circleci/.go_workspace/src/github.com/gruntwork-io/terratest/modules/terraform/output.go:56 +0x59
 github.com/gruntwork-io/terratest/modules/terraform.TestOutputList(0xc000184e00)
   /home/circleci/.go_workspace/src/github.com/gruntwork-io/terratest/modules/terraform/output_test.go:23 +0x19c
 testing.tRunner(0xc000184e00, 0x886d58)
   /usr/local/go/src/testing/testing.go:827 +0xbf
 created by testing.(*T).Run
   /usr/local/go/src/testing/testing.go:878 +0x353
 FAIL  github.com/gruntwork-io/terratest/modules/terraform 0.805s
```

PR #172, opened 2018-10-15
Add ability to SSH into GCP Compute Instances

josh-padnick:
As part of adding automated tests for Vault GCP, we needed the ability to SSH into a GCP Compute Instance and run commands. This PR adds that functionality. As part of the implementation, I added the ability to get and set Compute Instance metadata.

Assuming tests pass, this is ready to merge.

### DIFF
--- a/modules/gcp/compute.go
+++ b/modules/gcp/compute.go
@@ -514,3 +514,11 @@ func NewInstancesServiceE(t *testing.T) (*compute.InstancesService, error) {
 
 	return service.Instances, nil
 }
+
+// Return a random, valid name for GCP Compute Instances. Note that GCP requires Instance names to use lowercase letters only.
+func UniqueGcpInstanceName() string {
+	id := strings.ToLower(random.UniqueId())
+	instanceName := fmt.Sprintf("terratest-%s", id)
+
+	return instanceName
+}
--- a/modules/gcp/compute.go
+++ b/modules/gcp/compute.go
@@ -225,10 +225,12 @@ func (i *Instance) SetLabelsE(t *testing.T, labels map[string]string) error {
 	return nil
 }
 
+// GetMetadata gets the given Compute Instance's metadata
 func (i *Instance) GetMetadata(t *testing.T) []*compute.MetadataItems {
 	return i.Metadata.Items
 }
 
+// SetMetadata sets the given Compute Instance's metadata
 func (i *Instance) SetMetadata(t *testing.T, metadata map[string]string) {
 	err := i.SetMetadataE(t, metadata)
 	if err != nil {
@@ -236,7 +238,7 @@ func (i *Instance) SetMetadata(t *testing.T, metadata map[string]string) {
 	}
 }
 
-// SetLabelsE adds the tags to the given Compute Instance.
+// SetLabelsE adds the given metadata map to the existing metadata of the given Compute Instance.
 func (i *Instance) SetMetadataE(t *testing.T, metadata map[string]string) error {
 	logger.Logf(t, "Adding metadata to instance %s in zone %s", i.Name, i.Zone)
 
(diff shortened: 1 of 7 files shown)


modules/gcp/compute.go
```diff
@@ -224,6 +225,87 @@ func (i *Instance) SetLabelsE(t *testing.T, labels map[string]string) error {
...
+		t.Fatal(err)
+	}
+}
+
+// Add the given public SSH key to the Compute Instance. Users can SSH in with the given username.
+func (i *Instance) AddSshKeyE(t *testing.T, username string, publicKey string) error {
+	logger.Logf(t, "Adding SSH Key to Compute Instance %s for username %s\n", i.Name, username)
+
+	publicKeyFormatted := strings.TrimSpace(publicKey)
+	sshKeyFormatted := fmt.Sprintf("%s:%s %s", username, publicKeyFormatted, username)
```
brikis98: Seems very weird to have to do this formatting manually... Is this some GCP thing?

modules/gcp/compute_test.go
```diff
@@ -99,6 +100,105 @@ func TestGetAndSetLabels(t *testing.T) {
...
+	sleepBetweenRetries := 3 * time.Second
+
+	retry.DoWithRetry(t, "Read newly set metadata", maxRetries, sleepBetweenRetries, func() (string, error) {
+		instance := FetchInstance(t, projectID, instanceName)
+		metadataFromRead := instance.GetMetadata(t)
+		for _, metadataItem := range metadataFromRead {
+			for key, val := range metadataToWrite {
+				if metadataItem.Key == key && *metadataItem.Value == val {
+					return "", nil
+				}
```
brikis98: May want to log if the data isn't equal to make troubleshooting easier

modules/gcp/compute_test.go
```diff
@@ -99,6 +100,105 @@ func TestGetAndSetLabels(t *testing.T) {
...
+			}
+		}
+
+		return "", fmt.Errorf("Metadata that was written was not found in metadata that was read. Retrying.\n")
+	})
+}
+
+// Create a Compute Instance, and attempt to SSH in and run a command.
+func TestSshAccessToComputeInstance(t *testing.T) {
```
brikis98: I'd recommend writing this code in the `test` folder instead. That is, create Terraform code in the `examples` folder to deploy a compute instance and then show how to SSH to that instance in the `test` folder. Remember, the automated tests for Terratest are both validation that it works as expected *and* documentation for users that show how to use Terratest itself.

modules/gcp/compute_test.go
```diff
@@ -99,6 +100,105 @@ func TestGetAndSetLabels(t *testing.T) {
...
+	var instance *Instance
+	var publicIp string
+
+	maxRetries := 10
+	sleepBetweenRetries := 3 * time.Second
+
+	retry.DoWithRetry(t, "Attempting to get public IP address", maxRetries, sleepBetweenRetries, func() (string, error) {
+		instance = FetchInstance(t, projectID, instanceName)
+
+		publicIp = instance.GetPublicIp(t)
```
brikis98: Could be read via `terraform output` if you use the `examples`/`test` idea mentioned above.

modules/gcp/compute.go
josh-padnick: It is indeed a GCP-specific format. I added a comment to clarify and linked to the [relevant docs](https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys).

modules/gcp/compute_test.go
josh-padnick: This was a good recommendation. Doing this forced me to update the GCP example to output a public IP, so now my effort resolving the confusion can be enjoyed by example users. Added in c895d67e244e7088197801a261bc6a174ca7e8a0.

josh-padnick:
All feedback addressed, and rebased on master. This is ready to merge once tests pass.

Issue #170, opened 2018-10-11
How to load SSH key-pair from the filesystem?


sw-samuraj:
**Question:**
I went through the examples and `ssh` package source code, but I didn't find a function for loading of _existing_ SSH keys/key-pair. All the scenarios are about:

1. generating of a key-pair,
1. importing of the key-pair to _AWS_,
1. temporary saving of the key-pair for later use,
1. **loading of the saved key-pair**,
1. some tests,
1. deleting of the key-pair.

I'm missing a function which just performs _key-pair loading_, but not from the transient test data, but from the "real" keys stored somewhere on the filesystem.

brikis98:
Hahah, your timing is impeccable. We had just discussed a little while back how nice it would be if Terratest, when running locally, or perhaps when some env var was set, would import your existing public key (i.e., from `~/.ssh/id_rsa.pub`) into AWS, rather than creating a totally new one. That way, if a test failed, you could SSH to the relevant resources with your regular SSH key, which is probably already in ssh-agent. A PR for that is most welcome 😄

sw-samuraj:
With that goes another question - as I understand from the source code, all the private keys are _unencrypted_, right? Would be nice to be able to work with encrypted keys as well.

brikis98:
Ah, that's a good point. Not sure the best way to handle that. Interactively asking for a password could be a weird user experience. Perhaps providing the password via env var? Or is there some way to get the encrypted key from ssh-agent?

sw-samuraj:
I'll try to sketch a first draft and then discuss/develop it further.

james03160927:
Closing this issue due to 6 years of inactivity. Feel free to reopen if it still needs to be addressed.

PR #173, opened 2018-10-15
Design of key-pair loading

sw-samuraj:
Related to #170 

This is a first draft of the _key-pair loading_ functionality. **It's intended for a design discussion**, not for merge (yet).

Possible extension could be a function
```go
func LoadDefaultRSAKeyPair(t *testing.T) *KeyPair
```

which loads a key-pair from some default location, e.g. `~/.ssh`. Maybe the default location could be driven via some environment variable.

For now, an _encrypted_ private key is omitted from the design, but after solving of this topic, the next function could be

```go
func LoadRSAEncryptedKeyPair(t *testing.T, publicKeyPath string, privateKeyPath string, passPhrase string) *KeyPair
```

where the `passPhrase` would be providede via environment variable, e.g. `TF_VAR_pass_phrase`.

### DIFF
--- a/modules/ssh/key_pair.go
+++ b/modules/ssh/key_pair.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"io/ioutil"
 	"testing"
 
 	"github.com/gruntwork-io/terratest/modules/logger"
@@ -54,4 +55,31 @@ func GenerateRSAKeyPairE(t *testing.T, keySize int) (*KeyPair, error) {
 
 	// Return
 	return &KeyPair{PublicKey: sshPubKeyStr, PrivateKey: keyPem}, nil
+}
+
+// TODO Vit: document this
+func LoadRSAKeyPair(t *testing.T, privateKeyPath string, publicKeyPath string) *KeyPair {
+	keyPair, err := LoadRSAKeyPairE(t, privateKeyPath, publicKeyPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return keyPair
+}
+
+// TODO Vit: document this
+func LoadRSAKeyPairE(t *testing.T, publicKeyPath string, privateKeyPath string) (*KeyPair, error) {
+	publicKey, err := ioutil.ReadFile(publicKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	privateKey, err := ioutil.ReadFile(privateKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	return &KeyPair{
+		PublicKey:  string(publicKey),
+		PrivateKey: string(privateKey),
+	}, nil
 }


brikis98:
Please see https://github.com/gruntwork-io/terratest/pull/174

eak12913:
My apologies @sw-samuraj! I didn't see your PR before making my own today. It seems that we're taking a very similar approach here but one argument that I can make for my take on the problem is that I didn't want the user to have to write any extra go code in order to tell `terratest` to use their keys.

My ideal scenario would be to automatically detect that the test is not executing on a CI system and then default to using the keys of the user instead of randomly generated keys. Unfortunately, I'm not sure there's a CI agnostic way of detecting the presence of a CI system. I've seen some CI systems use the `CI=true` env variable, but I'm not sure that counting on that is good enough.

eak12913:
Few other thoughts: 

- Since neither one of us addressed the issue of dealing with encrypted private keys, we could also think about having the presence of some environment variable instruct `terratest` to take our randomly generated private key and then `ssh-add` it to ssh agent. Drawback: I'm not sure there's a great way to clean this up since that key may persist over many test runs where we want it to remain the same.
- My other idea was to just export the private key when the user calls `test_structure.SaveEc2KeyPair` and then have `terratest` log out the full `ssh -i [path_to_exported_private_key] -o 'IdentitiesOnly yes' [placeholder_ssh_name]@[placeholder_ip]`. This way, the user wouldn't have the convenience of using their own private keys, but would just have to copy the command and paste in the `user_name` and `ip` and then ssh would just use the exported private key.

sw-samuraj:
@eak12913 Your PR looks good - it basically mimics my suggestion of `LoadDefaultRSAKeyPair` function. 👍

My only point is to split:

* key-pair loading - it's generic,
* cloud import - it's cloud specific

I'm aiming to different cloud than AWS - that's why I placed my function in the `ssh` package. For example in [OCI](//cloud.oracle.com/cloud-infrastructure), the key import is not so straightforward and I will probably need to write a special function for it.

CpuID:
I'd love to see an integrated way to use encrypted private keys merged... aware there's 2 directions (with some overlap) in here and https://github.com/gruntwork-io/terratest/pull/174 - getting one of them over the line would be ideal :)

CpuID:
> For now, an _encrypted_ private key is omitted from the design, but after solving of this topic, the next function could be

this part looks like it might not be so straightforward...? since you don't get a PEM block back for encrypted keys using https://godoc.org/golang.org/x/crypto/ssh#ParsePrivateKey - you get a signer block?

means you can't just drop the private + public key into the `KeyPair` struct...

CpuID:
actually, looking further at this I can probably use the SSH agent functionality in `modules/ssh` instead of trying to load my private key within golang. much better

ellisonc:
@sw-samuraj, thanks for your contribution. We are removing old PRs. If this change is still required, we encourage you to re-open.

PR #176, opened 2018-10-19
Added packer multi artifact parallel build

eak12913:
Added: function: BuildArtifactsE can take a map of identifierName <-> Options and then parallelize the packer builds. Once all the packer builds have completed a map of identifierName <-> generated identifier is returned. If any artifact fails to build, the errors are accumulated and returned as a MultiError. The identifierName can be anything you want, it is only used so that you can know which generated artifact is which.

Added test to verify functionality.

### DIFF
--- a/modules/packer/packer.go
+++ b/modules/packer/packer.go
@@ -34,6 +34,8 @@ func BuildArtifactsE(t *testing.T, artifactNameToOptions map[string]*Options) (m
 	errorsOccurred := []error{}
 
 	for artifactName, curOptions := range artifactNameToOptions {
+		// The following is necessary to make sure testCase's values don't
+		// get updated due to concurrency within the scope of t.Run(..) below
 		artifactName := artifactName
 		curOptions := curOptions
 		go func() {
--- a/modules/packer/packer.go
+++ b/modules/packer/packer.go
@@ -5,8 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"sync"
 	"testing"
 
+	"github.com/gruntwork-io/terratest/modules/customerrors"
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/shell"
 )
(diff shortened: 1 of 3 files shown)


modules/packer/packer.go
```diff
@@ -19,6 +21,38 @@ type Options struct {
...
+func BuildArtifactsE(t *testing.T, artifactNameToOptions map[string]*Options) (map[string]string, error) {
+	var waitForArtifacts sync.WaitGroup
+	waitForArtifacts.Add(len(artifactNameToOptions))
+
+	var artifactNameToArtifactId = map[string]string{}
+	errorsOccurred := []error{}
+
+	for artifactName, curOptions := range artifactNameToOptions {
+		artifactName := artifactName
+		curOptions := curOptions
```
brikis98: Add comment explaining why you need this capture

test/packer_basic_example_test.go
```diff
@@ -51,6 +52,49 @@ func TestPackerBasicExample(t *testing.T) {
...
 
+func TestPackerMultipleConcurrentAmis(t *testing.T) {
+	t.Parallel()
+	i := 0
+
+	// Build a map of 3 randomId <-> packer.Options, in 3 random AWS Regions
+	// then build all of these AMIs in parallel and make sure that there are
+	// no errors.
+	var identifierToOptions = map[string]*packer.Options{}
+	for i < 3 {
```
brikis98: Why not a normal `for i := 0; i < 3; i++`?

test/packer_basic_example_test.go
```diff
@@ -51,6 +52,49 @@ func TestPackerBasicExample(t *testing.T) {
...
+		// Pick a random AWS region to test in. This helps ensure your code works in all regions.
+		awsRegion := terratest_aws.GetRandomRegion(t, nil, nil)
```
brikis98: If you happen to get the same region twice, will you get an error about duplicate AMI names? If so, the Packer template may want to take in a variable called `ami_base_name`, set by default to "terratest-packer-example" (the current name). At test time, you can configure `packer.Options` to use a different `random.UniqueId()` for each one as part of the name.

test/packer_basic_example_test.go
eak12913: I just saw that on the golang examples page - it seemed like fewer characters to type and was otherwise equivalent.

brikis98: But you set `i := 0` a couple lines above and do `i++` a few lines below... So # of characters is identical... but now spread over extra lines?

modules/packer/packer.go
eak12913: done

test/packer_basic_example_test.go
eak12913: fixed

test/packer_basic_example_test.go
eak12913: changed to standard for loop. I guess I was really looking for syntax like: `for i := range(0,2) ...`

modules/packer/packer.go
```diff
@@ -19,6 +21,54 @@ type Options struct {
...
+// know which generated artifact is which.
+func BuildArtifactsE(t *testing.T, artifactNameToOptions map[string]*Options) (map[string]string, error) {
+	var waitForArtifacts sync.WaitGroup
+	waitForArtifacts.Add(len(artifactNameToOptions))
+
+	var artifactNameToArtifactId = map[string]string{}
+	errorsOccurred := []error{}
+
+	for artifactName, curOptions := range artifactNameToOptions {
+		// The following is necessary to make sure testCase's values don't
```
brikis98: Not test cases

test/packer_basic_example_test.go
```diff
@@ -51,6 +52,43 @@ func TestPackerBasicExample(t *testing.T) {
...
+
+		packerOptions := &packer.Options{
+			// The path to where the Packer template is located
+			Template: "../examples/packer-basic-example/build.json",
+
+			// Variables to pass to our Packer build using -var options
+			Vars: map[string]string{
+				"aws_region":    awsRegion,
+				"ami_base_name": random.UniqueId(),
```
brikis98: Include some reasonable name as well as `UniqueId` so when debugging, we know where these images are coming from.

modules/packer/packer.go
eak12913: updated

test/packer_basic_example_test.go
eak12913: done

PR #177, opened 2018-10-19
Add GetInstances() and GetPublicIps() to GCP Instance Groups

josh-padnick:
With AWS, Terraform resources output a full collection of properties. For example, the [aws_autoscaling_group](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html#attributes-reference) outputs over 10 properties like `min_size`, `max_size`, etc.

But with GCP, Terraform outputs a surprisingly few number of properties. In particular, the [google_compute_region_instance_group_manager](https://www.terraform.io/docs/providers/google/r/compute_region_instance_group_manager.html#attributes-reference) includes just 3 properties. Therefore, nearly every user of Terratest that uses [Instance Groups](https://cloud.google.com/compute/docs/instance-groups/) will need functionality in Terratest to get public IPs and other instance properties.

In that spirit, this PR adds:

- `instanceGroup.GetInstances()`
- `instanceGroup.GetPublicIps()`

Please hold off on merging this for now, as I may add more functions here as I work through GCP tests in Vault.

### DIFF
--- a/examples/terraform-gcp-ig-example/main.tf
+++ b/examples/terraform-gcp-ig-example/main.tf
@@ -6,11 +6,11 @@
 # Create a Regional Managed Instance Group
 resource "google_compute_region_instance_group_manager" "example" {
   project = "${var.gcp_project_id}"
-  region = "${var.gcp_region}"
+  region  = "${var.gcp_region}"
 
-  name = "${var.cluster_name}-ig"
+  name               = "${var.cluster_name}-ig"
   base_instance_name = "${var.cluster_name}"
-  instance_template = "${google_compute_instance_template.example.self_link}"
+  instance_template  = "${google_compute_instance_template.example.self_link}"
 
   target_size = "${var.cluster_size}"
 }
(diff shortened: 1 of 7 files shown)


examples/terraform-gcp-ig-example/main.tf
```diff
@@ -0,0 +1,43 @@
...
+resource "google_compute_region_instance_group_manager" "example" {
+  project = "${var.gcp_project_id}"
+  region = "${var.gcp_region}"
+
+  name = "${var.cluster_name}-ig"
+  base_instance_name = "${var.cluster_name}"
+  instance_template = "${google_compute_instance_template.example.self_link}"
+
+  target_size = "${var.cluster_size}"
+}
```
brikis98: Please run `terraform fmt`

josh-padnick: Done in dc29013d3611839cc35ec1d06d3d899c477372f9.

PR #178, opened 2018-10-20
Add support for copying symbolic links.

josh-padnick:
While using Terratest on another project, this line:

```
exampleDir := test_structure.CopyTerraformFolderToTemp(t, "../", "examples/vault-cluster-private")
```

cased a strange error:

```
test_structure.go:59: open vendor/github.com/hashicorp/go-rootcerts/test-fixtures/capath-with-symlinks/securetrust.pem: no such file or directory
```

The issue turned out to be that this directory contained vendored files, which happened to include a symbolic link, which happened to point to anon-existent file. This PR adds support for for processing symlinks. Specifically, the symlink is copied exactly as is, regardless of whether it points to an existing file.


modules/files/files.go
```diff
@@ -124,3 +129,24 @@ func WriteFileWithSamePermissions(source string, destination string, contents []
...
+}
+
+// copySymLink copies the source symbolic link to the given destination.
+func copySymLink(source string, destination string) error {
+	symlinkPath, err := os.Readlink(source)
+	if err != nil {
+		return err
+	}
+
+	err = os.Symlink(symlinkPath, destination)
```
brikis98: Does this create a symlink to a symlink?

josh-padnick:
Tests pass and PR approved, so merging now!

PR #179, opened 2018-10-21
Terratest Log Parser

yorinasub17:
This is the newest version of the log parser we had been prototyping that is now ready for terratest core. The two new modules are:

- logger/parser : Contains the library code that powers the log parsing
- cmd/terratest_log_parser : Contains the CLI frontend into the log parser

The `cmd` directory comes from [a common project layout for go projects I found](https://github.com/golang-standards/project-layout), where library source is separated from CLI sources. I thought this made sense as they are inherently different things (one should be imported and called, while the other should not), so implemented here but let me know if you think we should have a different structure.

You can look at the [test fixtures](https://github.com/gruntwork-io/terratest/tree/3ae8f122c46368c8557b79312363e8a84e5a87d4/modules/logger/parser/fixtures) to see example input-outputs of the parser (used in [integration_test.go](https://github.com/gruntwork-io/terratest/blob/3ae8f122c46368c8557b79312363e8a84e5a87d4/modules/logger/parser/integration_test.go)).

I also updated the README with information on the new command and how it is meant to be used, as well as updating the circleCI config to:

- Build and use the command to parse out the logs
- Deploy and release the binaries to the GitHub releases when we build new versions

One question I have is if you know of a way to test the CircleCI deploy configs? I copied it from the gruntkms project, but I have no way of knowing if the workspace transfer worked until we actually do a tagged release with this code...


modules/logger/parser/store.go
```diff
@@ -0,0 +1,110 @@
...
+	"path/filepath"
+
+	"github.com/gruntwork-io/terratest/modules/files"
+	junitformatter "github.com/jstemmer/go-junit-report/formatter"
+	junitparser "github.com/jstemmer/go-junit-report/parser"
+	"github.com/sirupsen/logrus"
+)
+
+type LogWriter struct {
+	// Represents an open channel to a test writer for given test indexed by name
```
brikis98: Curious why you need channels for this?

modules/logger/parser/failed_test_marker_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	emptyMarkerStack := TestResultMarkerStack{}
+	fullMarkerStack := createTestStack()
+	assert.True(t, emptyMarkerStack.isEmpty())
+	assert.False(t, fullMarkerStack.isEmpty())
+}
+
+// Test that removeDedentedTestResultMarkers remove items that are dedented from the current level, assuming the stack
+// is ordered by indent level.
+func TestRemoveDedentedTestResultMarkers(t *testing.T) {
```
brikis98: `t.Parallel()` here and a few tests below?

modules/logger/parser/fixtures/panic_example_expected/summary.log
```diff
@@ -0,0 +1,65 @@
...
+--- PASS: TestCloseChannelsClosesAll (0.00s)
+--- PASS: TestEnsureDirectoryExistsHandlesExistingDirectory (0.00s)
+--- PASS: TestGetOrCreateChannelCreatesNewChannel (0.00s)
+--- FAIL: TestIsPanicLine (0.00s)
+    --- PASS: TestIsPanicLine/BaseCase (0.00s)
+    --- PASS: TestIsPanicLine/NotPanic (0.00s)
+panic: error [recovered]
+	panic: error
+
+goroutine 36 [running]:
```
brikis98: Great testing 👍

modules/logger/parser/integration_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	dirBAbs, err := filepath.Abs(dirB)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cmd := shell.Command{
+		Command: "diff",
+		Args:    []string{"-arq", dirAAbs, dirBAbs},
```
brikis98: Add comment for why we're using `diff` (as opposed to something written in Go).

modules/logger/parser/parser.go
```diff
@@ -0,0 +1,185 @@
...
+	"strings"
+	"sync"
+
+	junitparser "github.com/jstemmer/go-junit-report/parser"
+	"github.com/sirupsen/logrus"
+)
+
+// SpawnParsers will spawn the log parser and junit report parsers off of a single reader.
+func SpawnParsers(logger *logrus.Logger, reader io.Reader, outputDir string) {
+	pr, pw := io.Pipe()
```
brikis98: It's hard to guess what one and two letter variable names stand for. Longer, more descriptive names typically make the code easier to read & maintain.

modules/logger/parser/failed_test_marker.go
```diff
@@ -0,0 +1,64 @@
...
+// TestResultMarker tracks the indentation level of a test result line in go test output.
+// Example:
+// --- FAIL: TestSnafu
+//     --- PASS: TestSnafu/Situation
+//     --- FAIL: TestSnafu/Normal
+// The three markers for the above in order are:
+// TestResultMarker{TestName: "TestSnafu", IndentLevel: 0}
+// TestResultMarker{TestName: "TestSnafu/Situation", IndentLevel: 4}
+// TestResultMarker{TestName: "TestSnafu/Normal", IndentLevel: 4}
+type TestResultMarker struct {
```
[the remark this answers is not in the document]
yorinasub17: I thought about this. Piecing out the terratest logs in the core lib are easy enough (since WE control when those are written). However, extracting the go testing logs for detecting pass/fail is hard. Specifically, I have yet to find a good way to get hooks into the test itself so that you could intercept the test entries. Without that, you would only get the terratest logs but not the result entries (still useful). The lack of hooks argument is (sort of) supported by the community implementations of junit reporting. I found two implementations, [one for gradle](https://github.com/echocat/gradle-golang-plugin/commit/8464789852c305fb36ad5d4a3cba03bc13ddb046#diff-94b9a204193aa8994563f6f4e6f93474R1), and the other is [a command, which I am using here](https://github.com/jstemmer/go-junit-report), both of which rely on after the fact parsing to get the results out rather than baking it in to `go test` as a plugin.

The best alternative way I found so far for capturing test results is redefining `T`, but that would require changing all tests to use the new `T` type instead of go's `T` type. This also requires providing a custom `TestRunner` so that it will build the new `T` type. At this point, we are basically reimplementing the testing infrastructure of go, and not sure if that is good.

Given that assumption (that we can't grab `go test` logs on the fly), here are some additional bullet list pros and cons for the current approach that I came up with to further inform the decision:

Pros:
- Robust to unexpected failures in testing code, like `ctrl+c`, panics, OOM kills and the like since the parser is not tied to the testing process. This approach is less likely to miss these entries, and can be surfaced to the summary view for easy viewing in CI engines (no need to scroll), like the panic example.
- Can combine `go test` output (e.g `--- PASS` entries) with the log entries for the test in a single log file.
- Can extract out the summary view (those are all `go test` log entries).
- Can build `junit.xml` report that CI engines can use for test insights.

Cons:
- Complicated implementation that is potentially brittle. E.g if someone decides to change the logging format then this will all break. If we hook during the test, then the implementation is easier because those logs are all emitted at certain points in code, the information of which is lost in the final log and have to parse out.
- Have to store all the logs twice (the full interleaved version, and the broken out version) because the parsing depends on logs being available. (NOTE: this is avoidable with a pipe, but some CI engines are not good with pipes. For example, I couldn't get the piped version to work properly in circleCI).

modules/logger/parser/store.go
yorinasub17: I was thinking about the writer as a task, that would need to open the file and write to it and then I landed on channels and goroutines where the task would open the file and wait for input.
... and then this comment made me realize that I can just open the file handle and store that here and write to that.... 

picarddoublefacepalm-1

modules/logger/parser/failed_test_marker.go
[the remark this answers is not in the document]
brikis98: That's a good analysis. You should probably add it to the docs! There doesn't seem to be a perfect solution here, so let's go with what you have, as it's 100x better than what we had before. Thanks!

modules/logger/parser/store.go
brikis98: 😄

brikis98: 😄

modules/logger/parser/failed_test_marker.go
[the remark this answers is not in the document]
brikis98: That's a good analysis. You should probably add it to the docs! There doesn't seem to be a perfect solution here, so let's go with what you have, as it's 100x better than what we had before. Thanks!

modules/logger/parser/store.go
brikis98: 😄

modules/logger/parser/failed_test_marker_test.go
yorinasub17: fixed

modules/logger/parser/parser.go
yorinasub17: fixed

modules/logger/parser/integration_test.go
yorinasub17: fixed

modules/logger/parser/store.go
yorinasub17: fixed

modules/logger/parser/failed_test_marker.go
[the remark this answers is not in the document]
Etiene: I would like to +1 the sentiment that it can be improved at a later point

yorinasub17:
@brikis98 Ok I have addressed all the comments and this is ready to go. Any objections to merging and releasing?

modules/logger/parser/parser.go
```diff
@@ -0,0 +1,180 @@
...
+	"strings"
+	"sync"
+
+	junitparser "github.com/jstemmer/go-junit-report/parser"
+	"github.com/sirupsen/logrus"
+)
+
+// SpawnParsers will spawn the log parser and junit report parsers off of a single reader.
+func SpawnParsers(logger *logrus.Logger, reader io.Reader, outputDir string) {
+	forked_reader, forked_writer := io.Pipe()
```
brikis98: camelCase

yorinasub17: oops too much python...

yorinasub17:
Fixed to camelCase. Waiting for build, then will merge and release!

yorinasub17:
Tests passed, so merging!

Issue #175, opened 2018-10-18
New OutputMap function for the Terraform module


jmfontaine:
I am considering adding a set of functions to return the Terraform output for a map, similar to the recently added [OutputList and OutputListE functions](https://github.com/gruntwork-io/terratest/pull/169).

That was actually [suggested in the PR](https://github.com/gruntwork-io/terratest/pull/169#discussion_r224589733) and I have a need for it: I have a Terraform module that generates standardized AWS tags and outputs them as a map.

The function signatures would be:

```go
OutputMap(t *testing.T, options *Options, key string) [string]string
OutputMapE(t *testing.T, options *Options, key string) ([string]string, error)
```

If you think it would be valuable to the community, I can submit a PR for that feature.

brikis98:
That would indeed be very handy. PR welcome. Thank you!

jmfontaine:
Will submit a PR soon then. Thanks for the feedback!

jmfontaine:
This was fixed by PR #180. Closing.

PR #180, opened 2018-10-21
Add functions to return Terraform output as a map

jmfontaine:
This PR fixes issue #175 (New OutputMap function for the Terraform module) and adds the following functions in the Terraform module:

* `OutputMap(t *testing.T, options *Options, key string) map[string]string`: Calls terraform output for the given variable and returns its value as a map. If the output value is not a map type, then it fails the test.
* `OutputMapE(t *testing.T, options *Options, key string) (map[string]string, error)`: Calls terraform output for the given variable and returns its value as a map. If the output value is not a map type, then it returns an error.

The output for the tests is [available as a gist](https://gist.github.com/jmfontaine/135b4d9291e14e8bfa41c30ee2222b04).

In order to keep implementation simple, the values are expected to be castable to string.

Supporting arbitrary types as `interface{}` is possible. I considered renaming the functions to `OutputMapString` and `OutputMapStringE` respectively, and add two `OutputMapInterface` and `OutputMapInterfaceE`. Is that something that we want now or do we want to wait until someone has a use case for that?

### DIFF
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -91,6 +91,47 @@ func OutputListE(t *testing.T, options *Options, key string) ([]string, error) {
 	return list, nil
 }
 
+// OutputMap calls terraform output for the given variable and returns its value as a map.
+// If the output value is not a map type, then it fails the test.
+func OutputMap(t *testing.T, options *Options, key string) map[string]string {
+	out, err := OutputMapE(t, options, key)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


jmfontaine:
You are welcome! Thanks for this great tool. I am glad I could modestly contribute to it.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.13.6

PR #182, opened 2018-10-24
Attempt to fix recent GCP test failures.

josh-padnick:
Recent CircleCI test builds ([538](https://circleci.com/gh/gruntwork-io/terratest/538), [542](https://circleci.com/gh/gruntwork-io/terratest/542), [543](https://circleci.com/gh/gruntwork-io/terratest/543), [544](https://circleci.com/gh/gruntwork-io/terratest/544), [545](https://circleci.com/gh/gruntwork-io/terratest/545)) had multiple intermittent failures on the GCP tests. Because these tests passed consistently before, I wonder if the [recent GCP incident](https://status.cloud.google.com/incident/cloud-networking/18017) is relevant?

In either case, this is an initial attempt to fix GCP issues. Please do not merge until tests are passing consistently. Please do feel free to comment with feedback and ideas.

'cc @yorinasub17

### DIFF
--- a/modules/gcp/region.go
+++ b/modules/gcp/region.go
@@ -2,7 +2,6 @@ package gcp
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"strings"
 	"testing"
@@ -10,7 +9,7 @@ import (
 	"github.com/gruntwork-io/terratest/modules/collections"
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/random"
-	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/compute/v1"
 )
 
 // You can set this environment variable to force Terratest to use a specific Region rather than a random one. This is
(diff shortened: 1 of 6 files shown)


yorinasub17:
@josh-padnick , after digging in, I found a correlation between failed tests and the selection of region `asia-east2`, which just [launched on Monday](https://cloud.google.com/compute/docs/release-notes). I scanned the failing tests referenced in this PR and only 2 of those were NOT in the `asia-east2` region. The rest (~5 or 6) were all in that region. Should we try black listing it?

josh-padnick:
Thanks, @yorinasub17! Your suggestion was spot on. Would you mind reviewing this before I merge?

modules/gcp/region.go
```diff
@@ -100,10 +101,19 @@ func GetRandomZoneE(t *testing.T, projectID string, approvedZones []string, forb
...
-	zone := random.RandomString(zonesToPickFrom)
 
-	logger.Logf(t, "Using Zone %s", zone)
-	return zone, nil
+	zone := ""
+	maxAttempts := 1000
+
+	for i := 0; i < maxAttempts; i++ {
+		zone = random.RandomString(zonesToPickFrom)
+		if !isInRegions(zone, forbiddenRegions) {
```
yorinasub17: Is it possible to just filter out `forbiddenRegions` from `zonesToPickFrom` directly before randomly picking? I.e create a new list that only includes those that are not forbidden? Similar to how we do it [for aws](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/region.go#L54).

yorinasub17:
I took a pass. LGTM except for one possible optimization to avoid the occasional "could not find region" error.

josh-padnick:
I'm an idiot. So much for coding during a talk! I fixed the issue and added a test to boot. Once tests pass, I'll merge and release.

Issue #181, opened 2018-10-23
DynamoDb support


rafaelmarques7:
Hello,

First of all, let me start by saying that I recently discovered this library, and that I truly appreciate it, as it is very useful! 

With that being said, I was trying to test a terraform module that created dynamoDb tables.
During the tutorials, I noticed it was possible to test RDS databases, but now, upon viewing the documentation  of the [aws package](https://godoc.org/github.com/gruntwork-io/terratest/modules/aws), I noticed that there was no function available for testing DynamoDb. 

Is this really the case, or did I missed the function completely?

brikis98:
We have not yet added helpers for working with DynamoDB. PR very welcome 😃

iamjohnnym:
@brikis98 Given that @rafaelmarques7 doesn't have the time, I'll see if I can tackle this.  Do you have a specific vision for this, anything that you can think of that I should keep in mind when doing?

brikis98:
Thanks @iamjohnnym! 

I imagine the most useful functionality would be:

- Create a DynamoDB table
- Wait until the table exists and is ready (remember, AWS APIs are async + eventually consistent) 
- Write some data
- Read some data
- Delete a DynamoDB table

iamjohnnym:
That's fair.

I'll take over @rafaelmarques7 original work.  From there, I'll address your questions from that PR.  I also have a more immediate needs to check metadata of a dynamoDB table.

So I understand a bit better.  What processes use the create/destroy methods of a given resource?  I understand looking at attributes on AWS that gets deployed by terraform but what does `terratest` do with methods that create resources?

brikis98:
> What processes use the create/destroy methods of a given resource? I understand looking at attributes on AWS that gets deployed by terraform but what does terratest do with methods that create resources?

Not sure I understand the question?

yorinasub17:
Initial set of DynamoDB functions were introduced in https://github.com/gruntwork-io/terratest/releases/tag/v0.15.11. If there are more to add, please file new tickets. Thanks!

PR #183, opened 2018-10-24
Added DynamoDB functions, and respective tests

rafaelmarques7:
This Pull Request is a response to [this issue](https://github.com/gruntwork-io/terratest/issues/181#issuecomment-432336879). 

It adds functions to handle DynamoDb, in particular:
  * Create a table
  * Delete a table

It also adds a test function to assert that these two cases work properly. 
The test is currently passing, and its logs can be found [here](https://gist.github.com/rafaelmarques7/40885518c36e093503f35d8b9cc0dd18).

If this PR does not reach your standards, please let me know how I can improve it, and I will do so.

Kind regards,
Rafael

### DIFF
--- a/modules/aws/dynamodb.go
+++ b/modules/aws/dynamodb.go
@@ -1,0 +1,89 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/gruntwork-io/terratest/modules/logger"
+)
+
+func CreateDynamodbTable(t *testing.T, region string, name string) {
+	err := CreateDynamodbTableE(t, region, name)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func CreateDynamodbTableE(t *testing.T, region string, name string) error {
+	logger.Logf(t, "Creating DynamoDb table '%s' in %s", name, region)
+
+	dynamodbClient, err := NewDynamodbClientE(t, region)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	params := &dynamodb.CreateTableInput{
+		AttributeDefinitions: []*dynamodb.AttributeDefinition{
+			{
+				AttributeName: aws.String("Terratest"),
+				AttributeType: aws.String("S"),
+			},
+		},
+		KeySchema: []*dynamodb.KeySchemaElement{
+			{
+				AttributeName: aws.String("Terratest"),
+				KeyType: aws.String("HASH"),
+			},
+		},
+		ProvisionedThroughput: &dynamodb.ProvisionedThroughput{
+			ReadCapacityUnits:  aws.Int64(1),
+			WriteCapacityUnits: aws.Int64(1),
+		},
+		TableName: aws.String(name),
+	}
+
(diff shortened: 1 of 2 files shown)


modules/aws/dynamodb.go
```diff
@@ -0,0 +1,89 @@
...
+		t.Fatal(err)
+	}
+}
+
+func CreateDynamodbTableE(t *testing.T, region string, name string) error {
+	logger.Logf(t, "Creating DynamoDb table '%s' in %s", name, region)
+
+	dynamodbClient, err := NewDynamodbClientE(t, region)
+	if err != nil {
+		t.Fatal(err)
```
brikis98: Methods that end with `E` should return errors, not call `t.Fatal`

modules/aws/dynamodb.go
```diff
@@ -0,0 +1,89 @@
...
+			{
+				AttributeName: aws.String("Terratest"),
+				AttributeType: aws.String("S"),
+			},
+		},
+		KeySchema: []*dynamodb.KeySchemaElement{
+			{
+				AttributeName: aws.String("Terratest"),
+				KeyType: aws.String("HASH"),
+			},
```
brikis98: What are these attributes? Should they be configurable?

modules/aws/dynamodb.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+		ProvisionedThroughput: &dynamodb.ProvisionedThroughput{
+			ReadCapacityUnits:  aws.Int64(1),
+			WriteCapacityUnits: aws.Int64(1),
```
brikis98: Should this be configurable?

modules/aws/dynamodb.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+		TableName: aws.String(name),
+	}
+
+	_, err = dynamodbClient.CreateTable(params)
```
brikis98: The DynamoDB table API is asynchronous and after this returns, the table doesn't actually exist for another 5-30 seconds. You need to add a wait as shown here or many uses of this API will end up hitting errors when they try to use that table too early: https://github.com/gruntwork-io/terragrunt/blob/master/dynamodb/dynamo_lock_table.go#L170

modules/aws/dynamodb.go
rafaelmarques7: This method is called CreateDynamodbTable, and returns Fatal upon encountering an error.
The method CreateDynamodbTableE (which is called by the previous one) returns an error, and does not throw fatal. I think this part is correct, but please reevaluate it.

iamjohnnym:
:wave: Hey all,

Curious as to what the status is here.  I realize some changes are required, its been a little over a month since there was any last contact.  Any help needed?

brikis98:
@rafaelmarques7, let us know if you'll have time to wrap this one up. If not, @iamjohnnym, I'd welcome you taking over the PR and finishing it. Thx!

rafaelmarques7:
Hello @brikis98 @iamjohnnym, unfortunately I do not have the time to proceed with those changes :s. Wish you all good luck. Cheers.

brikis98:
@iamjohnnym @rafaelmarques7 Do either of you have time to wrap up this PR?

PR #189, opened 2018-10-30
Found and fixed two issues

eak12913:
1. There was an issue propagating errors out of ssh.go. There was an erroneous declaration of the errors array that would clobber any existing errors that were in there. The result was that some kinds of errors would not be propagated to the calling functions as expected.
2. There was a problem in some bash versions when calling the find command with the "*" filter. The solution turned out that we have to wrap in single quotes, but _only_ if the filter is just for "*" (https://stackoverflow.com/questions/6495501/find-paths-must-precede-expression-how-do-i-specify-a-recursive-search-that)

### DIFF
--- a/modules/aws/ec2-files.go
+++ b/modules/aws/ec2-files.go
@@ -218,21 +218,15 @@ func FetchFilesFromAsgsE(t *testing.T, awsRegion string, spec RemoteFileSpecific
 
 			instanceIDs, err := GetInstanceIdsForAsgE(t, curAsg, awsRegion)
 			if err != nil {
-				return err
-			}
-
-			for _, instanceID := range instanceIDs {
-				err = FetchFilesFromInstanceE(t, awsRegion, spec.SshUser, spec.KeyPair, instanceID, spec.UseSudo, curRemoteDir, spec.LocalDestinationDir, fileFilters)
-
-				if err != nil {
-					errorsOccurred = append(errorsOccurred, err)
-					err = nil // since we reuse err, make sure not to double add the same error if this loop exits
-				}
-			}
-
-			if err != nil {
 				errorsOccurred = append(errorsOccurred, err)
-				err = nil
+			} else {
+				for _, instanceID := range instanceIDs {
+					err = FetchFilesFromInstanceE(t, awsRegion, spec.SshUser, spec.KeyPair, instanceID, spec.UseSudo, curRemoteDir, spec.LocalDestinationDir, fileFilters)
+
+					if err != nil {
+						errorsOccurred = append(errorsOccurred, err)
+					}
+				}
 			}
 		}
 	}
--- a/modules/aws/ec2-files.go
+++ b/modules/aws/ec2-files.go
@@ -221,18 +221,18 @@ func FetchFilesFromAsgsE(t *testing.T, awsRegion string, spec RemoteFileSpecific
 				return err
 			}
 
-			errorsOccurred := []error{}
-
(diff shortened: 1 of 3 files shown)


modules/aws/ec2-files.go
```diff
@@ -221,18 +221,18 @@ func FetchFilesFromAsgsE(t *testing.T, awsRegion string, spec RemoteFileSpecific
...
 				if err != nil {
 					errorsOccurred = append(errorsOccurred, err)
+					err = nil // since we reuse err, make sure not to double add the same error if this loop exits
 				}
 			}
 
 			if err != nil {
 				errorsOccurred = append(errorsOccurred, err)
+				err = nil
 			}
```
yorinasub17: I'm actually not sure this whole block is necessary, given that you are checking `err` right after each call.

eak12913: Hmm - good point. I will remove this

modules/aws/ec2-files.go
```diff
@@ -221,18 +221,18 @@ func FetchFilesFromAsgsE(t *testing.T, awsRegion string, spec RemoteFileSpecific
 				return err
```
eak12913: This feels inconsistent to me. Below, when I encounter errors I add them to an array, but here I bounce out. I'm going to change behavior such that we try to scp files from however many nodes we can. Then return all the errors we encountered along the way.

modules/aws/ec2-files.go
eak12913: removed.

modules/aws/ec2-files.go
eak12913: Updated the logic to be more consistent as I described above.

modules/ssh/ssh.go
```diff
@@ -350,6 +350,9 @@ func listFileInRemoteDir(t *testing.T, sshSession *SshSession, options ScpDownlo
 
 		findCommandArgs = append(findCommandArgs, "\\(")
 		for i, curFilter := range options.FileNameFilters {
+			// due to inconsistent bash behavior we need to wrap the wildcard if it
+			// is just on its own.
```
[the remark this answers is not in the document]
eak12913: Added a test to explicitly test the "*" case - this was what exposed the problem when I found the issue initially.

modules/aws/ec2-files.go
```diff
@@ -218,22 +218,16 @@ func FetchFilesFromAsgsE(t *testing.T, awsRegion string, spec RemoteFileSpecific
...
-			}
-
-			errorsOccurred := []error{}
-
-			for _, instanceID := range instanceIDs {
-				err = FetchFilesFromInstanceE(t, awsRegion, spec.SshUser, spec.KeyPair, instanceID, spec.UseSudo, curRemoteDir, spec.LocalDestinationDir, fileFilters)
+				errorsOccurred = append(errorsOccurred, err)
+			} else {
+				for _, instanceID := range instanceIDs {
+					err = FetchFilesFromInstanceE(t, awsRegion, spec.SshUser, spec.KeyPair, instanceID, spec.UseSudo, curRemoteDir, spec.LocalDestinationDir, fileFilters)
```
brikis98: This is still reusing the `err` from before, but I suppose now it will happen to work... I still very much prefer immutable variables that never change from declaration to avoid these pitfalls, but this may be good enough for now.

eak12913: agreed

PR #190, opened 2018-11-01
Remove broken symlink from fixtures

Etiene:
Fixture was causing errors on packer builds that copy terratest folder.
Replaced by creating the test file on the go during test.

### DIFF
--- a/modules/files/files_test.go
+++ b/modules/files/files_test.go
@@ -74,6 +74,19 @@ func TestCopyFolderContentsWithSymLinks(t *testing.T) {
 func TestCopyFolderContentsWithBrokenSymLinks(t *testing.T) {
 	t.Parallel()
 
+	// Creating broken symlink
+	pathToFile := filepath.Join(copyFolderContentsFixtureRoot, "symlinks-broken/nonexistent-folder/bar.txt")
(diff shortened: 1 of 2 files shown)


modules/files/files_test.go
```diff
@@ -74,6 +74,19 @@ func TestCopyFolderContentsWithSymLinks(t *testing.T) {
 func TestCopyFolderContentsWithBrokenSymLinks(t *testing.T) {
 	t.Parallel()
 
+	// Creating broken symlink
+	fPath := filepath.Join(copyFolderContentsFixtureRoot, "symlinks-broken/nonexistent-folder/bar.txt")
+	sPath := filepath.Join(copyFolderContentsFixtureRoot, "symlinks-broken/bar.txt")
```
eak12913: what do the `s` and `f` stand for in these names? `soft` and `file`?

Etiene: file and symlink. Could've named it `filePath`, instead of `fPath`, but thought it could be confusing given the library being used is `filepath`.

brikis98: Perhaps `pathToFile` and `pathToSymlink`?

Issue #186, opened 2018-10-26
http_helper package is covering HTTP GET methods only and doesn't allow to send HTTP Headers


sw-samuraj:
The `http_helper` package currently supports the HTTP `GET` method only. Additionally, it doesn't allow to send any HTTP Header.

**Questions:**

1) Is it planned to support other HTTP methods, e.g. `POST`?

2) Is it planned to support HTTP Headers? This could be important if a service expects some mandatory headers, e.g. a token etc.

brikis98:
We add methods when we use them. So far, most of our tests have needed GET and no headers, but if you have a use case that needs more, I'd definitely welcome a PR to add those!

robmorgan:
Fixed in https://github.com/gruntwork-io/terratest/releases/tag/v0.18.5

PR #193, opened 2018-11-06
Add generic HTTP helper methods

sw-samuraj:
Related to #186 (generic HTTP methods)

I've added couple of generic HTTP helper methods which mimic current `HttpGet*` methods. Those new methods allows to:

* add an HTTP header(s),
* add a request body,
* set a HTTP method,
* returns a response body.

All those methods have similar signature:

```go
func HttpDo(t *testing.T, method string, url string, body io.Reader, headers map[string]string) (int, string)
```

Additionaly, there is a brand new method `HttpDoWithRetry` which behaves similarly to `HttpGetWithRetry` with exception that the given `expectedStatus` has to match received `statusCode` to pass. The method has following signature:

```go
func HttpDoWithRetry(t *testing.T, method string, url string, body io.Reader, headers map[string]string, expectedStatus int, retries int, sleepBetweenRetries time.Duration) string
```

### DIFF
--- a/modules/http-helper/dummy_server.go
+++ b/modules/http-helper/dummy_server.go
@@ -44,24 +44,26 @@ func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
 	return listener, port, err
 }
 
-// RunDummyHandlerServer runs a dummy HTTP server on a unique port that will serve the given handler. Returns the Listener for the server, the
-// port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
+// RunDummyHandlerServer runs a dummy HTTP server on a unique port that will serve given handlers. Returns the Listener for the server,
+// the port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
 // the Close() method on the Listener when you're done!
-func RunDummyHandlerServer(t *testing.T, path string, handler func(http.ResponseWriter, *http.Request)) (net.Listener, int) {
-	listener, port, err := RunDummyHandlerServerE(t, path, handler)
+func RunDummyHandlerServer(t *testing.T, handlers map[string]func(http.ResponseWriter, *http.Request)) (net.Listener, int) {
+	listener, port, err := RunDummyHandlerServerE(t, handlers)
 	if err != nil {
 		t.Fatal(err)
 	}
 	return listener, port
 }
 
-// RunDummyHandlerServerE runs a dummy HTTP server on a unique port that will server the given handler. Returns the Listener for the server, the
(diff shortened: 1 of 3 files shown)


sw-samuraj:
@brikis98 I've added couple of unit tests. Let me know if you missing something, or expect to have different type of tests.

I use a new version of `DummyServer` which takes a list of handlers with defined behavior and then I run a set of tests against those handlers.

modules/http-helper/dummy_server.go
```diff
@@ -44,6 +44,39 @@ func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
...
+}
+
+// RunDummyHandlerServerE runs a dummy HTTP server on a unique port that will server given handlers. Returns the Listener for the server,
+// the port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
+// the Close() method on the Listener when you're done!
+func RunDummyHandlerServerE(t *testing.T, handlers map[string]func(http.ResponseWriter, *http.Request)) (net.Listener, int, error) {
+	port := getNextPort()
+
+	for path, handler := range handlers {
+		http.HandleFunc(path, handler)
```
brikis98: Isn't this a global call? If so, then using this across multiple tests in parallel will cause conflicts, right?

IIRC, there was some way with Go's HTTP server to define a local instance of the server and its handler functions.

brikis98:
@sw-samuraj What is the latest status of this PR?

sw-samuraj:
I put it on hold, because of switching for some priority stuff. :weary: I'm going to close it now and will resume it later once having some spare time.

laurentdutheil:
Hello

I am very interrested by this PR.
I contribute to an Auth0 provider and I want to request their API to make some assertions. But I need to be authentified to request the API and the possibility to specify the headers parameters will help me a lot.

How can I help ?

sw-samuraj:
Yes, @laurentdutheil - the authorization headers were actually the reason why I've started this pull request. Unfortunately, I'm currently allocated on something else for now, but I still plan to finish this feature.

Actually, I think that implemented methods are just fine and ready and I need to deal with tests only. So, any help is welcome. :smiley:

jendrikjoe:
@sw-samuraj, I took the liberty to rewrite the tests to use the httptest.Server. I have the changes locally. Would you provide me push rights to your fork so I can simply push on top of it? :slightly_smiling_face: Otherwise, I would fork the fork and open a new MR :slightly_smiling_face:

sw-samuraj:
@jendrikjoe :+1:

PR #199, opened 2018-11-26
Return ssh command output on error case

Etiene:
The output of running ssh commands was being thrown away on error situations.
This PR fixes that.

### DIFF
--- a/test/terraform_ssh_example_test.go
+++ b/test/terraform_ssh_example_test.go
@@ -126,6 +126,27 @@ func testSSHToPublicHost(t *testing.T, terraformOptions *terraform.Options, keyP
 
 		return "", nil
 	})
+
+	// Run a command on the server that results in an error,
+	expectedText = "Hello, World"
+	command = fmt.Sprintf("echo -n '%s' && exit 1", expectedText)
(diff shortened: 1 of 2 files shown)


Etiene:
I manually tested it because I was using this branch while working on google vault repo, and the correct string is returned. I can add the proper test here soon.

test/terraform_ssh_example_test.go
```diff
@@ -126,6 +126,23 @@ func testSSHToPublicHost(t *testing.T, terraformOptions *terraform.Options, keyP
...
+
+	// Run a command on the server that results in an error,
+	expectedText = "Hello, World"
+	command = fmt.Sprintf("echo -n '%s' && return 1", expectedText)
+	description = fmt.Sprintf("SSH to public host %s with error command", publicInstanceIP)
+
+	// Verify that we can SSH to the Instance, run the command and see the output
+	retry.DoWithRetry(t, description, maxRetries, timeBetweenRetries, func() (string, error) {
+
+		actualText, err := ssh.CheckSshCommandE(t, publicHost, command)
```
brikis98: Check that `err` is not `nil`.

PR #206, opened 2018-12-08
Expose k8s node api without test

yorinasub17:
I wanted to reuse some of the functions i wrote here in a non-test context, so reorganized some of the functions so that logging can happen outside of core functionality.


yorinasub17:
I have a whole bunch of changes now on a different branch that depends on this, so going to close this and open a new PR from there.

Issue #194, opened 2018-11-10
Add support for Kubernetes and Helm


yorinasub17:
In a similar vein to having the aws package with a number of utilities useful for AWS, Terratest should provide functionalities for interacting with Kubernetes and Helm for verification.

I will be working on the following functionalities as needed:

- [x] Open proxy to service in background, with ability to close it
- [x] Get helm releases
- [x] Apply helm chart
- [ ] Upgrade helm release
- [x] Delete helm release
- [x] Create a namespace
- [x] Delete YAML resource
- [x] Apply YAML resource to a namespace
- [x] Delete config cluster, user, and context
- [x] Get nodes in cluster
- [x] Get ready nodes in cluster
- [x] Wait for all nodes in cluster to be ready

yorinasub17:
I am going to close this now that we have the initial support for helm released: https://github.com/gruntwork-io/terratest/releases/tag/v0.13.28

PR #207, opened 2018-12-11
k8s enhancements

yorinasub17:
More functions for https://github.com/gruntwork-io/terratest/issues/194

- Expose some functions that can be used without `testing.T` to not take in `testing.T` so that they can be used outside testing context.
    - `GetPublicIpsOfEc2InstancesFromClientE`
    - `GetKubernetesClientFromFileE`
    - `GetNodesByClientE`

- Migrate `ContinuouslyCheckUrl` from module-ecs
- Expose `StoreConfigToFile` from the k8s module tests, as that can be used for storing a kubernetes resource config to disk to be used with `kubectl apply`
- Introduce wrapper functions for calling kubectl:
     - `RunKubectl`, `RunKubectlE`, `RunKubectlAndGetOutputE`
     - `KubectlApplyFromString`, `KubectlApplyFromStringE`
     - `KubectlDeleteFromString`, `KubectlDeleteFromStringE`

- Introduce functions for interacting with Kubernetes Service resources.

### DIFF
--- a/examples/kubernetes-basic-example/README.md
+++ b/examples/kubernetes-basic-example/README.md
@@ -1,0 +1,40 @@
+# Kubernetes Basic Example
+
+This folder contains a minimal Kubernetes resource config file to demonstrate how you can use Terratest to write
+automated tests for Kubernetes.
+
+This resource file deploys an nginx container as a single pod deployment with a node port service attached to it.
+
+See the corresponding terratest code for an example of how to test this resource config:
+- [kubernetes_basic_example_test.go](../test/kubernetes_basic_example_test.go) for the most basic verification
+- [kubernetes_basic_example_service_check_test.go](../test/kubernetes_basic_example_service_check_test.go) for a more
+  advanced version of checking the service.
+
+
+## Deploying the Kubernetes resource
+
+1. Setup a Kubernetes cluster. We recommend using a local version:
+    - [minikube](https://github.com/kubernetes/minikube)
+    - [Kubernetes on Docker For Mac](https://docs.docker.com/docker-for-mac/kubernetes/)
+    - [Kubernetes on Docker For Windows](https://docs.docker.com/docker-for-windows/kubernetes/)
+
+1. Install and setup [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) to talk to the deployed
+   Kubernetes cluster.
+1. Run `kubectl apply -f nginx-deployment.yml`
+
+
+## Running automated tests against this Kubernetes deployment
+
+1. Setup a Kubernetes cluster. We recommend using a local version:
+    - [minikube](https://github.com/kubernetes/minikube)
+    - [Kubernetes on Docker For Mac](https://docs.docker.com/docker-for-mac/kubernetes/)
+    - [Kubernetes on Docker For Windows](https://docs.docker.com/docker-for-windows/kubernetes/)
+
+1. Install and setup [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) to talk to the deployed
+   Kubernetes cluster.
+1. Install and setup [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
+1. Install [Golang](https://golang.org/) and make sure this code is checked out into your `GOPATH`.
+1. `cd test`
+1. `dep ensure`
+1. `go test -v -run TestKubernetesBasicExample`
+1. You can also run `TestKubernetesBasicExampleServiceCheck`
--- a/modules/aws/region.go
+++ b/modules/aws/region.go
@@ -44,8 +44,13 @@ var stableRegions = []string{
 // those that have been around for at least 1 year.
 // Note that regions in the approvedRegions list that are not considered stable are ignored.
 func GetRandomStableRegion(t *testing.T, approvedRegions []string, forbiddenRegions []string) string {
-	regionsToPickFrom := collections.ListIntersection(stableRegions, approvedRegions)
-	regionsToPickFrom = collections.ListSubtract(regionsToPickFrom, forbiddenRegions)
+	regionsToPickFrom := stableRegions
+	if len(approvedRegions) > 0 {
+		regionsToPickFrom = collections.ListIntersection(regionsToPickFrom, approvedRegions)
+	}
+	if len(forbiddenRegions) > 0 {
+		regionsToPickFrom = collections.ListSubtract(regionsToPickFrom, forbiddenRegions)
+	}
 	return GetRandomRegion(t, regionsToPickFrom, nil)
 }
 
--- a/modules/aws/region.go
+++ b/modules/aws/region.go
@@ -20,6 +20,35 @@ const regionOverrideEnvVarName = "TERRATEST_REGION"
 // this region as a default.
 const defaultRegion = "us-east-1"
 
+// Reference for launch dates: https://aws.amazon.com/about-aws/global-infrastructure/
+var stableRegions = []string{
+	"us-east-1",      // Launched 2006
+	"us-east-2",      // Launched 2016
+	"us-west-1",      // Launched 2009
+	"us-west-2",      // Launched 2011
+	"ca-central-1",   // Launched 2016
+	"sa-east-1",      // Launched 2011
+	"eu-west-1",      // Launched 2007
+	"eu-west-2",      // Launched 2016
+	"eu-west-3",      // Launched 2017
+	"eu-central-1",   // Launched 2014
+	"ap-southeast-1", // Launched 2010
+	"ap-southeast-2", // Launched 2012
+	"ap-northeast-1", // Launched 2011
+	"ap-northeast-2", // Launched 2016
+	"ap-south-1",     // Launched 2016
+}
+
+// GetStableRandomRegion gets a randomly chosen AWS region that is considered stable. Like GetRandomRegion, you can
+// further restrict the stable region list using approvedRegions and forbiddenRegions. We consider stable regions to be
+// those that have been around for at least 1 year.
+// Note that regions in the approvedRegions list that are not considered stable are ignored.
+func GetRandomStableRegion(t *testing.T, approvedRegions []string, forbiddenRegions []string) string {
+	regionsToPickFrom := collections.ListIntersection(stableRegions, approvedRegions)
+	regionsToPickFrom = collections.ListSubtract(regionsToPickFrom, forbiddenRegions)
+	return GetRandomRegion(t, regionsToPickFrom, nil)
+}
+
 // GetRandomRegion gets a randomly chosen AWS region. If approvedRegions is not empty, this will be a region from the approvedRegions
 // list; otherwise, this method will fetch the latest list of regions from the AWS APIs and pick one of those. If
 // forbiddenRegions is not empty, this method will make sure the returned region is not in the forbiddenRegions list.
--- a/modules/http-helper/continuous.go
+++ b/modules/http-helper/continuous.go
@@ -29,8 +29,12 @@ func ContinuouslyCheckUrl(
 				statusCode, body, err := HttpGetE(t, url)
 				logger.Logf(t, "Got response %d and err %v from URL at %s", statusCode, err, url)
 				if err != nil {
+					// We use Errorf instead of Fatalf here because Fatalf is not goroutine safe, while Errorf is. Refer
+					// to the docs on `T`: https://godoc.org/testing#T
 					t.Errorf("Failed to make HTTP request to the URL at %s: %s\n", url, err.Error())
 				} else if statusCode != 200 {
+					// We use Errorf instead of Fatalf here because Fatalf is not goroutine safe, while Errorf is. Refer
+					// to the docs on `T`: https://godoc.org/testing#T
 					t.Errorf("Got a non-200 response (%d) from the URL at %s, which means there was downtime! Response body: %s", statusCode, url, body)
 				}
 			}
--- a/modules/http-helper/continuous.go
+++ b/modules/http-helper/continuous.go
@@ -8,18 +8,27 @@ import (
 	"github.com/gruntwork-io/terratest/modules/logger"
 )
 
+type GetResponse struct {
+	StatusCode int
+	Body       string
+}
+
 // Continuously check the given URL every 1 second until the stopChecking channel receives a signal to stop.
-// This function will return a sync.WaitGroup that can be used to wait for the checking to stop.
+// This function will return a sync.WaitGroup that can be used to wait for the checking to stop, and a read only channel
+// to stream the responses for each check.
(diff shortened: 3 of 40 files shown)


brikis98:
> Expose some functions that can be used without testing.T to not take in testing.T so that they can be used outside testing context.

What's the use case? Do we really want production code depending on a test library?

> Migrate ContinuouslyCheckUrl from module-ecs

Nice 👍

modules/k8s/config.go
```diff
@@ -23,6 +24,19 @@ func LoadConfigFromPath(path string) clientcmd.ClientConfig {
 	return config
 }
 
+// LoadApiClientConfig will load a ClientConfig object from a file path that points to a location on disk containing a
+// kubectl config, with the requested context loaded.
+func LoadApiClientConfig(path string, context string) (*restclient.Config, error) {
```
brikis98: `LoadApiClientConfigE`

modules/k8s/kubectl.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	defer os.Remove(tmpfile)
+	return RunKubectlE(t, options, "apply", "-f", tmpfile)
+}
+
+// StoreConfigToTempFile will store the provided config data to a temporary file created on the os and return the
+// filename.
+func StoreConfigToTempFile(t *testing.T, configData string) string {
+	out, err := StoreConfigToTempFileE(t, configData)
+	require.NoError(t, err)
```
brikis98: That is cleaner. Will make mental note to use this elsewhere in Terratest in the future.

modules/k8s/service.go
```diff
@@ -0,0 +1,211 @@
...
+func GetServiceFromClientE(clientset *kubernetes.Clientset, namespace string, serviceName string) (*corev1.Service, error) {
+	return clientset.CoreV1().Services(namespace).Get(serviceName, metav1.GetOptions{})
+}
+
+// WaitUntilServiceAvailable waits until the service endpoint is ready to accept traffic.
+func WaitUntilServiceAvailable(t *testing.T, namespace string, serviceName string, retries int, sleepBetweenRetries time.Duration) {
+	clientset, err := GetKubernetesClientE(t)
+	require.NoError(t, err)
+	statusMsg := fmt.Sprintf("Wait for service %s to be provisioned.", serviceName)
+	message, err := retry.DoWithRetryE(
```
brikis98: Could use the non "E" version here

modules/k8s/service.go
```diff
@@ -0,0 +1,211 @@
...
+// IsServiceAvailable returns true if the service endpoint is ready to accept traffic.
+func IsServiceAvailable(service *corev1.Service) bool {
+	// Only the LoadBalancer type has a delay. All other service types are available if the resource exists.
+	switch service.Spec.Type {
+	case corev1.ServiceTypeLoadBalancer:
+		ingress := service.Status.LoadBalancer.Ingress
+		// The load balancer is ready if it has at least one ingress point
+		return len(ingress) > 0
+	default:
+		return true
```
brikis98: Shouldn't the default be `false`?

modules/k8s/service.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// pickRandomNode will pick a random node in the kubernetes cluster
+func pickRandomNode(clientset *kubernetes.Clientset) (corev1.Node, error) {
+	nodes, err := GetNodesByClientE(clientset, metav1.ListOptions{})
+	if err != nil {
+		return corev1.Node{}, err
+	}
+	if len(nodes) == 0 {
+		return corev1.Node{}, NewKubernetesError("There are no nodes in the cluster")
```
brikis98: What's the advantage of a single `NewKubernetesError` used throughout this code? It doesn't make it easy to type check for specific errors, unlike, for example, a `NoNodesInCluster` type that is specific to this exact error here.

modules/k8s/service.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// findAwsNodeHostname will return the public ip of the node, assuming the node is an AWS EC2 instance.
+// If the instance does not have a public IP, will return the internal hostname as recorded on the Kubernetes node
+// object.
+func findAwsNodeHostname(node corev1.Node, awsIDUri *url.URL) (string, error) {
+	// Path is /AVAILABILITY_ZONE/INSTANCE_ID
+	parts := strings.Split(awsIDUri.Path, "/")
+	instanceID := parts[2]
+	availabilityZone := parts[1]
```
brikis98: Check length of `parts`

yorinasub17:
> What's the use case? Do we really want production code depending on a test library?

While I agree this is a bit awkward, the wrappers we built around the clouds end up being useful abstractions for writing any code that works against the underlying systems.

For example, in the rollout command I ended up doing a lot of the same things I would do in a test context: waiting for nodes to be ready, initializing clients by config files, manipulating config files, calling `kubectl`. And I really wanted to avoid copy pasting the code and duplicating the tests.

I guess the most ideal situation would be to break out the cloud verification code into its own repo that then prod code and terratest can import, but I wasn't quite ready to shave the yak of figuring out the right abstraction, and also cause a lot of disruption.

modules/http-helper/continuous.go
```diff
@@ -0,0 +1,29 @@
...
+		for {
+			select {
+			case <-stopChecking:
+				logger.Logf(t, "Got signal to stop downtime checks for URL %s.\n", url)
+				return
+			case <-time.After(sleepBetweenChecks):
+				statusCode, body, err := HttpGetE(t, url)
+				logger.Logf(t, "Got response %d and err %v from URL at %s", statusCode, err, url)
+				if err != nil {
+					t.Fatalf("Failed to make HTTP request to the URL at %s: %s\n", url, err.Error())
```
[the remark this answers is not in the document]
yorinasub17: Yes this works. It's been causing my tests to fail (and me pulling hair) the past 2 business days 😄 Also, I copy pasted this with no modifications out of `module-ecs`.

brikis98:
> For example, in the rollout command I ended up doing a lot of the same things I would do in a test context: waiting for nodes to be ready, initializing clients by config files, manipulating config files, calling kubectl. And I really wanted to avoid copy pasting the code and duplicating the tests.

Sanity check: is this code _really_ the same or just _somewhat_ the same. I worry that Terratest may update a lot of these functions over time to be more and more suited for testing, with little regard for potential prod uses of this code... And if it's not too much code, sometimes, trying to keep things perfectly DRY can be counter-productive.

modules/http-helper/continuous.go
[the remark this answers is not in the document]
brikis98: Hmm...

https://github.com/ipfs/go-ipfs/issues/2043
https://github.com/golang/go/issues/24678
https://github.com/golang/go/issues/15758

modules/k8s/service.go
yorinasub17: See comment at beginning of function:
> Only the LoadBalancer type has a delay. All other service types are available if the resource exists.

modules/k8s/service.go
yorinasub17: That's a good point. Will fix.

yorinasub17:
> is this code really the same or just somewhat the same.

The kubectl wrappers (`RunKubectl`) are exactly the same right now. Getting and waiting for nodes are also exactly the same. The AWS stuff is more questionable.

> I worry that Terratest may update a lot of these functions over time to be more and more suited for testing, with little regard for potential prod uses of this code...

Those are valid points, although as mentioned, it would be nice if they were isolated in a separate library to avoid that... but then there is overhead ⬇️ 

> And if it's not too much code, sometimes, trying to keep things perfectly DRY can be counter-productive.


Ok I will revert those and copy paste and see how much duplication it ends up being, and if overtime it really does get out of hand, we can discuss merging then.

modules/k8s/config.go
yorinasub17: Fixed

modules/k8s/service.go
yorinasub17: Fixed

modules/k8s/service.go
yorinasub17: Added check

yorinasub17:
- Reverted the "FromClient" versions
- Addressed all suggested changes:
    - Added more unique error types
    - Use the `E` suffix convention

modules/http-helper/continuous.go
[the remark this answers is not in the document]
yorinasub17: Ah ok I see. So the bug is that there are race conditions in there. Ok I'll see what I can do to fix this.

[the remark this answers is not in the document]
brikis98: Ah, nice find! You should add a comment that indicates why we specifically use `Errorf` to each spot in the code so we don't accidentally break that.

One other way to do this would be for the method to return a channel and to write to that channel an error if something went wrong or `nil` if everything finished successfully. The calling code could then read from that channel to see how things went. But `Errorf` is probably good enough for now that it's not worth adding this complexity.

yorinasub17:
UPDATES:
- c7853c2 - Made `ContinuouslyCheckUrl` use `Errorf`. It will also now return a `sync.WaitGroup` that can be used to wait for the continuous check to shutdown.
- ad2c714 - Found and fixed a bug in `DummyServer`, where we try to register multiple handlers to route `/` on the `DefaultServeMux`. Caught this when I wrote a test for `ContinuouslyCheckUrl`.
- 4f285bf - Added more k8s functionalities, related to dealing with namespaces. Also changed the function APIs to take in a `KubectlOptions` object so that you can control which context to use as well as which namespace.
- 8d4160d - Added a basic k8s example and test

Currently, I am waiting to get a green build.

yorinasub17:
UPDATE:
- 0091cc6 - Introduce `GetRandomStableRegion` for `aws`, which only selects from regions that have been around for more than a year. The list is hardcoded for now, but added the reference so that it can be evaluated on a yearly basis. This way, we won't suddenly have a bunch of test failures everytime a new region pops up.

examples/kubernetes-basic-example/README.md
```diff
@@ -0,0 +1,40 @@
+# Kubernetes Basic Example
```
brikis98: Nice 👍

brikis98: BTW, you may want to add to our newsletter that Terratest now supports automated testing for Kubernetes as a top-level item. SEO gold.

modules/aws/region.go
```diff
@@ -20,6 +20,35 @@ const regionOverrideEnvVarName = "TERRATEST_REGION"
 // this region as a default.
 const defaultRegion = "us-east-1"
 
+// Reference for launch dates: https://aws.amazon.com/about-aws/global-infrastructure/
```
brikis98: Nice addition 👍

brikis98: BTW, is there any easy way to add a yearly reminder to update this? Something in GitHub? Calendar invite?

modules/http-helper/dummy_server.go
```diff
@@ -28,7 +28,9 @@ func RunDummyServer(t *testing.T, text string) (net.Listener, int) {
 func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
 	port := getNextPort()
 
-	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+	// Create new serve mux so that multiple handlers can be created
+	server := http.NewServeMux()
+	server.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
```
brikis98: Nice fix. I could've sworn we did this already at some point... But it must've been in some other code base...

modules/http-helper/dummy_server_test.go
```diff
@@ -22,6 +23,25 @@ func TestRunDummyServer(t *testing.T) {
...
+
+	listener, port := RunDummyServer(t, text)
+
+	url := fmt.Sprintf("http://localhost:%d", port)
+	wg := ContinuouslyCheckUrl(t, url, stopChecking, 1*time.Second)
+	defer func() {
+		stopChecking <- true
+		wg.Wait()
+		shutDownServer(t, listener)
+	}()
```
brikis98: In theory, this test could pass if `ContinuouslyCheckUrl` was a noop... Worth adding a negative test that takes the server down part of the way through and checks that a failure happened? This might be tricky to do given that `ContinuouslyCheckUrl` calls `t.Errorf` explicitly... Unless you pass it a mock `t`...

modules/k8s/errors.go
```diff
@@ -0,0 +1,84 @@
+package k8s
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ServiceNotAvailable is returned when a Kubernetes service is not yet available to accept traffic.
+type ServiceNotAvailable struct {
```
brikis98: These are much clearer errors, thanks 👍

test/kubernetes_basic_example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	// At the end of the test, run `kubectl delete -f RESOURCE_CONFIG` to clean up any resources that were created.
+	defer k8s.KubectlDelete(t, options, kubeResourcePath)
+
+	// This will run `kubectl apply -f RESOURCE_CONFIG` and fail the test if there are any errors
+	k8s.KubectlApply(t, options, kubeResourcePath)
+
+	// This will get the service resource and verify that it exists and was retrieved successfully. This function will
+	// fail the test if the there is an error retrieving the service resource from Kubernetes.
+	service := k8s.GetService(t, options, "nginx-service")
+	require.Equal(t, service.Name, "nginx-service")
```
brikis98: Does this test actually try to connect to nginx?

examples/kubernetes-basic-example/README.md
yorinasub17: Yup was thinking of doing that as soon as this merges

modules/aws/region.go
yorinasub17: Set a reminder in slack for now. Used 1/23 every year.

test/kubernetes_basic_example_test.go
yorinasub17: This one, no. I wanted to keep this example as basic as possible. There is another test that will actually connect and check nginx: `kubernetes_basic_example_service_check_test.go`. You can see there that there is some added complexity from getting the endpoint and verifying it.

modules/http-helper/dummy_server_test.go
yorinasub17: Ok what do you think of? 4b54663 Not perfect, but I think is slightly better.

I know we can do the error channel instead like you suggested above, but I like the fact that the test will fail without any additional logic if there is one request that failed in the current model.

brikis98: LGTM!

yorinasub17:
Ok build passed, and review checked off. Going to merge, release, and start drafting an entry in the newsletter!

PR #210, opened 2018-12-13
Stream stdout and stderr better

yorinasub17:
The current method will not interleave stdout and stderr. Additionally, if a command only outputs to stderr, then this will not emit any logs until the command finishes, because it will block on `stdoutScanner.Scan()`.

The solution proposed here is to use goroutines for reading the data from each output and wait until they terminate.

### DIFF
--- a/modules/shell/command.go
+++ b/modules/shell/command.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 	"syscall"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


yorinasub17:
The main difference is that we don't have writer objects in terratest, since `logger.Log` is a normal function. To use multiwriter, we need to either create a writer object that will translate the standard write methods to `logger.Log`, or create a pipe and use a goroutine/function that reads off the reader to feed to `logger.Log`. Both seemed like it will add more complexity, so I opted for this goroutine approach. We could just pass straight to stdout and stderr, but then we lose the logging prefix which is important for `terratest_log_parser` to function.

modules/shell/command.go
```diff
@@ -115,6 +109,17 @@ func readStdoutAndStderr(t *testing.T, stdout io.ReadCloser, stderr io.ReadClose
...
 }
 
+func readData(t *testing.T, scanner *bufio.Scanner, wg *sync.WaitGroup, mutex *sync.Mutex, allOutput *[]string) {
+	defer wg.Done()
+	for scanner.Scan() {
+		text := scanner.Text()
+		logger.Log(t, text)
+		mutex.Lock()
+		*allOutput = append(*allOutput, text)
+		mutex.Unlock()
```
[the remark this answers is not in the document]
yorinasub17: I feel like this is going to bite us later so I will just extract this bit into a separate function.

modules/shell/command_test.go
```diff
@@ -19,3 +23,72 @@ func TestRunCommandAndGetOutput(t *testing.T) {
...
+	bashCode := fmt.Sprintf(`
+echo_stderr(){
+	sleep .$[ ( $RANDOM %% 10 ) + 1 ]s
+	(>&2 echo "%s")
+}
+echo_stdout(){
+	sleep .$[ ( $RANDOM %% 10 ) + 1 ]s
+	echo "%s"
+}
+for i in {1..5}
```
brikis98: Increase from 5 to something way bigger (500?). Make `sleep` short enough so the test doesn't take forever.

yorinasub17: Done

yorinasub17: Testing time was 2 seconds on my mac.

yorinasub17:
Ok build passed, so merging and releasing!

modules/shell/command_test.go
brikis98: Haha, OK, that's probably fast enough 😁

Issue #211, opened 2018-12-13
Function to get private IP of EC2 instances


drmmarsunited:
Hello!  I'd like to contribute some code functions to retrieve the `PrivateIpAddress` from a given EC2 instance.  In my use case, I launch nodes into an AWS ASG without public IPs and no ALB/ELB up front.  However, I have services that should be exposed via HTTP/HTTPS on these nodes that should be reachable from my test instance over direct connect.

Happy to contribute the code myself!  Didn't see any other open issues around this, but please let me know if there are.

drmmarsunited:
In preparation for carrying out the work, I don't see a `_test` file for `ec2.go` in the *aws* module.  Where are the tests (if there are any) for those functions?  Sorry if I'm missing something terribly obvious!

brikis98:
Sounds like a great addition!

We have tests in two places:

1. `_test.go` files. These are mostly "unit tests" (they're not purely unit tests in many cases, as this is infra code, but close enough).

1. Tests in the `/test` folder. These are automated tests for the examples in the `/examples` folder. We try to have as many of these as they are effectively executable documentation that shows how Terratest is meant to be used—and tests Terratest at the same time!

drmmarsunited:
Worth an out-loud brainstorm:

Incorporating an example of my use case wouldn't necessarily work, given that it's dependent upon routing setups of the network in which the test instance relies on.  I was trying to think of a simple way to incorporate a test that would run publicly without much fuss, and the simplest one I could conjure up was:

1. Modify the basic HTTP Terraform example to output the instance's `private_ip` attribute
2. Within the `_test.go` file, run the newly created `GetPrivateIpOfEc2Instance` function and compare it to the output attribute to ensure a match

The inverse to that was to write a negative test to ensure that attempting to connect to the private IP of an instance over SSH and/or an HTTP port would fail, but this feels wrong given that it *could* be possible with the right network setup, and I don't want to confuse Terratest users.

Thoughts?

brikis98:
The SSH code in Terratest can use a "jump host": https://github.com/gruntwork-io/terratest/blob/master/modules/ssh/ssh.go#L237. So, one option:

1. Launch 2 EC2 instances
1. Get the public IP of one
1. Get the private IP of the other
1. SSH to the private IP using the public IP as a jump host

drmmarsunited:
Will do.  Thanks.

drmmarsunited:
PR out: #212

[...]

PR #212, opened 2018-12-14
Adding GetPrivateIpInstance(s) AWS helper functions

drmmarsunited:
Issue: https://github.com/gruntwork-io/terratest/issues/211
Test Pass Gist: https://gist.github.com/drmmarsunited/d95744bbc049d040b1b17b3c90eda796

This PR adds the following helper functions to the `ec2.go` file in the `aws` module:

- GetPrivateIpOfEc2Instance
- GetPrivateIpOfEc2InstanceE
- GetPrivateIpsOfEc2Instances
- GetPrivateIpsOfEc2InstancesE

It also updates the [existing tests](https://github.com/gruntwork-io/terratest/blob/master/test/terraform_ssh_example_test.go) for the basic SSH example to use the newly created `GetPrivateIpOfEc2Instance` function to grab the private instance's IP address, instead of using the output from Terraform.

There was also a reversal of the public/private IP order in the string interpolation, where the public instance's IP address was being written into a string where the private instance's IP should have gone.  I've updated that as well.

### DIFF
--- a/test/terraform_ssh_example_test.go
+++ b/test/terraform_ssh_example_test.go
@@ -154,7 +154,9 @@ func testSSHToPrivateHost(t *testing.T, terraformOptions *terraform.Options, key
 	publicInstanceIP := terraform.Output(t, terraformOptions, "public_instance_ip")
 
 	// Get IP of private instance from AWS helper function instead of Terraform output
-	privateInstanceIP := aws.GetPrivateIpOfEc2Instance(t, terraform.Output(t, terraformOptions, "private_instance_id"), terraformOptions.Vars["aws_region"].(string))
+	privateInstanceID := terraform.Output(t, terraformOptions, "private_instance_id")
+	deployedAWSRegion := terraformOptions.Vars["aws_region"].(string)
+	privateInstanceIP := aws.GetPrivateIpOfEc2Instance(t, privateInstanceID, deployedAWSRegion)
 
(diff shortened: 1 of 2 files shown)


test/terraform_ssh_example_test.go
```diff
@@ -152,7 +152,9 @@ func testSSHToPublicHost(t *testing.T, terraformOptions *terraform.Options, keyP
 func testSSHToPrivateHost(t *testing.T, terraformOptions *terraform.Options, keyPair *aws.Ec2Keypair) {
 	// Run `terraform output` to get the value of an output variable
 	publicInstanceIP := terraform.Output(t, terraformOptions, "public_instance_ip")
-	privateInstanceIP := terraform.Output(t, terraformOptions, "private_instance_ip")
+
+	// Get IP of private instance from AWS helper function instead of Terraform output
+	privateInstanceIP := aws.GetPrivateIpOfEc2Instance(t, terraform.Output(t, terraformOptions, "private_instance_id"), terraformOptions.Vars["aws_region"].(string))
```
brikis98: Could you break this up into a few lines? One to read the `instanceId`, one to read the `awsRegion`, and then one to get the `privateInstanceIP` using the previous two.

drmmarsunited: Sure

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.13.18

Issue #208, opened 2018-12-12
function to check multiple outputs with just one output command


deeptirev:
I'd like to be able to get outputs for multiple keys without having to run output command multiple times. Is there provision to do this? Or even getting the entire output blob in json format would help.

PR #213, opened 2018-12-18
PR for Issue #208 - Returning multiple outputs

deeptirev:
PR to address issue #208 
Added 2 functions : 
- Function to return all output as map[string]interface{}
- Function to return outputs for requested keys as map[string]interface{}

Tests updated as well.


brikis98:
@Etiene Don't forget to release and add to newsletter!

PR #214, opened 2018-12-19
PrivateHostname of EC2 Instance + CapacityInfo of ASG

yorinasub17:
I had a need for getting the PrivateHostname of the EC2 instances in EKS world, because the kubernetes node names default to the private DNS name of the worker nodes.

I also had a need to get capacity information of ASGs when testing EKS rolling deployments.

Both functions seemed like they belong in terratest, so added them here.

### DIFF
--- a/modules/aws/asg.go
+++ b/modules/aws/asg.go
@@ -1,11 +1,56 @@
 package aws
 
 import (
+	"fmt"
 	"testing"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/stretchr/testify/require"
+
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/retry"
 )
+
+type AsgCapacityInfo struct {
+	MinCapacity     int64
+	MaxCapacity     int64
+	CurrentCapacity int64
+	DesiredCapacity int64
+}
+
+// GetCapacityInfoForAsg returns the capacity info for the queried asg as a struct, AsgCapacityInfo.
+func GetCapacityInfoForAsg(t *testing.T, asgName string, awsRegion string) AsgCapacityInfo {
+	capacityInfo, err := GetCapacityInfoForAsgE(t, asgName, awsRegion)
+	require.NoError(t, err)
+	return capacityInfo
+}
+
+// GetCapacityInfoForAsgE returns the capacity info for the queried asg as a struct, AsgCapacityInfo.
+func GetCapacityInfoForAsgE(t *testing.T, asgName string, awsRegion string) (AsgCapacityInfo, error) {
+	asgClient, err := NewAsgClientE(t, awsRegion)
+	if err != nil {
+		return AsgCapacityInfo{}, err
+	}
+
+	input := autoscaling.DescribeAutoScalingGroupsInput{AutoScalingGroupNames: []*string{aws.String(asgName)}}
+	output, err := asgClient.DescribeAutoScalingGroups(&input)
+	if err != nil {
+		return AsgCapacityInfo{}, err
+	}
+	groups := output.AutoScalingGroups
+	if len(groups) == 0 {
+		return AsgCapacityInfo{}, NewNotFoundError("ASG", asgName)
+	}
+	capacityInfo := AsgCapacityInfo{
+		MinCapacity:     *groups[0].MinSize,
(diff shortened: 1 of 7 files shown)


modules/aws/asg.go
```diff
@@ -1,12 +1,57 @@
...
+		return AsgCapacityInfo{}, err
+	}
+
+	input := autoscaling.DescribeAutoScalingGroupsInput{AutoScalingGroupNames: []*string{aws.String(asgName)}}
+	output, err := asgClient.DescribeAutoScalingGroups(&input)
+	if err != nil {
+		return AsgCapacityInfo{}, err
+	}
+	groups := output.AutoScalingGroups
+	if len(groups) == 0 {
```
brikis98: Is it possible `len(groups) > 1`?

modules/aws/asg_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/gruntwork-io/terratest/modules/random"
+)
+
+func TestGetCapacityInfoForAsg(t *testing.T) {
```
brikis98: Do we have an ASG example in the `?examples` folder? If so, perhaps add these tests to the corresponding tests in the `/tests` folder so users can discover this functionality more easily?

modules/aws/ec2.go
```diff
@@ -63,6 +63,63 @@ func GetPrivateIpsOfEc2InstancesE(t *testing.T, instanceIDs []string, awsRegion
...
+	}
+
+	return hostname, nil
+}
+
+// GetPrivateHostnamesOfEc2Instances gets the private IP address of the given EC2 Instance in the given region. Returns a map of instance ID to IP address.
+func GetPrivateHostnamesOfEc2Instances(t *testing.T, instanceIDs []string, awsRegion string) map[string]string {
+	ips, err := GetPrivateHostnamesOfEc2InstancesE(t, instanceIDs, awsRegion)
+	if err != nil {
+		t.Fatal(err)
```
brikis98: `require.NoError`?

modules/aws/ec2.go
```diff
@@ -63,6 +63,63 @@ func GetPrivateIpsOfEc2InstancesE(t *testing.T, instanceIDs []string, awsRegion
...
+	}
+	return ip
+}
+
+// GetPrivateHostnameOfEc2InstanceE gets the private IP address of the given EC2 Instance in the given region.
+func GetPrivateHostnameOfEc2InstanceE(t *testing.T, instanceID string, awsRegion string) (string, error) {
+	hostnames, err := GetPrivateHostnamesOfEc2InstancesE(t, []string{instanceID}, awsRegion)
+	if err != nil {
+		return "", err
```
brikis98: `require.NoError`?

modules/aws/ec2.go
```diff
@@ -63,6 +63,63 @@ func GetPrivateIpsOfEc2InstancesE(t *testing.T, instanceIDs []string, awsRegion
...
+}
+
+// GetPrivateHostnamesOfEc2InstancesE gets the private IP address of the given EC2 Instance in the given region. Returns a map of instance ID to IP address.
+func GetPrivateHostnamesOfEc2InstancesE(t *testing.T, instanceIDs []string, awsRegion string) (map[string]string, error) {
+	ec2Client, err := NewEc2ClientE(t, awsRegion)
+	if err != nil {
+		return nil, err
+	}
+	input := ec2.DescribeInstancesInput{InstanceIds: aws.StringSlice(instanceIDs)}
+	output, err := ec2Client.DescribeInstances(&input)
```
brikis98: Is this a paginated API?

modules/aws/asg.go
yorinasub17: No, because we only request one AutoScalingGroupName and the ASG name is unique per region.

modules/aws/ec2.go
yorinasub17: Yes, but this is actually a copy paste job so all the similar methods (get private/public ip) have this problem. I'll add a global comment to mention that we need pagination here, although I suspect that most people using terratest won't hit the limit ([1000](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#DescribeInstancesInput)).

yorinasub17:
Merging and releasing

PR #217, opened 2018-12-21
Don't depend on terragrunt

yorinasub17:


yorinasub17:
Yes. Right now, there is no binary for the latest release (https://github.com/gruntwork-io/terragrunt/releases/tag/v0.17.4) so everything that doesn't specify a terragrunt version is broken right now.

PR #218, opened 2019-01-10
Functions for interacting with pods

yorinasub17:
### Modules affected

* `k8s`

### Description

New functions added:

* `GetPod` and `GetPodE`: Look up a pod by name and return it.
* `WaitUntilPodAvailable` and `WaitUntilPodAvailableE`: Wait for the pod to be running on the cluster.

### DIFF
--- a/modules/k8s/pod.go
+++ b/modules/k8s/pod.go
@@ -30,10 +30,17 @@ func GetPodE(t *testing.T, options *KubectlOptions, podName string) (*corev1.Pod
 	return clientset.CoreV1().Pods(options.Namespace).Get(podName, metav1.GetOptions{})
 }
 
-// WaitUntilPodAvailable waits until the pod is running.
+// WaitUntilPodAvailable waits until the pod is running, retrying the check for the specified amount of times, sleeping
+// for the provided duration between each try. This will fail the test if there is an error or if the check times out.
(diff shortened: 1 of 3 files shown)


modules/k8s/pod.go
```diff
@@ -0,0 +1,58 @@
...
+func GetPodE(t *testing.T, options *KubectlOptions, podName string) (*corev1.Pod, error) {
+	clientset, err := GetKubernetesClientFromOptionsE(t, options)
+	if err != nil {
+		return nil, err
+	}
+	return clientset.CoreV1().Pods(options.Namespace).Get(podName, metav1.GetOptions{})
+}
+
+// WaitUntilPodAvailable waits until the pod is running.
+func WaitUntilPodAvailable(t *testing.T, options *KubectlOptions, podName string, retries int, sleepBetweenRetries time.Duration) {
```
brikis98: Should there be an `E` version of this method?

yorinasub17: Ah good catch. I had intended to add it in and forgot. Added it in.

yorinasub17:
Addressed the comment. Waiting on GCP fixes before merging.

yorinasub17:
Test passing now, so merging and will release!

PR #221, opened 2019-01-14
Fix curl based terratest_log_parser installation instructions

yorinasub17:
`curl` does not follow redirects by default, so we need to pass the option `-L`. Addresses https://github.com/gruntwork-io/terratest/issues/220

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ to install version 0.13.13 of `terratest_log_parser`:
 ```bash
 # This example assumes a linux 64bit machine
 # Use curl to download the binary
-curl -Lo terratest_log_parser https://github.com/gruntwork-io/terratest/releases/download/v0.13.13/terratest_log_parser_linux_amd64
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,7 +89,7 @@ to install version 0.13.13 of `terratest_log_parser`:
 ```bash
 # This example assumes a linux 64bit machine
 # Use curl to download the binary
-curl -o terratest_log_parser https://github.com/gruntwork-io/terratest/releases/download/v0.13.13/terratest_log_parser_linux_amd64
+curl -Lo terratest_log_parser https://github.com/gruntwork-io/terratest/releases/download/v0.13.13/terratest_log_parser_linux_amd64
```
brikis98: We typically use:

```
curl --location --silent --fail --show-error <URL>
```

That seems to be a good mix of handling redirects and errors and showing error messages but not other noise.

Issue #219, opened 2019-01-13
Add a cleanup state function


innovia:
When I run a random S3 bucket creation with random region i get a bad state for consequent runs

how do i delete the state from terratest before the test run.

brikis98:
> When I run a random S3 bucket creation with random region i get a bad state for consequent runs

What do you mean by bad state? What's the error?

innovia:
@brikis98  @bik
Hi Yevgeni first of all the error is because of you! you made me watch the 300,000 lines of code lessons and got me started refactoring my code. :-p

So the error is a 403 as you can see by the logs below, and I beleive its because of a region mismatch once a region was init on the provider. I'm probably doing something wrong here...

anyhow here's my setup:

I have a provider that set the region up

```
provider "aws" {
  region  = "${var.region}"
  version = "~> 1.29"
}
```

then i have my s3 resource:

```
resource "aws_s3_bucket" "s3_bucket_standard" {
  count         = "${var.bucket_type == "standard" ? 1 : 0}"
  bucket        = "${local.bucket_name}"
  region        = "${var.region}"
  force_destroy = "${var.force_destroy}"
  acl           = "private"

  versioning = {
    enabled = "${var.versioning_enabled}"
  }

  tags = "${local.tags}"
}
```

and my test:

```golang
import (
	"fmt"

	"github.com/gruntwork-io/terratest/modules/aws"
	"github.com/gruntwork-io/terratest/modules/random"
	"github.com/gruntwork-io/terratest/modules/terraform"
	"testing"
)

// TestS3StandardBucketCreate
func TestS3StandardBucketCreate(t *testing.T) {
	// CleanUpState(t)
	t.Parallel()

	// Give this bucket a unique id.
	bucketRandomName := fmt.Sprintf("terratest-s3-bucket-%x", random.UniqueId())

	// Pick a random AWS region to test in. This helps ensure your code works in all regions.
	awsRegion := aws.GetRandomStableRegion(t, nil, nil)

	terraformOptions := &terraform.Options{
		// The path to where our Terraform code is located
		TerraformDir: "../modules/aws/s3/bucket",

		// Variables to pass to our Terraform code using -var options
		Vars: map[string]interface{}{
			"cloudfront_bucket": false,
			"organization":      "org",
			"name":              bucketRandomName,
			"region":            awsRegion,
			"service_name":      "tf-tests",
			"product_name":      "terraform",
			"force_destroy":     true,
		},

[...]

PR #222, opened 2019-01-16
Improve documentation for test_structure.CopyTerraformFolderToTemp

yorinasub17:
Context is https://github.com/gruntwork-io/terratest/issues/219.

I know we have examples of usage of this, but it is dug deep in the test stages example and it isn't immediately obvious how to extract it out, so I decided to embed an example in the docs.

Also, I couldn't come up with a use case where we don't want to log the new temp folder, so added in the log in the function.


yorinasub17:
Ok merging and releasing!

PR #229, opened 2019-01-25
Extend aws/cloudwatch module

whage:
Added `NewCloudWatchClient` and `NewCloudWatchClientE`methods to the `cloudwatch` module.

The AWS go sdk provides the `cloudwatch` service for manipulating Alarms, which is different from `cloudwatchlogs`.
The new methods allow you to get a `cloudwatch` client.

Not sure if it is worth adding to the code, feel free to decline if not.

### DIFF
--- a/modules/aws/cloudwatch.go
+++ b/modules/aws/cloudwatch.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


yorinasub17:
Thanks for your contribution! The additions make sense.

Before accepting the submission, we need some tests to help maintain the functionality going forward. Can you come up with a unit test and/or integration test (the tests in the `/test` folder) that tests the functionality?

PR #230, opened 2019-01-29
[WIP] Add Support for Authenticating against GKE clusters

robmorgan:
This PR allows the Kubernetes module to authenticate correctly against GKE clusters by loading the GCP plugin.

Inspired by: https://github.com/kubernetes/client-go/blob/53c7adfd0294caa142d961e1f780f74081d5b15f/examples/out-of-cluster-client-configuration/main.go#L31

Reference: https://github.com/kubernetes/client-go/issues/242

### TODO
 - [x] Currently the tests are failing

### DIFF
--- a/modules/k8s/node.go
+++ b/modules/k8s/node.go
@@ -9,6 +9,9 @@ import (
 	"github.com/gruntwork-io/terratest/modules/retry"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	// The following line loads the gcp plugin which is required to authenticate against GKE clusters.
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
 
 // GetNodes queries Kubernetes for information about the worker nodes registered to the cluster. If anything goes wrong,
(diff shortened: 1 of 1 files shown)


robmorgan:
cc @rileykarson

modules/k8s/node.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/gruntwork-io/terratest/modules/retry"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	// The following line loads the gcp plugin which is required to authenticate against GKE clusters.
```
yorinasub17: Be sure to add link to the ticket from the description.

yorinasub17:
It looks like the test failure was transient. I am not exactly sure why it was failing in the first place though. I hope it isn't some race condition...

In any case, once you add the link in the comments, feel free to merge and release. Be sure to read our basecamp guide "Release process for new features, modules, and products" (https://3.basecamp.com/3989840/buckets/7257411/messages/1227462490) on the release notes format.

modules/k8s/node.go
robmorgan: Ok

robmorgan:
> UPDATE: I actually just set up the release notes drafter, so let's see if that works!

Ok, sure!

PR #231, opened 2019-01-30
RBAC testing

yorinasub17:
This adds functions that help with testing Kubernetes RBAC roles:

- Ability to auth as a `ServiceAccount`
- Ability to check what actions the user can perform
- A top level example demonstrating how one would test an RBAC role

TODO:
- [x] Investigate test instability. I have a suspicion that all the kubernetes actions against minikube are overloading the system. The docker test is also failing now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,22 +1,131 @@
+defaults: &defaults
+  machine: true
+  environment:
+    GRUNTWORK_INSTALLER_VERSION: v0.0.21
+    MODULE_CI_VERSION: v0.13.3
+    TERRAFORM_VERSION: 0.11.11
+    TERRAGRUNT_VERSION: NONE
+    PACKER_VERSION: 1.3.3
+    GOLANG_VERSION: 1.11
+    K8S_VERSION: v1.10.0  # Same as EKS
+    HELM_VERSION: v2.12.2
+    KUBECONFIG: /home/circleci/.kube/config
+    MINIKUBE_VERSION: v0.28.2  # See https://github.com/kubernetes/minikube/issues/2704
+    MINIKUBE_WANTUPDATENOTIFICATION: "false"
+    MINIKUBE_WANTREPORTERRORPROMPT: "false"
+    MINIKUBE_HOME: /home/circleci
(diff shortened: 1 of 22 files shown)


autero1:
Took a look at the test failures. Especially the docker test `TestPackerDockerExampleLocal` seems flaky. Packer doesn't seem to consistently tag the image. 

On successful builds, you see log statements such as this:
```
TestPackerDockerExampleLocal ... Tagging image: sha256:xxx
TestPackerDockerExampleLocal ... (docker-tag): Repository: gruntwork/packer-docker-example:latest
TestPackerDockerExampleLocal ... Build 'ubuntu-docker' finished.
```

On failed builds, you don't see those. Instead `docker-compose` fails with 
```
TestPackerDockerExampleLocal ... pull access denied for gruntwork/packer-docker-example, repository does not exist or may require 'docker login'
```
Which is a bit strange, as those tests have remained untouched for some time. 

> I have a suspicion that all the kubernetes actions against minikube are overloading the system. The docker test is also failing now.

I don't think you're far from the truth here. I have found `docker` and especially `docker-compose` -tests really unstable in CircleCI.

yorinasub17:
@autero1 Ok I think my hunch was correct. The build is passing now after I switched to running all the kubernetes tests separately. You can see the changes I made 145916d 

Specifically:
- Refactor the circleci config so that there are two tracks.
- Use golang build tags to control when kubernetes tests are run.

Since I made some surgery to circleci, I am going to cut a temp release to test the deployment step.

yorinasub17:
Also good to know re:docker in circleci. That is probably something we should keep an eye out for.

autero1:
> Also good to know re:docker in circleci. That is probably something we should keep an eye out for.

In some cases we have had to disable docker -based tests in CircleCI. Not really an option with `terratest` 😄

yorinasub17:
Ok build passed and I got a confirmed release with the deploy step working (the binaries are uploaded) so going to merge this. Pasted screenshot below since I will delete the test release.

release_v0_13_24-alpha_ _gruntwork-io_terratest

robmorgan:
Also LGTM 👍

PR #234, opened 2019-02-12
Template tests for helm charts

yorinasub17:
This adds enough functionality to implement helm template tests. Helm template tests are a concept I saw introduced in [the consul helm chart](https://github.com/hashicorp/consul-helm#test-details) which make a lot of sense as unit tests.

In addition to the functionality, I added a top line example to demonstrate how to write these tests:

- [A basic helm chart in `helm-basic-example`](https://github.com/gruntwork-io/terratest/tree/dcb8f8dbc1fc1247bee281397a3e3dcc1fd46368/examples/helm-basic-example)
- [The corresponding test file](https://github.com/gruntwork-io/terratest/blob/dcb8f8dbc1fc1247bee281397a3e3dcc1fd46368/test/helm_basic_example_template_test.go)


examples/kubernetes-basic-example/README.md
```diff
@@ -36,5 +36,5 @@ See the corresponding terratest code for an example of how to test this resource
 1. Install [Golang](https://golang.org/) and make sure this code is checked out into your `GOPATH`.
 1. `cd test`
 1. `dep ensure`
-1. `go test -v -run TestKubernetesBasicExample`
+1. `go test -v -tags kubernetes -run TestKubernetesBasicExample`
```
Etiene: the note on `test/helm_basic_example_template_test.go` could also be here so it's clearer why this tag is necessary

modules/helm/format.go
```diff
@@ -0,0 +1,81 @@
...
+	"testing"
+
+	"github.com/gruntwork-io/gruntwork-cli/collections"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/require"
+)
+
+// FormatSetValuesAsArgs formats the given values as command line args for helm using set (e.g of the format
+// --set key=value)
+func FormatSetValuesAsArgs(setValues map[string]string) []string {
```
brikis98: Should this be public?

modules/helm/format.go
```diff
@@ -0,0 +1,81 @@
+package helm
+
+import (
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	"github.com/gruntwork-io/gruntwork-cli/collections"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
```
brikis98: Should terratest depend on `gruntwork-cli`?

modules/helm/format.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// FormatValuesFilesAsArgsE formats the given list of values file paths as command line args for helm (e.g of the format
+// -f path)
+func FormatValuesFilesAsArgsE(t *testing.T, valuesFiles []string) ([]string, error) {
+	args := []string{}
+
+	for _, valuesFilePath := range valuesFiles {
+		// Pass through filepath.Abs to make sure this file exists
+		absValuesFilePath, err := filepath.Abs(valuesFilePath)
```
brikis98: I don't think `Abs` checks for existence of the file...

modules/helm/format.go
```diff
@@ -0,0 +1,81 @@
...
+// -f path). This will fail the test if one of the paths do not exist.
+func FormatValuesFilesAsArgs(t *testing.T, valuesFiles []string) []string {
+	args, err := FormatValuesFilesAsArgsE(t, valuesFiles)
+	require.NoError(t, err)
+	return args
+}
+
+// FormatValuesFilesAsArgsE formats the given list of values file paths as command line args for helm (e.g of the format
+// -f path)
+func FormatValuesFilesAsArgsE(t *testing.T, valuesFiles []string) ([]string, error) {
```
brikis98: What is the difference between this method and `FormatSetFilesAsArgsE`? Is it just the argument (`-f` vs `--set-file`)?

test/helm_basic_example_template_test.go
```diff
@@ -0,0 +1,120 @@
...
+	// arg to demonstrate how to select individual templates to render.
+	output := helm.RenderTemplate(t, options, helmChartPath, []string{"templates/deployment.yaml"})
+
+	// Now we use kubernetes/client-go library to render the template output into the Deployment struct. This will
+	// ensure the Deployment resource is rendered correctly.
+	// NOTE: the client-go library can only decode json, so we will first convert the yaml to json
+	jsonData, err := yaml.YAMLToJSON([]byte(output))
+	require.NoError(t, err)
+	var deployment appsv1.Deployment
+	require.NoError(t, json.Unmarshal(jsonData, &deployment))
```
brikis98: Is this something you'll need to do regularly when using yaml? If so, should we add a helper method to make this easier? Or even a `RenderTemplateToJson` method?

examples/helm-basic-example/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
+{{/* vim: set filetype=mustache: */}}
```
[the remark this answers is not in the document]
yorinasub17: Yup that is correct. Vim parses comments of any file and parses `vim:` as a vim command to execute. I think it is also worth mentioning that this file is autogenerated by helm when you do `helm create` to create a new chart.

examples/helm-basic-example/templates/deployment.yaml
```diff
@@ -0,0 +1,28 @@
...
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "helm-basic-example.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      containers:
+        - name: app
+          {{- $repo := required "containerImageRepo is required" .Values.containerImageRepo }}
+          {{- $tag := required "containerImageTag is required" .Values.containerImageTag }}
```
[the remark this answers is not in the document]
yorinasub17: This is part of the "stdlib" that helm provides in templates. See https://docs.helm.sh/developing_charts/#know-your-template-functions

modules/helm/cmd.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/gruntwork-io/terratest/modules/shell"
+)
+
+// GetCommonArgs extracts common helm options. In this case, these are:
+// - kubeconfig path
+// - kubeconfig context
+// - namespace
+// - helm home path
+func GetCommonArgs(options *Options, args ...string) []string {
```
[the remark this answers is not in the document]
yorinasub17: TBH I followed what `terraform/cmd.go` was doing. I was actually wondering the same thing, but err'd on the side of conservative.

modules/helm/format.go
yorinasub17: Well both of these are useful libraries in tests and main code and I don't think they change that much between tests and prod. Honestly, I look at `gruntwork-cli` more as a Gruntwork utils stdlib and less as a "production CLI lib" given some of the modules it provides.

e.g The stacktrace errors functionality is something that is very useful in test code.

I am open to suggestions. We can reimplement these functions in terratest (cringe), or we can create a new gruntwork-stdlib or something like that.

brikis98: No problem, we can keep using it, I was just checking if it was intentional.

modules/helm/format.go
yorinasub17: Ah you are right! Fixed to explicitly check for existence

modules/helm/format.go
yorinasub17: There's one more difference, where `--set-file` is key value pairs.

If you want to know what these args are: `-f`, which is short for `--values`, passes in a `values.yaml` file which is a yaml file listing out all the inputs and replaces `--set`. E.g let's say you have

```yaml
# example.yaml
containerImage:
  repo: nginx
  tag: 1.15.4
```

the following are equivalent:

```
helm install -f example.yaml
helm install --set containerImage.repo=nginx --set containerImage.tag=1.15.4
```

`--set-file` is a special version of `--set`, where the value is pulled from a file as opposed to passed in. Let's say that you also had the file `containerrepo.txt` which contains `nginx`. Then the following is also equivalent:

```
helm install --set-file containerImage.repo=./containerrepo.txt --set containerImage.tag=1.15.4
```

modules/helm/cmd.go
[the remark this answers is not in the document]
yorinasub17: Made private since I don't see anyone using this directly.

modules/helm/format.go
yorinasub17: Made private since I don't see anyone using this directly.

modules/helm/format.go
brikis98: Make sure to update this in all the places you use `Abs`!

test/helm_basic_example_template_test.go
yorinasub17: Yes. I actually caught this bug later when I tried using it in `k8s-service`, and then I realized we will do this quite a lot. I think I am thinking along the lines of `UnmarshalK8SYaml`.

modules/helm/format.go
yorinasub17: Did a quick grep and I think I was the only idiot who thought this.

test/helm_basic_example_template_test.go
yorinasub17: Ok added an `UnmarshalK8SYaml` function. This section now looks like:

```go
var deployment appsv1.Deployment
helm.UnmarshalK8SYaml(t, out, &deployment)
```

yorinasub17:
Ok build is passing! Going to merge and release

PR #237, opened 2019-02-17
Helm integration test framework

yorinasub17:
This introduces helm integration tests by providing `helm.Install` and `helm.Delete` functions. Additionally, this extends the `k8s` module with additional functions that map better to helm. Specifically, when you install a chart, it is not obvious what name is chosen for the chart. However, you are always guaranteed to have a bunch of labels defined on the resources, so it is more common to query Kubernetes resources using the labels than directly by name with helm.

Here is the list of new functions introduced:
- `helm.Install`: Install a helm chart onto a Kubernetes cluster
- `helm.Delete`: Delete a deployed release from Tiller
- `k8s.ListPods`: List pods in a given namespace
- `k8s.ListServices`: List services in a given namespace
- `k8s.WaitUntilNumPodsCreated`: Waits until the number of pods matching the given filters reaches the desired count.
- `k8s.Tunnel`: A struct that can be used to create and manage port forward tunnel to a Pod.

Finally, I had to further separate out the helm tests because Tiller overloaded minikube when running with the kubernetes tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ setup_minikube: &setup_minikube
     )
 
     # Install helm
-    helm init --wait
+    # helm init --wait
 
 
 install_gruntwork_utils: &install_gruntwork_utils
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,12 +30,6 @@ setup_minikube: &setup_minikube
     mkdir -p ${HOME}/.kube
     touch ${HOME}/.kube/config
 
-    # install helm
-    curl -Lo helm.tar.gz https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-amd64.tar.gz
-    tar -xvf helm.tar.gz
-    chmod +x linux-amd64/helm
-    sudo mv linux-amd64/helm /usr/local/bin/
-
     # Install minikube
     curl -Lo minikube https://github.com/kubernetes/minikube/releases/download/${MINIKUBE_VERSION}/minikube-linux-amd64
     chmod +x minikube
@@ -57,6 +51,22 @@ setup_minikube: &setup_minikube
       done
       true
     )
+
+
+install_helm: &install_helm
+  name: install helm
+  command: |
+    # install helm
+    curl -Lo helm.tar.gz https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-amd64.tar.gz
+    tar -xvf helm.tar.gz
+    chmod +x linux-amd64/helm
+    sudo mv linux-amd64/helm /usr/local/bin/
+
+    # Grant the default service account cluster admin rights so helm works
+    kubectl create clusterrolebinding default-service-account-cluster-admin-binding --clusterrole cluster-admin --serviceaccount kube-system:default
+
+    # Deploy Tiller
+    helm init --wait
 
 
 install_gruntwork_utils: &install_gruntwork_utils
@@ -204,7 +214,8 @@ jobs:
       - run:
           <<: *setup_minikube
 
-      - run: helm init --wait
+      - run:
+          <<: *install_helm
 
       # Run the Helm tests. These tests are run because the helm build tag is included, and we explicitly
       # select the helm tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,9 +58,6 @@ setup_minikube: &setup_minikube
       true
     )
 
-    # Install helm
-    # helm init --wait
-
 
 install_gruntwork_utils: &install_gruntwork_utils
   name: install gruntwork utils
@@ -176,10 +173,42 @@ jobs:
       - run:
           command: |
             mkdir -p /tmp/logs
-            # Run the unit tests first, then the integration tests. They are separate because the integration tests
-            # require additional filtering.
-            run-go-tests --packages "-tags kubernetes ./modules/k8s" | tee /tmp/logs/test_output.log
-            run-go-tests --packages "-tags kubernetes -run \"(TestKubernetes)|(TestHelm)\" ./test" | tee -a /tmp/logs/test_output.log
+            run-go-tests --packages "-tags kubernetes ./test ./modules/k8s" | tee /tmp/logs/test_output.log
+
+      - run:
+          command: ./cmd/bin/terratest_log_parser_linux_amd64 --testlog /tmp/logs/test_output.log --outputdir /tmp/logs
+          when: always
+
+      # Store test result and log artifacts for browsing purposes
+      - store_artifacts:
+          path: /tmp/logs
+      - store_test_results:
+          path: /tmp/logs
+
+
+  helm_test:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /home/circleci
+
+      - run:
+          <<: *install_gruntwork_utils
+
+      # The weird way you have to set PATH in Circle 2.0
+      - run: echo 'export PATH=$HOME/.local/bin:$HOME/terraform:$HOME/packer:$PATH' >> $BASH_ENV
+
+      - run:
+          <<: *setup_minikube
+
+      - run: helm init --wait
+
+      # Run the Helm tests. These tests are run because the helm build tag is included, and we explicitly
+      # select the helm tests
+      - run:
+          command: |
+            mkdir -p /tmp/logs
+            run-go-tests --packages "-tags helm -run ./test" | tee -a /tmp/logs/test_output.log
 
       - run:
           command: ./cmd/bin/terratest_log_parser_linux_amd64 --testlog /tmp/logs/test_output.log --outputdir /tmp/logs
@@ -226,10 +255,18 @@ workflows:
             tags:
               only: /^v.*/
 
+      - helm_test:
+          requires:
+            - setup
+          filters:
+            tags:
+              only: /^v.*/
+
       - deploy:
           requires:
             - test
             - kubernetes_test
+            - helm_test
           filters:
             tags:
               only: /^v.*/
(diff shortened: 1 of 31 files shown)


modules/helm/cmd.go
```diff
@@ -19,34 +18,34 @@ func getCommonArgs(options *Options, args ...string) []string {
...
+		return args, errors.WithStackTrace(err)
 	}
 	args = append(args, setFilesArgs...)
+	return args, nil
+}
 
+// RunHelmCommandAndGetOutputE runs helm with the given arguments and options and returns stdout/stderr.
+func RunHelmCommandAndGetOutputE(t *testing.T, options *Options, cmd string, additionalArgs ...string) (string, error) {
+	args := []string{cmd}
+	args = getCommonArgs(options, args...)
```
yorinasub17: This refactor was necessary because only `install` and `upgrade` take the `--set`, `--set-file`, and `-f` options.

modules/k8s/tunnel.go
```diff
@@ -0,0 +1,225 @@
...
+// Tunnel is the main struct that configures and manages port forwading tunnels to Kubernetes resources.
+type Tunnel struct {
+	KubectlOptions *KubectlOptions
+	LocalPort      int
+	RemotePort     int
+	ResourceType   KubeResourceType
+	ResourceName   string
+	Out            io.Writer
+	StopChan       chan struct{}
+	ReadyChan      chan struct{}
```
brikis98: Are all these fields are used outside of this package (esp the channels)? If not, you can make them lowercase to make them private.

modules/k8s/tunnel.go
```diff
@@ -0,0 +1,225 @@
...
+// not find an avilable port.
+func GetAvailablePort(t *testing.T) int {
+	port, err := GetAvailablePortE(t)
+	require.NoError(t, err)
+	return port
+}
+
+// GetAvailablePortE retrieves an available port on the host machine. This delegates the port selection to the golang net
+// library by starting a server and then checking the port that the server is using.
+func GetAvailablePortE(t *testing.T) (int, error) {
```
brikis98: If you run this in a concurrent context, couldn't a race condition lead to the same port being selected? i.e., goroutine one listens on port 12345, it's available, so it returns the port number, and in the `defer` stops listening. Right after the `defer`, but before the port number is returned, another goroutine happens to listen on 12345, see it's available, and returns it.

modules/k8s/tunnel_test.go
```diff
@@ -0,0 +1,114 @@
...
+	options := NewKubectlOptions("", "")
+	options.Namespace = uniqueID
+	configData := fmt.Sprintf(EXAMPLE_POD_YAML_TEMPLATE, uniqueID, uniqueID)
+	defer KubectlDeleteFromString(t, options, configData)
+	KubectlApplyFromString(t, options, configData)
+	WaitUntilPodAvailable(t, options, "nginx-pod", 60, 1*time.Second)
+
+	// Open a tunnel to pod from any available port locally
+	localPort := GetAvailablePort(t)
+	tunnel := NewTunnel(options, ResourceTypePod, "nginx-pod", localPort, 80)
```
brikis98: Nice, powerful API hiding a lot of complexity 👍

test/helm_basic_example_integration_test.go
```diff
@@ -0,0 +1,97 @@
...
+	endpoint := k8s.GetServiceEndpoint(t, service, 80)
+	// Test the endpoint for up to 5 minutes. This will only fail if we timeout waiting for the service to return a 200
+	// response.
+	http_helper.HttpGetWithRetryWithCustomValidation(
+		t,
+		fmt.Sprintf("http://%s", endpoint),
+		30,
+		10*time.Second,
+		func(statusCode int, body string) bool {
+			return statusCode == 200
```
brikis98: Check the body contains the text "nginx"?

.circleci/config.yml
```diff
@@ -223,10 +269,18 @@ workflows:
             tags:
               only: /^v.*/
 
+      - helm_test:
```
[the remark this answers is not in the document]
yorinasub17: Yes. See https://circleci.com/workflow-run/3daf8d7f-f7fe-4daa-a1a7-0679026b0a7e

examples/helm-basic-example/README.md
```diff
@@ -11,22 +11,25 @@ There are two kinds of tests you can perform on a helm chart:
...
+1. `go test -v -tags helm -run TestHelmBasicExampleTemplate` for the template test
+1. `go test -v -tags helm -run TestHelmBasicExampleDeployment` for the integration test
 
-**NOTE:** we have build tags to differentiate kubernetes tests from non-kubernetes tests. This is done because minikube
-is heavy and can interfere with docker related tests in terratest. To avoid overloading the system, we run the
-kubernetes tests separately from the others.
+**NOTE**: we have build tags to differentiate kubernetes tests from non-kubernetes tests, and further differentiate helm
+tests. This is done because minikube is heavy and can interfere with docker related tests in terratest. Similarly, helm
+can overload the minikube system and thus interfere with the other kubernetes tests. To avoid overloading the system, we
+run the kubernetes tests and helm tests separately from the others.
```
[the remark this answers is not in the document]
yorinasub17: I'll update the comment, but here is an example failing test: https://circleci.com/gh/gruntwork-io/terratest/869

Of the 28 tests, 19 fail. They all fail with:

```
Post https://10.142.2.94:8443/apis/authorization.k8s.io/v1/selfsubjectaccessreviews: dial tcp 10.142.2.94:8443: connect: connection refused
```

This indicates that either:

- minikube crashed
- minikube failed to respond

I don't think this would be a problem on beefy machines, as this does NOT happen when I run the kubernetes+helm tests locally, so I believe this is an issue with CircleCI. My best guess is memory. The executors have [8GB of RAM](https://circleci.com/docs/2.0/executor-types/#using-machine), but my laptop is 16GB. I don't think it's CPU because I am on a 13inch no touchbar model, which is a dual core machine.

modules/k8s/kubectl_options.go
```diff
@@ -14,3 +16,16 @@ func NewKubectlOptions(contextName string, configPath string) *KubectlOptions {
 		ConfigPath:  configPath,
 	}
 }
+
+// GetConfigPath will return a sensible default if the config path is not set on the options.
+func (kubectlOptions *KubectlOptions) GetConfigPath(t *testing.T) (string, error) {
+	var err error
```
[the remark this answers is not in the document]
yorinasub17: I'll add comment, but this is so that I can update `kubeConfigPath` in the if block below. Otherwise, it complains saying `err` is undefined.

modules/k8s/pod.go
```diff
@@ -30,6 +52,55 @@ func GetPodE(t *testing.T, options *KubectlOptions, podName string) (*corev1.Pod
...
+	desiredCount int,
+	retries int,
+	sleepBetweenRetries time.Duration,
+) error {
+	statusMsg := fmt.Sprintf("Wait for num pods created to match desired count %d.", desiredCount)
+	message, err := retry.DoWithRetryE(
+		t,
+		statusMsg,
+		60,
+		1*time.Second,
```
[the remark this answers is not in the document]
yorinasub17: Woops copy pasta from when I was experimenting in `helm-kubernetes-services`. Will fix.

modules/k8s/tunnel.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (resourceType KubeResourceType) String() string {
+	switch resourceType {
+	case ResourceTypePod:
+		return "pod"
+	case ResourceTypeService:
+		return "svc"
+	default:
+		// This should not happen
+		return ""
```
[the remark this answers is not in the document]
yorinasub17: Ok

modules/k8s/tunnel.go
yorinasub17: Ah ok. I didn't know that convention. Will update!

modules/k8s/tunnel.go
yorinasub17: I think you are right. Let me see if there is a way to make this work correctly.

modules/k8s/tunnel_test.go
yorinasub17: Can't take credit for it since it is copied from `helm` 😄  But yes I like this API.

test/helm_basic_example_integration_test.go
yorinasub17: Ok

examples/helm-basic-example/README.md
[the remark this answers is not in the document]
yorinasub17: New verbiage:
```
// NOTE: we have build tags to differentiate kubernetes tests from non-kubernetes tests. This is done because minikube
// is heavy and can interfere with docker related tests in terratest. Specifically, many of the tests start to fail with
// `connection refused` errors from `minikube`. To avoid overloading the system, we run the kubernetes tests and helm
// tests separately from the others. This may not be necessary if you have a sufficiently powerful machine.  We
// recommend at least 4 cores and 16GB of RAM if you want to run all the tests together.
```

yorinasub17:
Ok I believe I addressed all the comments except for the concurrency issue with `GetAvailablePort`. I will look into that tomorrow.

yorinasub17:
Also verified the private var conversion for `Tunnel` struct still works with `helm-kubernetes-services`.

modules/k8s/tunnel.go
yorinasub17: Ok resolved this by changing it so that the available port selection happens in `ForwardPortE` when local port is 0. This way, we can enforce a mutex lock between getting the available port and spawning the listener. See 7440db7 .

yorinasub17:
Ok I believe I addressed all the review comments, and build is passing, so going to go ahead and merge + release this!

Issue #239, opened 2019-02-22
Add more test-structure helpers for saving and loading *Options?


kirecek:
Are PRs with save/load data helpers welcomed? For example [kubectl options](https://godoc.org/github.com/gruntwork-io/terratest/modules/k8s#KubectlOptions). Or should I keep using generic `SaveTestData` in my tests?

yorinasub17:
Yup PRs welcome! save/load data helpers for kubectl options sounds like it will be a great addition!

PR #240, opened 2019-02-22
Add test-structure helper for save/load kubectl options

kirecek:
Related issue: https://github.com/gruntwork-io/terratest/issues/239

### DIFF
--- a/modules/test-structure/save_test_data_test.go
+++ b/modules/test-structure/save_test_data_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gruntwork-io/terratest/modules/k8s"
 	"github.com/gruntwork-io/terratest/modules/terraform"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type testData struct {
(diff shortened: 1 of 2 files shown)


modules/test-structure/save_test_data_test.go
```diff
@@ -218,3 +219,23 @@ func TestSaveAndLoadNamedInts(t *testing.T) {
...
 	assert.Equal(t, expectedData2, actualData2)
 }
+
+func TestSaveAndLoadKubectlOptions(t *testing.T) {
+	t.Parallel()
+
+	tmpFolder, err := ioutil.TempDir("", "save-and-load-kubectl-options")
+	if err != nil {
+		t.Fatalf("Failed to create temp dir: %v", err)
+	}
```
yorinasub17: Can you switch this to using `require.NoError(t, err)`? `require` is imported as `github.com/stretchr/testify/require`. We've been (slowly) transitioning to using `require` instead of `if => t.Fatal`, so for new code we use this pattern.

modules/test-structure/save_test_data_test.go
```diff
@@ -218,3 +219,23 @@ func TestSaveAndLoadNamedInts(t *testing.T) {
...
+
+	expectedData := &k8s.KubectlOptions{
+		ContextName: "terratest-context",
+		ConfigPath:  "~/.kube/config",
+		Namespace:   "default",
+		Env:         map[string]string{},
```
yorinasub17: Can you also populate `Env` with a few entries to make sure the nested object serializes correctly?

yorinasub17:
Ok the new changes look good, so I will go ahead and merge this in. I'll report back the test results here.

PR #246, opened 2019-03-01
Standardize modules/k8s functions to take in KubectlOptions

yorinasub17:
__NOTE: This is a backwards incompatible change__

This standardizes all the functions in `modules/k8s` to explicitly take in a `KubectlOptions` as opposed to assuming the default so that you can control which cluster actions perform. This is necessary to support test parallelism, so that you can either select a specific context for the action OR create a temp kubeconfig file and use that.

Additional features:
- The GKE auth plugin import is moved to the `client.go` file.

### DIFF
--- a/modules/k8s/client.go
+++ b/modules/k8s/client.go
@@ -5,6 +5,10 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 
+	// The following line loads the gcp plugin which is required to authenticate against GKE clusters.
+	// See: https://github.com/kubernetes/client-go/issues/242
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+
 	"github.com/gruntwork-io/terratest/modules/logger"
 )
 
--- a/modules/k8s/node.go
+++ b/modules/k8s/node.go
@@ -5,30 +5,27 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gruntwork-io/terratest/modules/logger"
-	"github.com/gruntwork-io/terratest/modules/retry"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	// The following line loads the gcp plugin which is required to authenticate against GKE clusters.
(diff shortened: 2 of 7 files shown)


robmorgan:
@yorinasub17 strange the tests failed. I guess its a glitch in the matrix?

modules/k8s/client.go
```diff
@@ -5,6 +5,10 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 
+	// The following line loads the gcp plugin which is required to authenticate against GKE clusters.
+	// See: https://github.com/kubernetes/client-go/issues/242
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+
```
robmorgan: Okay, cool 👍

yorinasub17:
The two tests that failed are known to be unstable. Rerunning now...

yorinasub17:
I think this is ready to merge now. I am going to merge this and cut a minor version bump.

modules/k8s/node.go
```diff
@@ -5,30 +5,27 @@ import (
...
 )
 
 // GetNodes queries Kubernetes for information about the worker nodes registered to the cluster. If anything goes wrong,
 // the function will automatically fail the test.
-func GetNodes(t *testing.T) []corev1.Node {
-	nodes, err := GetNodesE(t)
-	if err != nil {
-		t.Fatal(err)
-	}
+func GetNodes(t *testing.T, options *KubectlOptions) []corev1.Node {
```
brikis98: If there's a reasonable default you can fall back to without an `options` object, then a trick you can use in the future to make this truly optional and backwards compatible is:

```go
func GetNodes(t *testing.T, options ...*KubectlOptions) []corev1.Node {
  // ...
}

GetNodes(t) // valid
GetNodes(t, *KubectlOptions { ... }) // also valid
```

Of course, there are draw backs too: it only works for the last argument in a function and you have to do a little extra work as the module maintainer to see if 0, 1, or multiple values were passed (the latter case probably benefits from an error, which is still more code to write).

yorinasub17: Ah I see that makes sense. I wasn't looking at variadic parameters as a way to do optional args but totally makes sense. Thanks for the tip!

PR #248, opened 2019-03-01
 Fix #242: Support installation of remote charts 

### LINKED ISSUE #242: helm.Install should support charts from a repository
Right now `helm.Install` is focused on testing local charts, but sometimes you might want to install a chart from stable as a prerequisite/dependency to the local chart.

We should support being able to install arbitrary charts using `helm.Install`.

yorinasub17:
This fixes https://github.com/gruntwork-io/terratest/issues/242 by relaxing the constraint that the passing chart must exist on the file system to run `helm.Install`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,6 +222,9 @@ jobs:
       - run:
           command: |
             mkdir -p /tmp/logs
+            # Run the unit tests first, then the integration tests. They are separate because the integration tests
+            # require additional filtering.
+            run-go-tests --packages "-tags helm ./modules/helm" | tee /tmp/logs/test_output.log
             run-go-tests --packages "-tags helm -run TestHelm ./test" | tee -a /tmp/logs/test_output.log
 
       - run:
--- a/modules/helm/install.go
+++ b/modules/helm/install.go
@@ -12,22 +12,25 @@ import (
 
 // Install will install the selected helm chart with the provided options under the given release name. This will fail
 // the test if there is an error.
-func Install(t *testing.T, options *Options, chartDir string, releaseName string) {
-	require.NoError(t, InstallE(t, options, chartDir, releaseName))
+func Install(t *testing.T, options *Options, chart string, releaseName string) {
+	require.NoError(t, InstallE(t, options, chart, releaseName))
 }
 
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -222,6 +222,9 @@ jobs:
       - run:
           command: |
             mkdir -p /tmp/logs
+            # Run the unit tests first, then the integration tests. They are separate because the integration tests
+            # require additional filtering.
```
brikis98: Additional filtering? Not sure what that means.

modules/helm/install.go
```diff
@@ -12,22 +12,25 @@ import (
...
-	absChartDir, err := filepath.Abs(chartDir)
-	if err != nil {
-		return errors.WithStackTrace(err)
-	}
-	if !files.FileExists(chartDir) {
-		return errors.WithStackTrace(ChartNotFoundError{chartDir})
+func InstallE(t *testing.T, options *Options, chart string, releaseName string) error {
+	// If the chart refers to a path, convert to absolute path. Otherwise, pass straight through as it may be a remote
+	// chart.
+	if files.FileExists(chart) {
```
brikis98: Could this lead to an ambiguous situation where there's a chart called, say, `mysql` and a local file called `mysql`?

.circleci/config.yml
yorinasub17: Since all the tests are not tagged, if you do `go test -v -tag kubernetes ./test`, it will run all the integration tests. To run only the helm tests, I am cheating and doing `go test -v -tag helm -run TestHelm ./test`. Note that I can't do that all the time, because the unit tests aren't called that.

One way to address this is to have a separate integration test folder for k8s and helm but that might be confusing.

I am actually beginning to wonder if we should split terratest...

The real solution here, is to setup Jenkins.

modules/helm/install.go
yorinasub17: That can happen, but probably unlikely. This is because charts are always namespaced on the repo, so `mysql` is actually `stable/mysql` and our charts will be `gruntwork/k8s-service`. So unless you have a `stable` folder checked out locally, and those happen to have charts, you will reference the remote one.

yorinasub17:
Ok going to merge and release this.

PR #253, opened 2019-03-06
Initial ECS cluster support

thoo5ieb:


thoo5ieb:
I was thinking about the encapsulation of `ecs.Cluster` to make the interface simpler. But before, I would like to hear what you think once you get time to take a look at this.

modules/aws/ecs.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+	}
+	output, err := client.DescribeClusters(input)
+	if err != nil {
+		return nil, err
+	}
+
+	numClusters := len(output.Clusters)
+	if numClusters != 1 {
+		return nil, fmt.Errorf("Couldn't find ECS cluster %v in region %v", name, region)
```
brikis98: Might be clearer to say: "Expected to find 1 ECS cluster named '%s" in region '%v', but found '%d'

modules/aws/ecs_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateEcsClusterE(t *testing.T) {
+	t.Parallel()
+
+	region := GetRandomStableRegion(t, nil, nil)
+	ecs, err := CreateEcsClusterE(t, region, "terratest")
+	defer DeleteEcsCluster(t, region, ecs)
```
brikis98: Call `GetEcsCluster` in between?

thoo5ieb: Done.

modules/aws/ecs.go
thoo5ieb: Updated, thanks!

modules/aws/ecs_test.go
thoo5ieb: Thanks!

thoo5ieb:
Great, happy to hear it! :-)

thoo5ieb:
Tested it locally on my account, I had no problem. In any case, you will encounter anything, just let me know!

PR #254, opened 2019-03-06
Lock and Locktimeout parameter support added

jallentech:


modules/terraform/format.go
```diff
@@ -23,6 +24,20 @@ func FormatTerraformVarsAsArgs(vars map[string]interface{}) []string {
 	return formatTerraformArgs(vars, "-var")
 }
 
+// FormatTerraformLockAsArgs formats the lock and lock-timeout variables
+// -lock, -lock-timeout
+func FormatTerraformLockAsArgs(lockCheck bool, lockTimeout string) []string {
+	var lockArgs []string
+	if lockCheck == true {
```
brikis98:
```suggestion
	if lockCheck {
```

brikis98: Also, if this is `false`... Shouldn't we set `-lock=false`?

jallentech: Good point, thanks for looking over!
I've updated the code so that it adds a -lock=false if lock is set to false and changed the conditional to the preferred syntax

modules/terraform/format.go
```diff
@@ -14,15 +14,32 @@ func FormatArgs(options *Options, args ...string) []string {
...
+	var lockArgs []string
+	if lockCheck {
+		lockArgs = append(lockArgs, "-lock=true")
+		if lockTimeout != "" {
+			lockTimeoutValue := fmt.Sprintf("%s=%s", "-lock-timeout", lockTimeout)
+			lockArgs = append(lockArgs, lockTimeoutValue)
+		}
+	} else {
+		lockArgs = append(lockArgs, "-lock=false")
+	}
```
brikis98: Simpler:

```go
lockArgs := []string{fmt.Sprintf("-lock=%v", lockCheck)}
if lockTimeout != "" { 
  # etc
}
return lockArgs
```

modules/terraform/format.go
```diff
@@ -14,15 +14,32 @@ func FormatArgs(options *Options, args ...string) []string {
...
-// -var key=value).
+// FormatTerraformVarsAsArgs formats the given variables as command-line args for Terraform
+// (e.g. into the format -var key=value).
 func FormatTerraformVarsAsArgs(vars map[string]interface{}) []string {
 	return formatTerraformArgs(vars, "-var")
 }
 
+// FormatTerraformLockAsArgs formats the lock and lock-timeout variables
+// -lock, -lock-timeout
+func FormatTerraformLockAsArgs(lockCheck bool, lockTimeout string) []string {
```
brikis98: I think there's a bug here: this method will be called for all terraform commands, but not all of them support the `-lock` parameter. I think you'll need this method to know what Terraform command is being called and only add it self to those that support locking. See https://github.com/gruntwork-io/terragrunt/blob/master/config/config_helpers.go#L27.

brikis98:
Apologies, I lost track of this PR. Have all the comments been resolved?

jallentech:
> Apologies, I lost track of this PR. Have all the comments been resolved?

Hi @brikis98 yes, all of the comments have been resolved, thanks

modules/terraform/format.go
```diff
@@ -167,3 +199,13 @@ func primitiveToHclString(value interface{}, isNested bool) string {
 		return fmt.Sprintf("%v", v)
 	}
 }
+
+// Check whether string exists in a slice
+func sliceContains(s []string, e string) bool {
```
brikis98: NIT: Use https://github.com/gruntwork-io/terratest/blob/master/modules/collections/lists.go#L32 instead.

jallentech: Thanks @brikis98, I've updated the code to use ListContains instead of sliceContains

brikis98: Don't forget to remove the now unused `sliceContains` method!

jallentech: Sorry, thanks for pointing it out, it's been removed now!

brikis98:
Thx! Just kicked off tests.

brikis98:
Getting an unrelated test failure with the K8S / Helm tests. Some issue with MiniKube install:

```
E: Version '18.06.1-0ubuntu1.2~16.04.1' for 'docker.io' was not found
```

I unfortunately don't have time to dig into this right now, so will have to come back to it later.

yorinasub17:
> Getting an unrelated test failure with the K8S / Helm tests. Some issue with MiniKube install:

This was resolved in commit cf0c4c521bc299d6814455dfe0a81d056ae39aa5#diff-1d37e48f9ceff6d8030570cd36286a61 back in July. I just checked your branch and it looks like it is quite a bit behind (last commit from master is from May), and thus doesn't have this fix. Can you rebase on master to pull in the recent changes? I believe we've had a number of other test fixes as well so will be more likely to get a green build that way.

jallentech:
> > Getting an unrelated test failure with the K8S / Helm tests. Some issue with MiniKube install:
> 
> This was resolved in commit [cf0c4c5](https://github.com/gruntwork-io/terratest/commit/cf0c4c521bc299d6814455dfe0a81d056ae39aa5)#diff-1d37e48f9ceff6d8030570cd36286a61 back in July. I just checked your branch and it looks like it is quite a bit behind (last commit from master is from May), and thus doesn't have this fix. Can you rebase on master to pull in the recent changes? I believe we've had a number of other test fixes as well so will be more likely to get a green build that way.

Thanks, I've pulled in the latest changes and will keep an eye on the tests

brikis98:
OK, thx, re-running tests now!

brikis98:
Tests passed. Merging now.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.23.1

jallentech:
Thanks!

PR #256, opened 2019-03-11
reorder of terraform arguments var and varFiles to reflect common usage

matt-smith:
Switching the order of Vars and VarFiles in terraformArgs to reflect a more common scenario. 

It's likely that a user would like to override one or two variables from a var-file using single variables, currently the behaviour is the opposite way round. 

link to terraform [variable precedence](https://www.terraform.io/docs/configuration/variables.html#variable-definition-precedence) documentation.

### DIFF
--- a/modules/terraform/format.go
+++ b/modules/terraform/format.go
@@ -11,8 +11,8 @@ import (
 func FormatArgs(options *Options, args ...string) []string {
 	var terraformArgs []string
 	terraformArgs = append(terraformArgs, args...)
-	terraformArgs = append(terraformArgs, FormatTerraformVarsAsArgs(options.Vars)...)
 	terraformArgs = append(terraformArgs, FormatTerraformArgs("-var-file", options.VarFiles)...)
+	terraformArgs = append(terraformArgs, FormatTerraformVarsAsArgs(options.Vars)...)
 	terraformArgs = append(terraformArgs, FormatTerraformArgs("-target", options.Targets)...)
 	return terraformArgs
 }


brikis98:
Thanks for the PR! How did you test this? Could you add a test case that specifically verifies this behavior?

aldojimenez:
we have come across a new for this as well. But this PR has been open for quite some time, is there any chance this will be reviewed and actioned upon soon ?

Nuru:
We use `-var-file` to set the bulk of the variables, including defaults, and then would like to use `-var` to override them, for example to see what happens when a variable changes. This is currently not possible because changes provided via `-var` are clobbered by `-var-file`. Any chance you will accept this PR or provide an alternative soon?

Nuru:
I use `-var-file` to set the bulk of the variables, including defaults, and then would like to use `-var` to override them, for example to see what happens when a variable changes. This is currently not possible because changes provided via `-var` are clobbered by `-var-file`. Any chance you will accept this PR or provide an alternative soon?

archetypalsxe:
Also hoping that this gets merged soon

tobias-edwards:
If you locate your `terraform.tfvars` file in the same directory as the Terraform module you apply `terraform.InitAndApply`, then Terraform will automatically pick up on the variables defined in that file. You can then just use `Vars` to override the values set in your `terraform.tfvars` file:

```go
  TerraformDir: "../location/of/terraform/module/",
  Vars:         tfvarsOverridesGoHere,
}
```

sebastianczech:
As far as I was checking, #1217 resolves that issue. If you define ``SetVarsAfterVarFiles: true``, then values passed in ``Vars`` take precedence over values from ``VarFiles`` e.g.:

```go
terraformOptions := terraform.WithDefaultRetryableErrors(t, &terraform.Options{
	TerraformDir: ".",
	VarFiles:     []string{"example.tfvars"},
	Vars: map[string]interface{}{
		"example": "value_passed_from_test",
	},
	SetVarsAfterVarFiles: true,
})
```

matt-smith:
closing as resolved in #1217

PR #260, opened 2019-03-12
Correct ineffassign warnings

gliptak:


modules/terraform/output_test.go
```diff
@@ -126,13 +126,15 @@ func TestOutputsForKeys(t *testing.T) {
 	//List value
 	expectedListLen := 3
 	outputInterfaceList, ok := out["stars"].([]interface{})
+	assert.True(t, ok, "data type wrong")
```
Etiene: Thanks for the PR! But I think these messages could be more descriptive and indicate what was the expected and the actual date type instead of just saying something went wrong.

gliptak: This follows the pattern from earlier in the same file https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/output_test.go#L123
PS I'm not sure what other details are available, except that the cast failed ...

Etiene: Even hard coded mentioning which cast failed based on the comment above the block would be helpful, such as `assert.True(t, ok, "Wrong data type, expected list type value")` in this case. Getting just `data type wrong` in the logs would be confusing about where exactly the tests are failing, since now this exact message is thrown at multiple places. 

If instead of just the expected type you also want to mention which type you actually got, you can [check the type of the value](https://stackoverflow.com/a/27160765/2863960).

Feel free to update the entries found earlier in the file to reflect your new pattern.

gliptak: I updated the asserts with the variable names

modules/terraform/output_test.go
```diff
@@ -120,19 +120,21 @@ func TestOutputsForKeys(t *testing.T) {
 	//String value
 	expectedString := "Sun"
 	str, ok := out["our_star"].(string)
-	assert.True(t, ok, "data type wrong")
+	assert.True(t, ok, "data type wrong for str")
```
Etiene:
```suggestion
	assert.True(t, ok, "Wrong data type for 'our star', expected string")
```

gliptak: updated

PR #264, opened 2019-03-20
Add go module

msvechla:
Hello, 

Thanks for this awesome project!

As `go mod` is the new way of managing dependencies with go, it would be great to also have go module support with terratest.

Currently we are unable to use terratest with go modules, as the dependencies are not pinned correctly. This leads to errors like:

```
../../../../go/pkg/mod/k8s.io/client-go@v10.0.0+incompatible/kubernetes/scheme/register.go:22:2: unknown import path "k8s.io/api/admissionregistration/v1alpha1": cannot find module providing package k8s.io/api/admissionregistration/v1alpha1
``` 

This PR enables go modules support by generating a `go.mod` based on the existing `Gopkg.lock`.
Let me know whether this can get merged, or you need anything else.

Cheers!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,15 @@
+module github.com/gruntwork-io/terratest
+
+require (
+	github.com/aws/aws-sdk-go v1.12.74
+	github.com/boombuler/barcode v1.0.0
+	github.com/davecgh/go-spew v1.1.0
+	github.com/go-ini/ini v1.36.0
+	github.com/google/uuid v0.0.0-20161128191214-064e2069ce9c
+	github.com/jmespath/go-jmespath v0.0.0-20160202185014-0b12d6b521d8
+	github.com/pmezard/go-difflib v1.0.0
+	github.com/pquerna/otp v1.0.0
+	github.com/stretchr/testify v1.2.1
+	golang.org/x/crypto v0.0.0-20180426230345-b49d69b5da94
+	golang.org/x/net v0.0.0-20180420171651-5f9ae10d9af5
+)
--- a/go.sum
+++ b/go.sum
@@ -1,0 +1,11 @@
+github.com/aws/aws-sdk-go v1.12.74/go.mod h1:ZRmQr0FajVIyZ4ZzBYKG5P3ZqPz9IHG41ZoMu1ADI3k=
+github.com/boombuler/barcode v1.0.0/go.mod h1:paBWMcWSl3LHKBqUq+rly7CNSldXjb2rDl3JlRe0mD8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/go-ini/ini v1.36.0/go.mod h1:ByCAeIL28uOIIG0E3PJtZPDL8WnHpFKFOtgjp+3Ies8=
+github.com/google/uuid v0.0.0-20161128191214-064e2069ce9c/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
+github.com/jmespath/go-jmespath v0.0.0-20160202185014-0b12d6b521d8/go.mod h1:Nht3zPeWKUH0NzdCt2Blrr5ys8VGpn0CEB0cQHVjt7k=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/pquerna/otp v1.0.0/go.mod h1:Zad1CMQfSQZI5KLpahDiSUX4tMMREnXw98IvL1nhgMk=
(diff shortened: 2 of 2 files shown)


brikis98:
Nice, thanks for the PR!

1. Should we remove the `go dep` files? Or will that cause issues for users who haven't yet migrated to `go mod`? I'm worried it would be tricky to keep both `go dep` and `go mod` files in sync...
1. [According to the docs](https://github.com/golang/go/wiki/Modules), `go mod` won't be finalized until Go 1.13. Should we wait until then to make the full switch over?

msvechla:
As always the questions are on point.

> Should we remove the go dep files? Or will that cause issues for users who haven't yet migrated to go mod? I'm worried it would be tricky to keep both go dep and go mod files in sync...

1. I did not yet remove the `go dep` files on purpose, to keep the backwards compatibility. It will take some effort to keep them in sync, however currently the amount of dependencies looks manageable. At the end this should be decided by the maintainers of course.

> According to the docs, go mod won't be finalized until Go 1.13. Should we wait until then to make the full switch over?

2. Fully agreed. The question is if "switch over" means adding the `go mod` and removing the `go dep` with Go 1.13, or if we want to add the `go mod` now and remove the `go dep` with Go 1.13

I would tend to go with adding the `go mod` now and removing the `go dep` with Go 1.13, however all other approaches are also fine. What is your suggestion?

brikis98:
So from another scan of the docs, Go 1.13 isn't due until August. That's roughly 5 months to have to keep the Go dep and Go mod dependency files in sync. Odds of success if we do it manually are ~0. In fact, anyone who is using Go mod probably doesn't have `GOPATH` set up in the first place, and won't be able to use Go dep without lots of additional effort.  

So it seems we need to pick one or the other. I guess the main question for me is, if we remove the Go dep stuff now, will current Terratest users who rely on dep get all sorts of errors? Or does dep version the code, so they'd still be using the previous releases that were still configured with dep? Is there any easy way to test this?

msvechla:
I thought about adding a simple test to check if `go mod` and `dep` are in sync, but this is not easily possible, as the [modconv](https://github.com/golang/go/blob/master/src/cmd/go/internal/modconv/dep.go) package is internal.

As to my understanding removing the `Gopkg.lock` and `Gopkg.toml` files should not break the setup for existing users. I did some research and it looks like go will use the `vendor` directory automatically if present: https://golang.org/cmd/go/#hdr-Vendor_Directories

So my current assumption is, that things would only break, once dependencies are updated / added and the user did not yet switch to `go mod`.

However, it looks like there are some unknowns here, so postponing this until Go 1.13 is released, might be the safer choice.

brikis98:
> So my current assumption is, that things would only break, once dependencies are updated / added and the user did not yet switch to go mod.

That seems like something very likely to happen over the next ~5 months... 

> However, it looks like there are some unknowns here, so postponing this until Go 1.13 is released, might be the safer choice.

Agreed. I think moving to `go mod` is a great idea, and appreciate the PR, but it's probably best to wait until it's the official tool, and then go all in on it. Thanks!

tmc:
As we're nearing August could this be revisited? Initial attempts to use terratest in a bazel project appears to be complicated by the lack of module use.

brikis98:
As stated before, the plan is to wait for Go 1.13 to drop, with the official, built-in, enabled-by-default support for `go mod`.

DanielRis:
Go 1.13 has been released. It would be great to see a updated version of Terratest to simplify my test setup.

davewongillies:
@brikis98 Go 1.13 has dropped

simon-ibars-cko:
Hello,
indeed thanks for this awesome project 🙏 
I am also having this error when importing in test_structure in my go module. 
Indeed this is about the dependency on tunnel.go from k8s library [see this issue](https://github.com/gruntwork-io/terratest/issues/346) 
transitive = -> terratest/modules/test_structure -> terratest/modules/k8s -> k8s.io/kubernetes/pkg/kubectl/generate -> k8s.io/kubernetes/pkg/kubectl/generate not found

here is my 2 cents :
- Complete the support of module and relase this [existing PR](https://github.com/gruntwork-io/terratest/pull/264) as the only blocking point identified is go 1.13 released (and the current version of go is 1.13.1)
- having a new look on the [PR 240](https://github.com/gruntwork-io/terratest/pull/240) as it brings a dependency on k8s in the test_structure terratest module for only 3 methods related to k8s (SaveKubectlOptions, LoadKubectlOptions, formatKubectlOptionsPath),
would it make sense to have those methods in the k8s module folder instead of the test_structure module folder? Does those methods are used a lot in there actual form (not working when imported in a go module) ?

I would be available for the PR to move elsewhere those k8s helper methods, let me know 

Regards

brikis98:
@msvechla Could you update this PR to the latest? i.e., rebase, make sure new dependency versions are pulled in, and remove the `Gopkg.*` files?

simon-ibars-cko:
Ok @brikis98 I can do that but in a different pr for the sake of simplicity. here [it is ](https://github.com/gruntwork-io/terratest/pull/384)

msvechla:
Agreed, starting on a clean PR might be quicker. Feel free to close my pr in favor of #384. Alternatively I can rebase and make it work as well.

yorinasub17:
This has been merged with #384

PR #267, opened 2019-03-26
Add support for GCP's OS Login API

rileykarson:
This allows users to configure a single SSH key for access to Google Compute Engine VM instances using their Google identity (eg: Gmail, GSuite account).

### DIFF
--- a/modules/gcp/oslogin.go
+++ b/modules/gcp/oslogin.go
@@ -3,6 +3,7 @@ package gcp
 import (
 	"context"
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"testing"
 
 	"github.com/gruntwork-io/terratest/modules/logger"
@@ -11,14 +12,17 @@ import (
 	"google.golang.org/api/oslogin/v1"
 )
 
-// Add an OS Login SSH Key
+// ImportSSHKey will import an SSH key to GCP under the provided user identity.
+// The `user` parameter should be the email address of the user.
+// The `key` parameter should be the public key of the SSH key being uploaded.
+// This will fail the test if there is an error.
 func ImportSSHKey(t *testing.T, user, key string) {
-	err := ImportSSHKeyE(t, user, key)
-	if err != nil {
-		t.Fatalf("Could not add SSH Key to user %s: %s", user, err)
-	}
+	require.NoErrorf(t, ImportSSHKeyE(t, user, key), "Could not add SSH Key to user %")
 }
 
+// ImportSSHKeyE will import an SSH key to GCP under the provided user identity.
+// The `user` parameter should be the email address of the user.
+// The `key` parameter should be the public key of the SSH key being uploaded.
 func ImportSSHKeyE(t *testing.T, user, key string) error {
 	logger.Logf(t, "Importing SSH key for user %s", user)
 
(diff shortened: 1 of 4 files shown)


modules/gcp/oslogin.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/oslogin/v1"
+)
+
+// Add an OS Login SSH Key
+func ImportSSHKey(t *testing.T, user, key string) {
```
yorinasub17: Our public functions in terratest should be documented in the style:

```
ImportSSHKey will import an SSH key to GCP under the provided user identity. The `key` parameter should be the public key of the SSH key being uploaded. This will fail the test if there is an error.
```

and 


```
ImportSSHKey will import an SSH key to GCP under the provided user identity. The `key` parameter should be the public key of the SSH key being uploaded.
```

modules/gcp/oslogin.go
```diff
@@ -0,0 +1,90 @@
...
+	err := ImportSSHKeyE(t, user, key)
+	if err != nil {
+		t.Fatalf("Could not add SSH Key to user %s: %s", user, err)
+	}
```
yorinasub17: We have a new pattern now that uses require, so this should be:

```
require.NoErrorf(t, ImportSSHKeyE(t, user, key), "Could not add SSH Key to user %s: %s", user, err)
```

rileykarson: Done- not sure how to use this in cases with returns like https://github.com/gruntwork-io/terratest/pull/267/files#diff-672950984cb105f6fbcfa972d0119285R54

yorinasub17: You can do `require.NoErrorf(t, err, ...)`. It's not a big deal, but some of us have been finding it more readable when we drop the indent.

yorinasub17:
Also our CI build failed, because we don't have `GOOGLE_IDENTITY_TERRATEST` defined. What should this be?

rileykarson:
Ah, that should be the email of whatever GCP service account is being used as a test runner; presumably, the one these creds are used for: https://github.com/gruntwork-io/terratest/blob/master/.circleci/config.yml#L143

robmorgan:
> Also our CI build failed, because we don't have `GOOGLE_IDENTITY_TERRATEST` defined. What should this be?

@yorinasub17 @rileykarson Is this approach I took of any use? https://github.com/gruntwork-io/terraform-google-gke/blob/b90352e9f9d2b2d6e0192e79fc7c5f5264388cca/test/test_helpers.go#L19. I'm currently using a temporary account I created for use with GKE. I could look at moving these methods to Terratest.

rileykarson:
That's effectively the same approach, just a different name! Our usage there was superseded in Terraform-land by https://www.terraform.io/docs/providers/google/d/datasource_google_client_openid_userinfo.html so I didn't bother generalising.

With that datasource solving the GKE -> K8S issue, OS Login is a little unique in that it's managed by the user (not necessarily human- just a Google identity) and permissions can't be granted to act on behalf of another, so we need to know the test runner email.

yorinasub17:
@robmorgan @rileykarson Is there anything else here? Can I merge and release?

rileykarson:
Afaik I was waiting on Rob to set a variable in CI, no action items on my end.

yorinasub17:
As discussed, this is good to go so going to merge it.

PR #270, opened 2019-04-01
Fix k8s dep

yorinasub17:
Because Kubernetes doesn't support dep, the dependencies of apimachinery and client-go for Kubernetes are free floating. This caused some issues when bootstrapping a new project on terratest, where one of the dependencies had a breaking change that was pulled in.

This fixes the current breaking change, and also introduces a pattern for fixing it. Specifically:

- Transitive dependencies should be locked using `constraint` and not `override` (if you use `override`, it isn't pulled in for projects depending on terratest).
- To make `dep` work, each transitive dependency locked using `constraint` need to be forced into a direct dependency by adding it into the `required` list.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -24,6 +24,17 @@
 #   go-tests = true
 #   unused-packages = true
 
+# These dependencies are not directly required by terratest, but are defined here to directly hook them up so that they
+# are version locked when it is pulled in by projects depending on terratest.
+# This is necessary because kubernetes apimachinery and client-go do not support dep natively, and so overrides are
+# necessary to make it work.
+# See for more info:
+# - https://github.com/kubernetes/client-go/blob/kubernetes-1.11.3/INSTALL.md#dep-not-supported-yet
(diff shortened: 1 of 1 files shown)


yorinasub17:
> Why doesn't Gopkg.lock do, uh, locking?

It turns out that it only locks in `terratest`, but not for all the derivatives of `terratest`. So if `module-foo` depends on `terratest`, it builds the dependency version list by walking all the imports, pulling in all the versions defined as `[[constraint]]`, pretty much from scratch (as if the lock file didn't exist). However, it **ignores all `[[override]]` versions**. This is why to force the derivatives to get the right version, we need to force it into a `[[constraint]]` clause.

> Also, does the K8S stuff use Go.mod?

Sadly, no. They support `glide` and `godeps`.

I will need to look into how to make `gomod` work when we get there.

brikis98:
Ah, bummer. OK, thx for clarifying!

yorinasub17:
Any objections to merging and releasing this? I have a branch on `module-server` that depends on this.

brikis98:
No objection, ship!

yorinasub17:
Merging

yorinasub17:
Thanks!

PR #272, opened 2019-04-04
Change EC2 to RDS and point to correct example

scottsuch:
The rds example looks like it has a few copy/paste artifacts. Cleaning those up to ensure consistency. I'm unable to actually get this test running as it continually times out at 10 mins.


yorinasub17:
The relevant tests passed (there were some failures, but they are unrelated to this and are known to be brittle). Going to merge. Thanks again for your contribution!

PR #273, opened 2019-04-05
k8s ingress functions

yorinasub17:
__NOTE: This is blocked on a feature being implemented in `module-ci` to share a common script for setting up minikube__

This implements the following functions in the `k8s` module for helping with testing Ingress resources:
- `ListIngresses` and `ListIngressesE` for listing all the Ingress resources available in the cluster.
- `GetIngress` and `GetIngressE` for getting a single Ingress resource by name from the cluster.
- `IsIngressAvailable` which tells you if an Ingress is available to accept requests.
- `WaitUntilIngressAvailable` which will wait until the given ingress resource (by name) is available to accept requests (with timeout).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,67 +12,6 @@ defaults: &defaults
     K8S_VERSION: v1.10.0  # Same as EKS
     HELM_VERSION: v2.12.2
     KUBECONFIG: /home/circleci/.kube/config
-    MINIKUBE_VERSION: v0.28.2  # See https://github.com/kubernetes/minikube/issues/2704
-    MINIKUBE_WANTUPDATENOTIFICATION: "false"
-    MINIKUBE_WANTREPORTERRORPROMPT: "false"
-    MINIKUBE_HOME: /home/circleci
-    CHANGE_MINIKUBE_NONE_USER: "true"
-
-
-# Install and setup minikube
-# https://github.com/kubernetes/minikube#linux-continuous-integration-without-vm-support
-setup_minikube: &setup_minikube
-  name: install helm, kubectl and minikube
-  working_directory: /tmp
-  command: |
-    # Install yq so we can modify the configmap
-    sudo add-apt-repository ppa:rmescandon/yq
-    sudo apt-get update
-    sudo apt install -y yq
-
-    # install kubectl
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -74,6 +32,7 @@ install_gruntwork_utils: &install_gruntwork_utils
   command: |
     curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version "${GRUNTWORK_INSTALLER_VERSION}"
     gruntwork-install --module-name "gruntwork-module-circleci-helpers" --repo "https://github.com/gruntwork-io/module-ci" --tag "${MODULE_CI_VERSION}"
+    gruntwork-install --module-name "kubernetes-circleci-helpers" --repo "https://github.com/gruntwork-io/module-ci" --branch yori-add-k8s
```
yorinasub17: NOTE: this needs to be updated to a released version when ready.

yorinasub17:
FYI this is now ready to be reviewed as the corresponding branch in `module-ci` is now merged.

modules/shell/command_test.go
```diff
@@ -33,9 +33,13 @@ func TestRunCommandAndGetOutputOrder(t *testing.T) {
 	bashCode := fmt.Sprintf(`
 echo_stderr(){
 	(>&2 echo "%s")
+	# Add sleep to stabilize the test
+	sleep .01s
 }
 echo_stdout(){
 	echo "%s"
+	# Add sleep to stabilize the test
+	sleep .01s
```
yorinasub17: This test was getting unstable due to wrong interleaving from timing issues, so added a delay to make this more consistent.

.circleci/config.yml
```diff
@@ -176,7 +135,7 @@ jobs:
       - run: echo 'export PATH=$HOME/.local/bin:$HOME/terraform:$HOME/packer:$PATH' >> $BASH_ENV
 
       - run:
-          <<: *setup_minikube
+          command: setup-minikube
```
Etiene: whats the difference between these two?

modules/shell/command_test.go
Etiene: Could you elaborate? I still dont understand what was causing the instability

PR #277, opened 2019-04-17
Initial ELBV2 support

AlvaroRacero:


modules/aws/elb_test.go
```diff
@@ -0,0 +1,43 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/stretchr/testify/assert"
+)
+
+func getSubnetIdsPerAZ(t *testing.T, vpc *Vpc) []*string {
```
thoo5ieb: @AlvaroRacero I think, it would make sense to add a comment that describes a reason behind this function. AFAIR, alb needs 2 subnets at least from a different availability zone

modules/aws/elbv2.go
```diff
@@ -0,0 +1,120 @@
...
+	return &LoadBalancer{
+		Name: aws.StringValue(elb.LoadBalancerName),
+		ARN:  aws.StringValue(elb.LoadBalancerArn),
+		CanonicalHostedZoneID: aws.StringValue(elb.CanonicalHostedZoneId),
+		DNSName:               aws.StringValue(elb.DNSName),
+	}, nil
+}
+
+// CreateElbV2 creates ELB in the given region under the given name and subnets list.
+func CreateElbV2(t *testing.T, region string, name string, subnets []*string) {
```
thoo5ieb: I think it should return a value, ain't?

modules/aws/elbv2.go
```diff
@@ -0,0 +1,127 @@
...
+// DeleteElbV2 deletes existing ELB in the given region.
+func DeleteElbV2(t *testing.T, region string, name string) {
+	err := DeleteElbV2E(t, region, name)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+// DeleteElbV2E deletes existing ELB in the given region.
+func DeleteElbV2E(t *testing.T, region string, name string) error {
```
thoo5ieb: Since you have created a structure that describes load balancer, instead of using `name`, you should use a pointer to your structure. The structure holds the `name` attribute.

modules/aws/elbv2.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, fmt.Errorf("Expected to create 1 ELB named '%s' in region '%v', but found '%d'",
+			name, region, numLb)
+	}
+	lb := resp.LoadBalancers[0]
+
+	return CreateLoadBalancerFromElbV2(lb), nil
+}
+
+// DeleteElbV2 deletes existing ELB in the given region.
+func DeleteElbV2(t *testing.T, region string, name string) {
```
thoo5ieb: Since you have created a structure that describes load balancer, instead of using name, you should use a pointer to your structure. The structure holds the name attribute.

modules/aws/elbv2.go
```diff
@@ -0,0 +1,127 @@
...
+// LoadBalancer is an Amazon load balancer.
+type LoadBalancer struct {
+	Name                  string // The name of the load balancer.
+	ARN                   string // The Amazon Resource Name (ARN) of the load balancer.
+	CanonicalHostedZoneID string // The ID of the Amazon Route 53 hosted zone associated with the load balancer.
+	DNSName               string // The public DNS name of the load balancer.
+}
+
+// CreateLoadBalancerFromElbV2 creates a LoadBalancer struct from provided elbv2.LoadBalancer
+func CreateLoadBalancerFromElbV2(elbv2 *elbv2.LoadBalancer) (lb *LoadBalancer) {
```
thoo5ieb: I am unsure. It looks like to be a helper that translate one format into another. If you really, want to go this path, you should make it private and don't expose as an API function.

thoo5ieb: In a perfect world, your test case should not interact with already created resources. It's better to create a resource you need for the test. You will make your test more independent. 

I don't think, you will need to have `Describe*` functions here. I would vote for Create* and Delete*. Once that done, you won't need `CreateLoadBalancerFromElbV2`.

thoo5ieb: Finally, start with small things, and once you expose your PR, ask maintainers if they're interested in additional functionality.

modules/aws/elbv2.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/stretchr/testify/require"
+)
+
+// LoadBalancer is an Amazon load balancer.
```
yorinasub17:
```suggestion
// LoadBalancer represents an Amazon ELB (Elastic Load Balancer).
```

thoo5ieb:
@yorinasub17 I will be happy to try to answer that question. Basically, for some set of services, we do use one ALB with path requests forward to specific backends. It influences the design of the module because the AWS ALB gets created by a different entity and by use of module output we attach different listener rules. 

So the module itself depends on the ALB. Does it answer your question?

yorinasub17:
Ah this is similar to a recent request: using terratest to provision temporary resources that is used by the module in the tests. While in the past we added in functions for this use case, in general we have been being more careful about which resource provisioning functions we should include. We have started opting for using wrapper modules in the form of examples that provision the necessary resources in Terraform. In addition to providing a self-contained test, this also serves the purpose of highlighting how one is intended to use the module in question. Take a look at my previous response to this for more details on our thought process: https://github.com/gruntwork-io/terratest/issues/315#issuecomment-498509556

Given that, are there some things in your environment that would be incompatible with this approach of using wrapper modules?

thoo5ieb:
@yorinasub17 To be honest, that's the best option. We had a discussion about it internally. I am so happy you adopt this option!

thoo5ieb:
Thank you for taking the time on this!

yorinasub17:
Going to close this. Please comment here (and reopen PR) if the wrapper module approach doesn't work out for you and you need these functions.

PR #278, opened 2019-04-19
Add initial module/aws/ssm

salekseev:
```
Mon Apr 22 14:04:55 EDT 2019
=== RUN   TestParameterIsFound
=== PAUSE TestParameterIsFound
=== CONT  TestParameterIsFound
TestParameterIsFound 2019-04-22T14:04:57-04:00 region.go:106: Looking up all AWS regions available in this account
TestParameterIsFound 2019-04-22T14:04:58-04:00 region.go:91: Using region eu-central-1
TestParameterIsFound 2019-04-22T14:04:59-04:00 ssm_test.go:21: Created parameter with version 1
TestParameterIsFound 2019-04-22T14:04:59-04:00 ssm_test.go:23: Found key with name test-name-lQ58m3
--- PASS: TestParameterIsFound (2.76s)
PASS
ok  	_/Users/salekseev/src/github.com/gruntwork-io/terratest/modules/aws	2.785s
```

### DIFF
--- a/modules/aws/ssm.go
+++ b/modules/aws/ssm.go
@@ -1,0 +1,72 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ssm"
+)
+
+func GetParameter(t *testing.T, awsRegion string, keyName string) string {
+	keyValue, err := GetParameterE(t, awsRegion, keyName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return keyValue
+}
+
+func GetParameterE(t *testing.T, awsRegion string, keyName string) (string, error) {
+	ssmClient, err := NewSsmClientE(t, awsRegion)
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := ssmClient.GetParameter(&ssm.GetParameterInput{Name: aws.String(keyName), WithDecryption: aws.Bool(true)})
+	if err != nil {
+		return "", err
+	}
+
+	parameter := *resp.Parameter
+	return *parameter.Value, nil
+}
+
+func PutParameter(t *testing.T, awsRegion string, keyName string, keyDescription string, keyValue string) int64 {
+	version, err := PutParameterE(t, awsRegion, keyName, keyDescription, keyValue)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return version
+}
+
+func PutParameterE(t *testing.T, awsRegion string, keyName string, keyDescription string, keyValue string) (int64, error) {
+	ssmClient, err := NewSsmClientE(t, awsRegion)
+	if err != nil {
+		return 0, err
+	}
+
+	resp, err := ssmClient.PutParameter(&ssm.PutParameterInput{Name: aws.String(keyName), Description: aws.String(keyDescription), Value: aws.String(keyValue), Type: aws.String("SecureString")})
+	if err != nil {
+		return 0, err
+	}
+
+	return *resp.Version, nil
+}
+
+// NewSsmClient creates a ssm client.
+func NewSsmClient(t *testing.T, region string) *ssm.SSM {
+	client, err := NewSsmClientE(t, region)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return client
+}
+
+// NewSsmClientE creates an ssm client.
+func NewSsmClientE(t *testing.T, region string) (*ssm.SSM, error) {
+	sess, err := NewAuthenticatedSession(region)
+	if err != nil {
+		return nil, err
+	}
+
+	return ssm.New(sess), nil
+}
--- a/modules/aws/ssm.go
+++ b/modules/aws/ssm.go
@@ -8,12 +8,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


modules/aws/ssm.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ssm"
+)
+
+func GetParameter(t *testing.T, awsRegion string, keyName string) string {
+	keyValue, err := GetParameterE(t, awsRegion, keyName)
+	if err != nil {
+		t.Fatal(err)
+	}
```
brikis98: We've started shortening this to a one liner: `require.NoError(t, err)`

modules/aws/ssm.go
```diff
@@ -0,0 +1,72 @@
...
+		return "", err
+	}
+
+	resp, err := ssmClient.GetParameter(&ssm.GetParameterInput{Name: aws.String(keyName), WithDecryption: aws.Bool(true)})
+	if err != nil {
+		return "", err
+	}
+
+	parameter := *resp.Parameter
+	return *parameter.Value, nil
```
brikis98: Should this use `aws.StringValue`?

brikis98: And do you need a `nil` check for the previous line?

salekseev:
> Nice, thank you!
> 
> One sanity check: am I reading AWS's pricing docs correctly in that storing data in parameter store doesn't cost anything? I know that AWS Secrets Manager charges for each secret, so with hundreds of test runs per month, that can add up quickly, but it seems like SSM is free?

SSM Parameter Store is indeed free but has a hard limit on API requests. Parameter Store currently has 40 GET API (shared) requests/sec limit per account.

modules/aws/ssm.go
salekseev: Empty value versions are not allowed in SSM Parameter Store (https://forums.aws.amazon.com/message.jspa?messageID=849170) and you would instead get an error 400 is a version does not exist. I don't think it's necessary to check for `nul` values here in this case.

modules/aws/ssm.go
salekseev: I have updated PR to use the recommended pattern.

modules/aws/ssm.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"github.com/stretchr/testify/require"
+)
+
+func GetParameter(t *testing.T, awsRegion string, keyName string) string {
```
brikis98: One last thing: could you add comments to these new methods explaining that they are reading parameters from SSM?

salekseev: @brikis98 Totally. Sorry forgot to do it initially. Should be good to go.

brikis98:
Great, thanks! I just kicked off the automated tests in CI.

brikis98:
Tests are failing with some sort of Packer build issue. I don't have the bandwidth to dig into it right now, so I'll have to come back to this later 😕

salekseev:
Is there anything I can help with? It does not look to be related to my change but maybe I am misreading it.

yorinasub17:
It looks the primary test failure is actually in the `aws` module (specifically, there is a circular import somewhere in that module). The failing packer tests are known unstable tests that will eventually succeed.

Here is a minimal repro:
```
%~> cd modules/aws
%~> go test -v -run Foo .
# github.com/gruntwork-io/terratest/modules/aws
import cycle not allowed in test
package github.com/gruntwork-io/terratest/modules/aws (test)
        imports github.com/gruntwork-io/terratest/modules/aws
FAIL    github.com/gruntwork-io/terratest/modules/aws [setup failed]
```

salekseev:
> It looks the primary test failure is actually in the `aws` module (specifically, there is a circular import introduced in this code somewhere in that module). The failing packer tests are known unstable tests that will eventually succeed.
> 
> Here is a minimal repro:
> 
> ```
> %~> cd modules/aws
> %~> go test -v -run Foo .
> # github.com/gruntwork-io/terratest/modules/aws
> import cycle not allowed in test
> package github.com/gruntwork-io/terratest/modules/aws (test)
>         imports github.com/gruntwork-io/terratest/modules/aws
> FAIL    github.com/gruntwork-io/terratest/modules/aws [setup failed]
> ```

I am unable to reproduce the failure. I have rebased on latest `master` just in case.

modules/aws/ssm_test.go
```diff
@@ -0,0 +1,26 @@
+package aws
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/aws"
```
yorinasub17: This line is the issue. Since we are already in package `aws`, there is no need to import this. I am not sure why you don't get the import error though. Perhaps you have a different version of `go`?

salekseev: @yorinasub17 Good catch. I guess `go1.12.4` was hiding my error. Easy fix though. Thanks.

yorinasub17: Thanks for the fix! I just kicked off the build. Let's see if it passes this time!

PR #279, opened 2019-04-19
Update destory.go

abhaydiwan:
There is no option -auto-approve with destroy command.
It has to be -force.


brikis98:
I think `-force` is deprecated in favor of `-auto-approve`:

```
# terraform destroy --help
Usage: terraform destroy [options] [DIR]

  Destroy Terraform-managed infrastructure.

Options:

  -backup=path           Path to backup the existing state file before
                         modifying. Defaults to the "-state-out" path with
                         ".backup" extension. Set to "-" to disable backup.

  -auto-approve          Skip interactive approval before destroying.

  -force                 Deprecated: same as auto-approve.
```

PR #281, opened 2019-04-24
Use STS client for GetAccountId

### LINKED ISSUE #139: GetAccountId method should use GetCallerIdentity instead of GetUser
The `GetAccountId` method currently calls `iam.GetUser` and extracts the account ID from the user's ARN:

https://github.com/gruntwork-io/terratest/blob/master/modules/aws/account.go#L27

This means the method only works when authenticated as an IAM User, but will not work if you use, say, IAM Roles for auth. A better approach would be to use `sts.GetCallerIdentity`, which works with all auth mechanisms, and returns the account ID as a field, requiring no extra extraction.

brikis98:
Fixes #139. The `GetAccountIdE` method was originally using `iamClient.GetUser` to try to get the current AWS account, but this method doesn't work if you're authenticated via an IAM Role or something that isn't an IAM User. This PR updates the method to use `sts.GetCallerIdentity`, which should work no matter how you're authenticated.

### DIFF
--- a/modules/aws/account.go
+++ b/modules/aws/account.go
@@ -2,10 +2,10 @@ package aws
 
 import (
 	"errors"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"strings"
 	"testing"
-
-	"github.com/aws/aws-sdk-go/service/iam"
 )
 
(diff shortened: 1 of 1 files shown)


brikis98:
Looks like tests passed... So I'll risk the merge. Thx for the review!

PR #282, opened 2019-04-30
Disabling the verify of https for load balancer domains

geass:
**Problem**

If you want to get an endpoint of load balancer typed service and send a HTTP request to it, you might get TLS validation problem because of your cloud provider load balancer. 

**Sample Error**

x509: certificate is valid for *.abc.com,  not abc-foo.eu-central-1.elb.amazonaws.com. Sleeping for 10s and will try again.

**Solution**

`http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
`

best,

### DIFF
--- a/modules/http-helper/http_helper.go
+++ b/modules/http-helper/http_helper.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"testing"
 	"time"
+	"crypto/tls"
 
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/retry"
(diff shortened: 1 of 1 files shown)


modules/http-helper/http_helper.go
```diff
@@ -26,6 +27,8 @@ func HttpGet(t *testing.T, url string) (int, string) {
 func HttpGetE(t *testing.T, url string) (int, string, error) {
 	logger.Logf(t, "Making an HTTP GET call to URL %s", url)
 
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
yorinasub17: We actually have a few tests in our IaC library that checks the TLS cert is valid, so we definitely don't want to do this globally. I think it would be more flexible to accept the TLS config as a parameter to the function. This allows you to not only optionally disable TLS verification, but also support custom CAs.

This will be a backwards incompatible change, but I think we can live with it given the additional functionality it enables.

Note that this requires updating all the functions in this file to support accepting the parameter.

drmmarsunited: I just added your suggested fixes in #336 @yorinasub17 

This is blocking some of our testing as well, so I thought I would pick it up and finish it.

yorinasub17:
This PR has been superseded by https://github.com/gruntwork-io/terratest/pull/336.

PR #283, opened 2019-05-01
Extend AWS ECS support: service and task definition getters, example test case

shcherbin:
This PR extends the current AWS ECS module with:
- functions to get a particular AWS ECS service or task definition
- terraform module to deploy a service to an ECS cluster
- example test case to validate the module

the output of the integration test run can be found [here](https://gist.github.com/shcherbin/acfa89acdea5f99ab8e1632f7f4e1f4d).


modules/aws/ecs.go
```diff
@@ -36,7 +36,7 @@ func GetEcsClusterE(t *testing.T, region string, name string) (*ecs.Cluster, err
 
 	numClusters := len(output.Clusters)
 	if numClusters != 1 {
-		return nil, fmt.Errorf("Expected to find 1 ECS cluster named '%s' in region '%v', but found '%d'",
+		return nil, fmt.Errorf("expected to find 1 ECS cluster named '%s' in region '%v', but found '%d'",
```
yorinasub17: We like to keep error messages as sentences. Can you keep the capitalization here and also in all the new functions below?

modules/aws/ecs.go
```diff
@@ -90,6 +90,56 @@ func DeleteEcsClusterE(t *testing.T, region string, cluster *ecs.Cluster) error
 	return err
 }
 
+// GetEcsService fetches information about specified ECS service.
+func GetEcsService(t *testing.T, region, clusterName, serviceName string) *ecs.Service {
+	service, err := GetEcsServiceE(t, region, clusterName, serviceName)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yorinasub17: I know this file doesn't have this pattern yet, but we've started shortening this to a one liner: `require.NoError(t, err)`. Can you apply to these functions?

modules/aws/ecs.go
```diff
@@ -90,6 +90,56 @@ func DeleteEcsClusterE(t *testing.T, region string, cluster *ecs.Cluster) error
 	return err
 }
 
+// GetEcsService fetches information about specified ECS service.
+func GetEcsService(t *testing.T, region, clusterName, serviceName string) *ecs.Service {
```
yorinasub17: We actually like to set the type on each variable (e.g `region string, clusterName string, serviceName string`), even if they are the same. It is more verbose, but we find that it reads better. Can you update?

test/terraform_aws_ecs_example_test.go
```diff
@@ -0,0 +1,64 @@
...
+	// This will run `terraform init` and `terraform apply` and fail the test if there are any errors
+	terraform.InitAndApply(t, terraformOptions)
+
+	// Run `terraform output` to get the value of an output variable
+	taskDefinition := terraform.Output(t, terraformOptions, "task_definition")
+
+	// Look up the ECS cluster by name
+	cluster := aws.GetEcsCluster(t, awsRegion, expectedClusterName)
+
+	assert.Equal(t, int64(1), *cluster.ActiveServicesCount)
```
yorinasub17: It is safer to use `aws.Int64Value` and the like instead of directly dereferencing the pointers, as it handles the `nil` check for you. Can you update the file to use `aws.Int64Value` and its siblings wherever you dereference pointers?

modules/aws/ecs.go
shcherbin: I knew this would trigger a discussion! Intellij idea reported as a linting error so I fixed that. 
OK, will do. Following a convention makes much more sense.

modules/aws/ecs.go
shcherbin: Done.

modules/aws/ecs.go
shcherbin: Done.

test/terraform_aws_ecs_example_test.go
shcherbin: `github.com/aws/aws-sdk-go/aws` conflicts with `terratest/modules/aws`. I had to use an alias `awsSDK` if that's ok.

yorinasub17: Yup that is fine!

yorinasub17:
Ok the changes look good to me! I am going to kick off our build now. You should be able to view the progress once it is started.

test/terraform_aws_ecs_example_test.go
```diff
@@ -0,0 +1,66 @@
+package test
+
+import (
+	"fmt"
+	"terratest/modules/aws"
+	"terratest/modules/random"
+	"terratest/modules/terraform"
```
yorinasub17: The test failed due to an import error here. I missed this in the first go around, but when using `dep`, the source is checked out under `github.com/gruntwork-io/terratest`, which means these need to include that prefix. E.g `terratest/modules/aws` => `github.com/gruntwork-io/terratest/modules/aws`.

shcherbin: I've fixed the imports, but when I run `dep ensure` it adds terratest to the lock file. Is this behavior expected? 
See my reverted commit: https://github.com/gruntwork-io/terratest/pull/283/commits/8369f7e3520e4bcadf962fad242dddcf69d69133

yorinasub17: Hmm that is unexpected. It does not happen for me when I run `dep ensure`. What version of `dep` are you using? Run `dep version` to see.

shcherbin: @yorinasub17 nevermind, it works fine if the project is cloned into 
`${GOPATH}/src/github.com/gruntwork-io/terratest`

golang is not my primary language, sorry :)

yorinasub17: Ah yes that is a common gotcha in Go: the significance of `GOPATH`.

shcherbin: yup, I thought it was unexpected too, that's why I've reverted the commit. 
Any plans on migrating to `go mod`?

yorinasub17: You can see our recent discussion on `go mod` for our current thought process: https://github.com/gruntwork-io/terratest/pull/264

yorinasub17:
https://github.com/gruntwork-io/terratest/releases/tag/v0.15.1

PR #284, opened 2019-05-02
Add support for Terraform 12

tonerdo:
This PR updates the formatting for variables passed using the `-var` option by removing the surrounding quotes. Text that contain spaces are still passed properly. This change also works with Terraform 11.

### DIFF
--- a/modules/terraform/format.go
+++ b/modules/terraform/format.go
@@ -157,6 +157,6 @@ func primitiveToHclString(value interface{}) string {
 	//case int: return strconv.Itoa(v)
 
 	default:
-		return fmt.Sprintf("\"%v\"", v)
+		return fmt.Sprintf("%v", v)
 	}
 }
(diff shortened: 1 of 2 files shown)


yorinasub17:
Looks like unit tests are failing. Also, how hard is to add a simple tf12 test as well?

tonerdo:
> Looks like unit tests are failing

Yeah, looking into them. It's made me think of more advanced scenarios. If this is a Terraform 12 upgrade change then it probably goes beyond removing surrounding quotes. Will test out more extensive scenarios

brikis98:
> This change also works with Terraform 11.

Really? I remember putting in those quotes for a reason... I believe that when I originally wrote that code, Terraform would not parse HCL values (especially strings) that weren't wrapped in quotes... But I'm not sure if that was version 9 or 10 or 11...

BTW, if this turns out to behave differently with Terraform 11 and 12, we can have Terratest detect the version of Terraform that's installed (by running `terraform --version`) and adjust its behavior accordingly.

sumitmaggo:
Hi @tonerdo , @brikis98 , this change is not backward compatible with Terraform 0.11.x version. Is there any plan to fix this to support terratest with Terraform 0.11.x version?

tonerdo:
@sumitmaggo yeah, I'm looking into fixing this today

PR #174, opened 2018-10-15
Automatically use user's ssh keys in EC2 KeyPairs

eak12913:
Added a new feature where if you specify an environment variable: `USE_MY_KEYS_PLEASE=priave_key_name` then terratest will grab the specified `$HOME/.ssh/priave_key_name` and upload it to your test EC2 instances instead of generating random private keys.

The idea here is that when developing/running tests locally, users often bring up infrastructure with terratest and then leave that infrastructure up while they troubleshoot/make changes.

Terratest creates random keypairs each time and this makes it a little tricky to grab the randomly generated private key and set proper formatting and file permissions to use in a command line `ssh` command.

This PR seeks to make it easy to ssh to your locally deployed EC2 instances without having to change any of your existing test code.

PR #287, opened 2019-05-08
modules/aws: add support for SSH agent (as alternative to EC2 key pair)

CpuID:
Adds support for SSH Agent (system or terratest spawned) for fetching files in AWS module


CpuID:
Not directly related to https://github.com/gruntwork-io/terratest/pull/174 or https://github.com/gruntwork-io/terratest/pull/173 - and skimming over it looks like they are non-conflicting (touches different parts of the codebase).

CpuID:
For prior SSH Key use case:

```
  sshKeypair := ssh.KeyPair{
    PrivateKey: "PEM-encoded-private-key-string",
    PublicKey:  "PEM-encoded-public-key-string",
  }
  keyPair := aws.Ec2Keypair{
    KeyPair: &sshKeypair,
    Name:    "some key",
    Region:  "", 
  }
  log.Printf("%+v\n", aws.FetchContentsOfFileFromAsg(t, awsRegion, "username", &keyPair, asgName, true, "/etc/os-release"))
```

New usage:

```
  sshKeypair := ssh.KeyPair{
    PrivateKey: "PEM-encoded-private-key-string",
    PublicKey:  "PEM-encoded-public-key-string",
  }
  keyPair := aws.Ec2Keypair{
    KeyPair: &sshKeypair,
    Name:    "some key",
    Region:  "", 
  }
```

And the slightly different syntax:

```
  log.Printf("%+v\n", aws.FetchContentsOfFileFromAsg(t, awsRegion, "username", SshAuth{KeyPair: &keyPair}, asgName, true, "/etc/os-release"))
```

Or to use the SSH Agent instead:

```
  log.Printf("%+v\n", aws.FetchContentsOfFileFromAsg(t, awsRegion, "username", SshAuth{SshAgent: true}, asgName, true, "/etc/os-release"))
```

CpuID:
hold off on a review here, I want to tweak this a little further:

- rename `SshAuth` to `SshOptions`
- move SSH username into `SshOptions`
- add `SshPort` to `SshOptions` (need to augment in `modules/ssh` also, hard set to 22 right now)

CpuID:
OK, good for review now, thx

CpuID:
@brikis98 *bump* able to review? thx

modules/aws/ec2-files.go
```diff
@@ -15,85 +16,170 @@ type RemoteFileSpecification struct {
...
-	KeyPair                *Ec2Keypair
+	SshOptions             *SshOptions
 	LocalDestinationDir    string //base path where to store downloaded artifacts locally. The final path of each resource will include the ip of the host and the name of the immediate parent folder.
 }
 
+// Specify one of KeyPair, SshAgent, or OverrideSshAgent should be specified
+// Any more (or less) than one will result in error on use
+type SshOptions struct {
+	// Username
+	UserName string
```
brikis98: There's already an `SshUser` above. I assume one or the other should be removed?

modules/aws/ec2-files.go
```diff
@@ -15,85 +16,170 @@ type RemoteFileSpecification struct {
...
 	UseSudo                bool
 	SshUser                string
-	KeyPair                *Ec2Keypair
+	SshOptions             *SshOptions
 	LocalDestinationDir    string //base path where to store downloaded artifacts locally. The final path of each resource will include the ip of the host and the name of the immediate parent folder.
 }
 
+// Specify one of KeyPair, SshAgent, or OverrideSshAgent should be specified
+// Any more (or less) than one will result in error on use
+type SshOptions struct {
```
brikis98: Should this be specified in the `aws` package? Seems like a better fit for the `ssh` package...

modules/aws/ec2-files.go
```diff
@@ -15,85 +16,170 @@ type RemoteFileSpecification struct {
...
+	// EC2 Key Pair
+	KeyPair *Ec2Keypair
+	// If true, use default SSH agent on local system (started externally, available at SSH_AUTH_SOCK)
+	SshAgent bool
+	// Override SSH agent on local system
+	OverrideSshAgent *ssh.SshAgent
+
+	// The enabled authentication method
+	// Internally will be one of "keypair", "sshagent", or "overridesshagent"
+	enabledAuthMethod string
```
brikis98: Should this use an enum instead?

modules/aws/ec2-files.go
```diff
@@ -15,85 +16,170 @@ type RemoteFileSpecification struct {
...
+}
+
+func (s *SshOptions) Validate() error {
```
brikis98: Please use full names (e.g., `opts` or `sshOpts`) instead of single-character names.

modules/aws/ec2-files.go
```diff
@@ -15,85 +16,170 @@ type RemoteFileSpecification struct {
...
+	if s.UserName == "" {
+		return fmt.Errorf("SshOptions: UserName field cannot be empty")
```
brikis98: Please define custom, exported error types for these errors so it's possible for others to check the type of an error when handling it (rather than relying solely on string error messages).

modules/aws/ec2-files.go
```diff
@@ -15,85 +16,170 @@ type RemoteFileSpecification struct {
...
+			return multipleError
+		}
+		s.enabledAuthMethod = "overridesshagent"
+		return nil
+	}
+	return fmt.Errorf("Unexpected error validating SshOptions struct")
+}
+
+// Attaches the correct authentication method to an ssh.Host struct instance
+func addOptsToSshHost(t *testing.T, sshHost *ssh.Host, sshOpts *SshOptions) {
```
brikis98: Please rename to `addOptsToSshHostE` that returns errors rather than calling `t.Fatal` directly. See the README for details on `E` methods.

brikis98: Also, instead of modifying `sshHost` in place, is it possible to return a new one? Immutability tends to be easier to reason about, especially in an environment with many tests running concurrently.

modules/ssh/ssh.go
```diff
@@ -55,10 +63,11 @@ func ScpFileToE(t *testing.T, host Host, mode os.FileMode, remotePath, contents
 	}
 	dir, file := filepath.Split(remotePath)
 
+	host.setDefaultPort()
```
brikis98: Forgetting to call this seems like a VERY likely gotcha we'll hit in the future. Any ideas on how to avoid that issue?

modules/aws/ec2-files.go
brikis98: In fact, the `ssh` package already has most of these fields defined in a struct. Can that be reused?

modules/aws/ec2-files.go
CpuID: correct, good catch

modules/aws/ec2-files.go
CpuID: possibly, I can see the value in reuse there (other cloud providers would likely want the same thing)

let me review the code between the two modules in more detail, I'll get back to you. it definitely did feel a bit duplicated when writing it

modules/aws/ec2-files.go
CpuID: I'm going to double check before removing also, just to make sure nothing else touches `RemoteFileSpecification` that expects it

modules/aws/ec2-files.go
CpuID: probably, yea. I haven't used them to date but the use case makes sense. I'll refactor

modules/aws/ec2-files.go
CpuID: interesting thought, and yea I can do that. would make testing easier also so I can see the value

modules/ssh/ssh.go
CpuID: 👍 you kind of want it set upfront on `ssh.Host` instantiation. the only sane option I had was make sure it's called before the Port is used in the same method.

SO tells me the best options for constructors are: https://stackoverflow.com/a/18125763

- make the zero value a sensible default
- use a `New()` method

I'd be open to doing the `New() approach? I don't know if we can enforce using it though, without making `Host` private (a la `host`), but that means you can't touch it from your own codebase which is not optimal...

CpuID:
@brikis98 great feedback, thx. I'll likely try work through that early next week and let you know once it's ready for another review

CpuID:
(Mental notes for myself for refactoring here)

TODO:

- [ ] replace `type SshOptions struct` with `type Host struct` from `modules/ssh/ssh.go`
  - [ ] `SshOptions` may be a superset of `Host`? if so, extend `Host` to avoid redefining the underlying properties
    - `Ec2Keypair` comes to mind...
  - [ ] port across usage of `SshOptions` to either `Host` or the new parent type respectively (if a superset of properties)
- [ ] add a `func (host *Host) Validate() error {}` function to `ssh.go`?
  - take inspiration from `func (s *SshOptions) Validate() error {`
- [ ] PR changing `type Host struct` from `Port int` to `Port uint16` independently of this PR (to be scrutinised/merged async)
  - I can live with `int` for now, others may have opinions on this. Would prefer to not delay getting the core functionality out there.

CpuID:
Few errors to deal with here (notes for myself):

```
# github.com/gruntwork-io/terratest/test [github.com/gruntwork-io/terratest/test.test]
./terraform_gcp_example_test.go:141:3: cannot use promoted field Options.SshKeyPair in struct literal of type "github.com/gruntwork-io/terratest/modules/ssh".Host
./terraform_gcp_example_test.go:142:3: cannot use promoted field Options.SshUserName in struct literal of type "github.com/gruntwork-io/terratest/modules/ssh".Host
./terraform_redeploy_example_test.go:208:82: cannot use "ubuntu" (type string) as type *"github.com/gruntwork-io/terratest/modules/ssh".Options in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_redeploy_example_test.go:208:82: cannot use keyPair (type *"github.com/gruntwork-io/terratest/modules/aws".Ec2Keypair) as type string in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_redeploy_example_test.go:208:82: cannot use asgName (type string) as type bool in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_redeploy_example_test.go:208:82: cannot use true (type bool) as type string in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_scp_example_test.go:116:3: cannot use promoted field Options.SshKeyPair in struct literal of type "github.com/gruntwork-io/terratest/modules/ssh".Host
./terraform_scp_example_test.go:117:3: cannot use promoted field Options.SshUserName in struct literal of type "github.com/gruntwork-io/terratest/modules/ssh".Host
./terraform_scp_example_test.go:192:3: cannot use promoted field Options.SshKeyPair in struct literal of type "github.com/gruntwork-io/terratest/modules/ssh".Host
./terraform_scp_example_test.go:193:3: cannot use promoted field Options.SshUserName in struct literal of type "github.com/gruntwork-io/terratest/modules/ssh".Host
./terraform_scp_example_test.go:193:3: too many errors
```

CpuID:
Remaining errors to handle (notes for myself):

```
$ go test -run TestTerraformSshExample
# github.com/gruntwork-io/terratest/test [github.com/gruntwork-io/terratest/test.test]
./terraform_redeploy_example_test.go:208:82: cannot use "ubuntu" (type string) as type *"github.com/gruntwork-io/terratest/modules/ssh".Options in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_redeploy_example_test.go:208:82: cannot use keyPair (type *"github.com/gruntwork-io/terratest/modules/aws".Ec2Keypair) as type string in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_redeploy_example_test.go:208:82: cannot use asgName (type string) as type bool in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_redeploy_example_test.go:208:82: cannot use true (type bool) as type string in argument to "github.com/gruntwork-io/terratest/modules/aws".FetchContentsOfFilesFromAsg
./terraform_scp_example_test.go:251:3: unknown field 'SshUser' in struct literal of type "github.com/gruntwork-io/terratest/modules/aws".RemoteFileSpecification
./terraform_scp_example_test.go:254:4: cannot use keyPair (type *"github.com/gruntwork-io/terratest/modules/aws".Ec2Keypair) as type *"github.com/gruntwork-io/terratest/modules/ssh".KeyPair in field value
FAIL	github.com/gruntwork-io/terratest/test [build failed]
```

CpuID:
@brikis98 ready for review again, hopefully good this time :) I agree that a lot of this could reuse stuff in `modules/ssh`, so I've taken that approach this time.

Hoping to get this over the line so I can get off my fork :)

brikis98:
Thanks! I'm a bit buried today, but I'll take a look next week.

modules/aws/ec2-files.go
```diff
@@ -15,86 +15,87 @@ type RemoteFileSpecification struct {
...
+	SshOptions             *ssh.Options
 	LocalDestinationDir    string //base path where to store downloaded artifacts locally. The final path of each resource will include the ip of the host and the name of the immediate parent folder.
 }
 
 // FetchContentsOfFileFromInstance looks up the public IP address of the EC2 Instance with the given ID, connects to
-// the Instance via SSH using the given username and Key Pair, fetches the contents of the file at the given path
-// (using sudo if useSudo is true), and returns the contents of that file as a string.
-func FetchContentsOfFileFromInstance(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, instanceID string, useSudo bool, filePath string) string {
-	out, err := FetchContentsOfFileFromInstanceE(t, awsRegion, sshUserName, keyPair, instanceID, useSudo, filePath)
+// the Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or Override SSH Agent auth methods,
```
yorinasub17: NIT:
```suggestion
// the Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or Override SSH Agent auth methods),
```

modules/aws/ec2-files.go
```diff
@@ -15,86 +15,87 @@ type RemoteFileSpecification struct {
...
 		t.Fatal(err)
 	}
 	return out
 }
 
 // FetchContentsOfFileFromInstanceE looks up the public IP address of the EC2 Instance with the given ID, connects to
-// the Instance via SSH using the given username and Key Pair, fetches the contents of the file at the given path
-// (using sudo if useSudo is true), and returns the contents of that file as a string.
-func FetchContentsOfFileFromInstanceE(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, instanceID string, useSudo bool, filePath string) (string, error) {
+// the Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or Override SSH Agent auth methods,
```
yorinasub17:
```suggestion
// the Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or Override SSH Agent auth methods),
```

modules/aws/ec2-files.go
```diff
@@ -15,86 +15,87 @@ type RemoteFileSpecification struct {
...
 
 	return ssh.FetchContentsOfFileE(t, host, useSudo, filePath)
 }
 
 // FetchContentsOfFilesFromInstance looks up the public IP address of the EC2 Instance with the given ID, connects to
-// the Instance via SSH using the given username and Key Pair, fetches the contents of the files at the given paths
-// (using sudo if useSudo is true), and returns a map from file path to the contents of that file as a string.
-func FetchContentsOfFilesFromInstance(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, instanceID string, useSudo bool, filePaths ...string) map[string]string {
-	out, err := FetchContentsOfFilesFromInstanceE(t, awsRegion, sshUserName, keyPair, instanceID, useSudo, filePaths...)
+// the Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or Override SSH Agent auth methods,
```
yorinasub17:
```suggestion
// the Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or Override SSH Agent auth methods),
```

modules/aws/ec2-files.go
```diff
@@ -15,86 +15,87 @@ type RemoteFileSpecification struct {
...
 		t.Fatal(err)
 	}
 	return out
 }
 
 // FetchContentsOfFilesFromInstanceE looks up the public IP address of the EC2 Instance with the given ID, connects to
-// the Instance via SSH using the given username and Key Pair, fetches the contents of the files at the given paths
-// (using sudo if useSudo is true), and returns a map from file path to the contents of that file as a string.
-func FetchContentsOfFilesFromInstanceE(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, instanceID string, useSudo bool, filePaths ...string) (map[string]string, error) {
+// the Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or Override SSH Agent auth methods,
```
yorinasub17:
```suggestion
// the Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or Override SSH Agent auth methods),
```

modules/aws/ec2-files.go
```diff
@@ -15,86 +15,87 @@ type RemoteFileSpecification struct {
...
 }
 
 // FetchContentsOfFileFromAsg looks up the EC2 Instances in the given ASG, looks up the public IPs of those EC2
-// Instances, connects to each Instance via SSH using the given username and Key Pair, fetches the contents of the file
-// at the given path (using sudo if useSudo is true), and returns a map from Instance ID to the contents of that file
-// as a string.
-func FetchContentsOfFileFromAsg(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, asgName string, useSudo bool, filePath string) map[string]string {
-	out, err := FetchContentsOfFileFromAsgE(t, awsRegion, sshUserName, keyPair, asgName, useSudo, filePath)
+// Instances, connects to each Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or
+// Override SSH Agent auth methods, fetches the contents of the file at the given path (using sudo if useSudo is true),
```
yorinasub17:
```suggestion
// Instances, connects to each Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or
// Override SSH Agent auth methods), fetches the contents of the file at the given path (using sudo if useSudo is true),
```

modules/aws/ec2-files.go
```diff
@@ -15,86 +15,87 @@ type RemoteFileSpecification struct {
...
 	return out
 }
 
 // FetchContentsOfFileFromAsgE looks up the EC2 Instances in the given ASG, looks up the public IPs of those EC2
-// Instances, connects to each Instance via SSH using the given username and Key Pair, fetches the contents of the file
-// at the given path (using sudo if useSudo is true), and returns a map from Instance ID to the contents of that file
-// as a string.
-func FetchContentsOfFileFromAsgE(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, asgName string, useSudo bool, filePath string) (map[string]string, error) {
+// Instances, connects to each Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or
+// Override SSH Agent auth methods, fetches the contents of the file at the given path (using sudo if useSudo is true),
```
yorinasub17:
```suggestion
// Instances, connects to each Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or
// Override SSH Agent auth methods), fetches the contents of the file at the given path (using sudo if useSudo is true),
```

modules/aws/ec2-files.go
```diff
@@ -114,22 +115,22 @@ func FetchContentsOfFileFromAsgE(t testing.TestingT, awsRegion string, sshUserNa
 }
 
 // FetchContentsOfFilesFromAsg looks up the EC2 Instances in the given ASG, looks up the public IPs of those EC2
-// Instances, connects to each Instance via SSH using the given username and Key Pair, fetches the contents of the files
-// at the given paths (using sudo if useSudo is true), and returns a map from Instance ID to a map of file path to the
-// contents of that file as a string.
-func FetchContentsOfFilesFromAsg(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, asgName string, useSudo bool, filePaths ...string) map[string]map[string]string {
-	out, err := FetchContentsOfFilesFromAsgE(t, awsRegion, sshUserName, keyPair, asgName, useSudo, filePaths...)
+// Instances, connects to each Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or
+// Override SSH Agent auth methods, fetches the contents of the files at the given paths (using sudo if useSudo is true),
```
yorinasub17:
```suggestion
// Instances, connects to each Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or
// Override SSH Agent auth methods), fetches the contents of the files at the given paths (using sudo if useSudo is true),
```

modules/aws/ec2-files.go
```diff
@@ -114,22 +115,22 @@ func FetchContentsOfFileFromAsgE(t testing.TestingT, awsRegion string, sshUserNa
...
 	return out
 }
 
 // FetchContentsOfFilesFromAsgE looks up the EC2 Instances in the given ASG, looks up the public IPs of those EC2
-// Instances, connects to each Instance via SSH using the given username and Key Pair, fetches the contents of the files
-// at the given paths (using sudo if useSudo is true), and returns a map from Instance ID to a map of file path to the
-// contents of that file as a string.
-func FetchContentsOfFilesFromAsgE(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, asgName string, useSudo bool, filePaths ...string) (map[string]map[string]string, error) {
+// Instances, connects to each Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or
+// Override SSH Agent auth methods, fetches the contents of the files at the given paths (using sudo if useSudo is true),
```
yorinasub17:
```suggestion
// Instances, connects to each Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or
// Override SSH Agent auth methods), fetches the contents of the files at the given paths (using sudo if useSudo is true),
```

modules/aws/ec2-files.go
```diff
@@ -149,32 +150,31 @@ func FetchContentsOfFilesFromAsgE(t testing.TestingT, awsRegion string, sshUserN
 }
 
 // FetchFilesFromInstance looks up the EC2 Instances in the given ASG, looks up the public IPs of those EC2
-// Instances, connects to each Instance via SSH using the given username and Key Pair, downloads the files
-// matching filenameFilters at the given remoteDirectory (using sudo if useSudo is true), and stores the files locally
-// at localDirectory/<publicip>/<remoteFolderName>
-func FetchFilesFromInstance(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, instanceID string, useSudo bool, remoteDirectory string, localDirectory string, filenameFilters []string) {
-	err := FetchFilesFromInstanceE(t, awsRegion, sshUserName, keyPair, instanceID, useSudo, remoteDirectory, localDirectory, filenameFilters)
+// Instances, connects to each Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or
+// Override SSH Agent auth methods, downloads the files matching filenameFilters at the given remoteDirectory
```
yorinasub17:
```suggestion
// Instances, connects to each Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or
// Override SSH Agent auth methods), downloads the files matching filenameFilters at the given remoteDirectory
```

modules/aws/ec2-files.go
```diff
@@ -149,32 +150,31 @@ func FetchContentsOfFilesFromAsgE(t testing.TestingT, awsRegion string, sshUserN
...
 	}
 }
 
 // FetchFilesFromInstanceE looks up the EC2 Instances in the given ASG, looks up the public IPs of those EC2
-// Instances, connects to each Instance via SSH using the given username and Key Pair, downloads the files
-// matching filenameFilters at the given remoteDirectory (using sudo if useSudo is true), and stores the files locally
-// at localDirectory/<publicip>/<remoteFolderName>
-func FetchFilesFromInstanceE(t testing.TestingT, awsRegion string, sshUserName string, keyPair *Ec2Keypair, instanceID string, useSudo bool, remoteDirectory string, localDirectory string, filenameFilters []string) error {
+// Instances, connects to each Instance via SSH using the given username and one or more of: Key Pair, SSH Agent or
+// Override SSH Agent auth methods, downloads the files matching filenameFilters at the given remoteDirectory
```
yorinasub17:
```suggestion
// Instances, connects to each Instance via SSH (using the given username and one or more of Key Pair, SSH Agent or
// Override SSH Agent auth methods), downloads the files matching filenameFilters at the given remoteDirectory
```

modules/ssh/ssh.go
```diff
@@ -20,12 +20,18 @@ import (
...
 
-// Host is a remote host.
+// Host is a remote host, with some SSH options attached
 type Host struct {
-	Hostname    string // host name or ip address
+	*Options
+	Hostname string // host name or ip address
+}
+
+// SSH (connection) options, used for connecting to hosts
```
yorinasub17: NIT: This is confusing when there is also a `SshConnectionOptions` struct in the same module. Can we call this `AuthOptions` instead?

yorinasub17: Also, `CustomPort` probably belongs with `Host`

CpuID: @yorinasub17 valid feedback, and yea I was similarly torn re `SshConnectionOptions` and this struct. `AuthOptions` works4me, and I'll move `CustomPort` into `Host`. Give me a few mins

test/terraform_scp_example_test.go
```diff
@@ -112,9 +112,11 @@ func testScpDirFromHost(t *testing.T, terraformOptions *terraform.Options, keyPa
 	// as we know the Instance is running an Ubuntu AMI that has such a user
 	sshUserName := "ubuntu"
 	publicHost := ssh.Host{
-		Hostname:    publicInstanceIP,
-		SshKeyPair:  keyPair.KeyPair,
-		SshUserName: sshUserName,
+		Hostname: publicInstanceIP,
+		Options: &ssh.Options{
```
yorinasub17: It looks like you forgot to update these.

CpuID: on it

yorinasub17:
It looks like the following modules failed to build: `test`, `aws`, and `helm`. You can test these to get the compile errors (without running the tests) by going into each module and running:

```
go test -v -run TestFoo .
```

CpuID:
@yorinasub17 I've fixed the build errors mentioned, but I realised in `modules/aws/ec2-files.go` since moving `CustomPort` into `ssh.Host` I should expose the port so it can be specified in the `aws.Fetch*` methods respectively there. I'll sort that out then request another review - thanks.

CpuID:
@yorinasub17 ref https://github.com/gruntwork-io/terratest/pull/287#discussion_r397594813 - by moving `CustomPort` into `Host`, this means it needs to be explicitly added as an extra field to functions like this in `modules/aws`:

```
func FetchContentsOfFileFromInstance(t testing.TestingT, awsRegion string, sshOpts *ssh.AuthOptions, instanceID string, useSudo bool, filePath string) string {
func FetchContentsOfFileFromInstanceE(t testing.TestingT, awsRegion string, sshOpts *ssh.AuthOptions, instanceID string, useSudo bool, filePath string) (string, error) {
<and 8 more functions>
```

as there is no ability to specify a default function argument value or omit a function argument in Golang... that would require specifying `, int(22)` or `, 22` in the function calls above for all usage.

On the flipside, if it were in the `ssh.AuthOptions` struct it can optionally be specified, and if its not specified the `getPort()` function within `modules/ssh` handles setting it to a default value of 22.

Another "wildcard" use case to consider, if `CustomPort` is in `ssh.AuthOptions` it means you need a unique `ssh.AuthOptions{}` instance for different instances with different port numbers for connectivity.

In our use case, all hosts are available on SSH on the same port number (not 22), but I imagine that *could* differ by end user.

@yorinasub17 whats your feelings based on this, re which direction to go? I'm torn, as I can see the UX friction of both solutions :(

yorinasub17:
@CpuID thanks for your thoughtful take on the UX here! Those are good points. Perhaps there should be a struct like `EC2SSHOptions` which embeds `AuthOptions`, `useSudo` and `CustomPort`? That also gives us the flexibility to extend it to other cases like jump hosting through a bastion.

Issue #288, opened 2019-05-08
Feature Request: Add functionality to check S3 bucket versioning


drmmarsunited:
In the current `aws` module, the helper functions available for looking at / checking S3 bucket attributes are non-existent.  In the first of a few checks, I'd like to start adding support for people to ensure buckets they roll out meet their defined configuration.

PR link coming shortly

drmmarsunited:
PR linking complete

drmmarsunited:
Also going to be adding a PR to check for existence of bucket policies

zackproser:
Thanks for your contribution! I'm going to close this issue now as your PR was successfully merged.

PR #289, opened 2019-05-08
Adding support for checking S3 Bucket Versioning configuration

drmmarsunited:
Completed automated test run: https://gist.github.com/drmmarsunited/779269d656f5fe1d7ac0ec35772a0961
GH Issue: https://github.com/gruntwork-io/terratest/issues/288

This PR adds:
- A new basic AWS example that deploys a bucket with versioning enabled (terraform-aws-s3-example)
- Tests for the new example
- Functions in the `aws` module to support the creation, retrieval and assertion of S3 bucket versioning (following existing styles)
- Tests for the new functions


examples/terraform-aws-s3-example/locals.tf
```diff
@@ -0,0 +1,10 @@
+# ---------------------------------------------------------------------------------------------------------------------
```
yorinasub17: While I agree that this is good practice, can we have this at the bottom of `main.tf` so that it is consistent with the other examples?

modules/aws/s3.go
```diff
@@ -115,6 +116,35 @@ func CreateS3BucketE(t *testing.T, region string, name string) error {
 	return err
 }
 
+// PutS3BucketVersioningE creates an S3 bucket versioning configuration in the given region against the given bucket name, WITHOUT requiring MFA to remove versioning.
+func PutS3BucketVersioning(t *testing.T, region string, bucketName string) {
+	err := PutS3BucketVersioningE(t, region, bucketName)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yorinasub17: I know this file doesn't have this pattern yet, but we've started shortening this to a one liner: `require.NoError(t, err)`. Can you apply to these functions?

modules/aws/s3.go
```diff
@@ -208,6 +238,33 @@ func EmptyS3BucketE(t *testing.T, region string, name string) error {
...
+
+	res, err := s3Client.GetBucketVersioning(&s3.GetBucketVersioningInput{
+		Bucket: &bucket,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return *res.Status, nil
```
yorinasub17:
```suggestion
	return aws.StringValue(res.Status), nil
```

modules/aws/s3.go
```diff
@@ -230,6 +287,30 @@ func AssertS3BucketExistsE(t *testing.T, region string, name string) error {
...
+}
+
+// AssertS3BucketVersioningExistsE checks if the given S3 bucket has a versioning configuration enabled and returns an error if it does not.
+func AssertS3BucketVersioningExistsE(t *testing.T, region string, bucketName string) error {
+	status, err := GetS3BucketVersioningE(t, region, bucketName)
+	if err != nil {
+		return err
+	}
+
+	errorMsg := errors.New("bucket versioning is not enabled")
```
yorinasub17: Can you create an [error type](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/errors.go)? It makes checking whether the right error is being returned easier. E.g to differentiate between this error and an error from trying to retrieve the bucket.

modules/aws/s3.go
drmmarsunited: Implemented

modules/aws/s3.go
drmmarsunited: Implemented

modules/aws/s3.go
drmmarsunited: Implemented

drmmarsunited:
Committed changes to fix your callouts in comments!

modules/aws/s3.go
```diff
@@ -116,11 +116,11 @@ func CreateS3BucketE(t *testing.T, region string, name string) error {
...
 }
 
-// PutS3BucketVersioningE creates an S3 bucket versioning configuration in the given region against the given bucket name, WITHOUT requiring MFA to remove versioning.
+// PutS3BucketVersioning creates an S3 bucket versioning configuration in the given region against the given bucket name, WITHOUT requiring MFA to remove versioning.
 func PutS3BucketVersioning(t *testing.T, region string, bucketName string) {
 	err := PutS3BucketVersioningE(t, region, bucketName)
 	if err != nil {
-		t.Fatal(err)
+		require.NoError(t, err)
 	}
```
yorinasub17: Ah actually, the change is to replace the whole if block. As in, this can be:
```go
err := PutS3BucketVersioningE(t, region, bucketName)
require.NoError(t, err)
```

yorinasub17:
Thanks! One more fix and then I can kick off the build.

modules/aws/s3.go
drmmarsunited: Doh, that makes total sense.  I was rushing, sorry!

drmmarsunited:
Fixed require.NoError() usage in my functions

yorinasub17:
Ok this looks good! I just kicked off the build. Once that passes, we can go ahead and merge. Thanks!

drmmarsunited:
@yorinasub17 I am working on another PR for S3 stuff today.  Since I'm adding to the S3 module quite a bit, do you want me to refactor that file to use `require.NoError()` in all the places where `t.Fatal()` calls are currently made?

yorinasub17:
> do you want me to refactor that file to use require.NoError() in all the places where t.Fatal() calls are currently made?

That would be great! Thank you!

yorinasub17:
Ok build passed, so will go ahead to merge and release! Thanks again for your contribution!

modules/aws/errors.go
```diff
@@ -60,3 +60,20 @@ func (err AsgCapacityNotMetError) Error() string {
...
+// BucketVersioningNotEnabledError is returned when an S3 bucket that should have versioning does not have it applied
+type BucketVersioningNotEnabledError struct {
+	s3BucketName     string
+	awsRegion        string
+	versioningStatus string
+}
+
+func (err BucketVersioningNotEnabledError) Error() string {
+	return fmt.Sprintf(
+		"Versioning status for bucket #{err.s3BucketName} in the #{err.awsRegion} region is #{err.versioningStatus}",
```
yorinasub17: Ah I just realized this after merging. I don't think go actually supports string interpolation like this. Can you update this in your next PR?

drmmarsunited: I noticed some functions in that file already that used that syntax and made an (errant) assumption that this was some supported syntax I hadn't stumbled upon.  Sad day for me :( 

Looking at my latest merge from master, I see that syntax has all been corrected.  Holdover from Ruby I guess.  I will fix up.

PR #291, opened 2019-05-09
Adding support for S3 Bucket Policy assertion/retrieval

drmmarsunited:
Completed automated test runs: https://gist.github.com/drmmarsunited/079e5f28dbd51231b0a0184285c418be
GH Issue: https://github.com/gruntwork-io/terratest/issues/288

This PR adds:

- An updated basic AWS example that deploys a bucket with versioning enabled (terraform-aws-s3-example), with a custom variable to support deploying a bucket policy
- Split up tests for both deployment patterns (with and without policy assertion)
- Functions in the `aws` module to support the creation, retrieval and assertion of S3 bucket policies (following existing styles)
- Tests for the new functions

This PR fixes:

- Erroneous string interpolation syntax that was merged as part of [this PR for S3 bucket versioning](https://github.com/gruntwork-io/terratest/pull/289)
- Replaces the usage of `t.Fatal()` in tests to conform to the use of `require.NoError()` from other files in the `aws` module
- Several typos found during work


examples/terraform-aws-s3-example/README.md
```diff
@@ -3,7 +3,8 @@
 This folder contains a simple Terraform module that deploys resources in [AWS](https://aws.amazon.com/) to demonstrate
 how you can use Terratest to write automated tests for your AWS Terraform code. This module deploys an [S3
 Bucket](https://aws.amazon.com/s3/) and gives that Bucket a `Name` & `Environment` tag with the value specified in the
-`tag_bucket_name` and `tag_bucket_environment` variables, respectively.
+`tag_bucket_name` and `tag_bucket_environment` variables, respectively.  This module also contains a terraform variable 
+that will create a basic bucket policy that will restrict the bucket to only accept SSL connections.
```
drmmarsunited: This looks odd in this view, but the actual file looks just fine.

examples/terraform-aws-s3-example/variables.tf
```diff
@@ -17,6 +17,11 @@
 # These parameters have reasonable defaults.
 # ---------------------------------------------------------------------------------------------------------------------
 
+variable "with_policy" {
+  description = "If set to `true`, the bucket will be created with a bucket policy."
+  default = "false"
```
yorinasub17: Can you switch to using bools? (`false` without quotes).

examples/terraform-aws-s3-example/main.tf
```diff
@@ -17,6 +17,12 @@ resource "aws_s3_bucket" "test_bucket" {
   }
 }
 
+resource "aws_s3_bucket_policy" "bucket_access_policy" {
+  count = "${var.with_policy == "true" ? 1 : 0}"
```
yorinasub17: If you switch `with_policy` to bool, this also needs to be updated:

```
${var.with_policy ? 1 : 0}
```

examples/terraform-aws-s3-example/policies.tf
```diff
@@ -0,0 +1,29 @@
+data "aws_iam_policy_document" "s3_bucket_policy" {
```
yorinasub17: Just like the locals, can you add this to `main.tf`?

modules/aws/s3.go
```diff
@@ -15,9 +15,8 @@ import (
 // FindS3BucketWithTag finds the name of the S3 bucket in the given region with the given tag key=value.
 func FindS3BucketWithTag(t *testing.T, awsRegion string, key string, value string) string {
 	bucket, err := FindS3BucketWithTagE(t, awsRegion, key, value)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
yorinasub17: Thanks for updating these!

test/terraform_aws_s3_example_test.go
```diff
@@ -55,3 +55,52 @@ func TestTerraformAwsS3Example(t *testing.T) {
...
+
+	// Give this S3 Bucket an environment to operate as a part of for the purposes of resource tagging
+	expectedEnvironment := "Automated Testing"
+
+	// Pick a random AWS region to test in. This helps ensure your code works in all regions.
+	awsRegion := aws.GetRandomStableRegion(t, nil, nil)
+
+	terraformOptions := &terraform.Options{
+		// The path to where our Terraform code is located
+		TerraformDir: "../examples/terraform-aws-s3-example",
```
yorinasub17: Since this example is now shared across two tests, there will be conflicts in the terraform state (it uses local state, stored in the example folder). We could use `test_structure.CopyTerraformFolderToTemp` to resolve this (see https://github.com/gruntwork-io/terratest/blob/master/test/terraform_ssh_example_test.go#L22 for an example), but given the simplicity of the checks, I am inclined to keep our examples simple and merge the two tests. Can you consolidate the two tests and just run both bucket policy and versioning checks?

examples/terraform-aws-s3-example/variables.tf
drmmarsunited: Terraform best practices suggest this as an anti-pattern:

https://www.terraform.io/docs/configuration-0-11/variables.html#booleans

I will make the change, but just trying to stick with best practices.  Thoughts?

yorinasub17: Terraform 12 has a native bool type: https://www.terraform.io/docs/configuration/types.html#primitive-types. We're getting ready to upgrade to tf12, so just want to add these compatibilities in new code.

Also, the specific anti pattern is relying on the conversion rules of Terraform (e.g `true` => `1` and `false` => `0`). So relying on the bool in a conditional is actually a valid (and probably the only) case where bool makes sense.

test/terraform_aws_s3_example_test.go
drmmarsunited: Implemented

examples/terraform-aws-s3-example/policies.tf
drmmarsunited: Implemented

examples/terraform-aws-s3-example/main.tf
drmmarsunited: Implemented

examples/terraform-aws-s3-example/variables.tf
drmmarsunited: Sounds good.  Implemented.

drmmarsunited:
Pushed commit addressing all comments

yorinasub17:
Thanks for addressing the comments! I am going to go ahead and run the build now.

Issue #285, opened 2019-05-03
Feature request: Packer -var-file option


bflugon:
We're currently using -var-file in packer to build our images with different accounts in AWS. Would it be possible to add this support to terratest too, instead of having to specify each var separately?

yorinasub17:
That sounds like a useful addition. We haven't planned on implementing it, but a PR would be welcome!

This could be done by adding the option to [packer.Options](https://github.com/gruntwork-io/terratest/blob/master/modules/packer/packer.go#L16) and updating [formatPackerArgs](https://github.com/gruntwork-io/terratest/blob/master/modules/packer/packer.go#L136) to utilize it.

yorinasub17:
Released in https://github.com/gruntwork-io/terratest/releases/tag/v0.15.6

PR #294, opened 2019-05-11
Add VarFiles flag to packer Options

floresj:
Hello! This resolves the issue #285 that adds support for the `var-file` option for packer options. In lieu of modifying the [basic packer test](https://github.com/gruntwork-io/terratest/blob/master/test/packer_basic_example_test.go), I added an additional unit test to validate the output of the [formatPackerArgs](https://github.com/gruntwork-io/terratest/blob/master/modules/packer/packer.go#L136) function. However, if that does not suffice, I can update [test/packer_basic_example_test.go](https://github.com/gruntwork-io/terratest/blob/master/test/packer_basic_example_test.go) accordingly!

### DIFF
--- a/modules/packer/packer_test.go
+++ b/modules/packer/packer_test.go
@@ -2,6 +2,7 @@ package packer
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 )
 
@@ -115,5 +116,53 @@ func TestExtractArtifactINoIdPresent(t *testing.T) {
 
 	if err == nil {
 		t.Error("Expected to get an error when extracting an Artifact ID from text with no Artifact ID in it, but got nil")
+	}
+}
+
+func TestFormatPackerArgs(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		option   *Options
+		expected string
+	}{
+		{
+			option: &Options{
+				Template: "packer.json",
+			},
+			expected: "build -machine-readable packer.json",
+		},
+		{
+			option: &Options{
+				Template: "packer.json",
+				Vars: map[string]string{
+					"foo": "bar",
+				},
+				Only: "onlythis",
+			},
+			expected: "build -machine-readable -var foo=bar -only=onlythis packer.json",
+		},
+		{
+			option: &Options{
+				Template: "packer.json",
+				Vars: map[string]string{
+					"foo": "bar",
+				},
+				VarFiles: []string{
+					"foofile.json",
+				},
+			},
+			expected: "build -machine-readable -var foo=bar -var-file foofile.json packer.json",
+		},
+	}
+
+	for _, test := range tests {
+		args := formatPackerArgs(test.option)
+		if strings.Join(args, " ") != test.expected {
+			t.Errorf("Did not get expected formatted packer args. Expected: %s. Actual: %s",
+				test.expected,
+				args,
+			)
+		}
 	}
 }
--- a/modules/packer/packer_test.go
+++ b/modules/packer/packer_test.go
@@ -160,8 +160,6 @@ func TestFormatPackerArgs(t *testing.T) {
 
 	for _, test := range tests {
 		args := formatPackerArgs(test.option)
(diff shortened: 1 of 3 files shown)


modules/packer/packer_test.go
```diff
@@ -117,3 +118,51 @@ func TestExtractArtifactINoIdPresent(t *testing.T) {
...
+					"foofile.json",
+				},
+			},
+			expected: "build -machine-readable -var foo=bar -var-file foofile.json packer.json",
+		},
+	}
+
+	for _, test := range tests {
+		args := formatPackerArgs(test.option)
+		if strings.Join(args, " ") != test.expected {
```
yorinasub17: You can simplify this to `assert.Equal(t, strings.Join(args, " "), test.expected)`. When it is not equal, assert will show the expected and actual as well.

yorinasub17: The import is `github.com/stretchr/testify/assert`

modules/packer/packer_test.go
```diff
@@ -117,3 +118,51 @@ func TestExtractArtifactINoIdPresent(t *testing.T) {
 		t.Error("Expected to get an error when extracting an Artifact ID from text with no Artifact ID in it, but got nil")
 	}
 }
+
+func TestFormatPackerArgs(t *testing.T) {
```
yorinasub17: Nice test! 👍

modules/packer/packer_test.go
floresj: Thanks a lot for that suggestion! I updated it to use `assert.Equal`.

modules/packer/packer_test.go
floresj: Thanks again for the feedback! Here is a link to the output of my test run in AWS https://gist.github.com/floresj/9af415f4035f741a54bd1d22b83c888b

I initially started going the route of adding a `variables.json` to `terratest/examples/packer-basic-example/` but then decided to just dynamically generate that vars file within the go test in `test/packer_basic_example_test.go`. Since the `aws_region` is being randomly selected and used in subsequent calls, I figured that would keep it simple.

test/packer_basic_example_test.go
```diff
@@ -53,6 +55,61 @@ func TestPackerBasicExample(t *testing.T) {
...
+// for the `aws_region` variable.
+func TestPackerBasicExampleWithVarFile(t *testing.T) {
+	t.Parallel()
+
+	// Pick a random AWS region to test in. This helps ensure your code works in all regions.
+	awsRegion := terratest_aws.GetRandomStableRegion(t, nil, nil)
+
+	// Create temporary packer variable file to store aws region
+	varFile, err := ioutil.TempFile("", "*.json")
+	assert.NoError(t, err, "Did not expect temp file creation to cause error")
```
yorinasub17: Can you use `require.NoError` here? The import path is `github.com/stretchr/testify/require`. The difference is that `require` will fail and *exit* the test if it fails the assertion, vs `assert` will *continue* execution (but mark the test as failed).

test/packer_basic_example_test.go
```diff
@@ -53,6 +55,61 @@ func TestPackerBasicExample(t *testing.T) {
...
+
+	// Be sure to clean up temp file
+	defer os.Remove(varFile.Name())
+
+	// Write random generated aws region to temporary json file
+	varFileContent := []byte(fmt.Sprintf("{ \"aws_region\": \"%s\" }", awsRegion))
+	_, err = varFile.Write(varFileContent)
+	assert.NoError(t, err, "Did not expect writing to temp file %s to cause error", varFile.Name())
```
yorinasub17: Same as above: use `require`.

test/packer_basic_example_test.go
floresj: You got it! Thanks for the clarification. That is definitely an important distinction.

yorinasub17:
Thanks for fast turnaround! This looks good now so will kick off build.

PR #297, opened 2019-05-14
Fix T12 regression

### LINKED ISSUE #293: Error with passing variables as list of string to terraform commands
Hi!
I am trying to test my **AWS IAC** using terratest.
But whenever I'm trying to pass a list of strings from **terratest** to **terraform**. I'm getting error of wrong syntax from terraform.

Terraform version - Terraform v0.11.11

**some variables i'm using** : 
expectedSecurityGroupNames  := []string {"terratest_eth0_security_group", "terratest_eth1_security_group"}
subnetCidrBlocks     := map[string]string {   "terratest_eth0" : "10.1.1.0/24", "terratest_eth1" : "10.1.2.0/24" }

**this is what getting passed to terraform -** 
Running terraform [apply -input=false -lock=false -auto-approve -var vpc_name=terratest-vpc-XAZAD9 -var enable_dns_support=true -var az=ap-southeast-1c -var subnet_names=[terratest_eth0, terratest_eth1] -var map_public_ip_on_launch={terratest_eth0 = false, terratest_eth1 = false} -var source_dest_check_list=[true, false] -var region=ap-southeast-1 -var profile=default -var vpc_cidr_block=10.1.0.0/16 -var enable_dns_hostnames=true -var subnet_cidr_blocks={terratest_eth0 = 10.1.1.0/24, terratest_eth1 = 10.1.2.0/24} -var security_group_names=[terratest_eth0_security_group, terratest_eth1_security_group] -var private_ips_count_list=[1, 0]]

**This is expected -** 
Running terraform [apply -input=false -lock=false -auto-approve -var vpc_cidr_block="10.1.0.0/16" -var enable_dns_hostnames="true" -var enable_dns_support="true" -var subnet_cidr_blocks={terratest_eth0 = "10.1.1.0/24", terratest_eth1 = "10.1.2.0/24"} -var map_public_ip_on_launch={terratest_eth0 = "false", terratest_eth1 = "false"} -var source_dest_check_list=["true", "false"] -var vpc_name="terratest-vpc-psnigE" -var profile="default" -var az="ap-southeast-1c" -var subnet_names=["terratest_eth0", "terratest_eth1"] -var security_group_names=["terratest_eth0_security_group", "terratest_eth1_security_group"] -var private_ips_count_list=["1", "0"] -var region="ap-southeast-1"]

I'm facing this issue from last week only after few changes in [format.go](https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/format.go) file.

**Error I got**  -

invalid value "security_group_names=[terratest_eth0_security_group, terratest_eth1_security_group]" for flag -var: Cannot parse value for variable ("[terratest_eth0_security_group, terratest_eth1_security_group]") as valid HCL: At 1:6: unexpected token while parsing list: IDENT

tonerdo:
This PR addresses the issues introduced by #284. It works by introducing a new `isChild` argument to the `primitiveToHclString` function and creates a special case for the `string` type. If a string is a value in a map or an item in a list then it'll be quoted, otherwise, it remains unquoted. Behaviour for booleans and integers stay the same

E.g.

**Before**
-var name=Instance1
-var subnet_names=[terratest_eth0, terratest_eth1]
-var subnet_cidr_blocks={terratest_eth0 = 10.1.1.0/24, terratest_eth1 = 10.1.2.0/24}
-var private_ips_count_list=[1, 0]

**After**
-var name=Instance1
-var subnet_names=["terratest_eth0", "terratest_eth1"]
-var subnet_cidr_blocks={terratest_eth0 = "10.1.1.0/24", terratest_eth1 = "10.1.2.0/24"}
-var private_ips_count_list=[1, 0]

Fixes #293 

cc @s012sameer @sumitmaggo can you try importing this branch and confirm that this change fixes your issue before I go ahead to merge it

### DIFF
--- a/examples/terraform-basic-example/variables.tf
+++ b/examples/terraform-basic-example/variables.tf
@@ -11,7 +11,6 @@
 # You must provide a value for each of these parameters.
 # ---------------------------------------------------------------------------------------------------------------------
 
-
 # ---------------------------------------------------------------------------------------------------------------------
 # OPTIONAL PARAMETERS
 # These parameters have reasonable defaults.
@@ -25,4 +24,16 @@ variable "example" {
 variable "example2" {
   description = "Example variable 2"
   default     = ""
+}
+
+variable "example_list" {
+  description = "An example variable that is a list."
+  type        = "list"
+  default     = []
+}
+
+variable "example_map" {
+  description = "An example variable that is a map."
+  type        = "map"
+  default     = {}
 }
--- a/modules/terraform/format.go
+++ b/modules/terraform/format.go
@@ -57,7 +57,7 @@ func formatTerraformArgs(vars map[string]interface{}, prefix string) []string {
 // arbitrary Go types to an HCL string. Therefore, this method is a simple implementation that correctly handles
 // ints, booleans, lists, and maps. Everything else is forced into a string using Sprintf. Hopefully, this approach is
 // good enough for the type of variables we deal with in Terratest.
-func toHclString(value interface{}, isChild bool) string {
+func toHclString(value interface{}, isNested bool) string {
 	// Ideally, we'd use a type switch here to identify slices and maps, but we can't do that, because Go doesn't
 	// support generics, and the type switch only matches concrete types. So we could match []interface{}, but if
 	// a user passes in []string{}, that would NOT match (the same logic applies to maps). Therefore, we have to
@@ -68,7 +68,7 @@ func toHclString(value interface{}, isChild bool) string {
 	} else if m, isMap := tryToConvertToGenericMap(value); isMap {
 		return mapToHclString(m)
 	} else {
-		return primitiveToHclString(value, isChild)
+		return primitiveToHclString(value, isNested)
 	}
 }
 
@@ -140,7 +140,7 @@ func mapToHclString(m map[string]interface{}) string {
 
 // Convert a primitive, such as a bool, int, or string, to an HCL string. If this isn't a primitive, force its value
 // using Sprintf. See ToHclString for details.
-func primitiveToHclString(value interface{}, isChild bool) string {
+func primitiveToHclString(value interface{}, isNested bool) string {
 	switch v := value.(type) {
 
 	case bool:
@@ -150,8 +150,8 @@ func primitiveToHclString(value interface{}, isChild bool) string {
 		return "0"
 
 	case string:
-		// If string is a child of a larger data structure (e.g. list of string, map of string), ensure value is quoted
-		if isChild {
+		// If string is nested in a larger data structure (e.g. list of string, map of string), ensure value is quoted
+		if isNested {
 			return fmt.Sprintf("\"%v\"", v)
 		}
 
--- a/modules/terraform/format.go
+++ b/modules/terraform/format.go
@@ -44,7 +44,7 @@ func formatTerraformArgs(vars map[string]interface{}, prefix string) []string {
 	var args []string
 
 	for key, value := range vars {
-		hclString := toHclString(value)
+		hclString := toHclString(value, false)
 		argValue := fmt.Sprintf("%s=%s", key, hclString)
 		args = append(args, prefix, argValue)
 	}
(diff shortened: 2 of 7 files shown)


robmorgan:
I will also update `terraform-google-gke` to use this new version once it's released. It has a couple of nested string maps.

yorinasub17:
FYI, I'll pick this up to add the regression test @tonerdo.

yorinasub17:
UPDATE 91de246 

Added a regression test for the list and map encoding. I verified that this new test fails on `master`, but passes on this branch.

Note that this test fails on terraform 12 due to a bug in output:

```
--- FAIL: TestTerraformBasicExample (2.97s)
    output.go:58: json: cannot unmarshal array into Go value of type map[string]interface {}
FAIL
FAIL    github.com/gruntwork-io/terratest/test  3.007s
```

Looking into the output bug now.

yorinasub17:
Ah this is because the output format changed when using the json type:

TF11
```
{
     "sensitive": false,
     "type": "list",
     "value": [
         "test"
     ]
}
```

TF12
```
["test"]
```

I won't address this in this PR, but filed issue https://github.com/gruntwork-io/terratest/issues/299 to track the bug.

s012sameer:
Hi @tonerdo , these changes worked for me and fixed the issue #293 . Thanks... Are these changes going to be merged with the master branch ?

sumitmaggo:
Thanks @tornedo, the changes work for me, when will the new release be made?

modules/terraform/format.go
```diff
@@ -140,21 +140,22 @@ func mapToHclString(m map[string]interface{}) string {
...
 		}
 		return "0"
 
-	// Note: due to a Terraform bug (https://github.com/hashicorp/terraform/issues/7962), we can't use proper HCL
-	// syntax for ints have to wrap them as strings by falling through to the default case
-	//case int: return strconv.Itoa(v)
+	case string:
+		// If string is a child of a larger data structure (e.g. list of string, map of string), ensure value is quoted
+		if isChild {
+			return fmt.Sprintf("\"%v\"", v)
```
brikis98: Would it make more sense to do the inverse? That is, wrap _all_ strings, but then in the top-level method, if the value is returned, _unwrap_ the quotes there? That way, you don't have to pass this weird boolean all the way through.

examples/terraform-basic-example/variables.tf
```diff
@@ -26,3 +25,15 @@ variable "example2" {
...
+  description = "An example variable that is a list."
+  type        = "list"
+  default     = []
+}
+
+variable "example_map" {
+  description = "An example variable that is a map."
+  type        = "map"
+  default     = {}
+}
```
brikis98: Do we have tests for nested lists and maps?

modules/terraform/format.go
yorinasub17: IMO, I think this is actually easier to understand from an algorithm perspective. It is much easier to understand mentally in the context of `primitiveToHclString` what to do when nested in a larger structure, than it is to think about what we need to do about what `primitiveToHclString` returns in the context of the top most level, worrying about whether the nested value was a string or bool or int (which we don't know at the top level).

tonerdo: We could probably rename the parameter to `isNested` to make it clearer

yorinasub17: I like that name better, so went ahead and did that.

examples/terraform-basic-example/variables.tf
yorinasub17: Attempted to add some nested lists / maps testing: 6d1e95d .

yorinasub17:
> And are we running tests now against both TF 11 and 12?

Not yet. AFAIK, there are two things we need to do to be able to start testing on tf12:

- We need to update the examples to be tf12 compatible.
- https://github.com/gruntwork-io/terratest/issues/299

brikis98:
OK, so just to confirm:

1. We've manually tested this with TF11 and TF12 and it works.
1. We have automated tests in the CI job for TF11. We'll add TF12 later, when we update the examples.

Is that right? If so, anything blocking merge and release?

yorinasub17:
> so just to confirm:

Yes that's right.

> If so, anything blocking merge and release?

Nope! Will ship!

yorinasub17:
Thanks for checking in and the thorough review!

yorinasub17:
https://github.com/gruntwork-io/terratest/releases/tag/v0.15.8

PR #298, opened 2019-05-14
Delete SSH Public Keys imported during tests

robmorgan:
This PR adds two methods `DeleteSSHKey` and `DeleteSSHKeyE` for deleting SSH public keys that can be imported to a user identity. It seems GCP limits the maximum number of SSH keys to 70 per user and our tests were attaching a random key during each execution.

### DIFF
--- a/modules/gcp/oslogin_test.go
+++ b/modules/gcp/oslogin_test.go
@@ -15,6 +15,7 @@ func TestImportSSHKeyOSLogin(t *testing.T) {
 	user := GetGoogleIdentityEmailEnvVar(t)
 
 	ImportSSHKey(t, user, key)
+	DeleteSSHKey(t, user, key)
 }
 
 func TestGetLoginProfile(t *testing.T) {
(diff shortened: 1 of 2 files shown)


robmorgan:
cc @rileykarson

modules/gcp/oslogin_test.go
```diff
@@ -15,6 +15,7 @@ func TestImportSSHKeyOSLogin(t *testing.T) {
 	user := GetGoogleIdentityEmailEnvVar(t)
 
 	ImportSSHKey(t, user, key)
+	DeleteSSHKey(t, user, key)
```
yorinasub17: We should defer this before the call to `ImportSSHKey`. That way, we will run the `Delete` even if `Import` throws an error. I don't think the API is broken such that Import fails even when it successfully imports, but in the unlikely case that there is a bug there, this change will handle that.

modules/gcp/oslogin_test.go
```diff
@@ -33,6 +34,7 @@ func TestSetOSLoginKey(t *testing.T) {
 	user := GetGoogleIdentityEmailEnvVar(t)
 
 	ImportSSHKey(t, user, key)
+	defer DeleteSSHKey(t, user, key)
```
yorinasub17: Same as above: defer before calling Import.

PR #303, opened 2019-05-16
Use a retry loop to obtain the GCP OAuth2 token

robmorgan:
This PR is designed to improve the resiliency of the GCP methods for obtaining an OAuth2 token. In the past, we would often get errors such as:

```
region.go:77: Get https://www.googleapis.com/compute/v1/projects/terratest-214610/zones?alt=json&prettyPrint=false: oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: net/http: TLS handshake timeout
```
You can refer to the [Terratest log output](https://1429-53174289-gh.circle-artifacts.com/0/tmp/logs/TestPackerGCPBasicExample.log) from a recent build for more information. This patch is inspired by https://github.com/kubernetes/kubernetes/pull/56394.

### DIFF
--- a/modules/gcp/compute.go
+++ b/modules/gcp/compute.go
@@ -591,7 +591,7 @@ func NewComputeServiceE(t *testing.T) (*compute.Service, error) {
 	maxRetries := 6
 	timeBetweenRetries := 10 * time.Second
 	for i := 0; i < maxRetries; i++ {
-		logger.Log(t, "Attempting to request a Google OAuth2 token...")
+		logger.Logf(t, "Attempting to request a Google OAuth2 token (%d/%d)...", i+1, maxRetries)
 		client, err = google.DefaultClient(ctx, compute.CloudPlatformScope)
 		if err == nil {
 			logger.Log(t, "Got Google OAuth2 token, continuing.")
--- a/modules/gcp/compute.go
+++ b/modules/gcp/compute.go
@@ -590,6 +590,8 @@ func NewComputeServiceE(t *testing.T) (*compute.Service, error) {
 	var err error
 	maxRetries := 6
 	timeBetweenRetries := 10 * time.Second
+	// We directly implement the retry logic here instead of using the `retry` module so that we aren't relying on
+	// modifying variables in an inner function.
 	for i := 0; i < maxRetries; i++ {
 		logger.Logf(t, "Attempting to request a Google OAuth2 token (%d/%d)...", i+1, maxRetries)
 		client, err = google.DefaultClient(ctx, compute.CloudPlatformScope)
--- a/modules/gcp/compute.go
+++ b/modules/gcp/compute.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/random"
-	"github.com/gruntwork-io/terratest/modules/retry"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/compute/v1"
 )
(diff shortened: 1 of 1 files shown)


modules/gcp/compute.go
```diff
@@ -581,10 +584,23 @@ func NewComputeService(t *testing.T) *compute.Service {
...
-	if err != nil {
-		return nil, fmt.Errorf("Failed to get default client: %v", err)
-	}
+	// Retrieve the Google OAuth token using a retry loop as it can sometimes return an error.
+	// e.g: oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: net/http: TLS handshake timeout
+	// This is loosely based on https://github.com/kubernetes/kubernetes/blob/7e8de5422cb5ad76dd0c147cf4336220d282e34b/pkg/cloudprovider/providers/gce/gce.go#L831.
+	maxRetries := 6
+	timeBetweenRetries := 10 * time.Second
+	var client *http.Client
+	var err error
```
yorinasub17: Can you add a log message with the retry params, so we log how long this might try?

modules/gcp/compute.go
```diff
@@ -581,10 +584,23 @@ func NewComputeService(t *testing.T) *compute.Service {
...
+	retry.DoWithRetry(t, "Attempting to request a Google OAuth2 token...", maxRetries, timeBetweenRetries, func() (string, error) {
+		client, err = google.DefaultClient(ctx, compute.CloudPlatformScope)
```
yorinasub17: I am not quite a fan of reusing the client var here, where we are updating a var outside the scope of the inner function. Can we scope this to the function, and then call it again after success? Alternatively, can we just implement the for loop version here?

```
for i := 0; i < maxRetries; i++ {
    client, err = google.DefaultClient(ctx, compute.CloudPlatformScope)
    if err == nil {
        break;
    }
    time.Sleep(timeBetweenRetries)
}
if err != nil {
    return nil, err
}
```

modules/gcp/compute.go
```diff
@@ -581,10 +584,23 @@ func NewComputeService(t *testing.T) *compute.Service {
...
-		return nil, fmt.Errorf("Failed to get default client: %v", err)
-	}
+	// Retrieve the Google OAuth token using a retry loop as it can sometimes return an error.
+	// e.g: oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: net/http: TLS handshake timeout
+	// This is loosely based on https://github.com/kubernetes/kubernetes/blob/7e8de5422cb5ad76dd0c147cf4336220d282e34b/pkg/cloudprovider/providers/gce/gce.go#L831.
+	maxRetries := 6
+	timeBetweenRetries := 10 * time.Second
+	var client *http.Client
+	var err error
+	retry.DoWithRetry(t, "Attempting to request a Google OAuth2 token...", maxRetries, timeBetweenRetries, func() (string, error) {
```
yorinasub17: If we do go with this approach, this should be `DoWithRetryE` right? Since this is the `E` version?

modules/gcp/compute.go
robmorgan: Sure

modules/gcp/compute.go
robmorgan: I'll move it to use the inline `for` loop.

modules/gcp/compute.go
robmorgan: Skipping as I'm going with the `for` loop.

modules/gcp/compute.go
```diff
@@ -581,9 +583,25 @@ func NewComputeService(t *testing.T) *compute.Service {
...
 
-	client, err := google.DefaultClient(ctx, compute.CloudPlatformScope)
+	// Retrieve the Google OAuth token using a retry loop as it can sometimes return an error.
+	// e.g: oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: net/http: TLS handshake timeout
+	// This is loosely based on https://github.com/kubernetes/kubernetes/blob/7e8de5422cb5ad76dd0c147cf4336220d282e34b/pkg/cloudprovider/providers/gce/gce.go#L831.
+	var client *http.Client
+	var err error
+	maxRetries := 6
+	timeBetweenRetries := 10 * time.Second
+	for i := 0; i < maxRetries; i++ {
```
yorinasub17: Add a comment about why we aren't using `retry` module:

```
// We directly implement the retry logic here instead of using the `retry` module so that we aren't relying on modifying variables in an inner function.
```

robmorgan:
> LGTM! Just one more code comment suggestion.

sure

PR #308, opened 2019-05-20
Fix backend config params

brikis98:
Terraform seems to require that backend config arguments are passed using the syntax `-backend-config=key=value`. That is, with an equals sign as a separator. The syntax we use elsewhere with a space as a separator (e.g., `-var key=value`) doesn't seem to work for backend config. Added example + test to verify too.


brikis98:
I guess? Or maybe the behavior changed in some recent version of Terraform? Not sure.

brikis98:

PR #310, opened 2019-05-21
Fix output parsing for TF 0.12

brikis98:
Terraform 0.11 and older, when you run `terraform output xxx` return JSON of the format:

```json
{
    "sensitive": false,
    "type": "<TYPE>",
    "value": <VALUE>
}
```

Terraform 0.12 and newer return `<VALUE>` directly for the exact same commad. This PR updates the parsing to handle either version. Tested manually with TF 0.12. Will let the automated tests validate TF 0.11 behavior.


modules/terraform/output.go
```diff
@@ -114,18 +137,21 @@ func OutputMapE(t *testing.T, options *Options, key string) (map[string]string,
 		return nil, err
 	}
 
+	// Terraform 0.11 or older return an object where the value we want is under the key "value". Terraform 0.12 and
+	// older return the value we want directly.
 	value, containsValue := outputMap["value"]
```
yorinasub17: What if there is an output that actually contains the string `value`? Might be more robust if we check all three vars `sensitive`, `type`, and `value`.

brikis98: Good point! Fixed in https://github.com/gruntwork-io/terratest/pull/310/commits/4aff44dc5762eed52a5bf240fa020ac8c8e02d1b.

brikis98:
Tests failed due to intermittent failure in a Packer build. We need to add retries to those. Will do so in a separate PR.

PR #311, opened 2019-05-22
Add support for retries to Packer. Use regex to match errors.

brikis98:
This PR does the following:

1. Add a `DoWithRetryableErrors` method that takes in a map of retryable errors and an action to execute, and if the action returns an error, retries it if the error or the action's stdout/stderr matches one of the retryable errors.

1. Update the Packer code to use this method so we can have flaky Packer builds retry on known errors.

1. Update Terraform to use this code. The Terraform code already supported retries; the main (backwards incompatible!) change here is that you can now match error messages using a regular expression instead of plain string.

### DIFF
--- a/test/packer_basic_example_test.go
+++ b/test/packer_basic_example_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"regexp"
 	"testing"
 	"time"
 
@@ -20,8 +19,8 @@ import (
 // Occasionally, a Packer build may fail due to intermittent issues (e.g., brief network outage or EC2 issue). We try
 // to make our tests resilient to that by specifying those known common errors here and telling our builds to retry if
 // they hit those errors.
-var DefaultRetryablePackerErrors = map[*regexp.Regexp]string{
-	regexp.MustCompile("Script disconnected unexpectedly"): "Occasionally, Packer seems to lose connectivity to AWS, perhaps due to a brief network outage",
+var DefaultRetryablePackerErrors = map[string]string{
+	"Script disconnected unexpectedly": "Occasionally, Packer seems to lose connectivity to AWS, perhaps due to a brief network outage",
 }
 var DefaultTimeBetweenPackerRetries = 15 * time.Second
 
--- a/test/packer_basic_example_test.go
+++ b/test/packer_basic_example_test.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"testing"
+	"time"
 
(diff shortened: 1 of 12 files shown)


test/packer_basic_example_test.go
```diff
@@ -15,6 +17,16 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// Occasionally, a Packer build may fail due to intermittent issues (e.g., brief network outage or EC2 issue). We try
+// to make our tests resilient to that by specifying those known common errors here and telling our builds to retry if
+// they hit those errors.
+var DefaultRetryablePackerErrors = map[*regexp.Regexp]string{
+	regexp.MustCompile("Script disconnected unexpectedly"): "Occasionally, Packer seems to lose connectivity to AWS, perhaps due to a brief network outage",
```
brikis98: This is an intermittent Packer error I hit recently that motivated this PR. However, I've hit many others before, but just can't find the CI builds to copy down the error messages.

Can anyone else find error messages for intermittent Packer build failures so we can include retries for them here?

brikis98:
(Looks like CircleCI or GitHub is having a bit of an outage, so the tests for this PR have not yet kicked off)

brikis98:
@robmorgan Could I get a review on this?

test/packer_basic_example_test.go
robmorgan: Yeah here are two from my side:

1. Build: https://circleci.com/gh/gruntwork-io/terratest/1610
Failed Test Log: https://1610-53174289-gh.circle-artifacts.com/0/tmp/logs/TestPackerMultipleConcurrentAmis.log

2. Build: https://circleci.com/gh/gruntwork-io/terratest/1606
Failed Test Log: https://1606-53174289-gh.circle-artifacts.com/0/tmp/logs/TestPackerGCPBasicExample.log

The second one should be resolved by one of the patches I pushed, but there seems to be a bug with it.

robmorgan:
Added one comment, but approving anyway.

test/packer_basic_example_test.go
brikis98: The first error, I believe, is `Script disconnected unexpectedly`, which I already have in `DefaultRetryablePackerErrors`. The second one happens before the Packer build, so this PR won't be able to help there.

brikis98: Also, I just updated the retry logic for the second error to use the existing `retry.DoWithRetry` function: https://github.com/gruntwork-io/terratest/pull/311/commits/096157c5b62cd7d2c9d9c495c3bc9c225c70a11b

brikis98:
Thanks for the review Rob. Merging now!

Issue #312, opened 2019-05-29
[aws] Support for different authentication methods 


klebediev:
Hi!

I can see `NewAuthenticatedSessionFromRole` function implemented in [modules/aws/auth.go](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/auth.go), which handles authentication usung sts assume role method, but `NewAuthenticatedSession` is used for authentication in the rest of `modules/aws/*.go` 

In my environment using assume role method is a preferred way to authenticate. 
Are there plans to do this or how could we add support for different authentication methods in aws-related functionality?  

[Thinking of something like adding `*session.Session` as an "optional" parameter for aws functions (via variadic args, for example) and if not set, then create a session using  `NewAuthenticatedSession()` inside the function]

Sample function signature:
```go
func GetPrivateIpOfEc2Instance(t *testing.T, instanceID string, awsRegion string, sess ...*session.Session) string {
...
}
```

yorinasub17:
We generally recommend setting up the authentication outside of terratest, before executing the `go test` call. For example if you were using `aws-vault`, you can do:

```
aws-vault exec sandbox -- go test -timeout 60m .
```

The main reason for this is that it is one less thing to manage in your test, while leveraging the plethora of tools out there for AWS CLI authentication.

klebediev:
thanks for quick reply @yorinasub17!
My point is that Terraform AWS Provider gives us [a number of options for AWS auth](https://www.terraform.io/docs/providers/aws/index.html), what means that people will use them.
So, as for me, it would be reasonable if testing framework supported the same options directly

yorinasub17:
Fair point.

However, requiring the session to be passed in to all the aws functions is a major backwards incompatible change that adds extra calls to all test code. Can you come up with a different design that doesn't require passing through the session?

klebediev:
Yes, understand your concerns. 
Although, the solution above with passing session via variadic argument `...*session.Session` doesn't isn't backwards incompatible, such notion of optional parameters isn't supported by the philosophy of Golang. Plus such API isn't clear as for me..

Currently, `NewAuthenticatedSession(region)` is used everywhere in `aws` module. 
```
account.go:	sess, err := NewAuthenticatedSession(region)
acm.go:	sess, err := NewAuthenticatedSession(awsRegion)
asg.go:	sess, err := NewAuthenticatedSession(region)
cloudwatch.go:	sess, err := NewAuthenticatedSession(region)
[...]

PR #314, opened 2019-06-01
Add support for assume_role during authentication to aws module

klebediev:
- Addresses issue #312 by adding ability to assume IAM Role, specifying its ARN by environment variable `TERRATEST_IAM_ROLE` 
 - Changes: `NewAuthenticatedSession` renamed to`NewAuthenticatedSessionFromDefaultCredentials`, new `NewAuthenticatedSession` is a wrapper which checks OS environment and calls corresponding `NewAuthenticatedSession*` function
 - Backward compatible
 - It's hard to write a unified unit-test for  `TERRATEST_IAM_ROLE` behavior, because `os.Setenv("TERRATEST_IAM_ROLE", ....)` once set, will affect all consecutive calls to AWS in other tests running in parallel. Tested locally.

### DIFF
--- a/modules/aws/auth.go
+++ b/modules/aws/auth.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


yorinasub17:
Thanks for discussing things first on the issue! This looks good! Going to kick off the build now.

modules/aws/auth.go
```diff
@@ -13,8 +14,18 @@ import (
 	"github.com/pquerna/otp/totp"
 )
 
-// NewAuthenticatedSession gets an AWS Session, checking that the user has credentials properly configured in their environment.
+// NewAuthenticatedSession creates an AWS session following to standard AWS authentication workflow.
+// If TERRATEST_IAM_ROLE environment variable is set, assumes IAM role specified in it.
 func NewAuthenticatedSession(region string) (*session.Session, error) {
+	if assumeRoleArn, ok := os.LookupEnv("TERRATEST_IAM_ROLE"); ok {
```
yorinasub17: NIT: Can you extract this string constant to a global `const` at the top of the file so it is easier to change later, even if the file evolves overtime?

klebediev: good point, fixed

Issue #203, opened 2018-12-06
Make terratest_log_parser token buffer size configurable


yorinasub17:
If the log contains a word that is longer than 64*1024 characters, `terratest_log_parser` will crash with `FATA[2018-12-06T01:00:04Z] Error reading from scanner: bufio.Scanner: token too long `

trobert2:
I am also having this problem when trying to do `	terraform.InitAndApply(t, terraformOptions)` and the module I am testing has long files to pass in.
```
    apply.go:13: FatalError{Underlying: bufio.Scanner: token too long}
```

yorinasub17:
Reopening as that PR doesn't actually fix this issue.

pearcec:
This fixed it for me.  Did you add OutputMaxLineSize to your terraform.Options
tfOptions := &terraform.Options{
	TerraformDir: "../examples/definition",
	OutputMaxLineSize: 128 * 1024,
}

yorinasub17:
This is now fixed in https://github.com/gruntwork-io/terratest/releases/tag/v0.26.6

PR #317, opened 2019-06-05
Allow to have a custom scanner buffer size

thirdeyenick:
When terraform produces a lot of output in one line, the default max size of 64KB of bufio.NewScanner is not enough. This PR makes the size configurable. It should fix https://github.com/gruntwork-io/terratest/issues/203.


modules/shell/command.go
```diff
@@ -17,10 +17,11 @@ import (
...
 type Command struct {
-	Command    string            // The command to run
-	Args       []string          // The args to pass to the command
-	WorkingDir string            // The working directory
-	Env        map[string]string // Additional environment variables to set
+	Command           string            // The command to run
+	Args              []string          // The args to pass to the command
+	WorkingDir        string            // The working directory
+	Env               map[string]string // Additional environment variables to set
+	OutputMaxLineSize *int              // The max line size of stdout and stderr (in bytes)
```
yorinasub17: Does this need to be a pointer? Or can we get away with `int`, and the default behavior treats `0` as a special case?

thirdeyenick: Hi, thanks for the information. I will provide an update soon.

yorinasub17:
Test changes look good. Thanks for adding that! Just kicked off our CI build. If that goes well, will merge and release.

yorinasub17:
So the build failed because `terratest_log_parser` hasn't been updated to handle the output from the new test case. I see two options for resolution:

- Expose a CLI arg to the `terratest_log_parser` that propagates down to the parser.
- Run the large output test separately from the main test, so that it doesn't have to go through the log parser. This can be done using environment variables that can be used to control skipping.

Can you implement one of these? The first option is ideal, but can take time, so it is ok to just do option 2.

thirdeyenick:
Thanks for the feedback. Will have a look.

thirdeyenick:
So I tried to separate the test by using build tags instead of environment variables. I am not sure about the terratest_log_parser issue and if the output of the separate test should go into another log file...

.circleci/config.yml
```diff
@@ -110,6 +110,7 @@ jobs:
       # kubernetes_test build steps.
       - run: mkdir -p /tmp/logs
       - run: run-go-tests --packages "-p 1 ./..." | tee /tmp/logs/test_output.log
+      - run: run-go-tests --packages "-p 1 -tags=special -run TestRunCommandWithHugeLineOutput ./modules/shell"
```
yorinasub17: Can you reverse the order so that we run this test first? That way we short circuit and don't run the massive test suite if there is a bug here.

yorinasub17:
> So I tried to separate the test by using build tags instead of environment variables.

This is good enough for me! Just have one nit on the testing order. Once you address that, I'll kick off another build and hopefully then we can merge!

yorinasub17:
Thanks for taking care of that! Kicking off the build now.

modules/shell/command_special_test.go
```diff
@@ -0,0 +1,47 @@
+// +build special
+
+package shell
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
```
yorinasub17: Looks like this violated the rules in `goimports`. Specifically, thirdparty libs should be listed separately from stdlibs. Can you update this to be:

```
import (
        "bytes"
        "fmt"
        "testing"

        "github.com/stretchr/testify/assert"
)
```

thirdeyenick: Thought my vim plugin does that automatically, but apparently I did not configure this :-)

yorinasub17:
Thanks for the continued updates! Just kicked off another build.

yorinasub17:
The build passed, but looks like there was a merge conflict. I resolved it for you.

Going to kick off another build and when that passes, I'll be merging this in.

yorinasub17:
Ok build passed. Going to merge this. Thanks for your contribution!

PR #324, opened 2019-06-16
Fix output commands to only read from stdout

brikis98:
The `Output` commands in Terratest were running `terraform output` and then parsing the results from both `stdout` and `stderr`. This works with Terraform, as its `output` commands solely right to `stdout`, though it's a risky way to do it and may well break in the future; more to the point, it's broken with `terragrunt output` commands. Added failing test case and then fixed the code to make the test case pass.

### DIFF
--- a/modules/shell/command.go
+++ b/modules/shell/command.go
@@ -61,6 +61,7 @@ func RunCommandAndGetOutputE(t *testing.T, command Command) (string, error) {
 
 // RunCommandAndGetStdOut runs a shell command and returns solely its stdout (but not stderr) as a string. The stdout
 // and stderr of that command will also be printed to the stdout and stderr of this Go program to make debugging easier.
+// If there are any errors, fail the test.
 func RunCommandAndGetStdOut(t *testing.T, command Command) string {
 	output, err := RunCommandAndGetStdOutE(t, command)
 	require.NoError(t, err)
--- a/modules/shell/command.go
+++ b/modules/shell/command.go
@@ -78,7 +78,8 @@ func RunCommandAndGetStdOutE(t *testing.T, command Command) (string, error) {
 	return output, err
 }
 
-// RunCommandAndGetOutputE runs a shell command and returns its stdout and stderr as a string. The stdout and stderr of that command will also
+// runCommandAndStoreOutputE runs a shell command and stores each line from stdout and stderr in the given
+// storedStdout and storedStderr variables, respectively. The stdout and stderr of that command will also
 // be printed to the stdout and stderr of this Go program to make debugging easier.
 func runCommandAndStoreOutputE(t *testing.T, command Command, storedStdout *[]string, storedStderr *[]string) error {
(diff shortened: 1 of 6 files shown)


modules/shell/command.go
```diff
@@ -50,6 +52,35 @@ func RunCommandAndGetOutput(t *testing.T, command Command) string {
...
+	stdout := []string{}
+	stderr := []string{}
+	err := runCommandAndStoreOutputE(t, command, &stdout, &stderr)
+
+	output := strings.Join(stdout, "\n")
+	return output, err
+}
+
+// RunCommandAndGetOutputE runs a shell command and returns its stdout and stderr as a string. The stdout and stderr of that command will also
+// be printed to the stdout and stderr of this Go program to make debugging easier.
```
yorinasub17: NIT: This needs to be updated to reflect the new function.

modules/shell/command.go
```diff
@@ -50,6 +52,35 @@ func RunCommandAndGetOutput(t *testing.T, command Command) string {
...
 func RunCommandAndGetOutputE(t *testing.T, command Command) (string, error) {
+	allOutput := []string{}
+	err := runCommandAndStoreOutputE(t, command, &allOutput, &allOutput)
+
+	output := strings.Join(allOutput, "\n")
+	return output, err
+}
+
+// RunCommandAndGetStdOut runs a shell command and returns solely its stdout (but not stderr) as a string. The stdout
+// and stderr of that command will also be printed to the stdout and stderr of this Go program to make debugging easier.
```
yorinasub17: NIT: I've been updating these functions without `E` to also mention that the test will fail if there are any errors in running the command. See `k8s` module for example: https://github.com/gruntwork-io/terratest/blob/master/modules/k8s/cluster_role.go#L11

modules/terraform/cmd.go
```diff
@@ -61,6 +61,24 @@ func RunTerraformCommandE(t *testing.T, additionalOptions *Options, additionalAr
 	})
 }
 
+// RunTerraformCommandE runs terraform with the given arguments and options and returns solely its stdout (but not
```
yorinasub17: NIT: `s/RunTerraformCommandE/RunTerraformCommandAndGetStdoutE`

modules/shell/command.go
brikis98: https://github.com/gruntwork-io/terratest/pull/324/commits/3a650a70078fefc3af67ab2f35e1052acc15c12e

modules/shell/command.go
brikis98: https://github.com/gruntwork-io/terratest/pull/324/commits/4faa99995986c111de3296cd604c3a5421e6ed3b

modules/terraform/cmd.go
brikis98: https://github.com/gruntwork-io/terratest/pull/324/commits/5948b8d903086df32211b336922e7bd30a4d6d78

brikis98:
Transient test failures seem gone, so finally merging. Thx for the review!

PR #326, opened 2019-06-20
Lock down k8s.io/api so that it doesn't break over time

yorinasub17:
As I was upgrading to the latest terratest on a few of our modules, I ran into an issue with `dep` where it was pulling in the latest version of `k8s.io/api` that has some backwards incompatible changes which broke `k8s.io/client-go`.

We fix this by defining `k8s.io/api` as a transitive dependency and lock it down to a specific version: the last known revision locked down in `terratest` that is known to work.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -649,10 +649,10 @@
   version = "v2.2.1"
 
 [[projects]]
-  branch = "master"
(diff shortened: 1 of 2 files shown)


yorinasub17:
Build passed, so going to merge. Thanks Rob for review!

Issue #316, opened 2019-06-03
Tests are complicated... is it possible to re-write basic test to support new approach without $GOPATH?


mrPsycho:
Hello,

i'm trying to understand all this stuff with GO, and have confuse:
* GOPATH is not necessary due to new approach: https://stackoverflow.com/questions/21001387/how-do-i-set-the-gopath-environment-variable-on-ubuntu-what-file-must-i-edit/53026674#53026674
and when i'm trying to run basic test (https://github.com/gruntwork-io/terratest/tree/master/examples/terraform-basic-example)  - i see error connected to GOPATH.

* are there any best practicies for pointing that GOPATH for this project terratest?  
This would be great to hear, cause i would like to introduce terratest to our team, but... i can't understand how to setup it correctly, so everyone will use it by somewhat template.

mrPsycho:
```
mrpsy@LVAGPLTP2798:~/repos/terratest/test$ go test -v -run TestTerraformBasicExample
# _/home/mrpsy/repos/terratest/test
packer_basic_example_test.go:10:2: cannot find package "github.com/aws/aws-sdk-go/aws" in any of:
	/usr/lib/go-1.10/src/github.com/aws/aws-sdk-go/aws (from $GOROOT)
	/home/mrpsy/repos/terraform_conf/src/src/github.com/aws/aws-sdk-go/aws (from $GOPATH)
FAIL	_/home/mrpsy/repos/terratest/test [setup failed]
```

this what i see.

mrPsycho:
ok, i figured it out. 

but:
```
mrpsy@LVAGPLTP2798:~/src/github.com/gruntwork-io/terratest$ dep ensure
dep: WARNING: Unknown field in manifest: prune
dep: WARNING: Unknown field in manifest: require
ensure Solve(): No versions of google.golang.org/api met constraints:
	master: unable to update checked out version: : command failed: [git checkout 19ff8768a5c0b8e46ea281065664787eefc24121]: exit status 1
	v0.5.0: Could not introduce google.golang.org/api@v0.5.0, as it is not allowed by constraint master from project github.com/gruntwork-io/terratest.
	v0.4.0: Could not introduce google.golang.org/api@v0.4.0, as it is not allowed by constraint master from project github.com/gruntwork-io/terratest.
	v0.3.2: Could not introduce google.golang.org/api@v0.3.2, as it is not allowed by constraint master from project github.com/gruntwork-io/terratest.
[...]

PR #329, opened 2019-06-25
[skip ci] Improve docs on getting started

yorinasub17:
This improves the docs on the getting started experience with terratest. We have had a few inquiries and sources of confusion in terms of how one should get started with terratest. This attempts to address that by having detailed step-by-step guide on how one should start using terratest with their modules.

Reference: https://github.com/gruntwork-io/terratest/issues/316

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,13 +74,22 @@ Terratest uses the Go testing framework. To use terratest, you need to install:
 
 ### Setting up your project
 
-1. Golang requires go source files to be available in the `GOPATH`. By default this is typically `$HOME/go/src`. Create a new blank folder in the `GOPATH` to hold your terraform module and terratest code. For example, if you were developing a VPC module named `terraform-aws-vpc`, create the folder `$HOME/go/src/terraform-aws-vpc` to hold your module code.
+1. Golang requires go source files to be available in the `GOPATH`. By default this is typically `$HOME/go/src`. Create
+   a new blank folder in the `GOPATH` to hold your terraform module and terratest code. For example, if you were
+   developing a VPC module named `terraform-aws-vpc`, create the folder `$HOME/go/src/terraform-aws-vpc` to hold your
+   module code.
 1. In the project folder, create three subfolders:
     1. `modules`: This folder should contain your terraform modules that will be tested.
-    1. `examples`: This folder should contain examples of how to use the modules. These should be self-contained deployable examples. Meaning, it should provision all the resources that are necessary to run the modules in the `modules` folder.
+    1. `examples`: This folder should contain examples of how to use the modules. These should be self-contained
+       deployable examples. Meaning, it should provision all the resources that are necessary to run the modules in the
+       `modules` folder.
     1. `test`: This folder should contain your terratest code.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,21 +62,43 @@ validateServerIsWorking(t, terraformOptions)
...
 
-Prerequisite: install [Go](https://golang.org/).
+### Install requirements
 
-To add Terratest to your projects, we recommend using a Go dependency manager such as
-[dep](https://github.com/golang/dep) to add the packages you wish to use (see [package by package overview](#package-by-package-overview) for the list). For example, to add the `terraform` package:
+Terratest uses the Go testing framework. To use terratest, you need to install:
+
+- [Go](https://golang.org/) (requires version >=1.10)
+- [dep](https://github.com/golang/dep) (requires version >=0.5.1)
```
yorinasub17: With the `kubernetes` repos, it is no longer a good idea to use `go get`.

README.md
```diff
@@ -62,21 +62,43 @@ validateServerIsWorking(t, terraformOptions)
...
+
+
+### Setting up your project
+
+1. Golang requires go source files to be available in the `GOPATH`. By default this is typically `$HOME/go/src`. Create a new blank folder in the `GOPATH` to hold your terraform module and terratest code. For example, if you were developing a VPC module named `terraform-aws-vpc`, create the folder `$HOME/go/src/terraform-aws-vpc` to hold your module code.
```
yorinasub17: Once we have go modules, life would be a lot better. Esp with all the k8s stuff since k8s now officially supports go modules.

README.md
```diff
@@ -62,21 +62,43 @@ validateServerIsWorking(t, terraformOptions)
...
+  version = "0.17.4"
 ```
 
-Alternatively, you can use `go get`:
+Now you should be able to run the example test. To run the test:
 
-```bash
-go get github.com/gruntwork-io/terratest/modules/terraform
+1. Change your working directory to the `test` folder.
+1. Run `dep ensure`
```
brikis98: Perhaps add a note that this downloads dependencies, so you only need to run it once?

README.md
```diff
@@ -62,21 +62,43 @@ validateServerIsWorking(t, terraformOptions)
...
+1. Run `go test -v .`
```
brikis98: Then rephrase this to "Each time you want to run tests, use".

README.md
```diff
@@ -62,21 +62,43 @@ validateServerIsWorking(t, terraformOptions)
...
+
+Note that `go` has a default test timeout of 10 minutes. With infrastructure testing, your tests will surpass the 10 minutes very easily. To extend the timeout, you can pass in the `-timeout` option, which takes a `go` duration string (e.g `10m` for 10 minutes or `1h` for 1 hour). For example, to run the tests with a 90 minute timeout:
```
brikis98: 1. Make `-timeout` part of the default command above. It should *always* be there for infrastructure testing.
1. Expand this explanation to explain that if you hit the timeout, Go kills the test, preventing it from finishing, and preventing the cleanup code from running.

yorinasub17:
Addressed all comments in fd73724

yorinasub17:
Going to go ahead and merge this. Thanks for the reviews!

PR #330, opened 2019-06-25
Expose FindNodePortE and allow filtering nodes

yorinasub17:
This improves support for testing multi nodegroup kubernetes clusters. Specifically, right now `GetServiceEndpoint` will pick a random node from the full list of nodes. This is problematic for multi nodegroup clusters because some nodes might have restricted network settings, and thus are not intended to be accessed.

On the otherhand, it isn't clear to me what the right interface for `GetServiceEndpoint` would be to support restricting the node selection. Since this is a pretty specific use case, rather than supporting it in the generic `GetServiceEndpoint` function, I decided to expose the building blocks to support a user to implement this. This entails:

- Exposing `findNodePortE` as a public function so that users can easily find the associated node port for a service even outside the module.
- Exposing `findNodeHostnameE` as a public function
- Exposing `GetNodesByFilterE`, which can be used to filter nodes

### DIFF
--- a/modules/k8s/node.go
+++ b/modules/k8s/node.go
@@ -23,6 +23,12 @@ func GetNodes(t *testing.T, options *KubectlOptions) []corev1.Node {
 
 // GetNodesE queries Kubernetes for information about the worker nodes registered to the cluster.
 func GetNodesE(t *testing.T, options *KubectlOptions) ([]corev1.Node, error) {
+	return GetNodesByFilterE(t, options, metav1.ListOptions{})
+}
+
+// GetNodesByFilterE queries Kubernetes for information about the worker nodes registered to the cluster, filtering the
+// list of nodes using the provided ListOptions.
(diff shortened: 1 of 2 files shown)


yorinasub17:
Welp I just tried running the build three times and all three times a different test failure happened.

FWIW, this only touches the k8s module, and all three times the `kubernetes_test` and `helm_test` portion of the build passed: https://circleci.com/gh/gruntwork-io/terratest/tree/yori-expose-findnodeport

robmorgan:
@yorinasub17 ah dang! I've retriggered the build now, I saw some of these problematic GCP tests failing. I think I might need to do another pass at improving their reliability/retry logic.

robmorgan:
Seemed to work okay.

yorinasub17:
Ok going to merge this in. Thanks for the review! I do need to go through the endless cycle of testing to get it to pass on the tag so the binaries show up, but I am confident it will eventually pass.

PR #336, opened 2019-07-19
Fixes to PR #282

drmmarsunited:
Based on the comments to complete PR #282 I am submitting the following changes:

Added a TLS configuration pointer argument to all HTTP methods
Split out custom http_helper error types to `errors.go` as in other modules
Updated all references to existing HTTP methods to include blank TLS config struct
Updated all example files that called HTTP methods to include blank TLS config struct

### DIFF
--- a/modules/http-helper/http_helper.go
+++ b/modules/http-helper/http_helper.go
@@ -16,8 +16,8 @@ import (
 
 // HttpGet performs an HTTP GET, with an optional pointer to a custom TLS configuration, on the given URL and
 // return the HTTP status code and body. If there's any error, fail the test.
-func HttpGet(t *testing.T, url string, TLSConfig *tls.Config) (int, string) {
-	statusCode, body, err := HttpGetE(t, url, TLSConfig)
+func HttpGet(t *testing.T, url string, tlsConfig *tls.Config) (int, string) {
+	statusCode, body, err := HttpGetE(t, url, tlsConfig)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -26,14 +26,19 @@ func HttpGet(t *testing.T, url string, TLSConfig *tls.Config) (int, string) {
 
 // HttpGetE performs an HTTP GET, with an optional pointer to a custom TLS configuration, on the given URL and
 // return the HTTP status code, body, and any error.
-func HttpGetE(t *testing.T, url string, TLSConfig *tls.Config) (int, string, error) {
+func HttpGetE(t *testing.T, url string, tlsConfig *tls.Config) (int, string, error) {
 	logger.Logf(t, "Making an HTTP GET call to URL %s", url)
 
-	http.DefaultTransport.(*http.Transport).TLSClientConfig = TLSConfig
+	// Set HTTP client transport config
+	tr := &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
 
 	client := http.Client{
 		// By default, Go does not impose a timeout, so an HTTP connection attempt can hang for a LONG time.
 		Timeout: 10 * time.Second,
+		// Include the previously created transport config
+		Transport: tr,
 	}
 
 	resp, err := client.Get(url)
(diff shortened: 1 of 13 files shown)


modules/http-helper/http_helper.go
```diff
@@ -13,19 +14,23 @@ import (
...
 }
 
-// HttpGetE performs an HTTP GET on the given URL and return the HTTP status code, body, and any error.
-func HttpGetE(t *testing.T, url string) (int, string, error) {
+// HttpGetE performs an HTTP GET, with an optional pointer to a custom TLS configuration, on the given URL and
+// return the HTTP status code, body, and any error.
+func HttpGetE(t *testing.T, url string, TLSConfig *tls.Config) (int, string, error) {
 	logger.Logf(t, "Making an HTTP GET call to URL %s", url)
 
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = TLSConfig
```
yorinasub17: I'm not sure this is a good idea, as this updates the http library globally. This can have negative side effects for other libraries that depend on the http library.

It would be better to only set it on the client object being created in the context of this function. You can do that by wrapping the config on the transport:

```
tr := &http.Transport{
    TLSClientConfig: tlsConfig,
}

c := &http.Client{Transport: tr}
```

modules/http-helper/http_helper.go
```diff
@@ -13,19 +14,23 @@ import (
...
 		t.Fatal(err)
 	}
 	return statusCode, body
 }
 
-// HttpGetE performs an HTTP GET on the given URL and return the HTTP status code, body, and any error.
-func HttpGetE(t *testing.T, url string) (int, string, error) {
+// HttpGetE performs an HTTP GET, with an optional pointer to a custom TLS configuration, on the given URL and
+// return the HTTP status code, body, and any error.
+func HttpGetE(t *testing.T, url string, TLSConfig *tls.Config) (int, string, error) {
```
yorinasub17: It is unconventional in go to name local variables starting with caps. Use `tlsConfig`.

modules/http-helper/http_helper.go
drmmarsunited: Good shout.  Will update.

drmmarsunited:
Comments addressed with new commit

yorinasub17:
The updated changes look good! I am going to kick off the build now.

yorinasub17:
Ah looks like you missed a few imports. If you run `goimports` on the files `test/helm_basic_example_integration_test.go` and `test/kubernetes_basic_example_service_check_test.go`, it will take care of it.

drmmarsunited:
Gonna have to go smack my IDE around, it's supposed to do running that for me heh.  Fixed!

yorinasub17:
Ah the build failed again because there are a few compile errors. I just realized that your IDE is probably being confused because of all the build tags...

I know it isn't ideal, but you can reproduce it from the command line by running off the root of the repo:

```
(cd modules/k8s && go test -v -timeout 60m -count 1 -tags kubeall -run TestFoo .)
(cd modules/helm && go test -v -timeout 60m -count 1 -tags kubeall -run TestFoo .)
(cd test && go test -v -timeout 60m -count 1 -tags kubeall -run TestFoo .)
```

drmmarsunited:
Ok, squashed a more references in there I previously missed.  You are definitely right about my IDE not playing nice with those build tags.

yorinasub17:
I'm really sorry. My sloppiness in `modules/k8s` led to another build error here :(

```
# github.com/gruntwork-io/terratest/modules/k8s [github.com/gruntwork-io/terratest/modules/k8s.test]
modules/k8s/service_test.go:91:50: not enough arguments in call to http_helper.HttpGetWithRetryWithCustomValidation
	have (*testing.T, string, number, time.Duration, func(int, string) bool)
	want (*testing.T, string, *tls.Config, int, time.Duration, func(int, string) bool)
```

I forgot to add the `kubeall` tag to one of the files, so it wasn't being picked up in the previous command. The following command should uncover it for you:
```
(cd modules/k8s && go test -v -timeout 60m -count 1 -tags kubernetes -run TestFoo .)
```

drmmarsunited:
@yorinasub17 I fixed what you called out but stumbled upon something else merged into master:

https://github.com/gruntwork-io/terratest/blob/master/modules/helm/upgrade_test.go#L21

Typically you guys use the Testify library for your tests, but here a different `assert` library was used.  Do you want me to change this up to use Testify while I'm making changes?

yorinasub17:
Gah I completely missed that in the review. Thanks for catching that. No need to address that - I'll open a separate PR for it. I don't want to delay this PR further than it already has with unrelated issues.

drmmarsunited:
@yorinasub17 I saw the helm module fail tests in CircleCI, saw that I missed one last damn HTTP helper call.  FML.  Pushed a new commit to fix.

yorinasub17:
Build passed, so going to merge this. Thanks for working through the long back and forth!

PR #340, opened 2019-07-24
Improve unit test code in namespace.go

tariq1890:
Some minor code improvements in the unit tests.

### DIFF
--- a/modules/k8s/namespace.go
+++ b/modules/k8s/namespace.go
@@ -31,22 +31,22 @@ func CreateNamespaceE(t *testing.T, options *KubectlOptions, namespaceName strin
 }
 
 // GetNamespace will query the Kubernetes cluster targeted by the provided options for the requested namespace. This will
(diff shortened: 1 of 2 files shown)


modules/k8s/namespace.go
```diff
@@ -32,21 +32,21 @@ func CreateNamespaceE(t *testing.T, options *KubectlOptions, namespaceName strin
 
 // GetNamespace will query the Kubernetes cluster targeted by the provided options for the requested namespace. This will
 // fail the test if there is an error in creating the namespace.
```
yorinasub17: Can you update the docs to mention that this will also fail the test if the namespace doesn't exist?

tariq1890: Sure!

tariq1890: Done

yorinasub17:
Build passed, so going to merge this. Thanks for your contribution!

PR #335, opened 2019-07-18
Adding support for upgrade and rollback operation

Saurabh-16:
As of now helm module only support install and delete operations. Adding support for upgrade and rollback operations also.

PR #341, opened 2019-07-24
Update assert library used in test

yorinasub17:
The test introduced in https://github.com/gruntwork-io/terratest/pull/335 was using a different assertion library from what we normally use. This PR updates that to use the one we normally depend on.


yorinasub17:
Merging. Thanks for the review!

PR #343, opened 2019-07-26
SSH with password as auth method

### LINKED ISSUE #339: SSHing with username and password - unclear as to how/if possible
I'm trying to set up a test around a Terraform module which handles Linux VMs.
Public/private keys aren't in use by this client; they join Linux machines to domains and run auth that way, so the module doesn't have key functionality written into it.

From looking at the [Terratest module](https://godoc.org/github.com/gruntwork-io/terratest/modules/ssh) and [x/crypto/ssh](https://godoc.org/golang.org/x/crypto/ssh) godocs on SSH, I can see there is a [Password AuthMethod](https://godoc.org/golang.org/x/crypto/ssh#Password) but it's not clear to me if this can be leveraged by Terratest.

I'd also like to say thanks very much for the Terratest package as a whole, it's a valuable part of our pipeline!

jlucktay:
This resolves #339 by adding passwords as an additional alternative auth method to `ssh.Host`.

I duplicated the files and patterns from the existing SSH test and decided to set up in a separate `.go` test file and `/examples/` sub-directory as it seemed cleaner than tiptoeing around all of the key pair plumbing.

Hope this is of some use!


jlucktay:
@yorinasub17 is this going to be OK to merge now after I made the `main.tf` change you asked for?

yorinasub17:
Hi @jlucktay ,

Apologies for the delay here. I have been pulled in for a few other things that has got me delayed on PR reviews. With that said, this is good to go to merge and release, but we haven't had the time to make the motion yet. I am hoping to pick this up tomorrow.

yorinasub17:
Apologies for the delay and thanks for your patience! Going to merge this now.

yorinasub17:
https://github.com/gruntwork-io/terratest/releases/tag/v0.18.1

PR #347, opened 2019-08-01
Fix more dep issues

yorinasub17:
Found another dep issue, related to `k8s.io/kubernetes`. This locks that version down. Also replaces `override` with `constraint`, because `override` doesn't propagate down to projects depending on `terratest`.


yorinasub17:
Got the build to pass finally. Added another packer retry to try to make it more robust. Going to merge and release now. Thanks for the review!

PR #333, opened 2019-07-13
Adding a basic framework for a new Azure module

drmmarsunited:
Includes support for:
- Getting authorized access based on env variables or arguments
- 2 basic tests against Azure Virtual Machines, just as a starting point

The idea here was just to socialize an approach to the beginnings of an Azure module.  In true open source fashion, someone else contributed an approach to the Azure model 2 hours before me here haha!

I'm trying to approach the testing methods in a way that somewhat closely aligns to the existing methods for AWS, just to keep the style in line with the rest of the suite.  Happy to take consolidated feedback from you (the maintainers) and work together with the author of the other PR on an approach that will accelerate Azure capabilities in Terratest.

Issue link: #89 
Related PR: #332

PR #353, opened 2019-08-23
Fix build

yorinasub17:
This fixes the failing build introduced from merging https://github.com/gruntwork-io/terratest/pull/333

- We don't have CircleCI hooked to azure yet, so use build tags to disable the azure tests temporarily.
- Run `goimports`
- Run `dep ensure` and check in lock file


robmorgan:
I'm getting these errors in another PR, so I'm going to merge this one now.

PR #355, opened 2019-08-27
Generic http helper methods

jendrikjoe:
This MR is related to #186 and based upon @sw-samuraj work on #193.

In the original MR @sw-samuraj added HTTP helper methods which mimic current HttpGet* methods. Those new methods allow to:

    add an HTTP header(s),
    add a request body,
    set a HTTP method,
    returns a response body.

I changed the signature from the original proposal to now be:

func HTTPDo(t *testing.T, method string, url string, body io.Reader, headers map[string]string) (int, string)

I additionally changed the original tests to now use httptest.Server. This allows to run the tests in parallel without any interference. Finally, I expanded the test coverage to `HTTPDoWithCustomValidation` and `HTTPDoWithValidation`.

I tried to keep the git history intact to keep @sw-samuraj contributions visible. If there is a better way to do this, please note it and I will happily adapt to it :slightly_smiling_face:

### DIFF
--- a/modules/http-helper/dummy_server.go
+++ b/modules/http-helper/dummy_server.go
@@ -46,24 +46,26 @@ func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
 	return listener, port, err
 }
 
-// RunDummyHandlerServer runs a dummy HTTP server on a unique port that will serve the given handler. Returns the Listener for the server, the
-// port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
+// RunDummyHandlerServer runs a dummy HTTP server on a unique port that will serve given handlers. Returns the Listener for the server,
+// the port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
 // the Close() method on the Listener when you're done!
-func RunDummyHandlerServer(t *testing.T, path string, handler func(http.ResponseWriter, *http.Request)) (net.Listener, int) {
-	listener, port, err := RunDummyHandlerServerE(t, path, handler)
+func RunDummyHandlerServer(t *testing.T, handlers map[string]func(http.ResponseWriter, *http.Request)) (net.Listener, int) {
+	listener, port, err := RunDummyHandlerServerE(t, handlers)
 	if err != nil {
 		t.Fatal(err)
 	}
 	return listener, port
 }
 
-// RunDummyHandlerServerE runs a dummy HTTP server on a unique port that will server the given handler. Returns the Listener for the server, the
-// port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
+// RunDummyHandlerServerE runs a dummy HTTP server on a unique port that will server given handlers. Returns the Listener for the server,
+// the port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
 // the Close() method on the Listener when you're done!
-func RunDummyHandlerServerE(t *testing.T, path string, handler func(http.ResponseWriter, *http.Request)) (net.Listener, int, error) {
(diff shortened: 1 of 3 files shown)


brikis98:
@robmorgan Could you review this one?

modules/http-helper/dummy_server.go
```diff
@@ -46,6 +46,39 @@ func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
 	return listener, port, err
 }
 
+// RunDummyHandlerServer runs a dummy HTTP server on a unique port that will serve given handlers. Returns the Listener for the server,
+// the port it's listening on, or an error if something went wrong while trying to start the listener. Make sure to call
+// the Close() method on the Listener when you're done!
+func RunDummyHandlerServer(t *testing.T, handlers map[string]func(http.ResponseWriter, *http.Request)) (net.Listener, int) {
```
robmorgan:
```suggestion
func RunDummyServerWithHandlers(t *testing.T, handlers map[string]func(http.ResponseWriter, *http.Request)) (net.Listener, int) {
```
I propose renaming the functions to `RunDummyServerWithHandlers` and `RunDummyServerWithHandlersE` as I feel it follows the other Terratest functions more closely and the verb-style function names in the go `http` package: https://golang.org/pkg/net/http/#example_ListenAndServe. What do you think?

modules/http-helper/http_helper.go
```diff
@@ -56,7 +57,7 @@ func HttpGetE(t *testing.T, url string, tlsConfig *tls.Config) (int, string, err
 	return resp.StatusCode, strings.TrimSpace(string(body)), nil
 }
 
-// HttpGetWithValidation performs an HTTP GET on the given URL and verify that you get back the expected status code and body. If either
+// HTTPGetWithValidation performs an HTTP GET on the given URL and verify that you get back the expected status code and body. If either
```
robmorgan:
```suggestion
// HttpGetWithValidation performs an HTTP GET on the given URL and verify that you get back the expected status code and body. If either
```
These comments should match the function name 1:1. e.g `HttpGetWithValidation`

modules/http-helper/http_helper.go
```diff
@@ -65,7 +66,7 @@ func HttpGetWithValidation(t *testing.T, url string, tlsConfig *tls.Config, expe
 	}
 }
 
-// HttpGetWithValidationE performs an HTTP GET on the given URL and verify that you get back the expected status code and body. If either
+// HTTPGetWithValidationE performs an HTTP GET on the given URL and verify that you get back the expected status code and body. If either
```
robmorgan: Same here

jendrikjoe: Fixed, too.

modules/http-helper/http_helper.go
jendrikjoe: Should have caught that myself by looking at the linter, sorry :wink:

modules/http-helper/dummy_server.go
jendrikjoe: Makes sense :+1: Adapted to it.

modules/http-helper/dummy_server.go
```diff
@@ -46,6 +46,39 @@ func RunDummyServerE(t *testing.T, text string) (net.Listener, int, error) {
 	return listener, port, err
 }
 
+// RunDummyHandlerServer runs a dummy HTTP server on a unique port that will serve given handlers. Returns the Listener for the server,
```
robmorgan: Ok just fix these comments then I think we're good to run the tests.

jendrikjoe: Fixed :wink:

robmorgan:
Okay I've trigged a build: https://circleci.com/gh/gruntwork-io/terratest/2075

robmorgan:
Okay I managed to make the tests past, so I'll merge this to master onw.

Issue #359, opened 2019-09-02
Helper for AWS Vpcs: GetVpcById


kasia-kittel:
Hi. While testing my infra, I needed to write some helpers functions for AWS VPCs: `GetVpcByIdE` `GetVpcById`. Happy to contribute the code.

PR #360, opened 2019-09-02
[#359] Extend AWS Vpc helpers: GetVpcById

kasia-kittel:
This is the code for the issue: #359
It extends the AWS Vpc module with `GetVpcById`, `GetVpcByIdE`, and `GetVpcsE`. 
Backward compatibility is maintained. The test temporarily creates a Vpc.

Tests output: https://gist.github.com/kasia-kittel/ee1389f395fc85853105c47ad2022e1f

### DIFF
--- a/modules/aws/vpc.go
+++ b/modules/aws/vpc.go
@@ -83,7 +83,7 @@ func GetVpcsE(t *testing.T, filters []*ec2.Filter, region string) ([]*Vpc, error
 
 	numVpcs := len(vpcs.Vpcs)
 	retVal := make([]*Vpc, numVpcs)
-	
+
 	for i, vpc := range vpcs.Vpcs {
 		subnets, err := GetSubnetsForVpcE(t, aws.StringValue(vpc.VpcId), region)
 		if err != nil {
@@ -91,7 +91,7 @@ func GetVpcsE(t *testing.T, filters []*ec2.Filter, region string) ([]*Vpc, error
 		}
 		retVal[i] = &Vpc{Id: aws.StringValue(vpc.VpcId), Name: FindVpcName(vpc), Subnets: subnets}
 	}
-	
+
 	return retVal, nil
 }
 
--- a/modules/aws/vpc.go
+++ b/modules/aws/vpc.go
@@ -9,6 +9,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
 )
 
 // Vpc is an Amazon Virtual Private Cloud.
@@ -31,9 +32,7 @@ var isDefaultFilterValue = "true"
 // GetDefaultVpc fetches information about the default VPC in the given region.
 func GetDefaultVpc(t *testing.T, region string) *Vpc {
 	vpc, err := GetDefaultVpcE(t, region)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	return vpc
 }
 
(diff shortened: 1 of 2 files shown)


modules/aws/vpc.go
```diff
@@ -39,30 +39,60 @@ func GetDefaultVpc(t *testing.T, region string) *Vpc {
...
-	vpcs, err := client.DescribeVpcs(&ec2.DescribeVpcsInput{Filters: []*ec2.Filter{&defaultVpcFilter}})
+	return vpcs[0], err
+}
+
+func GetVpcById(t *testing.T, vpcId string, region string) *Vpc {
+	vpc, err := GetVpcByIdE(t, vpcId, region)
 	if err != nil {
-		return nil, err
+		t.Fatal(err)
 	}
```
yorinasub17: We have a new pattern for this now, where we use `require.NoError(t, err)`. I know this file hasn't been updated to use the new pattern yet, but we like to update whenever we notice the old pattern. Can you go ahead to update this here, along with adding in the relevant import? Thanks!

modules/aws/vpc.go
```diff
@@ -39,30 +39,60 @@ func GetDefaultVpc(t *testing.T, region string) *Vpc {
...
 
-	numVpcs := len(vpcs.Vpcs)
+func GetVpcByIdE(t *testing.T, vpcId string, region string) (*Vpc, error) {
+	vpcIdFilter := ec2.Filter{Name: &vpcIDFilterName, Values: []*string{&vpcId}}
+	vpcs, err := GetVpcsE(t, []*ec2.Filter{&vpcIdFilter}, region)
+
+	numVpcs := len(vpcs)
 	if numVpcs != 1 {
-		return nil, fmt.Errorf("Expected to find one default VPC in region %s but found %s", region, strconv.Itoa(numVpcs))
+		return nil, fmt.Errorf("Expected to find one VPC in region %s but found %s", region, strconv.Itoa(numVpcs))
```
yorinasub17:
```suggestion
		return nil, fmt.Errorf("Expected to find one VPC with ID %s in region %s but found %s", vpcId, region, strconv.Itoa(numVpcs))
```

modules/aws/vpc_test.go
```diff
@@ -17,6 +21,37 @@ func TestGetDefaultVpc(t *testing.T) {
...
+}
+
+func TestGetVpcsE(t *testing.T) {
+	t.Parallel()
+
+	region := GetRandomStableRegion(t, nil, nil)
+	azs := GetAvailabilityZones(t, region)
+
+	var isDefaultFilterName = "isDefault"
+	var isDefaultFilterValue = "true"
```
yorinasub17: Canonical way to declare and initialize vars in golang is `:=`. E.g `isDefaultFilterValue := "true"`

modules/aws/vpc_test.go
```diff
@@ -17,6 +21,37 @@ func TestGetDefaultVpc(t *testing.T) {
...
+
+	defaultVpcFilter := ec2.Filter{Name: &isDefaultFilterName, Values: []*string{&isDefaultFilterValue}}
+	vpcs, _ := GetVpcsE(t, []*ec2.Filter{&defaultVpcFilter}, region)
+
+	assert.Equal(t, len(vpcs), 1)
```
yorinasub17: This should be `require.Equal` so that the line below doesn't `panic` when there are no VPCs.

yorinasub17: To clarify, `assert` marks the test as `failed` when the assertion fails, but the test continues. `require` does the same thing and also halts the execution of the function.

modules/aws/vpc.go
kasia-kittel: done

kasia-kittel:
I addressed all comments in folling commits.

yorinasub17:
Looks good! Will kick off our build now.

yorinasub17:
Build passed, so going to merge this in and release. Thanks for your contribution!

yorinasub17:
https://github.com/gruntwork-io/terratest/releases/tag/v0.18.6

Issue #364, opened 2019-09-11
http_helper with retries have empty body after first request


jendrikjoe:
The way http_helper is implement, it is using io.Reader for all functions, even the retry ones.
If the retry is actually triggered, the body of the request is empty from the second request onwards as the reader is empty. 
I missed that case when opening #355. I will take care of implementing the retry functions such that they take `[]byte` as arguments and writing a test that is making sure this doesn't happen again :slightly_smiling_face:

yorinasub17:
Fixed in https://github.com/gruntwork-io/terratest/releases/tag/v0.19.0

PR #365, opened 2019-09-11
Use []byte for HTTP methods with retry

jendrikjoe:
This MR is addressing #364.
It is changing the type of the body argument from `io.Reader` to `[]byte` for all HTTP methods with retry functionality and a body.
This ensures that the body is not read upon the first request and then gone for good.
I additionally added a test, ensuring that this is indeed the case.

### DIFF
--- a/modules/http-helper/http_helper.go
+++ b/modules/http-helper/http_helper.go
@@ -138,7 +138,10 @@ func HttpGetWithRetryWithCustomValidationE(t *testing.T, url string, tlsConfig *
 
 // HTTPDo performs the given HTTP method on the given URL and return the HTTP status code and body.
 // If there's any error, fail the test.
-func HTTPDo(t *testing.T, method string, url string, body io.Reader, headers map[string]string) (int, string) {
+func HTTPDo(
+	t *testing.T, method string, url string, body io.Reader,
+	headers map[string]string,
+) (int, string) {
 	statusCode, respBody, err := HTTPDoE(t, method, url, body, headers)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


modules/http-helper/http_helper.go
```diff
@@ -173,8 +174,11 @@ func HTTPDoE(t *testing.T, method string, url string, body io.Reader, headers ma
 // HTTPDoWithRetry repeatedly performs the given HTTP method on the given URL until the given status code and body are
 // returned or until max retries has been exceeded.
 // The function compares the expected status code against the received one and fails if they don't match.
-func HTTPDoWithRetry(t *testing.T, method string, url string, body io.Reader, headers map[string]string, expectedStatus int, retries int, sleepBetweenRetries time.Duration) string {
-	out, err := HTTPDoWithRetryE(t, method, url, body, headers, expectedStatus, retries, sleepBetweenRetries)
+func HTTPDoWithRetry(t *testing.T, method string, url string,
+	body []byte, headers map[string]string, expectedStatus int,
+	retries int, sleepBetweenRetries time.Duration) string {
```
yorinasub17: Can you format this to:

```
func HTTPDoWithRetry(
    t *testing.T,
    method string,
    ... and so on ...,
) string {
```

modules/http-helper/http_helper_test.go
```diff
@@ -165,6 +170,8 @@ func retryHandler(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusServiceUnavailable)
```
yorinasub17: Can you read the body here as well, so that we read from the body multiple times? I believe the HTTP functions read it internally as well, but this at least makes it explicit.

jendrikjoe: Added the readAll here :+1:

modules/http-helper/http_helper_test.go
```diff
@@ -176,5 +184,7 @@ func failRetryHandler(w http.ResponseWriter, r *http.Request) {
...
```
yorinasub17: I had implicitly meant to read the body here as well but forgot to add a note. Can you do that? Sorry for the extra round trip this caused!

jendrikjoe: Well, now your mentioning it, it is kind of obvious, so no worries for the extra round trip :slightly_smiling_face: Changed it :+1:

yorinasub17:
Thanks for the changes! Just kicked off a build.

yorinasub17:
Build passed and this change looks good, so going to merge this in. Thanks for your contribution!

jendrikjoe:
Thanks for reviewing it and in general for this great repo :slightly_smiling_face:

Issue #369, opened 2019-09-19
stack overflow exception when using the terraform.OutputList  


ajayk:
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0x14c7806, 0xe)
	/usr/local/Cellar/go/1.13/libexec/src/runtime/panic.go:774 +0x72
runtime.newstack()
	/usr/local/Cellar/go/1.13/libexec/src/runtime/stack.go:1046 +0x6e9
runtime.morestack()
	/usr/local/Cellar/go/1.13/libexec/src/runtime/asm_amd64.s:449 +0x8f

goroutine 20 [running]:
runtime.assertE2I2(0x1448a60, 0x14363a0, 0xc008352680, 0x0, 0x0, 0x0)
	/usr/local/Cellar/go/1.13/libexec/src/runtime/iface.go:482 +0x9e fp=0xc0204fe360 sp=0xc0204fe358 pc=0x100bf9e
fmt.(*pp).handleMethods(0xc00835a1a0, 0x71, 0xc0204fe600)
	/usr/local/Cellar/go/1.13/libexec/src/fmt/print.go:594 +0x88 fp=0xc0204fe570 sp=0xc0204fe360 pc=0x10c6da8
fmt.(*pp).printArg(0xc00835a1a0, 0x14363a0, 0xc008352680, 0x71)
	/usr/local/Cellar/go/1.13/libexec/src/fmt/print.go:713 +0x206 fp=0xc0204fe608 sp=0xc0204fe570 pc=0x10c75e6
fmt.(*pp).doPrintf(0xc00835a1a0, 0x14d7451, 0x2d, 0xc0204fe780, 0x1, 0x1)
	/usr/local/Cellar/go/1.13/libexec/src/fmt/print.go:1030 +0x15b fp=0xc0204fe6f0 sp=0xc0204fe608 pc=0x10cafab
fmt.Sprintf(0x14d7451, 0x2d, 0xc0204fe780, 0x1, 0x1, 0x14489e0, 0x102f44a)
	/usr/local/Cellar/go/1.13/libexec/src/fmt/print.go:219 +0x66 fp=0xc0204fe748 sp=0xc0204fe6f0 pc=0x10c3e86
github.com/gruntwork-io/terratest/modules/terraform.OutputKeyNotFound.Error(...)
	/Users/zzz/go/src/github.com/gruntwork-io/terratest/modules/terraform/errors.go:20
[...]

PR #370, opened 2019-09-19
Fixes issue #369 : wrap the error in string

ajayk:


brikis98:
Ohhh, nice fix!

Could you add a test case for this? There should already be tests for `OutputList` in the repo, so it's a matter of adding one more check that intentionally looks up an invalid key and ensure you get the expected error. Thanks!

modules/terraform/errors.go
```diff
@@ -17,7 +17,7 @@ func (err TgInvalidBinary) Error() string {
 type OutputKeyNotFound string
 
 func (err OutputKeyNotFound) Error() string {
-	return fmt.Sprintf("output doesn't contain a value for the key %q", err)
+	return fmt.Sprintf("output doesn't contain a value for the key %q", string(err))
```
robmorgan, line 20:
```suggestion
	return fmt.Sprintf("output doesn't contain a value for the key %s", err)
```
Could we potentially simplify this?

tarndt: So the bug here is that we don't want to call `Error()` inside of `Error()` (recursion). Chaging the `fmt` verb from `%q` to `%s` does nothing to prevent that as that will result in `Error()` being called (as `Error()` is called if a type has it and does not have `String()`. The approach here is exactly the canonical approach [documented in the fmt pkg](https://golang.org/pkg/fmt/):

> To avoid recursion in cases such as
> 
> type X string
> func (x X) String() string { return Sprintf("<%s>", x) }
> 
> convert the value before recurring:
> 
> func (x X) String() string { return Sprintf("<%s>", string(x)) }
>

ellisonc (approved):

PR #371, opened 2019-09-24
using a interface instead of *testing.T

fr0stbyte:
I would like to use `terratest` with [ginkgo](https://github.com/onsi/ginkgo).
Because most of the terratest functions expect a `*testing.T` object, I thought I could use `GinkgoT()` however this doesn't work because terratest doesn't use an interface but uses the concrete impl.

This PR attempts to fix that by creating a `testing.TestingT` template that is then used throughout the project instead of `*testing.T` object.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,9 +72,6 @@ jobs:
             pre-commit install
             pre-commit run --all-files
 
-      # verify that the testing.TestingT interface is used instead of testing.T
-      - run: grep -Rw "*testing\.T" modules | grep -v _test.go
-
       # Build any binaries that need to be built
       # We always want to build the binaries, because we will use the terratest_log_parser to parse out the test output
       # during a failure.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ jobs:
             pre-commit run --all-files
 
       # verify that the testing.TestingT interface is used instead of testing.T
-      - run: grep -Rw "*testing\.T" modules | grep -v _test
+      - run: grep -Rw "*testing\.T" modules | grep -v _test.go
 
       # Build any binaries that need to be built
       # We always want to build the binaries, because we will use the terratest_log_parser to parse out the test output
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,9 @@ jobs:
             pre-commit install
             pre-commit run --all-files
 
+      # verify that the testing.TestingT interface is used instead of testing.T
+      - run: grep -Rw "*testing\.T" modules | grep -v _test
+
       # Build any binaries that need to be built
       # We always want to build the binaries, because we will use the terratest_log_parser to parse out the test output
       # during a failure.
--- a/modules/aws/account.go
+++ b/modules/aws/account.go
@@ -3,14 +3,15 @@ package aws
 import (
 	"errors"
 	"strings"
-	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/sts"
+
+	_ "github.com/gruntwork-io/terratest/modules/testing"
 )
 
 // GetAccountId gets the Account ID for the currently logged in IAM User.
-func GetAccountId(t *testing.T) string {
+func GetAccountId(t TestingT) string {
 	id, err := GetAccountIdE(t)
 	if err != nil {
 		t.Fatal(err)
@@ -19,7 +20,7 @@ func GetAccountId(t *testing.T) string {
 }
 
 // GetAccountIdE gets the Account ID for the currently logged in IAM User.
-func GetAccountIdE(t *testing.T) (string, error) {
+func GetAccountIdE(t TestingT) (string, error) {
 	stsClient, err := NewStsClientE(t, defaultRegion)
 	if err != nil {
 		return "", err
@@ -46,7 +47,7 @@ func extractAccountIDFromARN(arn string) (string, error) {
 }
 
 // NewStsClientE creates a new STS client.
-func NewStsClientE(t *testing.T, region string) (*sts.STS, error) {
+func NewStsClientE(t TestingT, region string) (*sts.STS, error) {
 	sess, err := NewAuthenticatedSession(region)
 	if err != nil {
 		return nil, err
--- a/modules/aws/account.go
+++ b/modules/aws/account.go
@@ -7,11 +7,11 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/sts"
 
-	_ "github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
 // GetAccountId gets the Account ID for the currently logged in IAM User.
-func GetAccountId(t TestingT) string {
+func GetAccountId(t testing.TestingT) string {
 	id, err := GetAccountIdE(t)
 	if err != nil {
 		t.Fatal(err)
@@ -20,7 +20,7 @@ func GetAccountId(t TestingT) string {
 }
 
 // GetAccountIdE gets the Account ID for the currently logged in IAM User.
-func GetAccountIdE(t TestingT) (string, error) {
+func GetAccountIdE(t testing.TestingT) (string, error) {
 	stsClient, err := NewStsClientE(t, defaultRegion)
 	if err != nil {
 		return "", err
@@ -47,7 +47,7 @@ func extractAccountIDFromARN(arn string) (string, error) {
 }
 
 // NewStsClientE creates a new STS client.
-func NewStsClientE(t TestingT, region string) (*sts.STS, error) {
+func NewStsClientE(t testing.TestingT, region string) (*sts.STS, error) {
 	sess, err := NewAuthenticatedSession(region)
 	if err != nil {
 		return nil, err
--- a/modules/ssh/agent.go
+++ b/modules/ssh/agent.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 
 	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/gruntwork-io/terratest/modules/logger"
 	"golang.org/x/crypto/ssh/agent"
 )
 
@@ -57,7 +58,7 @@ func (s *SshAgent) run(t testing.TestingT) {
 					return
 					// When s.ln.Accept() returns a legit error, we print it and continue accepting further requests
 				default:
-					t.Logf("could not accept connection to agent %v", err)
+					logger.Logf(t, "could not accept connection to agent %v", err)
 					continue
 				}
 			} else {
@@ -65,7 +66,7 @@ func (s *SshAgent) run(t testing.TestingT) {
 				go func(c io.ReadWriter) {
 					err := agent.ServeAgent(s.agent, c)
 					if err != nil {
-						t.Logf("could not serve ssh agent %v", err)
+						logger.Logf(t, "could not serve ssh agent %v", err)
 					}
 				}(c)
 			}
@@ -111,7 +112,7 @@ func SshAgentWithKeyPairs(t testing.TestingT, keyPairs []*KeyPair) *SshAgent {
 // Instantiates and returns an in-memory ssh agent with the given KeyPair(s) already added
 // You should stop the agent to cleanup files afterwards by calling `defer sshAgent.Stop()`
 func SshAgentWithKeyPairsE(t testing.TestingT, keyPairs []*KeyPair) (*SshAgent, error) {
-	t.Log("Generating SSH Agent with given KeyPair(s)")
+	logger.Logf(t, "Generating SSH Agent with given KeyPair(s)")
 
 	// Instantiate a temporary SSH agent
 	socketDir, err := ioutil.TempDir("", "ssh-agent-")
--- a/modules/ssh/agent.go
+++ b/modules/ssh/agent.go
@@ -9,8 +9,8 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/testing"
 	"golang.org/x/crypto/ssh/agent"
 )
 
--- a/modules/ssh/agent.go
+++ b/modules/ssh/agent.go
@@ -9,7 +9,7 @@ import (
 	"os"
 	"path/filepath"
 
-	_ "github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/gruntwork-io/terratest/modules/testing"
 	"golang.org/x/crypto/ssh/agent"
 )
 
@@ -24,7 +24,7 @@ type SshAgent struct {
 
 // Create SSH agent, start it in background and returns control back to the main thread
 // You should stop the agent to cleanup files afterwards by calling `defer s.Stop()`
-func NewSshAgent(t TestingT, socketDir string, socketFile string) (*SshAgent, error) {
+func NewSshAgent(t testing.TestingT, socketDir string, socketFile string) (*SshAgent, error) {
 	var err error
 	s := &SshAgent{make(chan bool), make(chan bool), socketDir, socketFile, agent.NewKeyring(), nil}
 	s.ln, err = net.Listen("unix", s.socketFile)
@@ -41,7 +41,7 @@ func (s *SshAgent) SocketFile() string {
 }
 
 // SSH Agent listener and handler
-func (s *SshAgent) run(t TestingT) {
+func (s *SshAgent) run(t testing.TestingT) {
 	defer close(s.stopped)
 	for {
 		select {
@@ -83,7 +83,7 @@ func (s *SshAgent) Stop() {
 
 // Instantiates and returns an in-memory ssh agent with the given KeyPair already added
 // You should stop the agent to cleanup files afterwards by calling `defer sshAgent.Stop()`
-func SshAgentWithKeyPair(t TestingT, keyPair *KeyPair) *SshAgent {
+func SshAgentWithKeyPair(t testing.TestingT, keyPair *KeyPair) *SshAgent {
 	sshAgent, err := SshAgentWithKeyPairE(t, keyPair)
 
 	if err != nil {
@@ -93,12 +93,12 @@ func SshAgentWithKeyPair(t TestingT, keyPair *KeyPair) *SshAgent {
 	return sshAgent
 }
 
-func SshAgentWithKeyPairE(t TestingT, keyPair *KeyPair) (*SshAgent, error) {
+func SshAgentWithKeyPairE(t testing.TestingT, keyPair *KeyPair) (*SshAgent, error) {
 	sshAgent, err := SshAgentWithKeyPairsE(t, []*KeyPair{keyPair})
 	return sshAgent, err
 }
 
-func SshAgentWithKeyPairs(t TestingT, keyPairs []*KeyPair) *SshAgent {
+func SshAgentWithKeyPairs(t testing.TestingT, keyPairs []*KeyPair) *SshAgent {
 	sshAgent, err := SshAgentWithKeyPairsE(t, keyPairs)
 
 	if err != nil {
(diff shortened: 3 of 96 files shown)


fr0stbyte:
Example of integration with `ginkgo`

```
package mypkg_test

import (
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"

	"github.com/gruntwork-io/terratest/modules/k8s"
)

var _ = Describe("ClusterPreconditions", func() {
	Context("New cluster", func() {
		options := k8s.NewKubectlOptions("", "")
		nodes := k8s.GetNodes(GinkgoT(), options)
		var masters, etcd, workers int

		for _, n := range nodes {
			for k := range n.ObjectMeta.Labels {
				switch k {
				case "node-role.kubernetes.io/master":
					masters++
				case "node-role.kubernetes.io/etcd":
					etcd++
				default:
					workers++
				}
			}
		}
		It("should have exactly 5 master nodes", func() {
			Expect(masters).Should(Equal(5))
		})
		It("should have exactly 5 etcd nodes", func() {
			Expect(etcd).Should(Equal(5))
		})
		It("should have more than 1 worker node", func() {
			Expect(workers).Should(BeNumerically(">", 1))
		})
	})
})
```

here is the result of running `ginkgo`
```./mypkg_test.go:13:32: cannot use ginkgo.GinkgoT() (type ginkgo.GinkgoTInterface) as type *testing.T in argument to k8s.GetNodes: need type assertion
```

Ofc, type assertion will fail b/c `GinkgoT()` doesn't return a `*testing.T` : https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L98-L123

modules/testing/types.go
```diff
@@ -0,0 +1,18 @@
+package testing
+
+type TestingT interface {
```
brikis98: Please add comments explaining why this interface exists

modules/testing/types.go
```diff
@@ -0,0 +1,18 @@
...
+	Fatal(args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Log(args ...interface{})
+	Logf(format string, args ...interface{})
+	Failed() bool
+	Parallel()
+	Skip(args ...interface{})
+	Skipf(format string, args ...interface{})
+	SkipNow()
+	Skipped() bool
```
brikis98: I suspect in Terratest we only use a couple off these, so this interface can probably be reduced quite a bit.

yorinasub17:
I agree that this is a good idea, but

> Also, I'm wondering if the Terratest methods that return errors (instead of calling t.Fail)—the methods that end in E—shouldn't omit the t parameter entirely?

I do like having the `E` and non-`E` versions take the same parameters for consistency. You also can't completely remove it for logging reasons (since we depend on `t.Name()` in the logger).

modules/testing/types.go
fr0stbyte: This is the interface as exported by ginkgo.
The only functions that I haven't seen referenced in the code are:
```
FailNow()
Skipf(fomat string, args ...interface{})
SkipNow()
Skipped() bool
```
Do you want them removed ?

modules/testing/types.go
fr0stbyte: is there something else you need me to add to the comments here or is this good ?

brikis98:
> I do like having the E and non-E versions take the same parameters for consistency. You also can't completely remove it for logging reasons (since we depend on t.Name() in the logger).

Ah, good point.

modules/testing/types.go
brikis98: I searched the `modules` folder for usages of `t.XXX` in non-test files (i.e., ignoring `xxx_test.go`) files, and the only usages I found were:

1. `t.Fatal()` / `t.Fatalf()`. These are by far the most common.
1. `t.Error()` / `t.Errorf()`. Only used a few times in goroutines because this method is goroutine-safe, unlike `t.Fatal`.
1. `t.Name()`: Used a few times.
1. `t.Logf()`: Used in just one place (`modules/ssh/agent.go`) and this is a bug. It should be replaced with `logger.Logf`. 

So, I think we can simplify this interface down to just 5 methods: two `Fatal` methods, two `Error` methods, and the `Name` method. If we want to be really minimalist, we could do just the `Fatalf` and `Errorf` method, as the ones without `f` are rarely used, and largely redundant. This much smaller interface would be far simpler to implement to integrate Terratest with other use cases!

modules/testing/types.go
brikis98: Once we've narrowed down the set of methods we want (as per the other discussion), we should copy over the docs for each method too, so users know what to implement for each one.

modules/testing/types.go
fr0stbyte: `Name()` method is not offered by `GinkgoT` interface
1. I can send a PR there to add it in the interface
2. Until that gets merged, if this interface requires .Name to exist, then it will not work with the current `gingko` versions

While this approach would be preferrable, it is highly likely that the feature will not be available (if at all) any time soon (given that it's a change at the API and might affect a lot of projects depending on it) 

these are the entries that make use of `t.Name()` in the `modules` dir not counting `_test.go` files:
```
docker/docker_compose.go:29:		Args:       append([]string{"--project-name", t.Name()}, args...),
logger/logger.go:47:	prefix := fmt.Sprintf("%s %s %s:", t.Name(), date.Format(time.RFC3339), CallerPrefix(callDepth+1))
test-structure/test_structure.go:75:	tmpRootFolder, err := files.CopyTerraformFolderToTemp(rootFolder, cleanName(t.Name()))
k8s/kubectl.go:114:	escapedTestName := url.PathEscape(t.Name())
```

Would you be open to the idea of replacing the use of `t.Name` in these files with something else ?

brikis98: Hm, `Name` is definitely important, especially for logging. What else would you propose replacing it with?

fr0stbyte: I think you're right, there isn't an easy way around this. I think I can have a workaround outside of terratest, until ginkgo updates their interface.
I have updated the interface to contain only the methods you mentioned ( including `Name()` )

modules/testing/types.go
fr0stbyte: I have added the documentation for the functions that remained in the interface.
Given that the documentation also talks about implementing certain functions in terms of others, I have added the referred functions as well, I hope that's ok.

modules/aws/account.go
```diff
@@ -3,14 +3,15 @@ package aws
 import (
 	"errors"
 	"strings"
-	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/sts"
+
+	_ "github.com/gruntwork-io/terratest/modules/testing"
```
brikis98: Why import with the `_`?

modules/testing/types.go
```diff
@@ -0,0 +1,38 @@
...
+	// and records the text in the error log. For tests, the text will be printed only if
+	// the test fails or the -test.v flag is set. For benchmarks, the text is always
+	// printed to avoid having performance depend on the value of the -test.v flag.
+	Log(args ...interface{})
+	// Logf formats its arguments according to the format, analogous to Printf, and
+	// records the text in the error log. A final newline is added if not provided. For
+	// tests, the text will be printed only if the test fails or the -test.v flag is
+	// set. For benchmarks, the text is always printed to avoid having performance
+	// depend on the value of the -test.v flag.
+	Logf(format string, args ...interface{})
```
brikis98: I think we can remove the `Log` functions. We use them in one place, but that should be replaced with `logger.Logf` instead.

modules/aws/account.go
fr0stbyte: so as not to qualify it with `testing.TestingT` (which may be confusing to others on the provenance of the interface)
I can change the pkg name on import, would that be more acceptable ?

brikis98: > which may be confusing to others on the provenance of the interface

Confusing in what sense?

modules/ssh/agent.go
```diff
@@ -110,7 +110,7 @@ func SshAgentWithKeyPairs(t *testing.T, keyPairs []*KeyPair) *SshAgent {
 
 // Instantiates and returns an in-memory ssh agent with the given KeyPair(s) already added
 // You should stop the agent to cleanup files afterwards by calling `defer sshAgent.Stop()`
-func SshAgentWithKeyPairsE(t *testing.T, keyPairs []*KeyPair) (*SshAgent, error) {
+func SshAgentWithKeyPairsE(t TestingT, keyPairs []*KeyPair) (*SshAgent, error) {
 	t.Log("Generating SSH Agent with given KeyPair(s)")
```
brikis98: It looks like you removed `Log` and `Logf` from the interface, but not the calls to `t.Log`/`t.Logf`?

modules/testing/types.go
brikis98: You can remove `Logf` too!

fr0stbyte:
@brikis98 sorry for the delay. Can you please take another look ?

modules/testing/types.go
brikis98: Looks like you updated the code to no longer use `t.Logf`, so I think this method can be removed from the interface now.

fr0stbyte:
@brikis98 can you please take another look ?

robmorgan:
Hi @fr0stbyte, @brikis98 is currently out on vacation this week. Given his involvement in this PR, I guess we should wait until he's back next week.

modules/testing/types.go
brikis98: This issue still stands.

fr0stbyte:
@brikis98 that's a great question, but I am not sure what the proper solution is. is code documentation enough ?

brikis98:
I suspect that'll be too error prone. Perhaps add a script in CI that does a `grep` in the `modules` folder for any imports of `testing.T` and fails if it finds any? Not sure if there's a Go linter we can sue to ban certain imports? @yorinasub17 Any ideas?

fr0stbyte:
@brikis98 can you point me to where the CI scripts reside ?

fr0stbyte:
@brikis98 how do I test that my CI directive works ?

.circleci/config.yml
```diff
@@ -66,6 +66,9 @@ jobs:
       - run: pre-commit install
       - run: pre-commit run --all-files
 
+      # verify that the testing.TestingT interface is used instead of testing.T
+      - run: grep -Rw "*testing\.T" modules | grep -v _test
```
brikis98: Should it be `_test.go`?

brikis98:
Ah, could you run `go imports -w .` in the `modules` folder and commit the changes?

fr0stbyte:
@brikis98 done

brikis98:
Looks like something about your `goimports` changes wasn't quite right. Here's the diff I just tried on a parallel branch: https://github.com/gruntwork-io/terratest/commit/1879d24bbc7081b27482ac71d63444740b6a4473 (`goimports` now passes; other tests are still running). You'll need the same in yours.

brikis98:
Tests failed on my branch with the `goimports` fix:

```
Not on branch master, setting the short flag to skip long-running tests.
Running Go tests in  with command: go test -v -timeout 45m -parallel 128 -p 1 ./... -short
go: finding github.com/gruntwork-io/terratest/modulest/testing latest
go: finding github.com/gruntwork-io/terratest/modulest latest
go: finding github.com/gruntwork-io/gruntwork-cli v0.5.1
go: finding github.com/Azure/azure-sdk-for-go v32.5.0+incompatible
go: finding github.com/Azure/go-autorest/autorest v0.9.1
go: finding github.com/aws/aws-sdk-go v1.23.8
go: finding github.com/go-errors/errors v1.0.2-0.20180813162953-d98b870cc4e0
go: finding github.com/sirupsen/logrus v1.4.2
go: finding github.com/Azure/go-autorest/autorest/azure/auth v0.3.0
go: finding cloud.google.com/go v0.38.0
go: finding github.com/urfave/cli v1.22.1
go: finding github.com/Azure/go-autorest/autorest/to v0.3.0
go: finding github.com/Azure/go-autorest/tracing v0.5.0
go: finding github.com/jstemmer/go-junit-report v0.9.1
go: finding github.com/ghodss/yaml v1.0.0
go: finding github.com/Azure/go-autorest/autorest/date v0.2.0
go: finding github.com/mattn/go-zglob v0.0.2-0.20190814121620-e3c945676326
go: finding github.com/googleapis/gax-go/v2 v2.0.5
go: finding golang.org/x/sys v0.0.0-20191110163157-d32e6e3b99c4
go: finding github.com/Azure/go-autorest/autorest/adal v0.6.0
go: finding github.com/cpuguy83/go-md2man/v2 v2.0.0
go: finding github.com/Azure/go-autorest/autorest/validation v0.2.0
go: finding google.golang.org/genproto v0.0.0-20190819201941-24fa4b261c55
go: finding github.com/Azure/go-autorest/logger v0.1.0
go: finding google.golang.org/api v0.9.1-0.20190821000710-329ecc3c9c34
go: finding google.golang.org/grpc v1.23.0
go: finding github.com/russross/blackfriday/v2 v2.0.1
go: finding github.com/jmespath/go-jmespath v0.0.0-20180206201540-c2b33e8439af
go: finding go.opencensus.io v0.21.0
go: finding golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
go: finding github.com/golang/protobuf v1.3.2
go: finding github.com/Azure/go-autorest/autorest/azure/cli v0.3.0
go: finding github.com/shurcooL/sanitized_anchor_name v1.0.0
go: finding github.com/oracle/oci-go-sdk v7.1.0+incompatible
go: finding golang.org/x/net v0.0.0-20190813141303-74dc4d7220e7
go: finding github.com/dimchansky/utfbom v1.1.0
go: finding golang.org/x/crypto v0.0.0-20190820162420-60c769a6c586
go: finding github.com/hashicorp/golang-lru v0.5.3
go: finding github.com/mitchellh/go-homedir v1.1.0
go: finding github.com/go-sql-driver/mysql v1.4.1
go: finding github.com/google/uuid v0.0.0-20161128191214-064e2069ce9c
go: finding github.com/dgrijalva/jwt-go v3.2.0+incompatible
go: finding k8s.io/api v0.0.0-20181110191121-a33c8200050f
go: finding github.com/gogo/protobuf v1.2.1
go: finding k8s.io/apimachinery v0.0.0-20190704094520-6f131bee5e2c
go: finding k8s.io/client-go v0.0.0-20190704095228-386e588352a4
go: finding github.com/pquerna/otp v1.2.0
go: finding github.com/boombuler/barcode v1.0.1-0.20190219062509-6c824513bacc
go: finding github.com/google/gofuzz v1.0.0
go: finding github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
go: finding github.com/googleapis/gnostic v0.3.1
go: finding gopkg.in/inf.v0 v0.9.1
go: finding github.com/imdario/mergo v0.3.7
go: finding golang.org/x/text v0.3.2
go: finding github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
go: finding github.com/spf13/pflag v1.0.3
go: finding github.com/peterbourgon/diskv v2.0.1+incompatible
go: finding github.com/json-iterator/go v1.1.7
go: finding github.com/modern-go/reflect2 v0.0.0-20180701023420-4b7aa43c6742
go: finding github.com/google/btree v1.0.0
go: finding github.com/docker/spdystream v0.0.0-20181023171402-6480d4af844c
go: finding github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd
go: finding golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
modules/k8s/kubectl.go:11:2: no matching versions for query "latest"
```

fr0stbyte:
@brikis98 sorry for the delay. Updated the PR and most of the tests pass locally, some fail because of missing AWS/Google credentials. Can you please take another look ?

brikis98:
Thx, re-running tests now!

brikis98:
Got some build failures:

```
# github.com/gruntwork-io/terratest/modules/k8s^M
modules/k8s/client.go:13:2: testing redeclared as imported package name^M
        previous declaration at modules/k8s/client.go:4:2^M
FAIL    github.com/gruntwork-io/terratest/modules/helm [build failed]^M
FAIL    github.com/gruntwork-io/terratest/modules/test-structure [build failed]^M
```

fr0stbyte:
@brikis98 updated the PR.
Is there a way for me to trigger a build to verify that things pass properly ?

brikis98:
Just kicked off the tests

brikis98:
Tests failed:

> #!/bin/bash -eo pipefail
> grep -Rw "*testing\.T" modules | grep -v _test.go
> 
> Exited with code exit status 1

.circleci/config.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       - run: pre-commit run --all-files
 
       # verify that the testing.TestingT interface is used instead of testing.T
-      - run: grep -Rw "*testing\.T" modules | grep -v _test.go
+      - run: [[ $(grep -Rw "*testing\.T" modules | grep -v _test.go | wc -l) -eq "0" ]]
```
stephencoe: I hope you don't mind me joining the review. it's great to see the work done enabling Terratest and gingko to work together!

Would it be better to add this check to the pre-commits as a local test? it means the check can happen prior to reaching CI

**Example (untested)**
```
- repo: local
  hooks:
    - id: test-interfaces-used
      name: test-interfaces-used
      entry: bash -c 'grep -Rw "*testing\.T" modules | grep -v _test.go | wc -l'
      language: system
      types: [go]
      pass_filenames: false
```

brikis98: That seems like a good idea!

stephencoe:
Hey @fr0stbyte 👋 Wondered if this is still something you are planning to work on, I am also looking for the same change so happy to help or pick this up if you no longer need it?

fr0stbyte:
@stephencoe thanks for the bump. Yes, I plan on completing it, was just busy at the day job :)

fr0stbyte:
@stephencoe @brikis98 can you please take another look ?

brikis98:
OK, will kick tests off again, shortly!

brikis98:
Some sort of dependency or compile failure?

```
go: finding golang.org/x/sys v0.0.0-20200107162124-548cf772de50
go: finding github.com/docker/spdystream v0.0.0-20181023171402-6480d4af844c
go: finding sigs.k8s.io/yaml v1.1.0
go: finding github.com/json-iterator/go v1.1.8
go: finding github.com/modern-go/reflect2 v1.0.1
go: finding github.com/pquerna/otp v1.2.0
go: finding github.com/boombuler/barcode v1.0.1-0.20190219062509-6c824513bacc
go: finding github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd
go: finding golang.org/x/text v0.3.2
go: finding golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
go: finding k8s.io/utils v0.0.0-20191114184206-e782cd3c129f
# github.com/gruntwork-io/terratest/modules/k8s
/home/circleci/project/modules/k8s/cluster_role_test.go:16:1: wrong signature for TestGetClusterRoleEReturnsErrorForNonExistantClusterRole, must be: func TestGetClusterRoleEReturnsErrorForNonExistantClusterRole(t *testing.T)
FAIL	github.com/gruntwork-io/terratest/modules/k8s [setup failed]
FAIL
```

fr0stbyte:
Not sure, found a typo in a module path so I've fixed that. the tests run locally. 
I used `go test -v -timeout 45m -parallel 128 -p 1 ./... -short` for that.

brikis98:
Still getting the same error:

```
go: finding golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
go: finding k8s.io/utils v0.0.0-20191114184206-e782cd3c129f
# github.com/gruntwork-io/terratest/modules/k8s
/home/circleci/project/modules/k8s/cluster_role_test.go:16:1: wrong signature for TestGetClusterRoleEReturnsErrorForNonExistantClusterRole, must be: func TestGetClusterRoleEReturnsErrorForNonExistantClusterRole(t *testing.T)
FAIL	github.com/gruntwork-io/terratest/modules/k8s [setup failed]
FAIL
```

I just realized these tests have a specific tag:

```
go test -v -timeout 45m -parallel 128 -tags kubernetes ./modules/k8s -short
```

fr0stbyte:
@brikis98 thanks for that tag, found the issue.

brikis98:
Alright, tests passed! Finally merging this in.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.26.0

fr0stbyte:
@brikis98 @stephencoe thanks for merging and for your patience with this !

stephencoe:
Awesome! 👌

lookas1:
@fr0stbyte I tried to look for any examples of tests using ginkgo and terratest and couldn't find anything. Do you have any examples you could share? Are you still using that approach for your IAC testing?

fr0stbyte:
@lookas1 sorry, I don't have an example handy, but did you try following the ginkgo documentation ?

Issue #372, opened 2019-09-27
test_structure failed to load data


njunxuan:
Hi there and thanks for this great tool. 

I was writing some tests on Windows 10 using the test-structure package and my tests kept failing due to the Save*Options functions failing:
1. `save_test_data.go:213: Failed to load value from ..\.test-data\TerraformOptions.json: open ..\.test-data\TerraformOptions.json: The system cannot find the path specified.`
2. `save_test_data.go:213: Failed to load test data from ..\.test-data\sqlYaml.json due to unexpected error: open ..\.test-data\sqlYaml.json: The system cannot find the file specified.`

I have taken a look at the code, and I believe this error is happening as the `IsTestDataPresent` function, which is called by the Save functions only checks for "no such file or directory". However, my system seems to be throwing slightly different errors as shown above. I have patched the function to use `os.Stat` to test whether the path exists instead and things seem to work.

I am wondering whats the best way forward to fix this issue and should I open a PR? Thanks.

EDIT: PR created https://github.com/gruntwork-io/terratest/pull/373

njunxuan:
Fixed by #373

PR #373, opened 2019-09-27
Change IsTestDataPresent to use os.Stat to check path

njunxuan:
The existing logic relies on error string detection which may not be consistent across systems and OSes. Using os.Stat should be more robust when checking for path existence.

https://github.com/gruntwork-io/terratest/issues/372

### DIFF
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -14,6 +14,16 @@ func FileExists(path string) bool {
 	return err == nil
 }
 
+// FileExistsE returns true if the given file exists
+// It will return an error if os.Stat error is not an ErrNotExist
+func FileExistsE(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil && !os.IsNotExist(err) {
+		 return false, err
+	}
+	return err == nil, nil
+}
+
 // CopyTerraformFolderToTemp creates a copy of the given folder and all its contents in a temp folder with a unique name and the given prefix.
 // This is useful when running multiple tests in parallel against the same set of Terraform files to ensure the
 // tests don't overwrite each other's .terraform working directory and terraform.tfstate files. This method returns
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -19,7 +19,7 @@ func FileExists(path string) bool {
 func FileExistsE(path string) (bool, error) {
 	_, err := os.Stat(path)
 	if err != nil && !os.IsNotExist(err) {
-		 return false, err
+		return false, err
 	}
 	return err == nil, nil
 }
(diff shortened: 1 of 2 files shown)


modules/test-structure/save_test_data.go
```diff
@@ -206,10 +205,16 @@ func LoadTestData(t *testing.T, path string, value interface{}) {
 
 // IsTestDataPresent returns true if a file exists at $path and the test data there is non-empty.
 func IsTestDataPresent(t *testing.T, path string) bool {
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			return false
```
yorinasub17: Good catch! Should we use `files.FileExists` instead (https://github.com/gruntwork-io/terratest/blob/master/modules/files/files.go#L11) here? The only difference is that it is not explicitly checking for `os.IsNotExist`, so it can also return `false` for other kinds of errors like permission failures. In most cases that error will surface at a later point in the test, but we should probably add:
```
func FileExistsE(path string) (bool, error) {
	_, err := os.Stat(path)
	if err != nil && !os.IsNotExist(err) {
		return false, err
	}
	return err == nil, nil
}
```

Can you update this PR to include that function and use that here?

njunxuan:
Hello, I have included the new function in the files package and it is now being used in `IsTestDataPresent`. Let me know if I have to make any other changes/squash the commits.

Thanks!

modules/test-structure/save_test_data.go
```diff
@@ -206,10 +205,17 @@ func LoadTestData(t *testing.T, path string, value interface{}) {
 
 // IsTestDataPresent returns true if a file exists at $path and the test data there is non-empty.
 func IsTestDataPresent(t *testing.T, path string) bool {
-	bytes, err := ioutil.ReadFile(path)
-	if err != nil && strings.Contains(err.Error(), "no such file or directory") {
+	exists, err := files.FileExistsE(path)
+	if err != nil {
+		t.Fatalf("Failed to load test data from %s due to unexpected error: %v", path, err)
+	}
+	if exists == false {
```
yorinasub17:
```suggestion
	if !exists {
```

yorinasub17:
Found one minor style nit, but everything else looks good so kicked off the build. Thank for your patience!

modules/files/files.go
```diff
@@ -14,6 +14,16 @@ func FileExists(path string) bool {
 	return err == nil
 }
 
+// FileExistsE returns true if the given file exists
+// It will return an error if os.Stat error is not an ErrNotExist
+func FileExistsE(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil && !os.IsNotExist(err) {
+		 return false, err
```
yorinasub17: Looks like this is indented too far out. Can you run `gofmt` on this file?

njunxuan:
Thanks for taking a look! I've made the necessary style changes.

yorinasub17:
Sorry for the delay in getting back to this! I just kicked off our build and if it passes, will be able to merge.

yorinasub17:
Tests passed, so going to merge and release this!

njunxuan:
Thank you so much for supporting and reviewing the code for this PR 😄

PR #376, opened 2019-09-29
tunnel: Break dependency on k8s.io/kubernetes

tmc:
This removes the dependency on k8s.io/kubernetes by pulling the label helper code in.

### DIFF
--- a/modules/k8s/tunnel.go
+++ b/modules/k8s/tunnel.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


modules/k8s/tunnel.go
```diff
@@ -47,6 +47,14 @@ func (resourceType KubeResourceType) String() string {
 	}
 }
 
+func makeLabels(labels map[string]string) string {
```
yorinasub17: Can you add a comment mentioning what this is doing, or alternatively point to where it is sourced from?

tmc: added a comment.

yorinasub17:
Since build passed and I saw the comment, going to merge this in. Thanks for your contribution!

Issue #89, opened 2018-05-15
Add support for Azure


brikis98:
Terratest has an [aws package](https://github.com/gruntwork-io/terratest/tree/master/modules/aws) with a number of utilities useful for AWS. It would be great to add similar utilities for testing code in Azure.

foreverXZC:
I am a member of Azure group and I am also thinking about this these days. I wonder what kind of features are important and thus really needed.

brikis98:
Hi @foreverXZC! Good question :)

I suspect many of the same utilities we have in the [aws package](https://github.com/gruntwork-io/terratest/tree/master/modules/aws) would be useful, such as functions to make it easier to work with [authentication](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/auth.go), [regions](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/region.go), [machine images](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/ami.go), [logging](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/cloudwatch.go), [user accounts](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/iam.go), and so on.

lewalt000:
Hi folks - thanks for introducing the Terratest tool! Have there been any updates on this issue? Azure support would be a great addition for those working with Azure.

lewalt000:
@brikis98 - thanks for the quick reply! So to use Azure, one would need to write the go code in their tests to talk with Azure, or develop their own helper library, correct?

brikis98:
You'd write a library of helpers similar to the [aws helpers](https://github.com/gruntwork-io/terratest/tree/master/modules/aws), but focused on authenticating to Azure and calling various services you're testing. You'd then use your helpers to test your Azure code.

nmiodice:
I'm contributing back some relevant work. See https://github.com/gruntwork-io/terratest/pull/332

drmmarsunited:
@nmiodice Would love to work with you on this, as I wrote the beginnings of one as well.  I was going to open a PR later tonight! My work is laid out a little differently than yours, so I might open a PR as well just to get the mainters' thoughts on how we could come together on approach.

nmiodice:
@drmmarsunited Absolutely. Looking through your PR I like the approach you took with the authorizer as it is more robust and works with non-service principal use cases. Other than that, these look very compatible.

[...]

PR #377, opened 2019-09-29
Feature/add azure network

MathieuBuisson:
Hi,

Following https://github.com/gruntwork-io/terratest/pull/333 , which has introduced Azure support, the present PR adds coverage for network-related resources in the azure module :

  - Virtual Networks
  - Subnets
  - Network Security Groups
  - Public IPs

This contains only get methods for now.  
If needed, methods to create and delete resources can be added in the future.

I opted to create a separate example module and a separate example test file from the existing one for Azure, because there are quite a few resources involved.  
If you prefer to "merge" this example (and its test) with the existing one for Azure, I'm open to change that.

To view the end of the output of running the `go test` against the example test `terraform_azure_network_example_test.go`, please refer to [this gist](https://gist.github.com/MathieuBuisson/696583a8e79f5bc4436c4872b3c6f240).

Issue link :  https://github.com/gruntwork-io/terratest/issues/89

Thanks.


brikis98:
@MathieuBuisson Thanks for the PR! As Terratest has grown (we just passed 3,000 stars!), we've had to think more about what types of helper functions we want in the library to make sure it doesn't become so big as to be unusable. Please check out the updated [contribution guidelines](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md), especially the part about [infrastructure and validation helpers](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md#infrastructure-and-validation-helpers). These Azure helpers meet requirements for "platforms" and "creating infrastructure," but what about "complexity" (i.e., are these just direct wrappers around the Azure SDK or are you doing something more complicated) and "popularity" (i.e., is this something most Azure/Terratest users will need or is it a relatively esoteric use case)?

faizan82:
> (i.e., is this something most Azure/Terratest users will need or is it a relatively esoteric use case)?

@brikis98 This would be a nice addition for azure. At present, I would need to wrap azure sdk with terratest to test the cases covered here. Would be good if we can have this feature as part of terratest package. 
Also as we add more features for different clouds, I believe this is bound to become big. Perhaps we need to then target terratest for different cloud platforms separately.

brikis98:
Please see https://github.com/gruntwork-io/terratest/issues/89 for the status on Terratest with Azure. In particular, we're [starting some work](https://github.com/gruntwork-io/terratest/issues/89#issuecomment-650052374) around it, and when that work is done, we'll be able to come back to this PR!

yorinasub17:
Thanks for the initial contribution, and apologies we never got around to working with you to get this merged.

As @brikis98 mentioned, we are working out a new contribution standard for Azure specifically and actively working on getting that finalized. In the process, we merged a few PRs that appear to overlap with this in functionality (e.g., https://github.com/gruntwork-io/terratest/pull/631).

We also have a few more PRs from the Microsoft team we plan on prioritizing before officially announcing our new Azure contribution guidelines.

Given that, we will be closing this PR out, but feel free to contribute any missing functions from the package that you identify after the announcement.

Thanks for your patience, and again apologies that we never were able to bring this to merge.

Issue #187, opened 2018-10-26
AWS Feature Request: Support a s3:putobject test


jolexa:
Hello,
I'd like to test that S3 replication is setup correctly. I can use `GetS3ObjectContents` to test the replica destination but I have no easy way to put an object. This looks pretty easy given that the get test already exists, open to this idea before I start work?

josh-padnick:
Sure! For your specific use case, I'm not sure how long it will take for a file written to S3 Bucket A to sync to S3 Bucket B, so you'll probably need to use the retry functions that Terratest has. But, yes, a PR would be most welcome!

mauricioklein:
Hello there!

I submitted the PR above implementing the PutObject functionality.
Hope everything is OK with the implementation. Otherwise, please feel free to drop a comment.

Thanks!

PR #380, opened 2019-10-01
Implement PutObject functionality for AWS S3

mauricioklein:
This PR implements the function `PutS3Object`, which inserts into an existing S3 bucket a given key and body.

Related issue: #187

### DIFF
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -163,6 +163,29 @@ func PutS3BucketVersioningE(t *testing.T, region string, bucketName string) erro
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


yorinasub17:
Hi thanks for your contribution!

While this functions seem reasonable, it appears to be a modification API. In general, the terratest philosophy is to deploy resources using terraform and then verifying that what you deployed is correct. The modules are meant to provide helper functions used for verification purposes, and modifying/creating resources should be optimized for those contexts.

So I want to understand your needs for modifying the security groups, and hearing a concrete test case scenario that you have which these functions would help would be a good way to understand that.

Issue #125, opened 2018-06-27
{Feature Request} Add aws module for s3 tagging check


ahsannaseem:
A module that will check specific set of keys like `companyTeam, companyCost` and such. This will help to make sure that terraform code must have mandatory tags.

brikis98:
Seems like a reasonable addition to the [s3 helpers](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/s3.go). PR welcome!

floresj:
I wouldn't mind taking a stab at this one. I'm thinking of adding the followings functions to `modules/aws/s3.go`.

Functions for retrieving S3 Bucket tags
```go
GetS3BucketTags(t *testing.T, awsRegion string, bucket string) []*s3.Tag
GetS3BucketTagsE(t *testing.T, awsRegion string, bucket string) ([]*s3.Tag, error)
```

Some assertions for checking the existence of keys `tagKeys`. 
```go
AssertS3BucketContainsTagKeys(*testing.T, awsRegion string, bucketName string, tagKeys []string)
AssertS3BucketContainsTagKeysE(*testing.T, awsRegion string, bucketName string, tagKeys []string) error
```

Some assertions for checking that key/values  `tagKeyValuePairs` exist for a particular Bucket. For instance, does a tag existence where `key = 'foo'` and `value = 'bar'`
```go
AssertS3BucketContainsTagKeyValuePairs(*testing.T, awsRegion string, bucketName string, tagKeyValuePairs map[string]string)
AssertS3BucketContainsTagKeyValuePairsE(*testing.T, awsRegion string, bucketName string, tagKeyValuePairs map[string]string) error
```
And this could apply at the object level as well.

```go
GetS3BucketObjectTags(...)
GetS3BucketObjectTagsE(...)
AssertS3BucketObjectContainsTagKeys(...)
// etc...
```

Does this sound reasonable? Just making sure I was understanding the intent behind this issue.

brikis98:
Methods to fetch the tags on a bucket and object makes sense. The `Assert` methods less so, as that should probably be done with a standard Golang assert library in your own test code.

mauricioklein:
@floresj 
Are you working on that?
Please let me know, otherwise, I can support with a PR 👍

floresj:
Hey! @mauricioklein Go for it! I haven't gotten around to it so please feel free to knock that out.

mauricioklein:
PRs created ☝️

james03160927:
[...]

PR #382, opened 2019-10-01
Implement GetBucketTagging functionality for AWS S3

mauricioklein:
This PR adds support for `S3::GetBucketTagging`, returning all the tags attached to a given S3 bucket.

Related issue: #125

### DIFF
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -326,7 +326,10 @@ func GetS3BucketTaggingE(t *testing.T, awsRegion string, bucket string) (map[str
 	)
 
 	for _, tag := range result.TagSet {
-		tags[*(tag.Key)] = *(tag.Value)
+		key := aws.StringValue(tag.Key)
+		value := aws.StringValue(tag.Value)
+
+		tags[key] = value
 	}
 
 	return tags, nil
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -302,6 +302,36 @@ func GetS3BucketPolicyE(t *testing.T, awsRegion string, bucket string) (string, 
 	return aws.StringValue(res.Policy), nil
 }
 
+// GetS3BucketTagging returns the tags associated to a given S3 bucket
+func GetS3BucketTagging(t *testing.T, awsRegion string, bucket string) map[string]string {
+	tags, err := GetS3BucketTaggingE(t, awsRegion, bucket)
+	require.NoError(t, err)
+
+	return tags
+}
+
+// GetS3BucketTaggingE returns the tags associated to a given S3 bucket
+func GetS3BucketTaggingE(t *testing.T, awsRegion string, bucket string) (map[string]string, error) {
+	tags := make(map[string]string)
+
+	s3Client, err := NewS3ClientE(t, awsRegion)
+	if err != nil {
+		return tags, err
+	}
+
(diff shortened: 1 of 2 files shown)


modules/aws/s3.go
```diff
@@ -302,6 +302,36 @@ func GetS3BucketPolicyE(t *testing.T, awsRegion string, bucket string) (string,
...
+	}
+
+	result, err := s3Client.GetBucketTagging(
+		&s3.GetBucketTaggingInput{
+			Bucket: aws.String(bucket),
+		},
+	)
+
+	for _, tag := range result.TagSet {
+		tags[*(tag.Key)] = *(tag.Value)
```
brikis98:
```suggestion
		tags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
```

modules/aws/s3_test.go
```diff
@@ -29,6 +30,37 @@ func TestCreateAndDestroyS3Bucket(t *testing.T) {
...
+		"keyB": "valueB",
+	}
+
+	addTagsToBucket(t, s3Client, region, s3BucketName, tags)
+
+	bucketTags := GetS3BucketTagging(t, region, s3BucketName)
+
+	if !reflect.DeepEqual(tags, bucketTags) {
+		t.Fatalf("Expect bucket to have tags '%+v', but got '%+v'", tags, bucketTags)
+	}
```
brikis98: Would `assert.Equal` work here?

modules/aws/s3_test.go
```diff
@@ -29,6 +30,37 @@ func TestCreateAndDestroyS3Bucket(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tags := map[string]string{
+		"keyA": "valueA",
+		"keyB": "valueB",
+	}
+
+	addTagsToBucket(t, s3Client, region, s3BucketName, tags)
```
brikis98: Is this an async operation? Could checking tags immediately after adding them fail due to eventual consistency?

modules/aws/s3_test.go
mauricioklein: Indeed it does.
Updated!

modules/aws/s3_test.go
mauricioklein: Looks like it's a sync operation since the response is only returned when the operation is fulfilled.
AWS Go SDK has another function called `s3.PutBucketTaggingRequest`, and this one yes, is async, and requires polling to check the operation status

brikis98:
@mauricioklein Apologies for the delay. As Terratest has grown (we just passed 3,000 stars!), we've had to think more about what types of helper functions we want in the library to make sure it doesn't become so big as to be unusable. Please check out the updated [contribution guidelines](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md), especially the part about [infrastructure and validation helpers](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md#infrastructure-and-validation-helpers). These S3 helpers meet requirements for "platforms" and "creating infrastructure," but what about "complexity" (i.e., are these just direct wrappers around the AWS SDK or are you doing something more complicated) and "popularity" (i.e., is this something most AWS/Terratest users will need or is it a relatively esoteric use case)?

PR #385, opened 2019-10-02
Implement GetObjectTagging functionality for AWS S3

mauricioklein:
This PR implements the function GetS3ObjectTagging, which returns the tags associated to an object in S3.

Related issue: #125

### DIFF
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -327,7 +327,10 @@ func GetS3ObjectTaggingE(t *testing.T, awsRegion string, bucket string, key stri
 	)
 
 	for _, tag := range result.TagSet {
-		tags[*(tag.Key)] = *(tag.Value)
+		key := aws.StringValue(tag.Key)
+		value := aws.StringValue(tag.Value)
+
+		tags[key] = value
 	}
 
 	return tags, nil
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -302,6 +302,37 @@ func GetS3BucketPolicyE(t *testing.T, awsRegion string, bucket string) (string, 
 	return aws.StringValue(res.Policy), nil
 }
 
+// GetS3ObjectTagging returns the tags associated to an object stored in a S3 bucket
+func GetS3ObjectTagging(t *testing.T, awsRegion string, bucket string, key string) map[string]string {
+	tags, err := GetS3ObjectTaggingE(t, awsRegion, bucket, key)
+	require.NoError(t, err)
+
+	return tags
+}
+
+// GetS3ObjectTaggingE returns the tags associated to an object stored in a S3 bucket
+func GetS3ObjectTaggingE(t *testing.T, awsRegion string, bucket string, key string) (map[string]string, error) {
+	tags := make(map[string]string)
+
+	s3Client, err := NewS3ClientE(t, awsRegion)
+	if err != nil {
+		return tags, err
+	}
+
+	result, err := s3Client.GetObjectTagging(
+		&s3.GetObjectTaggingInput{
+			Bucket: aws.String(bucket),
+			Key:    aws.String(key),
+		},
+	)
+
+	for _, tag := range result.TagSet {
+		tags[*(tag.Key)] = *(tag.Value)
+	}
+
+	return tags, nil
+}
+
 // AssertS3BucketExists checks if the given S3 bucket exists in the given region and fail the test if it does not.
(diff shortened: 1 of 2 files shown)


modules/aws/s3.go
```diff
@@ -302,6 +302,40 @@ func GetS3BucketPolicyE(t *testing.T, awsRegion string, bucket string) (string,
 	return aws.StringValue(res.Policy), nil
 }
 
+// GetS3ObjectTagging returns the tags associated to an object stored in a S3 bucket
```
yorinasub17:
```suggestion
// GetS3ObjectTagging returns the tags associated to an object stored in a S3 bucket. This will fail the test if there is an error.
```

modules/aws/s3_test.go
```diff
@@ -29,6 +30,38 @@ func TestCreateAndDestroyS3Bucket(t *testing.T) {
...
+	t.Parallel()
+
+	region := GetRandomStableRegion(t, nil, nil)
+	id := random.UniqueId()
+	logger.Logf(t, "Random values selected. Region = %s, Id = %s\n", region, id)
+
+	s3BucketName := "gruntwork-terratest-" + strings.ToLower(id)
+
+	CreateS3Bucket(t, region, s3BucketName)
+	defer DeleteS3Bucket(t, region, s3BucketName)
```
yorinasub17: defer the delete before the create, so that if there is an error in create that happens after the bucket is created, we still attempt to delete it.

modules/aws/s3_test.go
```diff
@@ -29,6 +30,38 @@ func TestCreateAndDestroyS3Bucket(t *testing.T) {
...
+
+	s3Client, err := NewS3ClientE(t, region)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yorinasub17: We've been using the pattern `require.NoError(t, err)` instead of the if check.

modules/aws/s3_test.go
```diff
@@ -29,6 +30,38 @@ func TestCreateAndDestroyS3Bucket(t *testing.T) {
...
+
+	key := fmt.Sprintf("test-%s", id)
+	tags := map[string]string{
+		"keyA": "valueA",
+		"keyB": "valueB",
+	}
+
+	addObjectToBucket(t, s3Client, region, s3BucketName, key, "The body")
+	addTagsToObject(t, s3Client, region, s3BucketName, key, tags)
+	defer EmptyS3Bucket(t, region, s3BucketName)
```
yorinasub17: Similar to comment above, defer the emptying before adding.

modules/aws/s3_test.go
```diff
@@ -212,3 +245,47 @@ func testEmptyBucket(t *testing.T, s3Client *s3.S3, region string, s3BucketName
...
+		Bucket: aws.String(s3BucketName),
+		Key:    aws.String(key),
+		Body:   aws.ReadSeekCloser(strings.NewReader(body)),
+	}
+
+	_, err := s3Client.PutObject(config)
+
+	if err != nil {
+		t.Fatalf("Failed to add key '%s' to bucket '%s': %s", key, s3BucketName, err.Error())
+	}
```
yorinasub17: `require.NoError(t, err)`

modules/aws/s3_test.go
```diff
@@ -212,3 +245,47 @@ func testEmptyBucket(t *testing.T, s3Client *s3.S3, region string, s3BucketName
...
+}
+
+func addTagsToObject(t *testing.T, s3Client *s3.S3, region string, s3BucketName string, key string, tags map[string]string) {
+	logger.Logf(t, "Adding the following tags to object '%s': %+v", key, tags)
+
+	// Convert custom tags in AWS tags
```
yorinasub17: I think you can remove this comment, as it is adding little value to the context here.

modules/aws/s3_test.go
```diff
@@ -212,3 +245,47 @@ func testEmptyBucket(t *testing.T, s3Client *s3.S3, region string, s3BucketName
...
+		Tagging: &s3.Tagging{
+			TagSet: awsTags,
+		},
+	}
+
+	_, err := s3Client.PutObjectTagging(config)
+
+	if err != nil {
+		t.Fatalf("Failed to add tags to object '%s': %s", key, err.Error())
+	}
```
yorinasub17: `require.NoError(t, err)`

Issue #354, opened 2019-08-27
[Feature]: Allow ability to specify Namespace in NewKubectlOptions


BrendanThompson:
If you look at the `struct` definition for `KubectlOptions` it has attributes for the `Namespace`, this is not implemented by the `NewKubectlOptions()` func ( _As can be seen below_ ). Is the premise here to make it the responsibility of the caller to create an instance of the struct without using the aforementioned func when they require explicit declaration of the `Namespace` attribute?

https://github.com/gruntwork-io/terratest/blob/f8167b3905e7b41b2cd5f20a8bd7b2696e7f93dd/modules/k8s/kubectl_options.go#L13

An option would be to extend the current implementation of the `NewKubectlOptions()` func to be something like:
```go
func NewKubectlOptions(contextName string, configPath string, namespace string, env map[string]string) *KubectlOptions {
	return &KubectlOptions{
		ContextName: contextName,
		ConfigPath:  configPath,
		Namespace: namespace,
		Env: env,
	}
}
```
This includes both `Namespace` and `Env`, the only issue here is the fact that the caller would then always need to include these values. My `go` knowledge isn't that great. The other alternative would be to remove `Env` and only allow `Namespace`, further as `Namespace` is a string it could also be defaulted in the struct.

yorinasub17:
This was to be backwards compatible with previous releases so that you didn't have to update all instances of `NewKubectlOptions`, which would be quite a lot. Originally many tests we were writing were done against the default Namespace so it was rare for us to specify a specific Namespace.

However, over time as we wrote more tests, it became clear that it would be nice to have the Namespace unique per test so nowadays almost all our tests have an immediate line to update the Namespace. For example:

```
options := NewKubectlOptions("", "")
options.Namespace = uniqueNamespace()
```

Given this pattern is now recommended, it probably makes sense to update `NewKubectlOptions` to accept the namespace as a param.

If you are curious about the design: in general, we don't want to take in all the params of the struct for the constructor because then it wouldn't be any different than manually constructing it using the struct call with `{}`. We usually only include the options that are almost always set, while excluding those that are only set occasionally and have them set to a sane default in the constructor function.

BrendanThompson:
@yorinasub17 that makes sense! Definitely agree though the standard with Kubernetes now is certainly around using `namespaces`.

In terms of your design, that absolutely makes sense! To be honest I didn't even think about just setting the attribute like you've demonstrated above.

j4ng5y:
[...]

PR #389, opened 2019-10-07
#354: Add namespace to the NewKubectlOptions constructor

j4ng5y:
Update the NewKubectlOptions constructor to require the namespace struct attribute as per #354

### DIFF
--- a/modules/k8s/config.go
+++ b/modules/k8s/config.go
@@ -28,7 +28,7 @@ func LoadConfigFromPath(path string) clientcmd.ClientConfig {
 	return config
 }
 
-// LoadApiClientConfig will load a ClientConfig object from a file path that points to a location on disk containing a
+// LoadApiClientConfigE will load a ClientConfig object from a file path that points to a location on disk containing a
 // kubectl config, with the requested context loaded.
 func LoadApiClientConfigE(configPath string, contextName string) (*restclient.Config, error) {
 	overrides := clientcmd.ConfigOverrides{}
@@ -161,7 +161,7 @@ func CopyHomeKubeConfigToTemp(t *testing.T) string {
 	return path
 }
 
-// CopyHomeKubeConfigToTemp will copy the kubeconfig in the home directory to a temp file.
+// CopyHomeKubeConfigToTempE will copy the kubeconfig in the home directory to a temp file.
 func CopyHomeKubeConfigToTempE(t *testing.T) (string, error) {
 	configPath, err := KubeConfigPathFromHomeDirE()
 	if err != nil {
--- a/modules/k8s/errors.go
+++ b/modules/k8s/errors.go
@@ -13,6 +13,7 @@ type IngressNotAvailable struct {
 	ingress *extensionsv1beta1.Ingress
 }
 
+// Error is a simple function to return a formatted error message as a string
 func (err IngressNotAvailable) Error() string {
 	return fmt.Sprintf("Ingress %s is not available", err.ingress.Name)
 }
@@ -33,6 +34,7 @@ type DesiredNumberOfPodsNotCreated struct {
 	DesiredCount int
 }
 
+// Error is a simple function to return a formatted error message as a string
 func (err DesiredNumberOfPodsNotCreated) Error() string {
 	return fmt.Sprintf("Desired number of pods (%d) matching filter %v not yet created", err.DesiredCount, err.Filter)
 }
@@ -42,6 +44,7 @@ type ServiceAccountTokenNotAvailable struct {
 	Name string
 }
 
+// Error is a simple function to return a formatted error message as a string
 func (err ServiceAccountTokenNotAvailable) Error() string {
 	return fmt.Sprintf("ServiceAccount %s does not have a token yet.", err.Name)
 }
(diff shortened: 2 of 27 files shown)


j4ng5y:
So, in order to maintain backwards compatibility, I changed the NewKubectlOptions function to accept a variable length of strings and if we see that the number is 3 now, rather than original 2 case, then we set the namespace as well.

modules/k8s/kubectl_options.go
```diff
@@ -2,18 +2,28 @@ package k8s
...
-		ContextName: contextName,
-		ConfigPath:  configPath,
+// NewKubectlOptions will return a pointer to new instance of KubectlOptions with the configured options
+func NewKubectlOptions(opts ...string) *KubectlOptions {
+	switch {
+	case len(opts) = 3:
+		return &KubectlOptions{
+			ContextName: opts[0],
+			ConfigPath:  opts[1],
+			Namespace:   opts[2],
```
yorinasub17: While you are in here, can you also initialize the map for `Env`? Basically add `Env: map[string]string{}` to the struct creation. The reason is that by default it initializes to `nil`, and you can't start adding to `Env`, so by initializing, it makes the experience better.

j4ng5y: Will do

j4ng5y: added this in 12eaf30

yorinasub17: Thanks!

yorinasub17:
Thanks! Kicking off the build now.

j4ng5y:
The helm stuff is probably going to fail, I forgot to update those. I'm adding the fixes now.

yorinasub17:
Ah it looks like it also failed on `goimports` in `modules/k8s/kubectl_options.go`, probably due to the misalignment in the struct initialization for `Env`. Can you run `gofmt` (or `goimports`) on that file?

yorinasub17:
Thanks! Kicked off another build.

yorinasub17:
Got a few more build failures resulting from:
```
# github.com/gruntwork-io/terratest/modules/k8s
../k8s/client.go:22:30: not enough arguments in call to NewKubectlOptions
        have (string, string)
        want (string, string, string)
```

j4ng5y:
Ok, in places that don't use the options, I'm going to use the "default" namespace, which seems to happen all over the place so that will take me just a minute

yorinasub17:
Ok kicking off another test build!

j4ng5y:
I'm having a fairly hard time figuring out why that one build ... looked like an AMI thing

yorinasub17:
The AMI failure is flaky and happens from time to time due to the number of packer builds we run in parallel. However, the `setup-minikube` failure is new to me and it's weird because I know this was working this morning in a different repo. I am investigating that now...

yorinasub17:
Ah it looks like this branch is quite a bit behind master and does not include the fixes for minikube (https://github.com/gruntwork-io/terratest/commit/cf0c4c521bc299d6814455dfe0a81d056ae39aa5). Can you rebase on master?

j4ng5y:
re-basing now

yorinasub17:
Thanks! Kicking off a new build.

j4ng5y:
S3 bucket failure this time

yorinasub17:
Looks like there are also a few more build failures that were not caught before, due to the build tags:

```
modules/helm/install_test.go:31:41: not enough arguments in call to k8s.NewKubectlOptions
	have (string, string)
	want (string, string, string)
modules/helm/upgrade_test.go:32:41: not enough arguments in call to k8s.NewKubectlOptions
	have (string, string)
	want (string, string, string)
modules/k8s/service_account_test.go:45:44: not enough arguments in call to NewKubectlOptions
	have (string, string)
	want (string, string, string)
```

j4ng5y:
Cool, i'll clean those up this morning.

modules/k8s/config.go
```diff
@@ -161,7 +161,7 @@ func CopyHomeKubeConfigToTemp(t *testing.T) string {
 	return path
 }
 
-// CopyHomeKubeConfigToTemp will copy the kubeconfig in the home directory to a temp file.
+// CopyHomeKubeConfigToTempE will copy the kubeconfig in the home directory to a temp file.
```
yorinasub17: Good catch! Thanks!

modules/k8s/errors.go
```diff
@@ -51,10 +54,12 @@ type PodNotAvailable struct {
 	pod *corev1.Pod
 }
 
+// Error is a simple function to return a formatted error message as a string
 func (err PodNotAvailable) Error() string {
 	return fmt.Sprintf("Pod %s is not available", err.pod.Name)
 }
 
+// NewPodNotAvailableError returnes a PodNodAvailable struct when Kubernetes deems a pod is not available
```
yorinasub17:
```suggestion
// NewPodNotAvailableError returnes a PodNotAvailable struct when Kubernetes deems a pod is not available
```

j4ng5y: Spelling is hard :smile:

j4ng5y:
I fixed the go-imports issue if you wouldn't mind building again

j4ng5y:
I miss another call to NewKubectlOptions, but I fixed that. Also, a few S3 tests failed. Not sure whats up with that :smile:

j4ng5y:
Horray! Everything built :)

yorinasub17:
Ok I just reviewed the change set again and this looks good to me! I am going to go ahead and merge this in. Thanks for your patience and responsiveness working with me through the issues!

yorinasub17:
https://github.com/gruntwork-io/terratest/releases/tag/v0.20.0

PR #391, opened 2019-10-08
338 wrong log message for gcp zonal instance groups

j4ng5y:


j4ng5y:
I now realized that I need to clean up my local repo 😄

PR #392, opened 2019-10-09
add daemonset support

ajayk:
* Adds support for daemonsets 
* allows to query for daemonsets 

`     daemonset := k8s.GetDaemonSet(t, options, "ebs-csi-node")  `
   `     require.Equal(t, daemonset.Name, "ebs-csi-node")`

### DIFF
--- a/modules/k8s/daemonset_test.go
+++ b/modules/k8s/daemonset_test.go
@@ -1,0 +1,103 @@
+package k8s
+
+import (
+	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
+	"strings"
+	"testing"
+)
+
+func TestGetDaemonSetEReturnsErrorForNonExistantDaemonSet(t *testing.T) {
+	t.Parallel()
+
+	options := NewKubectlOptions("", "")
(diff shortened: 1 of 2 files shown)


modules/k8s/daemonset_test.go
```diff
@@ -0,0 +1,103 @@
+package k8s
```
yorinasub17: Can you add the relevant build tags like we have in the other test code in this module? Otherwise these tests will fail in our CI environment.

modules/k8s/daemonset_test.go
```diff
@@ -0,0 +1,105 @@
...
+    metadata:
+      labels:
+        name: fluentd-elasticsearch
+    spec:
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      containers:
+      - name: fluentd-elasticsearch
+        image: gcr.io/fluentd-elasticsearch/fluentd:v2.5.1
```
yorinasub17: Can you use a different, simpler image that doesn't have external dependencies for the unit test? That way, it won't fail on boot up and the config can be simplified (e.g we don't need all the volumes).

E.g a simple alpine container with `sleep 999999`.

ajayk:
Thanks @yorinasub17 for the review ,  addressed the review comments

yorinasub17:
Build passed, so going to merge this. Thanks for the contribution!

PR #393, opened 2019-10-10
Add CleanupTestDataFolder for deleting .test-data

### LINKED ISSUE #383: How to use CleanupTestData to cleanup
Hi,

I am using `test_structure` and the `Save*Option` functions to manage my terraform and k8 options. At the end of the test, I want to delete the `.test-data` directory. I have tried something along the lines of:
`test_structure.CleanupTestData(workingDir)`, where `workingDir` is the path I passed to the `Save` functions.

However, I get errors about how the directory is not empty, which seems odd to me as I thought the purpose of the function is to clean up the directory. How should I go about using the CleanupTestData function?

Thanks.

njunxuan:
This function allows the user to easily cleanup the `.test-data` directory created by the `Save` functions. It optionally allows the user to delete the `workingDir` folder itself. I am still very new to Go, so let me know if there are any improvements required. Thanks!

Closes #383

### DIFF
--- a/modules/test-structure/save_test_data.go
+++ b/modules/test-structure/save_test_data.go
@@ -282,3 +282,30 @@ func CleanupTestData(t *testing.T, path string) {
 		logger.Logf(t, "%s does not exist. Nothing to cleanup.", path)
 	}
 }
+
+// CleanupTestData cleans up the .test-data folder inside the given folder.
+// If deleteParent is true, CleanupTestDataFolder will also delete the given folder.
+func CleanupTestDataFolder(t *testing.T, path string, deleteParent bool) {
+	if deleteParent {
+		deleteFolder(t, path)
+	} else {
+		testDataPath := filepath.Join(path, ".test-data")
+		deleteFolder(t, testDataPath)
+	}
+}
+
+// cleanupTestDataFolder deletes the given folder.
+func deleteFolder(t *testing.T, path string) {
+	exists, err := files.FileExistsE(path)
+	if err != nil {
+		t.Fatalf("Failed to clean up test data folder at %s: %v", path, err)
+	}
+
+	if exists {
+		if err := os.RemoveAll(path); err != nil {
+			t.Fatalf("Failed to clean up test data folder at %s: %v", path, err)
+		}
+	} else {
+		logger.Logf(t, "%s does not exist. Nothing to cleanup", path)
+	}
+}
--- a/modules/test-structure/save_test_data.go
+++ b/modules/test-structure/save_test_data.go
@@ -311,7 +311,7 @@ func cleanupTestDataFolder(t *testing.T, path string) error {
 	}
 
 	if !exists {
-		logger.Logf(t, "%s does not exist. Nothing to cleanup", path)
+		logger.Logf(t, "%s does not exist. Nothing to cleanup.", path)
 		return nil
 	}
 	if err := os.RemoveAll(path); err != nil {
--- a/modules/test-structure/save_test_data.go
+++ b/modules/test-structure/save_test_data.go
@@ -283,29 +283,27 @@ func CleanupTestData(t *testing.T, path string) {
 	}
 }
 
-// CleanupTestData cleans up the .test-data folder inside the given folder.
+// CleanupTestDataFolder cleans up the .test-data folder inside the given folder.
 // If deleteParent is true, CleanupTestDataFolder will also delete the given folder.
 func CleanupTestDataFolder(t *testing.T, path string, deleteParent bool) {
-	if deleteParent {
-		deleteFolder(t, path)
-	} else {
-		testDataPath := filepath.Join(path, ".test-data")
-		deleteFolder(t, testDataPath)
+	if !deleteParent {
+		path = filepath.Join(path, ".test-data")
 	}
+	cleanupTestDataFolder(t, path)
 }
 
 // cleanupTestDataFolder deletes the given folder.
-func deleteFolder(t *testing.T, path string) {
+func cleanupTestDataFolder(t *testing.T, path string) {
 	exists, err := files.FileExistsE(path)
 	if err != nil {
 		t.Fatalf("Failed to clean up test data folder at %s: %v", path, err)
 	}
 
-	if exists {
-		if err := os.RemoveAll(path); err != nil {
-			t.Fatalf("Failed to clean up test data folder at %s: %v", path, err)
-		}
-	} else {
+	if !exists {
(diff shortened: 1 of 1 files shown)


modules/test-structure/save_test_data.go
```diff
@@ -282,3 +282,30 @@ func CleanupTestData(t *testing.T, path string) {
 		logger.Logf(t, "%s does not exist. Nothing to cleanup.", path)
 	}
 }
+
+// CleanupTestData cleans up the .test-data folder inside the given folder.
```
bwhaley: The first line of a comment preceding a function should generally begin with exactly the function name. Looks like that was the intent here but perhaps the function name changed? In any case, I'd start with `CleanupTestDataFolder cleans up the...`.

modules/test-structure/save_test_data.go
```diff
@@ -282,3 +282,30 @@ func CleanupTestData(t *testing.T, path string) {
...
+func CleanupTestDataFolder(t *testing.T, path string, deleteParent bool) {
+	if deleteParent {
+		deleteFolder(t, path)
+	} else {
+		testDataPath := filepath.Join(path, ".test-data")
+		deleteFolder(t, testDataPath)
+	}
+}
+
+// cleanupTestDataFolder deletes the given folder.
```
bwhaley: Same comment as above applies here.

modules/test-structure/save_test_data.go
```diff
@@ -282,3 +282,30 @@ func CleanupTestData(t *testing.T, path string) {
 		logger.Logf(t, "%s does not exist. Nothing to cleanup.", path)
 	}
 }
+
+// CleanupTestData cleans up the .test-data folder inside the given folder.
+// If deleteParent is true, CleanupTestDataFolder will also delete the given folder.
+func CleanupTestDataFolder(t *testing.T, path string, deleteParent bool) {
+	if deleteParent {
```
bwhaley: This logic could be simplified to:
```
if !deleteParent {
    path = filepath.Join(path, ".test-data")
}
deleteFolder(t, path)
```

modules/test-structure/save_test_data.go
```diff
@@ -282,3 +282,30 @@ func CleanupTestData(t *testing.T, path string) {
...
+}
+
+// cleanupTestDataFolder deletes the given folder.
+func deleteFolder(t *testing.T, path string) {
+	exists, err := files.FileExistsE(path)
+	if err != nil {
+		t.Fatalf("Failed to clean up test data folder at %s: %v", path, err)
+	}
+
+	if exists {
```
bwhaley: Do you think the following is more or less readable?

```
if !exists {
    logger.Logf(t, "%s does not exist. Nothing to cleanup", path)
    return
}
if err := os.RemoveAll(path); err != nil {
    t.Fatalf("Failed to clean up test data folder at %s: %v", path, err)
}
```

or even

```
switch exists {
case true:
    if err := os.RemoveAll(path); err != nil {
        t.Fatalf("Failed to clean up test data folder at %s: %v", path, err)
    }
case false:
    logger.Logf(t, "%s does not exist. Nothing to cleanup", path)
}
```

Just trying to show some alternatives. Any of them would work. IMHO, the first example shown above is the cleanest/most idiomatic.

njunxuan:
Thanks for taking a look at the PR! The comment naming was definitely an oversight on my part after performing some refactoring. Your suggested edits do make the code more readable 😄

I have made the changes locally and will leave some time for other reviewers to provide any feedback before pushing.

modules/test-structure/save_test_data.go
```diff
@@ -282,3 +282,28 @@ func CleanupTestData(t *testing.T, path string) {
 		logger.Logf(t, "%s does not exist. Nothing to cleanup.", path)
 	}
 }
+
+// CleanupTestDataFolder cleans up the .test-data folder inside the given folder.
+// If deleteParent is true, CleanupTestDataFolder will also delete the given folder.
+func CleanupTestDataFolder(t *testing.T, path string, deleteParent bool) {
```
brikis98: Hm, `deleteParent` seems like a strange, and potentially dangerous, thing to toss into the API. What's the use case for that?

modules/test-structure/save_test_data.go
```diff
@@ -282,3 +282,28 @@ func CleanupTestData(t *testing.T, path string) {
...
+	if !deleteParent {
+		path = filepath.Join(path, ".test-data")
+	}
+	cleanupTestDataFolder(t, path)
+}
+
+// cleanupTestDataFolder deletes the given folder.
+func cleanupTestDataFolder(t *testing.T, path string) {
```
brikis98: Please check out these guidelines about how we handle errors in Terratest: https://github.com/gruntwork-io/terratest/#error-handling

modules/test-structure/save_test_data.go
njunxuan: Currently, I am using the `CopyTerraformFolderToTemp` and setting the output path as my working directory for all the `Save*` functions. I do this to ensure that all my parallel tests init with their own config. After the tests, I would like to clean up the multiple copies (using the `deleteParent` flag).

There are other situations where I do not use `CopyTerraformFolderToTemp` and thus I would only want to delete the `.test-data` (using `false` for `deleteParent`) folder inside the working directory that I passed to the `Save*` functions.

Would be happy to hear if there are better ways to achieve my use case. Thanks! :smile:

modules/test-structure/save_test_data.go
njunxuan: My bad, I will update the code accordingly to follow the error handling conventions.

modules/test-structure/save_test_data.go
brikis98: I think this is mixing two use cases into one function, which is probably not the right approach. `CleanupTestDataXXX` should only deal with the test data stuff in `.test-data`. Cleaning up `CopyTerraformFolderToTemp` contents should be done via a separate function dedicated to that use case. Mixing the two together is confusing and, given that we are talking about deleting whole folders of content, it's too risky to do anything that isn't crystal clear and obvious.

njunxuan: Sure that works too. I also realised that with the `deleteParent` flag, the function is more or less just a `os.RemoveAll` call. I will make the necessary changes.Thanks!

brikis98:
Tests passed. Merging now.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.20.1

njunxuan:
A big thanks to both reviewers!

PR #395, opened 2019-10-11
Fix build failure from merging conflicts

yorinasub17:
When we merged https://github.com/gruntwork-io/terratest/pull/392 and https://github.com/gruntwork-io/terratest/pull/389, one did not integrate with the other causing the build to fail. This fixes that.


yorinasub17:
Saw build pass, so going to merge this in. Thanks for review!

PR #397, opened 2019-10-18
Support specifying the chart namespace in RenderTemplate

mikegrass:
Currently, the `RenderTemplate` family of functions don't support specifying the release namespace for a chart:

```console
$ helm template --help | grep namespace
      --namespace string         namespace to install the release into
```

For charts that use the [`.Release.Namespace`](https://helm.sh/docs/chart_template_guide/#built-in-objects) object in chart templates, I'd like to be able to confirm the correct behavior via `RenderTemplate`.

`Install`/`InstallE` *do* support overriding the namespace parameter via `options.KubectlOptions`:

https://github.com/gruntwork-io/terratest/blob/3bd1244cbcb33e46f51549f97b028b530f752543/modules/helm/install.go#L35-L37

Permit this behavior for `RenderTemplate`/`RenderTemplateE` as well, using the same source for the namespace name as `Install` / `InstallE`.

### DIFF
--- a/test/helm_basic_example_template_test.go
+++ b/test/helm_basic_example_template_test.go
@@ -10,13 +10,17 @@
 package test
 
 import (
+	"fmt"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 
 	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/random"
 )
 
 // This file contains examples of how to use terratest to test helm chart template logic by rendering the templates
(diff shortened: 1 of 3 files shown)


test/helm_basic_example_template_test.go
```diff
@@ -35,8 +39,11 @@ func TestHelmBasicExampleTemplateRenderedDeployment(t *testing.T) {
 	helmChartPath, err := filepath.Abs("../examples/helm-basic-example")
 	require.NoError(t, err)
 
-	// Since we aren't deploying any resources, there is no need to setup kubectl authentication, helm home, or
-	// namespaces
+	// Since we aren't deploying any resources, there is no need to setup kubectl authentication or helm home.
+
+	// Set up the namespace; confirm that the template renders the expected value for the namespace.
+	namespaceName := "medieval-" + strings.ToLower(random.UniqueId())
+	fmt.Printf("Namespace: %s\n", namespaceName)
```
yorinasub17: Debug statements should be using `logger.Logf`. Otherwise, it won't be captured by the `terratest_log_parser`.

mikegrass: Thanks @yorinasub17! I've corrected this in https://github.com/gruntwork-io/terratest/pull/397/commits/e55d8ce7741626d58db1bf8f2407fe11ab180707. Mind taking another look?

yorinasub17:
Thanks for making that change! Just kicked off our build.

PR #399, opened 2019-10-25
k8s: Add Get funcs for RoleBinding and ClusterRoleBinding

kirecek:


yorinasub17:
Thanks for your contribution! However, looking at the code, these functions appear to be simple wrappers around the client-go library.

We used to accept these as is, but as Terratest has grown (we just passed 3,000 stars!), we've had to think more about what types of helper functions we want in the library to make sure it doesn't become so big as to be unusable. Please check out [the updated contribution guidelines](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md#types-of-contributions), especially the part about infrastructure and validation helpers.

These RBAC helpers meet requirements for "platforms," "creating infrastructure," and "popularity," but do not meet the "complexity" requirement.

With that said, I would like to understand your motivations for these functions. Is there something about the k8s client-go library that make it hard to use the RBAC functions directly?

kirecek:
Oh, yes k8s client is totally okay to use. I see your point.  Thank You :) Closing the MR

yorinasub17:
Thanks for understanding!

PR #405, opened 2019-11-09
helm: add ReleaseName to options and pass it to template rendering

### LINKED ISSUE #402: helm/testing: no release name definition in template rendering
There is no possibility to define release name when you render a template.

StupidScience:
Fixes #402

Atm I just pass new option to template rendering.

Should i rework install, delete, etc from positional arg to options as well?
It will be kinda breaking change.

### DIFF
--- a/modules/helm/template.go
+++ b/modules/helm/template.go
@@ -15,15 +15,15 @@ import (
 // RenderTemplate runs `helm template` to render the template given the provided options and returns stdout/stderr from
 // the template command. If you pass in templateFiles, this will only render those templates. This function will fail
 // the test if there is an error rendering the template.
-func RenderTemplate(t *testing.T, options *Options, chartDir string, templateFiles []string) string {
-	out, err := RenderTemplateE(t, options, chartDir, templateFiles)
+func RenderTemplate(t *testing.T, options *Options, chartDir string, releaseName string, templateFiles []string) string {
+	out, err := RenderTemplateE(t, options, chartDir, releaseName, templateFiles)
 	require.NoError(t, err)
 	return out
 }
 
 // RenderTemplateE runs `helm template` to render the template given the provided options and returns stdout/stderr from
 // the template command. If you pass in templateFiles, this will only render those templates.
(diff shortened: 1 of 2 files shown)


yorinasub17:
Hmm after thinking this through some more and looking at the implementation, the variability in how the different commands handle the release name makes it a bit annoying to implement it this way (e.g release name is a positional argument in `delete`).

You might be right that it might make more sense to just add a positional arg for `releaseName` to template rendering and just enhance that function. What do you think?

StupidScience:
As for me Options looks as more appropriate place for release name but you're right that it is passed in different ways for different helm commands. Also there will be less changes if it will be passed to render as positional arg. So I think reworked RenderTemplate functions will be more correct.

yorinasub17:
This LGTM, but you will need to update the tests. If you grep for `RenderTemplate`, you will see A few places (only one test file) where it is used:
```
%~> rg RenderTemplate
test/helm_basic_example_template_test.go
59:     // Run RenderTemplate to render the template and capture the output. Note that we use the version without `E`, since
63:     output := helm.RenderTemplate(t, options, helmChartPath, []string{"templates/deployment.yaml"})
125:                    _, err := helm.RenderTemplateE(t, options, helmChartPath, []string{})

modules/helm/template.go
15:// RenderTemplate runs `helm template` to render the template given the provided options and returns stdout/stderr from
18:func RenderTemplate(t *testing.T, options *Options, chartDir string, templateFiles []string) string {
19:     out, err := RenderTemplateE(t, options, chartDir, templateFiles)
24:// RenderTemplateE runs `helm template` to render the template given the provided options and returns stdout/stderr from
26:func RenderTemplateE(t *testing.T, options *Options, chartDir string, templateFiles []string) (string, error) {
```

StupidScience:
@yorinasub17 updated

yorinasub17:
Build passed and everything looks good, so will go ahead and merge this! Thanks for your contribution!

yorinasub17:
https://github.com/gruntwork-io/terratest/releases/tag/v0.23.0

PR #406, opened 2019-11-10
Add Docker build and run commands

brikis98:
This PR adds `docker.Build` and `docker.Run` commands to Terratest. These are quite handy when you need to build and run a Docker image at test time. I've written a few local versions of these in various tests, so I figured it was time to finally centralize the implementation in Terratest.


brikis98:
Need to use this in a project, so going to merge now. Feedback is very welcome!

PR #425, opened 2019-12-20
Correct Golint warning

gliptak:


modules/aws/ec2-files.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/gruntwork-io/terratest/modules/ssh"
 )
 
+// RemoteFileSpecification sdescribes which files you want to copy from your instances
```
brikis98:
```suggestion
// RemoteFileSpecification describes which files you want to copy from your instances
```

gliptak: Updated. Sorry for the typo @brikis98

PR #426, opened 2019-12-26
Enable gofmt pre-commit check

gliptak:


yorinasub17:
Thanks for your contribution! I'm actually going to close without merging though because `goimports` is a superset of `gofmt` so this is unnecessary. [From the docs](https://godoc.org/golang.org/x/tools/cmd/goimports):

> In addition to fixing imports, goimports also formats your code in the same style as gofmt

gliptak:
@yorinasub17 I'm not following

https://goreportcard.com/report/github.com/gruntwork-io/terratest#gofmt shows ```gofmt``` warnings which weren't caught by ```goimports```

yorinasub17:
See https://github.com/gruntwork-io/pre-commit/pull/17#issuecomment-569117010

Issue #421, opened 2019-12-18
Adding support to run and test docker stop command


dhwaneetbhatt:
I am using terratest to test images which run a node server, and at the end of every test I stop the container and run it again in another test using different combinations.

For it, I would need a way to stop the container, I currently do it by running `docker stop` using Go's built in exec command, but it does make sense to bake it into the framework.

I have implemented this locally and its working. If it is good enough to be added to the framework, I'll raise a PR for it.

yorinasub17:
Please check out [our contribution guide](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md#types-of-contributions). To avoid Terratest from becoming massive and bloated, we need to be picky about the helpers we add.

A `docker.Stop` helper function passes the litmus test for "platforms", "popularity", and "creating infrastructure", but I am not sure it passes the test for "complexity." If you look at the docker helpers we currently have, there is not a lot of code surrounding the `run` and `build` commands: most of the heavy lifting is done by the `shell` module.

So a quick sanity check: would it simplify your logic some more if you use the `shell` module as opposed to the `exec` function?

dhwaneetbhatt:
@yorinasub17 yes that's correct, it doesn't pass the test of complexity. I would be okay using terratest shell for the use case. Thanks for providing a good workaround, its much easier to use than Go's exec.

Also, I would sooner or later need to use docker exec as well to verify few things once the container is running. I could use shell for that too but do you think it passes litmus test for all the parameters to be added to the terratest docker module?

brikis98:
@yorinasub17 One thought: while we certainly don't want Terratest to get too bloated, I wonder if our contribution guidelines are missing a balance of "API is reasonably complete and coherent." E.g., if we have a `docker run` helper, a `docker stop` helper seems like it should be part of the API too, even if it's not overly complicated to write. That said, we don't need to support *every* Docker CLI method; just the major ones, and those that would often be used in tandem. WDYT?

dhwaneetbhatt:
For context, I mostly test containers which run services, run I have to run them in detach mode, and then test a couple of things, and then stop them before the test ends.

For some cases, I also use `docker exec` to run commands inside the container to assert a few things (proper versions of libraries installed, directories exist etc). I consider `docker stop` to be a core docker CLI command, but I do not consider `docker exec` to be a core CLI method, but `docker exec` is sufficiently more complex so it passes the litmus test for complexity.

[...]

PR #430, opened 2020-01-07
Adding support for docker stop

dhwaneetbhatt:
# Description
Adding support for `docker stop` command in the terratest docker API.

# Github Issue
https://github.com/gruntwork-io/terratest/issues/421

# pre-commit hook
```
git commit -m "Adding support for docker stop"
goimports................................................................Passed
[docker-stop-command 1cb9753] Adding support for docker stop
 2 files changed, 87 insertions(+)
 create mode 100644 modules/docker/stop.go
 create mode 100644 modules/docker/stop_test.go
```

# Test Run output
https://gist.github.com/dhwaneetbhatt/018f0775493143b30b7b463deaa45e37

### DIFF
--- a/modules/docker/stop.go
+++ b/modules/docker/stop.go
@@ -49,9 +49,7 @@ func formatDockerStopArgs(containers []string, options *StopOptions) ([]string, 
 		args = append(args, "--time", strconv.Itoa(options.Time))
 	}
 
-	for _, container := range containers {
-		args = append(args, container)
-	}
+	args = append(args, containers...)
 
 	return args, nil
 }
--- a/modules/docker/stop.go
+++ b/modules/docker/stop.go
@@ -1,0 +1,57 @@
+package docker
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/shell"
+	"github.com/stretchr/testify/require"
+)
+
+// StopOptions defines the options that can be passed to the 'docker stop' command
+type StopOptions struct {
+	// Seconds to wait for stop before killing the container (default 10)
+	Time int
+}
+
+// Stop runs the 'docker stop' command for the given containers and return the stdout/stderr. This method fails
+// the test if there are any errors
+func Stop(t *testing.T, containers []string, options *StopOptions) string {
+	out, err := StopE(t, containers, options)
+	require.NoError(t, err)
+	return out
+}
+
+// StopE runs the 'docker stop' command for the given containers and returns any errors.
+func StopE(t *testing.T, containers []string, options *StopOptions) (string, error) {
+	logger.Logf(t, "Running 'docker stop' on containers '%s'", containers)
+
+	args, err := formatDockerStopArgs(containers, options)
+	if err != nil {
+		return "", err
+	}
+
+	cmd := shell.Command{
+		Command: "docker",
+		Args:    args,
+	}
+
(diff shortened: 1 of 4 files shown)


modules/docker/stop.go
```diff
@@ -0,0 +1,57 @@
...
+func formatDockerStopArgs(containers []string, options *StopOptions) ([]string, error) {
+	args := []string{"stop"}
+
+	if options.Time != 0 {
+		args = append(args, "--time", strconv.Itoa(options.Time))
+	}
+
+	for _, container := range containers {
+		args = append(args, container)
+	}
```
yorinasub17: NIT: This loop can be simplified to:

```
args = append(args, container...)
```

modules/docker/stop_test.go
```diff
@@ -0,0 +1,30 @@
...
+		Detach: true,
+		Name:   name,
+		Remove: true,
+	}
+	Run(t, "nginx:1.17-alpine", runOpts)
+
+	// try to stop it now
+	stopOpts := &StopOptions{}
+	out := Stop(t, []string{name}, stopOpts)
+	require.Contains(t, out, name)
```
yorinasub17: NIT: Would be nice to enhance the test with a check to see if the container actually stopped. A few options for this:

- Use `shell` to call `docker ps` to see if this container stopped.
- Since we are using nginx, we can use `HttpGet` to see if the container is unreachable. This requires an equivalent `HttpGet` check after the `Run` call to make sure the container is up.

modules/docker/stop.go
dhwaneetbhatt: @yorinasub17 done

modules/docker/stop_test.go
dhwaneetbhatt: @yorinasub17 I would prefer to use checks that are independent of other tests, `docker ps` output may vary with other running tests, so I implemented by calling out nginx using `HttpGet`, thanks for suggesting.

dhwaneetbhatt: Updated gist with test run: https://gist.github.com/dhwaneetbhatt/018f0775493143b30b7b463deaa45e37

yorinasub17: Thanks for adding the http get calls. However, I am worried about the brittleness of this test, especially under the load of all the other tests. E.g I am pretty sure there is a possibility that there is a delay between `docker run` and the nginx server being available, and so there might be a race condition between the `HttpGet` call and nginx coming up.

Can you update those checks to instead use https://github.com/gruntwork-io/terratest/blob/master/modules/http-helper/http_helper.go#L120 for the confirmation check (wait until nginx comes up), and https://github.com/gruntwork-io/terratest/blob/master/modules/retry/retry.go#L53 for the negation check (wait until nginx comes down)? 2 minutes is probably enough for the wait (60 for max retries, 2*time.Second for sleep between retries).

Let me know if it isn't clear how to use those functions and happy to provide additional pointers.

dhwaneetbhatt: @yorinasub17 made the changes. For the nginx going down, the `docker stop` command is actually synchronous and will wait for container to die before returning. So, we can just verify by executing `docker ps` along with name filter and ensure it does not return any values.

Made another minor change regarding godoc comment in `run.go`, and added `.vscode` in .gitignore since I use vscode.

Let me know if it looks good.

yorinasub17:
This looks good! Thanks for your patience working with me on improving the tests.

Going to kick off another build now, and will merge if it passes.

yorinasub17:
Build passed, so going to merge this. Thanks again for your contribution!

PR #434, opened 2020-01-09
Add methods to test Google Cloud Build resources

robmorgan:
This PR adds methods to test [Google Cloud Build](https://cloud.google.com/cloud-build) resources.

**Update:** I also had to bump the Azure SDK as https://github.com/google/go-containerregistry also depends on it and the older versions were causing a conflict.


modules/gcp/cloudbuild_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if err := tw.WriteHeader(hdr); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := tw.Write([]byte(file)); err != nil {
+		t.Fatal(err)
+	}
+	if err := tw.Close(); err != nil {
+		t.Fatal(err)
+	}
```
yorinasub17: NIT: each of these should probably be `require.NoError`

modules/gcp/gcr.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	opts := gcrremote.WithAuth(auther)
+
+	if err := gcrremote.Delete(name, opts); err != nil {
+		return fmt.Errorf("Failed to delete %s. Got error: %v", name, err)
+	}
+
+	return nil
+}
```
yorinasub17: I am not sure these are interfaces we want to maintain, as they fail the litmus test for "Creating infrastructure": https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md#types-of-contributions

Is there a reason you can't use a terraform example for the GCR repositories?

modules/gcp/cloudbuild.go
```diff
@@ -0,0 +1,165 @@
...
+func CreateBuild(t *testing.T, projectID string, build *cloudbuildpb.Build) *cloudbuildpb.Build {
+	out, err := CreateBuildE(t, projectID, build)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return out
+}
+
+// CreateBuildE creates a new build blocking until the operation is complete.
+func CreateBuildE(t *testing.T, projectID string, build *cloudbuildpb.Build) (*cloudbuildpb.Build, error) {
```
yorinasub17: Same as above re:creating infrastructure litmus test

modules/gcp/gcr.go
yorinasub17: After reading through https://stackoverflow.com/questions/53714284/how-do-you-create-a-gcr-repo-with-terraform, this makes sense to me given that there is no way to explicitly manage GCR repositories on the fly using `terraform`. This means that to properly clean up a test, you have to manually delete the images that are pushed using `docker`, and these functions makes it easy to do that.

robmorgan: Ok

modules/gcp/cloudbuild_test.go
robmorgan: fixed.

robmorgan:
Merging

PR #440, opened 2020-01-14
Add lambda function "invoke" helper.

mjgpy3:
I tend to use this a lot in my infrastructure tests.


yorinasub17:
Hi thanks for your contribution! Looking at the code, I am not sure this by itself passes our litmus test for wrapper functions. See [our contribution guide](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md#types-of-contributions). I think these wrappers are simple enough that it doesn't pass the test for "complexity".

With that said, I think enhancing these functions to automatically serialize and deserialize the payload as json would make them more useful. E.g:

```
func InvokeFunctionE(t *testing.T, region string, functionNameOrArn string, payload interface{}) (FunctionResult, error) {
    var payloadJson []byte = nil
    if payload != nil {
        payloadJson, err := json.Marshal(payload)
        if err != nil {
            return FunctionResult{} err
        }
    }
    // ... rest of function omitted for brevity ...
}
```

Additionally, I think errors in lambda should be treated as errors in `InvokeFunctionE`. If we accept that, then the `FunctionResult` struct should actually be a custom error struct (`FunctionError`) that is returned as the error if `FunctionError` is set on the response.

I think if we implemented these two features, it would make this pass the complexity test.

mjgpy3:
@yorinasub17 thank you for the feedback! I'll push those changes soon.

mjgpy3:
@yorinasub17 I believe that I've implemented your recommendations. Let me know your thoughts!

modules/aws/lambda.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if payload != nil {
+		payloadJson, err := json.Marshal(payload)
+
+		if err != nil {
+			return nil, err
+		} else {
+			invokeInput.Payload = payloadJson
+		}
```
yorinasub17: NIT: No need for using `else` here.

modules/aws/lambda.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/aws/aws-sdk-go/service/lambda"
+)
+
+// InvokeFunction invokes a lambda function.
+func InvokeFunction(t *testing.T, region, functionName string, payload interface{}) []byte {
+	out, err := InvokeFunctionE(t, region, functionName, payload)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yorinasub17: NIT: we've been standardizing on using `require.NoError(t, err)` in place of `if => t.Fatal`. The import path is `"github.com/stretchr/testify/require"`.

modules/aws/lambda.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type FunctionError struct {
+	Message    string
+	StatusCode int64
+}
+
+func (err *FunctionError) Error() string {
+	return err.Message
+}
```
yorinasub17: You might want the Payload here. The `FunctionError` string actually doesn't contain anything useful, since it is either `Handled` or `Unhandled`. Outputting the payload is going to be more useful:

```
return fmt.Sprintf("%s error invoking lambda function: %v", err.Message, err.Payload)
```

modules/aws/lambda.go
```diff
@@ -0,0 +1,78 @@
...
+
+// NewLambdaClient creates a new Lambda client.
+func NewLambdaClient(t *testing.T, region string) *lambda.Lambda {
+	client, err := NewLambdaClientE(t, region)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yorinasub17: NIT: See above re: `require.NoError`

modules/aws/lambda.go
```diff
@@ -0,0 +1,78 @@
...
+	out, err := lambdaClient.Invoke(invokeInput)
+	if err != nil {
+		return nil, err
+	}
+
+	if out.FunctionError != nil {
+		return out.Payload, &FunctionError{Message: *out.FunctionError, StatusCode: *out.StatusCode}
+	}
+
+	return out.Payload, err
```
yorinasub17: NIT: Probably better to explicitly return `nil`:
```suggestion
	return out.Payload, nil
```

mjgpy3:
@yorinasub17 let me know if anything else needs to be done! Thanks for the feedback.

examples/terraform-aws-lambda-example/README.md
```diff
@@ -0,0 +1,25 @@
+# Terraform Lambda Example
+
+This folder contains a very simple Terraform module to demonstrate how you can use Terratest to deploy a lambda function
```
yorinasub17: NIT:
```suggestion
This folder contains a Terraform module to demonstrate how you can use Terratest to deploy a lambda function
```

It's usually not a good idea to say "very simple" or even "simple" because that can be subjective.

examples/terraform-aws-lambda-example/main.tf
```diff
@@ -0,0 +1,39 @@
...
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          Service = "lambda.amazonaws.com"
+        }
+      }
+    ]
+    Version : "2012-10-17"
+  })
```
yorinasub17: Use [`aws_iam_policy_document` data source](https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html).

examples/terraform-aws-lambda-example/main.tf
```diff
@@ -0,0 +1,39 @@
...
+# ---------------------------------------------------------------------------------------------------------------------
+
+provider "archive" {
+  version = "1.3"
+}
+
+data "archive_file" "zip" {
+  type        = "zip"
+  source_dir  = "${path.module}/src"
+  output_path = "/tmp/.${var.function_name}.zip"
```
yorinasub17: NIT: This won't work on all OS platforms (e.g Windows). Probably better to use `${path.module}`.

examples/terraform-aws-lambda-example/main.tf
```diff
@@ -0,0 +1,39 @@
+# ---------------------------------------------------------------------------------------------------------------------
```
yorinasub17: Ah I realized you are using tf12 syntax. We haven't updated our CI routine to use tf12 yet so this will probably fail. We have started the work though so will likely have a PR with the updates in the next week. Probably makes sense to do that instead of updating to tf11.

test/terraform_aws_lambda_example_test.go
```diff
@@ -0,0 +1,64 @@
...
+	response := aws.InvokeFunction(t, awsRegion, functionName, ExampleFunctionPayload{ShouldFail: false, Echo: "hi!"})
+
+	// This function just echos it's input as a JSON string when `ShouldFail` is `false``
+	assert.Equal(t, `"hi!"`, string(response))
+
+	// Invoke the function, this time causing it to error and capturing the error
+	response, err := aws.InvokeFunctionE(t, awsRegion, functionName, ExampleFunctionPayload{ShouldFail: true, Echo: "hi!"})
+
+	// Function-specific errors have their own special return
+	functionError := err.(*aws.FunctionError)
```
yorinasub17: Use the syntax:
```suggestion
	functionError, ok := err.(*aws.FunctionError)
	require.True(t, ok)
```

That way, we won't panic and shutdown ALL other running tests if it is not this error.

mjgpy3:
@yorinasub17 more good suggestions, thank you! I've implemented them and am happy to wait for the TF syntax to be updated.

yorinasub17:
We just merged in the tf12 update, so if you can rebase this branch on master, I can go ahead and kick off the tests. Thanks for waiting!

mjgpy3:
@yorinasub17 I just rebased. No problem in waiting.

yorinasub17:
Thanks! Just kicked off our build.

examples/terraform-aws-lambda-example/src/lambda.go
```diff
@@ -0,0 +1,24 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"github.com/aws/aws-lambda-go/lambda"
+)
```
yorinasub17: The build failed `goimports` formatting on this file. This is because the third party import is grouped with the stdlib imports. The build should move beyond that step once you add a new line between the `lambda` import and `fmt`.

mjgpy3: @yorinasub17 updated, thanks

yorinasub17:
The build failed, but it was for a known flaky test, so I am going to go ahead and merge + release this. Thanks for the contribution!

mjgpy3:
Thanks @yorinasub17!

PR #442, opened 2020-01-25
Azure Kubernetes Service sample and helper methods

TsuyoshiUshio:
Based on the conversation of the issue. https://github.com/gruntwork-io/terratest/issues/89
I create a sample and helper method for supporting Azure Kubernetes Service usage. 

This PR provide two helper methods. 

1. Fetch the Kubernetes Service metadata
1. Wait for the allocation of Public IP Address on Kubernetes Service

In case of Azure, Node Port is not good option for external access outside of the kubernetes cluster, we use ingress or LoadBalancer type. For both case, we need to wait until Public IPAddress is assigned to the service. It takes some time. That is why, the sample for kubernetes doesn't work for Azure users. So I'd like to contribute this to unblock it to Azure users.

### DIFF
--- a/examples/terraform-azure-aks-example/README.md
+++ b/examples/terraform-azure-aks-example/README.md
@@ -1,0 +1,59 @@
+# Terraform Azure AKS Example
+
+This folder contains a Terraform module that deploys a basic AKS cluster in [Azure](https://azure.microsoft.com/) to demonstrate how you can use Terratest to write automated tests for your Azure Terraform code. 
+
+This module deploys [Azure Kubenetes Service](https://azure.microsoft.com/en-us/services/kubernetes-service/), then deploys nginx by a kubernetes yaml file with a Public IP Address using the `Service` resource.
+
+Check out [test/terraform_azure_aks_example_test.go](/test/terraform/azure_aks_example_test.go) to see how you can write automated tests for this module and validate the configuration of the parameters and options. 
+
+**WARNING**: This module and the automated tests for it deploy real resources into your Azure account which can cost you money. 
+
+## Prerequisite: Setup Azure CLI access
+1. Sign up for [Azure](https://azure.microsoft.com/).
+1. Install [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest)
+1. Install [Terraform](https://www.terraform.io/) and make sure it's on your `PATH`.
+1. Install [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) and make sure it's on your `PATH`.
+1. Login to Azure on the CLI with `az login` or `az login --use-device`, and then configure the CLI.
+
+## Running this module manually
+1. Create [Service Principal](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli?view=azure-cli-latest) then set the value to the environment variables. 
+1. Run `terraform init`
+1. Run `terraform apply`
+1. Apply `nginx-deployment.yml`
+1. Watch the service until Public IPAddress is assigned.
+1. Send http request to the Public IPAddress, make sure it returns 200.
+1. When you're done, run `terraform destroy`.
+
+### Example
+
+```bash
+$ az login 
+$ export ARM_SUBSCRIPTION_ID={YOUR_SUBSCRIPTION_ID} 
+$ az ad sp create-for-rbac
+$ export TF_VAR_client_id={YOUR_SERVICE_PRINCIPAL_APP_ID}
+$ export TF_VAR_client_secret={YOUR_SERVICE_PRINCIPAL_PASSWORD}
+$ terraform init
+$ terraform apply
+$ kubectl --kubeconfig ./kubeconfig -f ./nginx-deployment.yml
+$ kubectl --kubeconfig ./kubeconfig get svc -w
+// Open browser and access the Nginx Service IPAddress
+$ terraform destroy
+```
+
+## Running automated tests against this module
+1. Create [Service Principal](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli?view=azure-cli-latest) then set the value to the environment variables. 
+1. Install [Golang](https://golang.org/) version `1.13+` required. 
+1. `cd test`
+1. `go test -v -timeout 60m -tags azure -run TestTerraformAzureAKS`
+
+
+### Example
+
+```bash
+$ az login 
+$ export ARM_SUBSCRIPTION_ID={YOUR_SUBSCRIPTION_ID} 
+$ export TF_VAR_client_id={YOUR_SERVICE_PRINCIPAL_APP_ID}
+$ export TF_VAR_client_secret={YOUR_SERVICE_PRINCIPAL_PASSWORD}
+$ cd test
+$ go test -v -timeout 60m -tags azure -run TestTerraformAzureAKS
+```
--- a/examples/terraform-azure-aks-example/main.tf
+++ b/examples/terraform-azure-aks-example/main.tf
@@ -1,0 +1,69 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY AN AZURE AKS CLUSTER
+# This is an example of how to deploy an Azure AKS cluster with load balancer in front of the service 
+# to handle providing the public interface into the cluster.
+# ---------------------------------------------------------------------------------------------------------------------
+
+# ------------------------------------------------------------------------------
+# CONFIGURE OUR AZURE CONNECTION
+# ------------------------------------------------------------------------------
+
+provider "azurerm" {
+  version = "1.40.0"
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY A RESOURCE GROUP
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "azurerm_resource_group" "k8s" {
+  name     = var.resource_group_name
+  location = var.location
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY AN AZURE KUBERNETES CLUSTER
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "azurerm_kubernetes_cluster" "k8s" {
+  name                = var.cluster_name
+  location            = azurerm_resource_group.k8s.location
+  resource_group_name = azurerm_resource_group.k8s.name
+  dns_prefix          = var.dns_prefix
+
+  linux_profile {
+    admin_username = "ubuntu"
+
+    ssh_key {
+      key_data = file(var.ssh_public_key)
+    }
+  }
+
+  default_node_pool {
+    name       = "agentpool"
+    node_count = var.agent_count
+    vm_size    = "Standard_DS1_v2"
+  }
+
+  service_principal {
+    client_id     = var.client_id
+    client_secret = var.client_secret
+  }
+
+  tags = {
+    Environment = "Development"
+  }
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# CREATE KUBECONFIG FILE
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "local_file" "kubeconfig" {
+  content  = azurerm_kubernetes_cluster.k8s.kube_config_raw
+  filename = "kubeconfig"
+
+  depends_on = [
+    azurerm_kubernetes_cluster.k8s
+  ]
+}
--- a/examples/terraform-azure-aks-example/variables.tf
+++ b/examples/terraform-azure-aks-example/variables.tf
@@ -1,0 +1,46 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# REQUIRED PARAMETERS
+# You must provide a value for each of these parameters.
+# ---------------------------------------------------------------------------------------------------------------------
+
+variable "client_id" {
+  description = "The Service Principal Client Id for AKS to modify Azure resources."
+}
+variable "client_secret" {
+  description = "The Service Principal Client Password for AKS to modify Azure resources."
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# OPTIONAL PARAMETERS
+# These parameters have reasonable defaults.
+# ---------------------------------------------------------------------------------------------------------------------
+
+variable "agent_count" {
+  description = "The number of the nodes of the AKS cluster."
+  default     = 3
+}
+
+variable "ssh_public_key" {
+  description = "The public key for the ssh connection to the nodes."
+  default     = "~/.ssh/id_rsa.pub"
+}
+
+variable "dns_prefix" {
+  description = "The prefix to set for the AKS cluster resoureces names."
+  default     = "k8stest"
+}
+
+variable cluster_name {
+  description = "The name to set for the AKS cluster."
+  default     = "k8stest"
+}
+
+variable resource_group_name {
+  description = "The name to set for the resource group."
+  default     = "azure-k8stest"
+}
+
+variable location {
+  description = "The location to set for the AKS cluster."
+  default     = "Central US"
+}
--- a/modules/azure/aks.go
+++ b/modules/azure/aks.go
@@ -1,0 +1,45 @@
+package azure
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2019-11-01/containerservice"
+)
+
+// GetManagedClustersClientE is a helper function that will setup an Azure ManagedClusters client on your behalf
+func GetManagedClustersClientE(subscriptionID string) (*containerservice.ManagedClustersClient, error) {
+	// Validate Azure subscription ID
+	subscriptionID, err := getTargetAzureSubscription(subscriptionID)
+	if err != nil {
+		return nil, err
+	}
+
+	managedServicesClient := containerservice.NewManagedClustersClient(subscriptionID)
+	authorizer, err := NewAuthorizer()
+
+	if err != nil {
+		return nil, err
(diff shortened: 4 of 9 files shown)


README.md
```diff
@@ -156,6 +156,8 @@ such as:
 1.  [Terraform Packer Example](/examples/terraform-packer-example): A more complicated example that shows how to use
     Packer to build an AMI with a web server installed and deploy that AMI in AWS using Terraform.
 1.  [Terraform GCP Example](/examples/terraform-gcp-example): A simple Terraform configuration that creates a GCP Compute Instance and Storage Bucket.
+1.  [Terraform Azure Example](/examples/terraform-azure-example): A simple Terraform configuration that creates an Azure VM.
+1.  [Terraform Azure AKS Example](/examples/terraform-azure-aks-example): A terraform configuration that create Azure Kubernetes Service Cluster
```
yorinasub17:
```suggestion
1.  [Terraform Azure AKS Example](/examples/terraform-azure-aks-example): A terraform configuration that creates an Azure Kubernetes Service (AKS) Cluster.
```

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
+# Terraform Azure AKS Example
+
+This folder contains a simple Terraform moudle that deploys a simple AKS cluster in [Azure](https://azure.microsoft.com/) to demonstrate how you can use Terratest to write automated tests for your Azure Terraform code.
```
yorinasub17:
```suggestion
This folder contains a Terraform module that deploys a basic AKS cluster in [Azure](https://azure.microsoft.com/) to demonstrate how you can use Terratest to write automated tests for your Azure Terraform code. 
```

We try to avoid the word "simple" because it can be intimidating if you struggle to understand it, and even the most basic examples can be complex to understand if you've never used Azure.

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
+# Terraform Azure AKS Example
+
+This folder contains a simple Terraform moudle that deploys a simple AKS cluster in [Azure](https://azure.microsoft.com/) to demonstrate how you can use Terratest to write automated tests for your Azure Terraform code. 
+
+This module deploies [Azure Kubenetes Service](https://azure.microsoft.com/en-us/services/kubernetes-service/), then deploies nginx by a kubernetes yaml file with Public IPAddress with a service.
```
yorinasub17:
```suggestion
This module deploys [Azure Kubenetes Service](https://azure.microsoft.com/en-us/services/kubernetes-service/), then deploys nginx by a kubernetes yaml file with a Public IP Address using the `Service` resource.
```

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+**WARNING**: This module and the automated tests for it deploy real resources into your Azure account which can cost you money. 
+
+## Running this module manually
+
+1. Sign up for [Azure](https://azure.microsoft.com/).
+1. Install [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest)
+1. Install [Terraform](https://www.terraform.io/) and make sure it's on your `PATH`.
+1. Install [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) and make sure it's on your `PATH`.
+1. Login Azure with `az login` or `az login --use-device` then configure
```
yorinasub17:
```suggestion
1. Login to Azure on the CLI with `az login` or `az login --use-device`, and then configure the CLI.
```

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. Create [Service Principal](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli?view=azure-cli-latest) then set the value to the environment variables. 
+1. Run `terraform init`
+1. Run `terraform apply`
+1. Apply nginx-deployment.yml
```
yorinasub17:
```suggestion
1. Apply `nginx-deployment.yml`
```

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. Watch the service until Public IPAddress is assigned.
+1. Send http request to the Public IPAddrees, make sure it returns 200.
```
yorinasub17:
```suggestion
1. Send http request to the Public IPAddress, make sure it returns 200.
```

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. When you're done, run `terraform destroy`.
+
+### Example
+
+```bash
+$ az login 
+$ export ARM_SUBSCRIPTION_ID={YOUR_SUBSCRIPTION_ID} 
+$ az ad sp create-for-rbac
+$ export TF_VAR_client_id={YOUR_SERVCIE_PRINIPAL_APP_ID}
```
yorinasub17:
```suggestion
$ export TF_VAR_client_id={YOUR_SERVICE_PRINCIPAL_APP_ID}
```

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
...
+// Open browser and access the Nginx Service IPAddress
+$ terraform destroy
+```
+
+## Running automated tests against this module
+1. Sign up for [Azure](https://azure.microsoft.com/).
+1. Install [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest)
+1. Install [Terraform](https://www.terraform.io/) and make sure it's on your `PATH`.
+1. Login Azure with `az login` or `az login --use-device` then configure
+1. Create [Service Principal](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli?view=azure-cli-latest) then set the value to the environment variables.
```
yorinasub17: NIT: Maybe extract these 4 steps out as a prerequisite step, labeled "Prerequisite: Setup Azure CLI access"? Then you can link to the same steps between running example and running test.

examples/terraform-azure-aks-example/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. Install [Golang](https://golang.org/) version `1.13+` required. 
+1. `cd test`
+1. `go test -v -timeout 30m -run TestTerraformAzureAKSExample`
+
+
+### Example
+
+```bash
+$ export ARM_SUBSCRIPTION_ID={YOUR_SUBSCRIPTION_ID} 
+$ export TF_VAR_client_id={YOUR_SERVCIE_PRINIPAL_APP_ID}
```
yorinasub17:
```suggestion
$ export TF_VAR_client_id={YOUR_SERVICE_PRINCIPAL_APP_ID}
```

examples/terraform-azure-aks-example/main.tf
```diff
@@ -0,0 +1,55 @@
+# ---------------------------------------------------------------------------------------------------------------------
```
yorinasub17: NIT: I know many of our examples in this repo don't do this, but for this module it might help to have a preamble with a high level description of what might be deployed. This is because there are more than 2 resources here, so it helps to give some context on the end result. See https://github.com/gruntwork-io/terraform-aws-influx/blob/master/examples/tick-multi-cluster/main.tf#L1 for example (this is one of our public modules that we maintain at Gruntwork).

examples/terraform-azure-aks-example/main.tf
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# CREATE KUBECONFIG FILE
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "local_file" "kubeconfig" {
+    content     = azurerm_kubernetes_cluster.k8s.kube_config_raw
+    filename = "kubeconfig"
```
yorinasub17: We like to keep our code formatted with `terraform fmt`

examples/terraform-azure-aks-example/main.tf
yorinasub17: NIT: We use two space tabbing for our tf files. It would be good to be consistent.

examples/terraform-azure-aks-example/provider.tf
```diff
@@ -0,0 +1,3 @@
+provider "azurerm" {
+    version = "1.40.0"
+}
```
yorinasub17: NIT: Can you put this in the `main.tf` file? We usually keep small files together to avoid file sprawl. Since the provider is always a single block, we don't typically split it out.

examples/terraform-azure-aks-example/variables.tf
```diff
@@ -0,0 +1,38 @@
+
+
```
yorinasub17: NIT: We like to avoid unnecessary whitespace.

yorinasub17: Also, we usually provide a description for all variables so that it is clear what should be put in here. E.g I do not know at a glance if `client_secret` refers to secret key for k8s, or for Azure.

modules/azure/aks.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2019-11-01/containerservice"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/retry"
+)
+
+// GetManagedClustersClient is a helper function that will setup an Azure ManagedClusters client on your behalf
+func GetManagedClustersClient(subscriptionID string) (*containerservice.ManagedClustersClient, error) {
```
yorinasub17: Terratest convention is to use `E` to denote functions that return an `error`. Functions without `E` should use `require` to fail the test if there is an error. This applies to all functions in this file.

modules/azure/aks.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	managedServicesClient.Authorizer = *authorizer
+
+	return &managedServicesClient, nil
+}
+
+// GetManagedCluster will return ContainerService
```
yorinasub17: This looks like it is returning a ManagedCluster object, and not ContainerService?

modules/azure/aks.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	managedCluster, err := client.Get(context.Background(), resourceGroupName, clusterName)
+	if err != nil {
+		return nil, err
+	}
+	return &managedCluster, nil
+}
+
+// WaitUntilServiceExternalIPsAvailable is waiting for allocation of External IP Address
+func WaitUntilServiceExternalIPsAvailable(t *testing.T, options *k8s.KubectlOptions, serviceName string, retries int, sleepBetweenRetries time.Duration) {
```
yorinasub17: This already exists in `k8s`: https://github.com/gruntwork-io/terratest/blob/master/modules/k8s/service.go#L58

test/terraform_azure_aks_example_test.go
```diff
@@ -0,0 +1,104 @@
...
+	defer terraform.Destroy(t, terraformOptions)
+
+	// This will run `terraform init` and `terraform apply` and fail the test if there are any errors
+	terraform.InitAndApply(t, terraformOptions)
+
+	// Look up the cluster node count
+	var ActualCount int32 = 0
+	cluster, err := azure.GetManagedCluster(t, expectedResourceGroupName, expectedClusterName, "")
+	assert.Nil(t, err)
+	ActualCount = *(*cluster.ManagedClusterProperties.AgentPoolProfiles)[0].Count
```
yorinasub17: We can probably simplify this as:
 
```suggestion
	actualCount := *(*cluster.ManagedClusterProperties.AgentPoolProfiles)[0].Count
```
and remove the `ActualCount` declaration above.

test/terraform_azure_aks_example_test.go
```diff
@@ -0,0 +1,104 @@
...
+	// At the end of the test, run `terraform destroy` to clean up any resources that were created
+	defer terraform.Destroy(t, terraformOptions)
+
+	// This will run `terraform init` and `terraform apply` and fail the test if there are any errors
+	terraform.InitAndApply(t, terraformOptions)
+
+	// Look up the cluster node count
+	var ActualCount int32 = 0
+	cluster, err := azure.GetManagedCluster(t, expectedResourceGroupName, expectedClusterName, "")
+	assert.Nil(t, err)
```
yorinasub17:
```suggestion
	require.NoError(t, err)
```

modules/azure/aks.go
TsuyoshiUshio: @yorinasub17 It doesn't work for Azure. In case of Azure, we need to use LoadBalancer or ingress type to accept external access. For the Load Balancer with Azure, the method go to the next line if it creates a service, however, we need to wait, until the service is actually get an IP Address. That is why, I create this method. The next question is should I keep it on Azure or should I move it to k8s namespace? If it is general, k8s might be useful for users, however, if it happens only on Azure, I'm wondering where is a good place to put this method.

yorinasub17: Hmm that seems like a bug if you are seeing this behavior, because it should be doing exactly that: wait for IP address

Can you see what is wrong with https://github.com/gruntwork-io/terratest/blob/master/modules/k8s/service.go#L81?

examples/terraform-azure-aks-example/main.tf
TsuyoshiUshio: Done.  Please review it.

examples/terraform-azure-aks-example/main.tf
TsuyoshiUshio: Done

examples/terraform-azure-aks-example/provider.tf
TsuyoshiUshio: Done

examples/terraform-azure-aks-example/variables.tf
TsuyoshiUshio: Done

modules/azure/aks.go
TsuyoshiUshio: Fixed. Done.

test/terraform_azure_aks_example_test.go
TsuyoshiUshio: Done

modules/azure/aks.go
TsuyoshiUshio: Done

modules/azure/aks.go
TsuyoshiUshio: @yorinasub17  I read the code, it looks exactly the same as mine. 
I suspect that the code is the same, so I remove mine, and point k8s namespace one. 
Maybe the problem is, Even after the IPAddress is allocated, it looks still take time to fetch it.
My test execute polling after the IPAddress readiness, so it might hide the issue. 
 
I'm testing my method ,then I'll replace the method to k8s's and replace it then observe the result, if it looks the same, I'm remove my method.

TsuyoshiUshio: Hi @yorinasub17 
I figure out what is the problem.  It is not about the `k8s.WaitUntilServiceAvailable()` It works well, however, the problem is, `k8s.GetServiceEndpoint()` The endpoint returns an empty string in case of LoadBalancer type of Azure. Since the method reference Hostname. 

image

The GetServiceEndpointE's documentation said that it return hostname, the behavior works correctly.  

I'm gonna remove WaitFor... method and use `k8s`  and try not to use `GetServiceEndpoint` for Azure.  I'll use 

```go 
	endpoint := service.Status.LoadBalancer.Ingress[0].IP
```
  instead.

examples/terraform-azure-aks-example/README.md
TsuyoshiUshio: Done

TsuyoshiUshio:
Hi @yorinasub17  I've done the all fix for your requests. Thank you for your kind review!

modules/azure/aks.go
TsuyoshiUshio: Absolutely, I've done it. Please make sure run CI test for Kubernetes.

TsuyoshiUshio:
Hi @yorinasub17  I've done your request!

TsuyoshiUshio:
@yorinasub17 I remove the merge conflict.

yorinasub17:
@TsuyoshiUshio Thanks! I think code wise this now looks good, but the main sticking point for us right now is the prospect of running automated tests to ensure we can maintain the code going forward. Will post an update in https://github.com/gruntwork-io/terratest/issues/89.

TsuyoshiUshio:
Thank you for the kind review @yorinasub17 , I put the comment on #89

yorinasub17:
We just merged in our tf12 update fix which stabilizes the build. Can you rebase this branch on master to pull in those changes? I can kick off the build once that is done!

TsuyoshiUshio:
Hi @yorinasub17 
Thank you!
I've done the rebase. Please kick off the  build!

TsuyoshiUshio:
Hi, @yorinasub17. I have a look at the root cause of the CI failure. 
It seems Packer related testing, and the main branch also has a failure of Packer related testing. 
Is it freaky testing?

yorinasub17:
@TsuyoshiUshio Yup it is flaky testing. The last two build failures contain tests that were mutually exclusive, so from a build perspective, this is good to go. Apologies that I haven't merged+released it yet. I need some time to formulate the release notes so I had been delaying doing that till later in the day. You should see this merged + released before EOD pacific time.

TsuyoshiUshio:
@yorinasub17 Thank you!  
BTW, we are working on the subscription issue internally. Sorry for the late reply. :)

PR #448, opened 2020-01-31
LoadOrCreateAndSaveString

MajorBreakfast:
I found that I needed a function that either loads a test data string from disk or if it does not exist yet, creates it. I chose a name for the function that contains both "load" and "save".  An alternative to the current name is `LoadOrSaveString()`.

### DIFF
--- a/modules/test-structure/save_test_data.go
+++ b/modules/test-structure/save_test_data.go
@@ -108,6 +108,18 @@ func LoadString(t *testing.T, testFolder string, name string) string {
 	return val
 }
 
+// LoadOrSaveString loads and unserializes a uniquely named string value from the given folder. If the string is not found,
+// it will create and save it instead.
+func LoadOrSaveString(t *testing.T, testFolder string, name string, create func() string) string {
+	if IsTestDataPresent(t, formatNamedTestDataPath(testFolder, name)) {
+		return LoadString(t, testFolder, name)
+	}
+
+	val := create()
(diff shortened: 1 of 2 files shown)


MajorBreakfast:
I used it for a random Kubernetes cluster name. Usually I want to get a new random name on each test run. However, if a name has already been stored (and not cleaned up because I purposefully skipped the cleanup via a `SKIP_cleanup` env var), I don't want to generate a new random name, but reuse the current one, to make the `terraform apply` cheap. The `LoadOrSaveString` function makes the code for this nice and short.

brikis98:
Oh, interesting. A pattern we've used for that is:

```go
test_structure.RunTestStage(t, "setup", func() {
  uniqueName := fmt.Sprintf("foo-%s", random.UniqueId())
  test_structure.SaveString(t, "/some/folder", "unique-name", uniqueName)
})

name := test_structure.LoadString(t, "/some/folder", "unique-name")

// Then use name in the rest of the test
```

If you run the test with no env vars, it creates and stores the name. If you skip the `setup` step, then it only loads the name. 

Would that accomplish what you're trying to do?

MajorBreakfast:
yes this works.

The difference in usage is that you need to run the tests with `SKIP_cleanup=true` on the first run and `SKIP_cleanup=true SKIP_setup=true` on subsequent runs. However, that slight complication is probably not worth creating `LoadOrSave...` functions for each data type. Thanks for the tip! The setup stage approach works

PR #450, opened 2020-02-04
Fix hello world example for k8s on minikube

yorinasub17:
The hello world example for k8s is broken because `LoadBalancer` services on `minikube` does not behave like any other k8s flavor. `LoadBalancer` services are automatically converted to `NodePort` under the hood in `minikube`, so it doesn't get the `ingress` property. Unfortunately, a lot of our functions assume the `ingress` property and use that to determine service endpoints and availability for that. Note that this is also happening in CI and our CI builds have been broken.

This fixes the behavior by having special handling of minikube in the various service functions.

### DIFF
--- a/modules/k8s/service.go
+++ b/modules/k8s/service.go
@@ -68,7 +68,15 @@ func WaitUntilServiceAvailable(t *testing.T, options *KubectlOptions, serviceNam
 			if err != nil {
 				return "", err
 			}
-			if !IsServiceAvailable(service) {
+
+			isMinikube, err := IsMinikubeE(t, options)
+			if err != nil {
+				return "", err
+			}
+
+			// For minikube, all services will be available immediately so we only do the check if we are not on
+			// minikube.
+			if !isMinikube && !IsServiceAvailable(service) {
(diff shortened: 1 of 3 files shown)


modules/k8s/service.go
```diff
@@ -68,7 +68,15 @@ func WaitUntilServiceAvailable(t *testing.T, options *KubectlOptions, serviceNam
...
 				return "", err
 			}
-			if !IsServiceAvailable(service) {
+
+			isMinikube, err := IsMinikubeE(t, options)
+			if err != nil {
+				return "", err
+			}
+
+			// For minikube, all services will be available immediately so we only do the check if we are not on
```
brikis98: Why are they available immediately? Is it not async / event-based under the hood?

yorinasub17: `NodePort` and `ClusterIP` are available immediately at creation time because all the information/control is available to the master components.

yorinasub17:
Going to merge and release this. Be advised that builds are still failing because tf12 update is not done yet.

Issue #445, opened 2020-01-28
Make each module independently consumable


rvasilevsf:
Hey folks,
Thanks for this wonderful collection of modules and examples! It definitely makes everybody's life easier!

With the rapid development of new modules and functionalities though, importing terratest becomes somewhat cumbersome, as it brings so many (sometimes conflicting) dependencies.
Is there any reason why you haven't enabled module consumption on per module basis, rather than on a repository basis? In other words, having go.mod in each module folder.
Would it be possible to add it? It will greatly help importing and using just the necessary dependencies.

yorinasub17:
That is actually not a bad idea! I think the only potential downside is disparate versions of common libraries if there is a library that is shared across the modules, but AFAIK, this is probably rare in the case of the modules within terratest since they are mostly isolated in scope.

What do you think @brikis98 ?

brikis98:
I like the idea too, but there are some logistics to think through. Would we have to break down the test suite by example too? So there would be an `examples` folder that's mostly unchanged, a `modules` folder that would contain each module with its own `go.mod` and its own `xxx_test.go` files, and no more `test` folder? 

Also, does this make the CI build significantly more complicated? Does importing multiple packages get more complicated? 

If someone wants to experiment with this and submit an RFC with ideas, I'd be very open to it!

yorinasub17:
> Would we have to break down the test suite by example too?

We would want to keep the current model, where the test folder is unified. This allows us to use it as an example of how to consume terratest. In that regard, users are not going to want to break up their test by terratest module.


> Also, does this make the CI build significantly more complicated?

The way this would work is we would use a local module reference. The `go.mod` for `test` would look like:

```
require (
    github.com/gruntwork-io/terratest/modules/shell v0.0.0 // Local
    github.com/gruntwork-io/terratest/modules/aws v0.0.0 // Local
    github.com/gruntwork-io/terratest/modules/k8s v0.0.0 // Local
)

replace (
    github.com/gruntwork-io/terratest/modules/shell => ../modules/shell
    github.com/gruntwork-io/terratest/modules/aws => ../modules/aws
    github.com/gruntwork-io/terratest/modules/k8s => ../modules/k8s
)
```

> Does importing multiple packages get more complicated?

[...]

Issue #418, opened 2019-12-10
Should the set of examples work with terraform v0.12.12


ggrover:
I noticed that running the test against `examples/terraform-aws-example` failed as it was missing ` tags = {` after the
Just wondering if the tests should be working against terraform v0.12.12?

brikis98:
@ggrover Hm, you're right, the examples haven't been updated for 0.12.x, even though the library itself has!

@yorinasub17 It looks like we're still using Terraform 0.11 in the CI env for Terratest?

yorinasub17:
> It looks like we're still using Terraform 0.11 in the CI env for Terratest?

Yes. This was because this place is the only place currently that validates Terratest works with tf11, since all our other repos validates tf12 support.

With that said, at this point, I would like to propose we drop support for tf11. This will involve:

- Updating examples to use tf12
- Drop forking of output functions based on tf version
- Update CI environment to tf12

brikis98:
@yorinasub17 Ah, gotcha. +1 on what you suggested.

brikis98:
Note that it's not super high priority, as we know Terratest works with TF12. Perhaps punt until after CIS work?

yorinasub17:
This is now resolved. Beginning with [v0.24.0](https://github.com/gruntwork-io/terratest/releases/tag/v0.24.0), all examples have been updated to tf12 and tf11 support was dropped.

PR #451, opened 2020-02-05
Tf12 upgrade

yorinasub17:
This updates the examples and our CI environment to use Terraform 12. As a part of that transition, this PR drops support for TF11 in the various output related functions.

Additional upgrades:

- Update examples to be compatible with packer 1.5
- Update GCP code to be compatible with `google` provider 3.x
- Update fixtures to be compatible with terragrunt 19


examples/packer-docker-example/build.json
```diff
@@ -27,7 +27,8 @@
     "name": "ubuntu-docker",
     "type": "docker",
     "image": "gruntwork/ubuntu-test:16.04",
-    "commit": true
+    "commit": true,
+    "changes": ["ENTRYPOINT [\"\"]"]
```
yorinasub17: This is to address https://github.com/hashicorp/packer/issues/7487

examples/terraform-basic-example/variables.tf
```diff
@@ -18,22 +18,30 @@
...
-  type        = "map"
+  type        = map(string)
   default     = {}
 }
+
+variable "example_any" {
+  description = "An example variable that is can be anything"
+  type        = any
+  default     = null
+}
```
yorinasub17: This was necessary to make all the format tests work, since it uses complex nested collections.

examples/terraform-remote-exec-example/main.tf
```diff
@@ -55,24 +64,26 @@ resource "aws_security_group" "example" {
...
-    user = "${var.ssh_user}"
-    port = "${var.ssh_port}"
+    type  = "ssh"
+    host  = aws_instance.example_public.public_ip
+    user  = var.ssh_user
+    port  = var.ssh_port
     agent = true
   }
 
+  // copy our example script to the server
```
brikis98: Doubled

examples/terraform-remote-exec-example/main.tf
```diff
@@ -55,24 +64,26 @@ resource "aws_security_group" "example" {
...
   // copy our example script to the server
   provisioner "file" {
     source      = "files/get-public-ip.sh"
     destination = "/tmp/get-public-ip.sh"
   }
 
+  // change permissions to executable and pipe its output into a new file
```
brikis98: Doubled

modules/terraform/output.go
```diff
@@ -70,17 +70,15 @@ func OutputListE(t *testing.T, options *Options, key string) ([]string, error) {
 		return nil, err
 	}
 
-	if outputMap, isMap := output.(map[string]interface{}); isMap {
-		return parseListOutputTerraform11OrOlder(outputMap, key)
-	} else if outputList, isList := output.([]interface{}); isList {
-		return parseListOutputTerraform12OrNewer(outputList, key)
+	if outputList, isList := output.([]interface{}); isList {
+		return parseListOutputTerraform(outputList, key)
```
brikis98: Sanity check this is intentional to no longer support TF 0.11 after this release?

test/fixtures/copy-folder-contents/no-state-files/terraform.tfvars
```diff
@@ -1 +0,0 @@
- terragrunt = {}
```
brikis98: Intentional to delete, same with other .tfvars files?

modules/terraform/output.go
yorinasub17: Yes, based on https://github.com/gruntwork-io/terratest/issues/418#issuecomment-564638173

Will wait until you approve this before I move forward. 

FWIW, I think this is not a huge deal if we implement https://github.com/gruntwork-io/terratest/issues/445. Then, one could use an older version of the `terraform` module that is still compatible with tf11 but newer versions of the other helpers (e.g `aws`, `k8s`, etc).

test/fixtures/copy-folder-contents/no-state-files/terraform.tfvars
yorinasub17: Yes, since we switched to tg19

examples/terraform-remote-exec-example/main.tf
yorinasub17: Good catch. fixed! b8d0ff8

examples/terraform-remote-exec-example/main.tf
yorinasub17: Fixed! b8d0ff8

modules/terraform/output.go
brikis98: Agreed on dropping TF 11 support. No need to block on #445 at this point.

yorinasub17:
Thanks for review! Going to go ahead and merge + release this!

PR #458, opened 2020-02-12
Fix pre-commit call

yorinasub17:
The build broke because `pre-commit` can no longer be called due to issues with `cfgv`. This fixes that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,9 @@ jobs:
           name: run precommit
           command: |
             go get golang.org/x/tools/cmd/goimports
+            # Oct 26, 2019: Install the last known working version of pre-commit. Also, we have to pin the version of a
+            # transitive dependency that is being pulled in (cfgv) which released a new version that is no longer compatible
+            # with any python < 3.6.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -61,10 +61,13 @@ jobs:
...
       # Run pre-commit hooks and fail the build if any hook finds required changes.
-      - run: go get golang.org/x/tools/cmd/goimports
-      - run: pip install pre-commit==1.11.2
-      - run: pre-commit install
-      - run: pre-commit run --all-files
+      - run:
+          name: run precommit
+          command: |
+            go get golang.org/x/tools/cmd/goimports
+            pip install pre-commit==1.11.2 cfgv==2.0.1
```
brikis98: NIT: add a comment about why we need that specific version of cfgv

yorinasub17: :+1: Shamelessly copying from `terraform-aws-couchbase` 0d4d791

yorinasub17:
Ok merging this in. Thanks for review!

PR #462, opened 2020-02-19
Added helper function to extract tags from an AWS TagSet

bincyber:
This PR adds a generic helper function to extract the tags from an AWS TagSet into a map.

A [TagSet](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Tagging.html) is of the following form:

```
"TagSet": [
    {
      "Key": "Name",
      "Value": "my-vpc"
    },
    {
      "Key": "CreatedBy",
      "Value": "Terraform"
    },
    {
      "Key": "Env",
      "Value": "test"
    },
]
```

An example of using this to extract the tags set on a VPC:

```go
vpcID := "vpc-a5b9c5e7"

req := client.DescribeVpcsRequest(&ec2.DescribeVpcsInput{
	VpcIds: []string{
		vpcID,
	},
})

resp, err := req.Send(context.Background())
if err != nil {
	logger.Logf(t, "Failed to lookup the VPC %q. Error: %s", vpcID, err.Error())
}
assert.Nil(t, err)

tags := aws.ExtractResourceTags(resp.Vpcs[0].Tags)
```

### DIFF
--- a/modules/aws/ec2.go
+++ b/modules/aws/ec2.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


brikis98:
Agreed with @yorinasub17. We are always grateful for contributions, but as Terratest grows both in size and popularity, we have to be more and more picky with what we add to the library to ensure it doesn't become too bloated. I think looping over the tags directly in your own code is simple and readable enough (and probably more type safe) than this helper, so I'm going to close this one out. Thanks!

bincyber:
Thank you for reviewing my PR.

I agree with you that this function is somewhat complex and that the alternative is easier and more readable, however if a user wanted to check the tags of all created resources, looping over the tagset leads to a lot of code duplication . Nonetheless, I appreciate the feedback.

PR #467, opened 2020-02-26
Helm3 support

yorinasub17:
This updates terratest to be compatible with helm3. Here are the relevant changes:

- `helm.DeleteE`: `--purge` has been removed and is now the default. Instead, there is `--keep-history` for the negative of that.
- `helm.DeleteE`: Releases are now tied to the namespace they were deployed in, so we need to pass in the corresponding namespace.
- `helm.UpgradeE`: Releases are now tied to the namespace they were deployed in, so we need to pass in the corresponding namespace.
- `helm.InstallE`: Namespace argument is now required and is a positional argument.
- `helm.RenderTemplateE`: Release name is now required and is a positional argument.
- `helm.RenderTemplateE`: `-x` no longer exists, but is replaced with `--show-only`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ install_helm: &install_helm
   name: install helm
   command: |
     # install helm
-    curl -Lo helm.tar.gz https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-amd64.tar.gz
+    curl -Lo helm.tar.gz https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz
     tar -xvf helm.tar.gz
     chmod +x linux-amd64/helm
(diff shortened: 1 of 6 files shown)


yorinasub17:
Test failure was `TestPackerGCPBasicExample` which is completely unrelated to this change and is known to be flaky. I am disinclined to kick off another build, but let me know if I should until it passes. What is most important is that the helm tests pass: https://circleci.com/gh/gruntwork-io/terratest/2878

yorinasub17:
> Do these changes imply that Terratest now only works with Helm 3 and above?

Yes. I think this is fine as what we have now is sufficient for the majority of helm chart testing use cases. New features that we add would be for testing the helm 3 features like library charts. So if anyone needs to test against helm 2, they can use the current version of terratest that still supports it.

---

Since approved, will go ahead and merge + do backwards incompatible release. Thanks for the review!

PR #469, opened 2020-02-28
Do not convert booleans to int

### LINKED ISSUE #468: Booleans within Objects break when translated to ints
Hi,

I have been stumbling upon weird behaviour from terraform and terratest.

## Situation

Since tf 0.12, one can define `object`s (a.k.a structs) within variables. 
Example:

```hcl
variable "boolean" {
  type = object({
    x = string
    y = bool
  })
}
```

The weird behaviour that I'm seeing is that when a variable's top-level type is bool, one can use `1` as a value for `true` (and `0` for `false`).
When defining an object with a boolean inside, `0` and `1` are not allowed though.

`variables.tf`:
```hcl
variable "x" {
  type = bool
}

variable "y" {
  type = object({
    y = bool
  })
}

output "a" {
  value = var.x
}

output "b" {
  value = var.y
}
```

```
$ terraform apply -var 'y={"y" = 1}' -var 'x=1'

Error: Invalid value for input variable

The argument -var="y=..." does not contain a valid value for variable
"y": attribute "y": bool required.

$ terraform apply -var 'boolean={"y" = true}' -var 'x=1'
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

a = true
b = {
  "y" = true
}
```

This seems to be an inconsistency within terraform, and i am planning on opening an issue on their side too.

However, you have merged a PR quite some time ago already, that translates booleans into ints:
https://github.com/gruntwork-io/terratest/pull/113
I don't think that conversion from bool to int is necessary anymore.

tommyknows:
Fixes #468
This commit partly reverts the PR [#113](https://github.com/gruntwork-io/terratest/pull/113). Terraform
seems to not allow integers as booleans when these boolean types are
nested within other types. Top-Level variables as booleans are not
affected.
In order to fix this, we should not tranlate the booleans to integers
anymore, but pass them through directly.

### DIFF
--- a/modules/terraform/format.go
+++ b/modules/terraform/format.go
@@ -3,6 +3,7 @@ package terraform
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/gruntwork-io/terratest/modules/collections"
@@ -184,10 +185,7 @@ func primitiveToHclString(value interface{}, isNested bool) string {
 	switch v := value.(type) {
 
 	case bool:
-		if v {
-			return "1"
-		}
-		return "0"
+		return strconv.FormatBool(v)
 
 	case string:
 		// If string is nested in a larger data structure (e.g. list of string, map of string), ensure value is quoted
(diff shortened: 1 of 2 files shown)


yorinasub17:
Thanks for the contribution! The change looks good to me, and confirmed terraform 12 handles the bool correctly on the CLI arg. Kicked off a build and if it passes, we can merge this in.

tommyknows:
Can I not look at the build / test log / failure? 
(I get a 404 saying that "You may have been logged out.").

I can't run the tests locally, they already fail on master... 😕

yorinasub17:
Sorry I had intended to merge this (the test failure is an unrelated change), but I haven't had the chance yet to come back. I will be merging + releasing this before the next 4 hours.

yorinasub17:
Released as https://github.com/gruntwork-io/terratest/releases/tag/v0.25.1

PR #470, opened 2020-02-28
Allow configuring all command options for packer

krishicks:
:wave:

I have a packer template that includes copying of files from a directory relative to the working dir via a file provisioner; this means I need to be able to tell terratest to run with a specific working dir. `shell.Command` already supported this, but the packer module wasn't taking advantage of that. This change allows passing through WorkingDir down to the `shell.Command` that runs packer.

I also added the OutputMaxLineSize option while I was in there; this means packer can set all possible options that are able to be set on the `shell.Command`.

I didn't see a test that I could add to for this, so I did not add one. I'm happy to add one if I can be pointed to where it should belong, though!

A possible improvement to this would be further changes that make the `Template` option not required as packer does not require it be set if you use the `packer-template.json` name for the template. For now, my usage of terratest specifies both `WorkingDir` and `Template`.

### DIFF
--- a/modules/packer/packer.go
+++ b/modules/packer/packer.go
@@ -26,6 +26,8 @@ type Options struct {
 	RetryableErrors    map[string]string // If packer build fails with one of these (transient) errors, retry. The keys are a regexp to match against the error and the message is what to display to a user if that error is matched.
 	MaxRetries         int               // Maximum number of times to retry errors matching RetryableErrors
 	TimeBetweenRetries time.Duration     // The amount of time to wait between retries
+	WorkingDir         string            // The directory to run packer in
+	OutputMaxLineSize  int               // The max line size of stdout and stderr (in bytes)
 }
 
 // BuildArtifacts can take a map of identifierName <-> Options and then parallelize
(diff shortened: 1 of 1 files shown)


brikis98:
Tests failed with some errors that look transient, such as:

```
--- FAIL: TestPackerGCPBasicExample (12.09s)
    region.go:77: Get https://compute.googleapis.com/compute/v1/projects/terratest-214610/zones?alt=json&prettyPrint=false: oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: write tcp 10.128.1.23:50250->172.217.214.95:443: write: broken pipe
```

Going to re-run.

brikis98:
Sigh. One test failed with another transient error:

```
TestPackerBasicExampleWithVarFile 2020-03-03T09:55:56Z command.go:158: 1583229356,,ui,error,==> amazon-ebs: Splitting up /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_xenial_InRelease into data and signature failedE: GPG error: http://archive.ubuntu.com/ubuntu xenial InRelease: Clearsigned file isn't valid%!(PACKER_COMMA) got 'NODATA' (does the network require authentication?)
```

We're going to have to go through and add retries in a few places to make this more stable.

brikis98:
Filed https://github.com/gruntwork-io/terratest/issues/471 to track this. I don't think it has anything to do with this PR, so I'm going to merge.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.25.2

krishicks:
Thanks for the quick turnaround!

PR #472, opened 2020-03-03
Initial docs for terraform plan info feature

fajpunk:


docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+
+Sometimes you can get the information you need in your tests from a `terraform plan`; you may not need to do an `apply`.  Terratest provides a way to use certain information from a `terraform plan` in your tests.
+
+The `InitAndPlanWithInfo` function will return a `PlanInfo` struct that will contain information about which resources would be changed and what specifically about them would change.
+
+For example, let's say you have some `web_server` resources that you're creating with [`for_each`](https://www.terraform.io/docs/configuration/resources.html#for_each-multiple-resource-instances-defined-by-a-map-or-set-of-strings), and you're using the `for_each` values in the name:
+
+```hcl
+resource "aws_instance" "web_server" {
+  for_each = ["env1", "env2", "env3"]
```
yorinasub17: NIT: This is technically the wrong syntax, as resource `for_each` takes a map. This should actually be:

```
  for_each = toset(["env1", "env2", "env3"])
```

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+  name = "web_server_${each.value}"
+  ...
+}
+
+if you want to assert that the names of `web_server` resources that are definied with a terraform `for_each`, you can test this by making assertions against info in this `PlanInfo` struct:
```
bwhaley:
```suggestion
if you want to assert that the names of `web_server` resources that are defined with a terraform `for_each`, you can test this by making assertions against info in this `PlanInfo` struct:
```

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+
+```go
+plan := terraform.InitAndPlanWithInfo(t, terraformOptions)
+
+for _, rc := range plan.PlannedValues {
+    if rc.Name == "web_server" {
+        expected := fmt.Sprintf("web_server_%s", rc.Index)
+        assert.Equal(rc.Values["name"], expected)
+    }
```
yorinasub17: NIT: You would probably also want to make sure that you reach this block. E.g this test will actually pass if `plan.PlannedValues` was empty, which is probably not what you intended.

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+
```
bwhaley:
```suggestion
```

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+  Testing specific changes described in Terraform plan output
+tags: ["testing-best-practices"]
+order: 202
+nav_title: Documentation
+nav_title_link: /docs/
+---
+
+Sometimes you can get the information you need in your tests from a `terraform plan`; you may not need to do an `apply`.  Terratest provides a way to use certain information from a `terraform plan` in your tests.
+
+The `InitAndPlanWithInfo` function will return a `PlanInfo` struct that will contain information about which resources would be changed and what specifically about them would change.
```
yorinasub17: Is there a way to make this work with the output from `InitAndPlan`? Or is the new function absolutely necessary because of how you need to extract the information?

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+  ...
+}
+
+if you want to assert that the names of `web_server` resources that are definied with a terraform `for_each`, you can test this by making assertions against info in this `PlanInfo` struct:
+
+```go
+plan := terraform.InitAndPlanWithInfo(t, terraformOptions)
+
+for _, rc := range plan.PlannedValues {
+    if rc.Name == "web_server" {
```
brikis98: What else is in `rc`? E.g., Would `rc.Type` contain `aws_instance`? If so, might want to include that in the example... And the if-statement in general, for completeness. Alternatively, could use Terraform "addresses" like `rc.Address == "aws_instance.web_server"`... The latter may be important resources nested in modules.

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+  name = "web_server_${each.value}"
+  ...
+}
+
+if you want to assert that the names of `web_server` resources that are definied with a terraform `for_each`, you can test this by making assertions against info in this `PlanInfo` struct:
+
+```go
+plan := terraform.InitAndPlanWithInfo(t, terraformOptions)
+
+for _, rc := range plan.PlannedValues {
```
brikis98: NIT: What's `rc`? Resources Changed? Might be more readable to spell it out a bit more.

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+The `InitAndPlanWithInfo` function will return a `PlanInfo` struct that will contain information about which resources would be changed and what specifically about them would change.
+
+For example, let's say you have some `web_server` resources that you're creating with [`for_each`](https://www.terraform.io/docs/configuration/resources.html#for_each-multiple-resource-instances-defined-by-a-map-or-set-of-strings), and you're using the `for_each` values in the name:
+
+```hcl
+resource "aws_instance" "web_server" {
+  for_each = ["env1", "env2", "env3"]
+  name = "web_server_${each.value}"
+  ...
+}
```
brikis98: Missing "```"?

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
+---
+layout: collection-browser-doc
+title: Testing plan output
+category: testing-best-practices
+excerpt: >-
+  Testing specific changes described in Terraform plan output
+tags: ["testing-best-practices"]
+order: 202
```
yorinasub17: `202` is already used by `testing-environment.md`, so everything needs to be shifted to be correct. You don't need to change this now, but would need to be updated prior to merge.

docs/_docs/02_testing-best-practices/testing-plan-output.md
brikis98: It's probably a good idea to show the actual `plan` output here and then explain how that is parsed into a `struct`. You can then show side by side how to read certain aspects of the plan from the struct and do checks on them.

A popular example to include here may be how to read security group settings from the `plan` and ensure there are no CIDR blocks set to `0.0.0.0/0`.

docs/_docs/02_testing-best-practices/testing-plan-output.md
```diff
@@ -0,0 +1,40 @@
...
+
+if you want to assert that the names of `web_server` resources that are definied with a terraform `for_each`, you can test this by making assertions against info in this `PlanInfo` struct:
+
+```go
+plan := terraform.InitAndPlanWithInfo(t, terraformOptions)
+
+for _, rc := range plan.PlannedValues {
+    if rc.Name == "web_server" {
+        expected := fmt.Sprintf("web_server_%s", rc.Index)
+        assert.Equal(rc.Values["name"], expected)
```
brikis98: I think Terraform calls them arguments or parameters?

fajpunk: This is where it starts to get hairy.  That `Values` struct member is a direct translation from the JSON Terraform plan output, where each resource object has a `"values"` key which is an object from Terraform resource arguments -> values:

```
{
  "address": "null_resource.test[\"val1\"]",
  "mode": "managed",
  "type": "null_resource",
  "name": "test",
  "index": "val1",
  "provider_name": "null",
  "schema_version": 0,
  "values": {
    "triggers": {
      "some_attribute": "attr-val1"
    }
  }
},
```
One question is if we want to go with the terminology of the plan output format as described [here](https://www.terraform.io/docs/internals/json-format.html#values-representation), or if we want to translate it into what we think would make more sense.

My vote, at least initially, would be to stick with the plan output terminology; I think it could be a slippery slope and get more and more confusing if we start straying too far from it.  But of course I'm totally open to disagreement and implementing this in whatever way folks think is best :)

bwhaley:
I don't have much to add beyond what Yori and Jim have already mentioned, but I'm wondering about the title and description of the PR. As currently titled, it suggests that there will be more than one PR... this one is the docs, then the code is to come. Would it make sense to have a single PR for this feature with a title like "Adds terraform plan info" and a corresponding description?

docs/_docs/02_testing-best-practices/testing-plan-output.md
brikis98: Oh, I gotcha.

My guess is most people will:

1. Run `terraform plan`.
1. Write a test checking what they visually see in the `plan` output.

So my initial instinct would be to focus on that model. In fact, reading the JSON output looks a bit painful, so I'm not sure we want to force everyone to do that...

But, I don't have a good mental model of how the JSON output maps to the human-friendly output, and therefore, how much of a rabbit hole it may be to translate between them. So, use your judgment and document whatever you decide :)

Also, sanity check: how stable is that JSON output likely to be?

docs/_docs/02_testing-best-practices/testing-plan-output.md
fajpunk: Gah!  Yep, I changed it in the example file, which I haven't committed yet, but not in the docs.

docs/_docs/02_testing-best-practices/testing-plan-output.md
fajpunk: Good call, updated.  This is where I think some helper methods on the PlanInfo struct could be helful,  stuff like
```
func (p *PlanInfo) GetResourcesByName(name string)...
func (p *PlanInfo) GetUpdatedResources()...
func (p *PlanInfo) GetResourcesByType(type string)...
etc
```

docs/_docs/02_testing-best-practices/testing-plan-output.md
fajpunk: I think we do need a new function.  The only way that I can see to get the JSON representation of the plan is to save the plan output to a file with `terraform plan -o some.plan` and then run `terraform show -json some.plan` on that file.

docs/_docs/02_testing-best-practices/testing-plan-output.md
fajpunk: Yeah, I saw that, that's on my todo list.  Not sure at the moment where it would go.  I'll add a TODO comment in the file

docs/_docs/02_testing-best-practices/testing-plan-output.md
fajpunk: D'oh!  Fixed.

docs/_docs/02_testing-best-practices/testing-plan-output.md
fajpunk: Cool, I'll start down the road of calling that "values" key "attributes".  If it gets too hairy, we can correct the other way.

The JSON output is likely to be stable enough that this would still be useful work, I think:
```
Note: The output includes a format_version key, which currently has major version zero to indicate that the format is experimental and subject to change. A future version will assign a non-zero major version and make stronger promises about compatibility. We do not anticipate any significant breaking changes to the format before its first major version, however.
```
From https://www.terraform.io/docs/internals/json-format.html#values-representation

docs/_docs/02_testing-best-practices/testing-plan-output.md
fajpunk: Maybe this new function could use the current plan functions and pass an `-out` arg to the `RunTerraformComand` functions with the path to a tmp file, and then maybe there would be separate `Show` functions, and `InitAndPlanWithInfo` would call the relevant `Plan` then `Show` functions and unmarshall the result from the `Show` function.

yorinasub17: Ah I see. Given that machinery I think having the `InitAndPlanWithInfo` as a helper would be good, but eventually we can probably break out each step:

```
InitAndPlan  // Update Options so that you can have it store a plan file
planInfo := ShowPlanInfo(planfile) // takes the plan file and returns the info
```

The nice thing about the second function is that you can actually inject that into a CI/CD pipeline for production code to smoke check changes!

docs/_docs/02_testing-best-practices/testing-plan-output.md
yorinasub17: Makes sense given that gnarly json!

modules/terraform/plan_info.go
```diff
@@ -0,0 +1,75 @@
...
+
+// NewPlanInfo returns a PlanInfo struct given the json-formatted output of a terraform plan.
+func NewPlanInfo(jsonOutput string) PlanInfo {
+	var v PlanInfo
+	err := json.Unmarshal([]byte(jsonOutput), &v)
+
+	if err != nil {
+		fmt.Println("TODO: Couldn't parse json")
+		fmt.Println(err)
+	}
```
yorinasub17: NIT: In go, it is common to just return the error. When in doubt, you can just default to returning the error in most functions.

modules/terraform/plan_info.go
```diff
@@ -0,0 +1,75 @@
...
+
+// TODO: The `Index` field can be a string or an int depending on if `count` or
+// `for_each` was use.  Doesn't parse as an int right now.
+type Resource struct {
+	Module       string
+	Address      string
+	Mode         string
+	Type         string
+	Name         string
+	Index        string
```
yorinasub17: NIT: It would help to have a `json` tag on all the attributes that will be extracted, even if it is not necessary. It looks like some of these are fields that are managed by terratest (e.g `Module`), although I might be misunderstanding your code.

modules/terraform/show.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+type Format int
+
+const (
+	JSON Format = iota
+	Human
+)
```
yorinasub17: Nice use of enum! 👍

modules/terraform/plan_info.go
fajpunk: Yeah, this is sorta messy.  All of these attributes would be extracted. I can put `json` tags on them for sure.  I can also nest these, there's really no reason why anyone would need to use them un-nested (I wasn't aware until recently that I could nest structs in Go)

modules/terraform/plan_info.go
```diff
@@ -0,0 +1,88 @@
...
+			Address   string `json:"address"`
+			Resources []KnownResource
+		} `json:"child_modules"`
+	} `json:"planned_values"`
+	ChangedResources []ChangedResource `json:"resource_changes"`
+	AllResources     []KnownResource
+}
+
+// NewPlanInfo returns a PlanInfo struct given the json-formatted output of a terraform plan.
+func NewPlanInfo(jsonOutput string) (PlanInfo, error) {
```
yorinasub17: This is not a go idiom, but in terratest we have the convention of marking functions that return errors with the suffix `E`.

test/terraform_plan_details_example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	terraform.InitAndPlan(t, terraformOptions)
+
+	// Outputs a JSON-formatted version of the terraform plan file
+	output := terraform.Show(t, terraformOptions, terraform.JSON, outputPath)
+
+	// Constructs a PlanInfo datastructure with information about the plan
+	info, err := terraform.NewPlanInfo(output)
+	assert.Nil(t, err)
```
yorinasub17: We like to use `require.NoError` when checking errors. The difference between `require` and `assert` is that `require` will fail the test and _end execution_ when the condition fails, where as `assert` only marks the test as failed but will continue execution. This can lead to test panic where you might reference a nil pointer.

Additionally, the difference between `Nil` and `NoError` is that `NoError` will show the error details when there is an error.

test/terraform_plan_details_example_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+// An example of how to test information in a terraform plan.  This tests that
+// a resource's name is correctly created from a value in a `for_each`
+// expression.
+func TestTerraformPlanDetailsExample(t *testing.T) {
+	outputPath := "/tmp/terraform-plan.plan"
```
yorinasub17: NIT: You can use https://golang.org/pkg/io/ioutil/#TempFile to generate a temporary file.

modules/terraform/show.go
yorinasub17: NIT: This should probably be called `PlanJSON` or `PlanHuman`, in case we have other `JSON` in the future related to terraform (since the namespace is shared with the whole package).

modules/terraform/plan_info_test.go
```diff
@@ -0,0 +1,16 @@
+package terraform
+
+import (
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"testing"
+)
+
+func TestNewPlanInfoFlattensResources(t *testing.T) {
+	raw, err := ioutil.ReadFile("testdata/plan_output/root_and_child_modules.json")
```
yorinasub17: Don't forget to check the error here (`require.NoError(t, err)`).

modules/terraform/plan_info.go
yorinasub17: Also, we typically return pointers instead of the struct when a function needs to return a complex data structure (as opposed to a primitive), so that we can return `nil` when there is an error during execution of the function.

modules/terraform/plan_info.go
```diff
@@ -0,0 +1,88 @@
...
+			Address   string `json:"address"`
+			Resources []KnownResource
+		} `json:"root_module"`
+		ChildModules []struct {
+			Address   string `json:"address"`
+			Resources []KnownResource
+		} `json:"child_modules"`
+	} `json:"planned_values"`
+	ChangedResources []ChangedResource `json:"resource_changes"`
+	AllResources     []KnownResource
```
yorinasub17: NIT: I think `AllResources` might be more usable if this was `map[string]KnownResource`, where the index is the resource address.

Issue #400, opened 2019-10-29
Add support to run remote commands using an SSM agent


remilapeyre:
We are using AWS SSM to connect to our instances and disable SSH so it would be great to have some equivalent of `ssh.CheckSshCommand()` and others that uses SSM so that we can test them.

I will start writing those if I can find the time for that.

yorinasub17:
As Terratest has grown, we've had to think more about what types of helper functions we want in the library to make sure it doesn't become so big as to be unusable. Please check out the updated contribution guidelines, especially the part about [infrastructure and validation helpers](https://github.com/gruntwork-io/terratest/blob/master/CONTRIBUTING.md#types-of-contributions).

Specifically, to decide if we want to add something like this to terratest, we need to pass four tests:

- [Platforms] Is the function for one of the supported platforms?
- [Creating infrastructure] Is the function validating infrastructure (as opposed to creating or modifying)?
- [Complexity] Is the function complex to implement directly with the existing APIs?
- [Popularity] Is the function a popular use case?

I think this passes the litmus test for "platforms", "creating infrastructure", and "popularity", but what about "complexity"? What is it about the current SSM APIs that make it difficult to implement this directly using the existing terratest + AWS SDK APIs?

remilapeyre:
Hi @yorinasub17, the SSM API to create an SSM session gives you an URL to a websocket that you can use to send commands or get results (https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_StartSession.html#API_StartSession_ResponseElements).

How to interact with this websocket is not documented and I could not find helper functions in the Go SDK. It may be easy but I did not have time to try to do it yet, I will and come back with the results.

There is a way to get an SSH connection through AWS SSM (https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html) but this requires to have a running SSH daemon and credentials set and we disable the daemon on our infrastructure. We could keep it for the tests and remove it later but we would like to avoid testing one AMI and deploying another.

remilapeyre:
Hi @yorinasub17, I spend some time to implement a working solution:

```go
package test

import (
	"context"
	"encoding/json"
	"fmt"
	"regexp"
	"testing"
	"time"

	"github.com/gruntwork-io/terratest/modules/logger"
	"github.com/gruntwork-io/terratest/modules/retry"
	"github.com/gruntwork-io/terratest/modules/terraform"

[...]

PR #483, opened 2020-03-25
Add support to run remote commands using an SSM agent

remilapeyre:
Closes https://github.com/gruntwork-io/terratest/issues/400


modules/aws/ssm.go
```diff
@@ -68,3 +75,141 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+		if err := req.Send(); err != nil {
+			return "", err
+		}
+
+		// Remove the SSM prefix from stderr
+		stdErr := *resp.StandardErrorContent
+		prefix := regexp.MustCompile(`/var/lib/amazon/ssm/.*/_script\.sh: line 1: `)
+		stdErr = prefix.ReplaceAllString(stdErr, "")
+
+		b, err := json.Marshal(result{
```
brikis98: Please use a full variable name and not one-character versions. After scrolling down a few lines, I couldn't remember what `b` was.

brikis98: Oh, also, I'm guessing you're creating a JSON string here so you can then return it from the retry function and unmarshall it back into a struct later... The reason the retry function returns a string is that Go doesn't support generics, so it was either `interface{}` and all our code casting, or `string`, and the function then works well in a few cases, and poorly in others. I think instead of marshalling to/from JSON, in this case, you may want to declare the return value variables outside the retry loop and set them to the values you want within the loop. You can then return `"", nil` to exit the loop.

test/terraform_aws_ssm_example_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"testing"
+	"time"
+
+	"github.com/gruntwork-io/terratest/modules/aws"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+)
+
+func TestTerraformAwsSsmExample(t *testing.T) {
+	t.Parallel()
+	region := "us-east-2"
```
brikis98: We don't usually hard-code regions. Instead, we pick a random one (using a Terratest helper) to ensure the code works in multiple regions.

test/terraform_aws_ssm_example_test.go
```diff
@@ -0,0 +1,42 @@
...
+		TerraformDir: "../examples/terraform-aws-ssm-example",
+	}
+	defer terraform.Destroy(t, terraformOptions)
+
+	terraform.InitAndApply(t, terraformOptions)
+
+	instanceID := terraform.Output(t, terraformOptions, "instance_id")
+	timeout, _ := time.ParseDuration("3m")
+
+	// website::tag::1:: Wait for the instance to appear in the SSM catalog
```
brikis98: These tags are only used for examples that appear directly on the website; those are usually the "Hello, World" stye examples. Not sure we need to list this one on there?

test/terraform_aws_ssm_example_test.go
```diff
@@ -0,0 +1,42 @@
...
+	aws.WaitForSsmInstance(t, region, instanceID, timeout)
+
+	// website::tag::2:: Run a command and check its result
+	stdout, stderr := aws.CheckSsmCommand(t, region, instanceID, "echo Hello, World", timeout)
+	if stdout != "Hello, World\n" {
+		t.Fatalf("Wrong value for stdout: %q", stdout)
+	}
+	if stderr != "" {
+		t.Fatalf("Wrong value for stderr: %q", stderr)
+	}
```
brikis98: NIT: use `require.Equals`

test/terraform_aws_ssm_example_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// website::tag::3:: Run a command and get the error
+	_, _, err := aws.CheckSsmCommandE(t, region, instanceID, "false", timeout)
+	if err.Error() != "Failed" {
+		t.Fatalf("Wrong value for error: %q", err.Error())
+	}
```
brikis98: Use `require.Error` to ensure `err` is not `nil` and avoid a panic... And then `require.Equals` to check the value.

test/terraform_aws_ssm_example_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/gruntwork-io/terratest/modules/aws"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+)
+
+func TestTerraformAwsSsmExample(t *testing.T) {
```
brikis98: Thank you for adding an example and automated tests for it! 👍

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
+provider "aws" {
+  region = "us-east-2"
```
brikis98: Take the region in as an input var. You can set the `default` to `us-east-2`.

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
...
+}
+
+resource "aws_iam_role" "example" {
+  name = "example"
+
+  assume_role_policy = jsonencode({
```
brikis98: Please use the `aws_iam_policy_document` data source to format IAM policies.

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
+provider "aws" {
+  region = "us-east-2"
+}
+
+resource "aws_iam_role" "example" {
+  name = "example"
```
brikis98: Since we run tests in parallel, this name needs to include a variable to namespace it. The test code can then set that variable using `random.UniqueId` to ensure all the resources get unique names and don't clash.

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
...
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "example-ssm" {
+  role       = aws_iam_role.example.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM"
+}
+
+resource "aws_iam_instance_profile" "example" {
+  name = "example"
```
brikis98: Same comment about `name` here

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
...
+  role = aws_iam_role.example.name
+}
+
+resource "aws_instance" "example" {
+  ami                         = "ami-04100f1cdba76b497"
```
brikis98: Please use the `aws_ami` data source to find the proper AMI ID for the region. Also, give the data source a name so we know what AMI this is! There are several examples in the repo of this you can copy.

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
...
+  instance_type               = "t2.micro"
+  associate_public_ip_address = true
+  iam_instance_profile        = aws_iam_instance_profile.example.name
+
+  # website::tag::1:: When the instance boots, install the SSM agent.
+  user_data = <<-EOF
+    #!/bin/bash
+    mkdir /tmp/ssm
+    curl https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/debian_amd64/amazon-ssm-agent.deb -o /tmp/ssm/amazon-ssm-agent.deb
```
brikis98: Doesn't the Amazon Linux AMI have this pre-installed?

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
...
+    sudo dpkg -i /tmp/ssm/amazon-ssm-agent.deb
+    rm -rf /tmp/ssm
+    sudo systemctl enable amazon-ssm-agent
```
brikis98: User Data scripts run as root, so `sudo` is prob not necessary

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
...
+  EOF
+}
+
+# website::tag::2:: Output the instance's public IP address.
+output "instance_id" {
+  value = aws_instance.example.id
+}
```
brikis98: Please put output variables in `outputs.tf`.

modules/aws/ssm.go
```diff
@@ -68,3 +75,141 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+			CommandId:  resp.Command.CommandId,
+			InstanceId: &instanceID,
+		})
+		if err := req.Send(); err != nil {
+			return "", err
+		}
+
+		// Remove the SSM prefix from stderr
+		stdErr := *resp.StandardErrorContent
+		prefix := regexp.MustCompile(`/var/lib/amazon/ssm/.*/_script\.sh: line 1: `)
```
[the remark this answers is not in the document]
remilapeyre: It would either never panic or always, wouldn't it?

Since we already know it's a valid regexp pattern it should be safe to keep that way right?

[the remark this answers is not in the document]
brikis98: We want to avoid panics anywhere we can, especially in Terratest, as even a single panic could prevent clean up and leave lots of expensive resources running. E.g., I wouldn't want to kick off our test suite with this code in there now due to the possibility of a panic (same for various nil pointer dereferences I mentioned elsewhere) and even if it works now, someone could make a harmless-looking change to it in the future that causes panics.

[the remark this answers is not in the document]
remilapeyre: I had written that because in some cases SSM adds a weird prefix when an error occurs but I can't get it to do it know. Maybe it only happens when we send multiple commands at once. Since I can't write a test case for this, let's remove it for now.

I think this was the last item of your review so everything should be good now 🎉

remilapeyre:
Hi @brikis98, is this good enough to be merged?

brikis98:
Thanks for the PR! I'm super booked this week, but will try to take a look in the next few days.

examples/terraform-aws-ssm-example/variables.tf
```diff
@@ -0,0 +1,3 @@
+variable "region" {
+  type = string
```
brikis98, line 2: Please add a `description` and `default`. See also all the other `variables.tf` files for the typical comments/structure we have in these files.

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,52 @@
+provider "aws" {
```
brikis98, line 1: Please see the other `main.tf` files for the typical comment structure we use in these files to help make them more readable.

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
 
 	return ssm.New(sess), nil
 }
+
+// WaitForSsmInstanceE waits until the instance get registered to the SSM inventory.
+func WaitForSsmInstanceE(t testing.TestingT, awsRegion, instanceID string, timeout time.Duration) error {
+	timeBetweenRetries := 2 * time.Second
+	maxRetries := int(timeout.Seconds() / timeBetweenRetries.Seconds())
```
brikis98, line 80: If someone sets `timeout` to 1 second, won't this end up with a `maxRetries` of 0?

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+			Filters: []*ssm.InventoryFilter{
+				{
+					Key:    key,
+					Type:   filterType,
+					Values: values,
+				},
+			},
+		})
+
+		if err := req.Send(); err != nil {
```
brikis98, line 98: NIT: we typically call `client.GetInventory` as a one-liner, rather than creating the request and calling `Send` separately.

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+	timeBetweenRetries := 2 * time.Second
+	maxRetries := int(timeout.Seconds() / timeBetweenRetries.Seconds())
+	description := fmt.Sprintf("Waiting for %s to appear in the SSM inventory", instanceID)
+
+	key := aws.String("AWS:InstanceInformation.InstanceId")
+	filterType := aws.String("Equal")
+	values := aws.StringSlice([]string{instanceID})
+	_, err := retry.DoWithRetryE(t, description, maxRetries, timeBetweenRetries, func() (string, error) {
+		client := NewSsmClient(t, awsRegion)
+		req, resp := client.GetInventoryRequest(&ssm.GetInventoryInput{
```
brikis98, line 88: NIT: probably makes sense to declare `input` as a variable outside the retry loop, with all the filter params defined in-line, rather than defining the filter params outside with separate named variables and the `input` inside.

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+	err := WaitForSsmInstanceE(t, awsRegion, instanceID, timeout)
+	require.NoError(t, err)
+}
+
+// CheckSsmCommand checks that you can run the given command on the given instance through AWS SSM. Returns stdout and stderr.
+func CheckSsmCommand(t testing.TestingT, awsRegion, instanceID, command string, timeout time.Duration) (string, string) {
+	stdout, stderr, err := CheckSsmCommandE(t, awsRegion, instanceID, command, timeout)
+	if err != nil {
+		t.Fatalf("failed to execute '%s' on %s (%v):]\n  stdout: %#v\n  stderr: %#v", command, instanceID, err, stdout, stderr)
+	}
```
brikis98, line 123: NIT: Use `require.NoError`.

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+	return stdout, stderr
+}
+
+type result struct {
```
brikis98, line 127: Any reason not to make this a public `struct` with a name like `CommandOutput` and return that from the `CheckSsmCommand` methods (instead of two strings that force you to always look up what each one is)?

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+	Stdout string
+	Stderr string
+}
+
+// CheckSsmCommandE checks that you can run the given command on the given instance through AWS SSM. Returns the stdout, stderr and an error if one occurs.
+func CheckSsmCommandE(t testing.TestingT, awsRegion, instanceID, command string, timeout time.Duration) (string, string, error) {
+	logger.Logf(t, "Running command '%s' on EC2 instance with ID '%s'", command, instanceID)
+
+	timeBetweenRetries := 2 * time.Second
+	maxRetries := int(timeout.Seconds() / timeBetweenRetries.Seconds())
```
brikis98, line 137: Same issue here

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+	documentName := "AWS-RunShellScript"
+	req, resp := client.SendCommandRequest(&ssm.SendCommandInput{
+		Comment:      &comment,
+		DocumentName: &documentName,
+		InstanceIds:  []*string{&instanceID},
+		Parameters: map[string][]*string{
+			"commands": []*string{&command},
+		},
+	})
+	if err := req.Send(); err != nil {
```
brikis98, line 154: Same comment about calling `client.SendCommand` as a one-liner.

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+		return "", "", err
+	}
+	comment := "Terratest SSM"
+	documentName := "AWS-RunShellScript"
+	req, resp := client.SendCommandRequest(&ssm.SendCommandInput{
+		Comment:      &comment,
+		DocumentName: &documentName,
+		InstanceIds:  []*string{&instanceID},
+		Parameters: map[string][]*string{
+			"commands": []*string{&command},
```
brikis98, line 151: NIT: The AWS Go SDK has lots of helpers for turning these into pointers. E.g., `aws.String(comment)`. Typically easier to read if you use those for all these params.

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+		"bad status: InProgress": "bad status: InProgress",
+		"bad status: Delayed":    "bad status: Delayed",
+	}
+	var stdout, stderr string
+	_, err = retry.DoWithRetryableErrorsE(t, description, retryableErrors, maxRetries, timeBetweenRetries, func() (string, error) {
+		req, resp := client.GetCommandInvocationRequest(&ssm.GetCommandInvocationInput{
+			CommandId:  resp.Command.CommandId,
+			InstanceId: &instanceID,
+		})
+		if err := req.Send(); err != nil {
```
brikis98, line 172: Same comment

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+		}
+
+		// Remove the SSM prefix from stderr
+		stderr = aws.StringValue(resp.StandardErrorContent)
+
+		stdout = aws.StringValue(resp.StandardOutputContent)
+
+		status := aws.StringValue(resp.Status)
+
+		if status == "Success" {
```
brikis98, line 183: NIT: Is this defined as a constant somewhere in the AWS Go SDK? Same with `"Failed"` below?

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+			return "", nil
+		}
+
+		if status == "Failed" {
+			return "", fmt.Errorf("Failed")
```
brikis98, line 188: Is there a reason provided when there's a failure?

modules/aws/ssm.go
```diff
@@ -68,3 +73,127 @@ func NewSsmClientE(t testing.TestingT, region string) (*ssm.SSM, error) {
...
+		}
+
+		return "", fmt.Errorf("bad status: %s", status)
+	})
+
+	if err != nil {
+		return stdout, stderr, err.(retry.FatalError).Underlying
```
brikis98, line 195: Do a type check on `err` to avoid a potential panic here.

test/terraform_aws_ssm_example_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"time"
+
+	"github.com/gruntwork-io/terratest/modules/aws"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTerraformAwsSsmExample(t *testing.T) {
+	t.Parallel()
+	region := aws.GetRandomStableRegion(t, []string{}, []string{})
```
brikis98, line 14:
```suggestion
	region := aws.GetRandomStableRegion(t, nil, nil)
```

test/terraform_aws_ssm_example_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	stdout, stderr := aws.CheckSsmCommand(t, region, instanceID, "echo Hello, World", timeout)
+	require.Equal(t, stdout, "Hello, World\n")
+	require.Equal(t, stderr, "")
+
+	stdout, stderr, err := aws.CheckSsmCommandE(t, region, instanceID, "cat /wrong/file", timeout)
+	require.Error(t, err)
+	require.Equal(t, err.Error(), "Failed")
+	require.Equal(t, "cat: /wrong/file: No such file or directory\nfailed to run commands: exit status 1", stderr)
+	require.Equal(t, "", stdout)
```
brikis98, line 39: Nice tests 👍

modules/aws/ssm.go
remilapeyre: Yes, so `DoWithRetryE()` will try once and return an error if it did not succeed which is what is expected right?

modules/aws/ssm.go
remilapeyre: I'm not sure I get what you mean here

modules/aws/ssm.go
remilapeyre: Yes, at least when the command is invalid you get the message back in `stderr`

test/terraform_aws_ssm_example_test.go
remilapeyre: Thanks :)

remilapeyre:
Hi @brikis98, thanks for the review. There is just two remarks left but otherwise everything should be good now.

modules/aws/ssm.go
brikis98: Ah, right. For some reason, I had in my mind that `maxRetries` of 0 means `DoWithRetry` won't run at all, but looking at it again, I realize it will run, but won't retry, which is what we want here. So all is well.

modules/aws/ssm.go
brikis98: In your AWS SDK usage, you are doing all the API calls in two steps:

1. Create the request: e.g., `client.GetInventoryRequest(input)`.
1. Send the request: e.g., ,`req.Send()`

What I'm saying is that most of these API calls can be done in one line: e.g., `client.GetInventory(input)`.

examples/terraform-aws-ssm-example/main.tf
```diff
@@ -0,0 +1,69 @@
...
+  role       = aws_iam_role.example.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM"
+}
+
+resource "aws_iam_instance_profile" "example" {
+  name_prefix = "example"
+  role        = aws_iam_role.example.name
+}
+
+data "aws_ami" "amazon-linux-2" {
```
brikis98, line 50: NIT: the convention is to use snake_cases (with underscores) rather than dash-case (with dashes).

modules/aws/ssm.go
remilapeyre: I had no idea this was possible with this version of the AWS SDK, thanks!

brikis98:
Thanks! Kicking off tests now.

brikis98:
Tests failed, but due to an unrelated GCP issue (https://github.com/gruntwork-io/terratest/issues/536), so I'm going to merge. Thanks for the contribution!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.28.1

Issue #280, opened 2019-04-24
Test Terraform plan


bitfield:
Firstly, thanks for Terratest—it's _fantastic_.

I've already used it successfully to write some integration tests for Terraform modules. However, these are _very slow_. That's not Terratest's fault; it just takes a while to create cloud resources, wait for them to be available, and destroy them again.

I'd like to also write some unit tests that I can run very quickly; all I need to do is get the Terraform plan, and verify one or two things about it. I can run `terraform.InitAndPlan()`, and get an exit code that indicates whether or not the plan is a no-op. This is very useful, but I'd like to do more. I'd like to get the list of target resources, for example, and check some of their attributes.

Here's a simple example. If my Terraform code creates an Azure VM named 'testvm', I'd like to test that:

 - [x] The code plans without errors
 - [x] The plan is not a no-op
 - [ ] The plan contains one target
 - [ ] The target is an `azurerm_virtual_machine` named `testvm`
 - [ ] The target's tags include the string `terraform`

(I've ticked the things I can already do with Terratest.)

Now, I can call `terraform.GetExitCodeForTerraformCommandE` myself with the required arguments to save a plan file, open the plan file and parse it myself using the Terraform library, and so on. But a way to do this directly in Terratest would be a big timesaver.

For example, I'd like to be able to write something like this:

```go
terraform.InitAndPlan(t, terraformOptions)
p := terraform.GetPlan()
if len(p.Targets) != 1 {
	t.Fatalf(...)
}
if p.Targets[0] != "azurerm_virtual_machine.testvm" {
	t.Fatalf(...)
}
...
```

brikis98:
This is a great idea 👍 A PR to add this is very welcome. 

There is a gotcha: the format of `plan` output changes quite often, and the folks at HashiCorp have said that they will not be committing to a stable format in the next few releases. However, there are a few possible solutions:

1. Open source libraries: https://github.com/palantir/tfjson and https://github.com/lifeomic/terraform-plan-parser look interesting.
1. `terraform show`: Once Terraform 12 is out, the `terraform show` command will be able to parse plan files (i.e., the output of `terraform plan -out=<FILE>`) and return JSON. Read more here: https://www.terraform.io/docs/internals/json-format.html

I suspect the second option will be the way to go: run `terraform plan -out=<FILE>`, run `terraform show -json <FILE>`, parse the resulting JSON, and return it as a Go struct.

bitfield:
[...]

PR #484, opened 2020-03-26
WIP: Terraform show testing

pysysops:
Good Morning!

Hope everything is going well for you all and everyone is coping in these trying times. 

I too have the need for a speedier testing pattern as mentioned in #280 - I thought that I'd have an attempt at making `terraform show` output available to test and wanted to open a PR early in my attempt to get some feedback and collaboration on it.

Whilst this PR is WIP there will be changes to it. I'm happy to discuss decisions or better approaches to make this attempt more useful :)


pysysops:
Ah I see that there is an ongoing effort in #472 - I hadn't seen any cross reference #280 from that PR. I'll have a catch up on that PR first.

solarmosaic-kflorence:
This looks good to me, slightly cleaner than what's in #472 IMO.

yorinasub17:
Thanks for your contribution! Our team is a bit buried at the moment and this is a bit of an involved PR so we haven't had a chance to look at this, but we're hoping to take a deeper look and compare with #472 in the near future (next 1-2 weeks). Thanks for your patience!

PS: for a change like this, it would be great if you had docs or example code that highlights how you intend these functions to be used. That helps reduce the review load (and thus faster turn around) since we don't have to piece those together. See our contribution guidelines for more details, specifically the section [on readme driven development](https://terratest.gruntwork.io/docs/community/contributing/#update-the-documentation).

js-timbirkett:
Hi @yorinasub17 - Thanks for the advice, I'll take a look at adding some documentation to this PR very soon...

modules/terraform/jsonplan/plan.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// The types in here are a slightly modified version of the types in:
+// https://github.com/hashicorp/terraform/tree/master/command/jsonplan
+// Modifications at:
+// - resource.go:17
+// - resource.go:52
```
yorinasub17, line 14: If you are referencing line numbers here, you should link to the commit sha instead of `master`. These line numbers look off now (L17 is a blank line for example).

Also, it would help if there is a description of what and why it changed as well.

modules/terraform/show.go
```diff
@@ -0,0 +1,19 @@
+package terraform
+
+import (
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// Show runs terraform show with the given options and returns stdout/stderr.
+// This will fail the test if there is an error in the command.
+func Show(t testing.TestingT, options *Options) string {
```
yorinasub17, line 10: Is there a reason why this isn't returning the `Plan` struct?

modules/terraform/show_test.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"io/ioutil"
+	"net"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/files"
+	"github.com/gruntwork-io/terratest/modules/terraform/jsonplan"
+	"github.com/stretchr/testify/require"
+)
+
```
yorinasub17, line 12: Thanks for adding unit tests! However, I have a nit: can you also provide instructions on how the fixtures were generated so that in the future when we need to update them to the newer plan file format versions, it will be easy to do?

nishkrishnan:
Hi, just wanted to know when this will be completed, this would be awesome for our usecase!

js-timbirkett:
Thanks for your feedback @yorinasub17 - I'll take a look at this PR again int he next week or so and tackle some of your observations. I hope you're safe and well in these strange times 😄

OwenTuz:
Hi @js-timbirkett are you still working on this? I think this could be a great feature so I'd be keen to pick it up and see if we can get something merged if not.

modules/terraform/show.go
OwenTuz: I can't answer for @pysysops but think this makes sense in line with other functions such as `terraform.Init`?
https://github.com/gruntwork-io/terratest/blob/56fc21222eaa395af081e5d629e5b81da49098b8/modules/terraform/init.go#L10

At present the `Plan` struct gets created like this:
https://github.com/gruntwork-io/terratest/blob/a97c8ce784c19935fd696b30b6becd3def889abf/modules/terraform/show_test.go#L38-L42

...so perhaps we could answer your comment by adding some syntactic sugar to wrap calling `terraform.Show` and `jsonplan.Unmarshal`?

Personally I prefer the current method, as I think it makes it clearer that we are taking Terraform's JSON plan output and turning it into a format that is internal to Terratest.

However, it's ultimately a trade-off between writeability (with syntactic sugar) and readability (current method), so I'm happy to implement this however maintainers would prefer.

yorinasub17: This makes sense to me. Thanks for the clarifying argument in favor of returning string.

modules/terraform/jsonplan/plan.go
yorinasub17: NOTE: Unfortunately, we won't be able to use this code as this would be considered a modification of the source file, which from my understanding, under MPL 2.0 (which [terraform is licensed as](https://github.com/hashicorp/terraform/blob/master/LICENSE)) means that this source file must also be licensed under MPL 2.0.*

With that said, there is now a public library from hashicorp with this content (https://github.com/hashicorp/terraform-json), so this is no longer necessary.

*IANAL

yorinasub17:
Hi @js-timbirkett / @pysysops,

Just wanted to let you know that we have a need for this now internally, so I will be building off of this and try to complete it in the next few days. Note that I will focus on implementing the limited set of functions here, specifically the ones to enable running `terraform show` to get the plan json out, but NOT the struct as there is now [terraform-json](https://github.com/hashicorp/terraform-json) for that purpose.

Note that I will be working off a branch that contains these exact commits, so this PR should automatically close as merged when that branch gets merged (so that we recognize your contribution). Thanks!

yorinasub17:
Extension PR is now open: https://github.com/gruntwork-io/terratest/pull/694. I will keep this open so that it gets closed as merged when it merges.

PR #490, opened 2020-04-03
Also delete the delete markers whem emptying a bucket

becoded:
EmptyS3BucketE isn't deleting everything if you have some delete markers in your bucket.

Those delete markers also need to be removed in order to be able to delete the bucket.

### DIFF
--- a/modules/aws/s3_test.go
+++ b/modules/aws/s3_test.go
@@ -148,8 +148,9 @@ func TestAssertS3BucketPolicyExists(t *testing.T) {
 
 func testEmptyBucket(t *testing.T, s3Client *s3.S3, region string, s3BucketName string) {
 	expectedFileCount := rand.Intn(10000)
-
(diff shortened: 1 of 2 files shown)


modules/aws/s3_test.go
```diff
@@ -169,6 +170,17 @@ func testEmptyBucket(t *testing.T, s3Client *s3.S3, region string, s3BucketName
 			t.Fatal(err)
 		}
 
+		if i < 10 {
```
brikis98, line 173: Could you add a comment explaining what you're doing here? It's obvious in context of this PR, but won't be for a maintainer a few months from now.

becoded: Sure, done

brikis98:
Tests failed due to transient error in the APT repos that caused a Packer build failure. That's totally unrelated to this PR, so I'm going to merge anyway. Thanks for the PR!

Issue #494, opened 2020-04-07
Assert failing if you are using local configuration for testing (i.e: localstack locally)


ffernandezcast:
I'm using localstack locally to test my terraform code:

This is my provider configuration

```
provider "aws" {
  access_key                  = "mock_access_key"
  region                      = "eu-west-2"
  s3_force_path_style         = true
  secret_key                  = "mock_secret_key"
  skip_credentials_validation = true
  skip_metadata_api_check     = true
  skip_requesting_account_id  = true

  endpoints {
    apigateway     = "http://localhost:4567"
    cloudformation = "http://localhost:4581"
    cloudwatch     = "http://localhost:4582"
    dynamodb       = "http://localhost:4569"
    es             = "http://localhost:4578"
    firehose       = "http://localhost:4573"
    iam            = "http://localhost:4593"
    kinesis        = "http://localhost:4568"
    lambda         = "http://localhost:4574"
    route53        = "http://localhost:4580"
    redshift       = "http://localhost:4577"
    s3             = "http://localhost:4572"
    secretsmanager = "http://localhost:4584"
    ses            = "http://localhost:4579"
    sns            = "http://localhost:4575"
    sqs            = "http://localhost:4576"
    ssm            = "http://localhost:4583"
    stepfunctions  = "http://localhost:4585"
    sts            = "http://localhost:4592"
  }
}
```

I.E: I'm using `aws.AssertS3BucketExists` method

But I constantly receiving an error about the credentials (In my case I have fake one and receive 403). This mean the aws import dependency is not using the terraform provider configuration and continue using the default aws cli config credentials.

I have checked the code and I realized the module `https://github.com/gruntwork-io/terratest/blob/master/modules/aws/auth.go` doesn't have the option to use local custom services endpoint

jmoghisi:
PR https://github.com/gruntwork-io/terratest/pull/1211 and https://github.com/gruntwork-io/terratest/pull/1264 use custom env variables to drive the configuration of the aws client. Whilst these are great contributions, they are specific to a particular provider configuration. For example, if above HCL configuration used `assume_role` for auth or a `http_proxy` then neither PR would work since the AWS client configuration would deviate from the terraform module.

Instead should the provider HCL for the module under test be parsed to derive the client configuration? i.e. the list of endpoints to configure the AWS client with would be derived from the provider HCL configuration. It would be used to derive other client settings too such as Auth, HTTP Proxy, etc.

This approach may have some other benefits:
[...]

Issue #453, opened 2020-02-07
No valid credential sources found for AWS Provider when running terratest in github actions


211217613:
I'm trying to run terratest against localstack in github actions. When running natively on my laptop everything works fine. But when run in github actions I get the following error.

```
TestTerraformS3 2020-02-07T15:48:45Z command.go:158: rerun this command to reinitialize your working directory. If you forget, other
TestTerraformS3 2020-02-07T15:48:45Z command.go:158: commands will detect it and remind you to do so if necessary.
TestTerraformS3 2020-02-07T15:48:45Z retry.go:72: terraform [get -update]
TestTerraformS3 2020-02-07T15:48:45Z command.go:87: Running command terraform with args [get -update]
TestTerraformS3 2020-02-07T15:48:45Z command.go:158: - s3_test in ../../../s3_bucket
TestTerraformS3 2020-02-07T15:48:45Z command.go:158: - s3_test.tagging in ../../../tagging
TestTerraformS3 2020-02-07T15:48:45Z retry.go:72: terraform [apply -input=false -lock=false -auto-approve]
TestTerraformS3 2020-02-07T15:48:45Z command.go:87: Running command terraform with args [apply -input=false -lock=false -auto-approve]
TestTerraformS3 2020-02-07T15:48:48Z command.go:158: 
TestTerraformS3 2020-02-07T15:48:48Z command.go:158: Error: No valid credential sources found for AWS Provider.
TestTerraformS3 2020-02-07T15:48:48Z command.go:158: 	Please see https://terraform.io/docs/providers/aws/index.html for more information on
TestTerraformS3 2020-02-07T15:48:48Z command.go:158: 	providing credentials for the AWS Provider
TestTerraformS3 2020-02-07T15:48:48Z command.go:158: 
TestTerraformS3 2020-02-07T15:48:48Z command.go:158:   on main.tf line 4, in provider "aws":
TestTerraformS3 2020-02-07T15:48:48Z command.go:158:    4: provider "aws" {
TestTerraformS3 2020-02-07T15:48:48Z command.go:158: 
TestTerraformS3 2020-02-07T15:48:48Z command.go:158: 
TestTerraformS3 2020-02-07T15:48:48Z retry.go:80: Returning due to fatal error: FatalError{Underlying: exit status 1}
```

The terraform file under test is:
```
provider "aws" {
  region                      = "us-east-1"
  skip_credentials_validation = true
  skip_metadata_api_check     = true
[...]

PR #495, opened 2020-04-07
Add aws custom configuration support for local testing (i.e: localstack use)

ffernandezcast:
Issues related:

-  https://github.com/gruntwork-io/terratest/issues/494

- https://github.com/gruntwork-io/terratest/issues/453

The goal is be able to use Terratest assert like AssertS3BucketExists without receive credentials issues because is trying to use aws cloud and not localstack terraform provider config

i.e:

include in your test something like:

**Global test config file**
```
var LocalEndpoints = map[string]string{
	"apigateway":     "http://localhost:4567",
	"cloudformation": "http://localhost:4581",
	"cloudwatch":     "http://localhost:4582",
	"dynamodb":       "http://localhost:4569",
	"es":             "http://localhost:4578",
	"firehose":       "http://localhost:4573",
	"iam":            "http://localhost:4593",
	"kinesis":        "http://localhost:4568",
	"lambda":         "http://localhost:4574",
	"route53":        "http://localhost:4580",
	"redshift":       "http://localhost:4577",
	"s3":             "http://localhost:4572",
	"secretsmanager": "http://localhost:4584",
	"ses":            "http://localhost:4579",
	"sns":            "http://localhost:4575",
	"sqs":            "http://localhost:4576",
	"ssm":            "http://localhost:4583",
	"stepfunctions":  "http://localhost:4585",
	"sts":            "http://localhost:4592",
}
```

**Inside Global Setup Tests Suite File**
```
func TestMain(m *testing.M) {
     aws.SetAwsEndpointsOverrides(LocalEndpoints)
}
```


modules/aws/auth.go
```diff
@@ -153,3 +153,48 @@ type CredentialsError struct {
 func (err CredentialsError) Error() string {
 	return fmt.Sprintf("Error finding AWS credentials. Did you set the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables or configure an AWS profile? Underlying error: %v", err.UnderlyingErr)
 }
+
+func getNewSession(c *aws.Config) (*session.Session, error) {
+	var config *aws.Config
+	customConfig := GetCustomConfig()
+
+	if customConfig != nil {
+		config = getAwsConfigFromCustomConfig(customConfig)
```
brikis98, line 162: This is ignoring the passed in config (`c`)... So tests that set a specific region or passed in custom credentials (see the usage of `getNewSession` above) won't end up actually using them, which is going to be quite confusing. I think you'll need to merge the passed-in config and the custom config.

modules/aws/config.go
```diff
@@ -0,0 +1,82 @@
...
+	STSRegionalEndpoint endpoints.STSRegionalEndpoint
+
+	S3UsEast1RegionalEndpoint endpoints.S3UsEast1RegionalEndpoint
+}
+
+var once sync.Once
+
+var defaultConfig *Config
+
+func NewCustomConfig(c *Config) *Config {
```
brikis98, line 72: Also, consider renaming this method to `SetAwsConfigOverrides` and please add comments explaining exactly what it does and when/why you might use it.

modules/aws/config.go
```diff
@@ -0,0 +1,82 @@
...
+	"sync"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+)
+
+type RequestRetryer interface{}
+
+type Config struct {
```
brikis98, line 16: This needs a clearer name, like `AwsConfigOverrides`. Also, is there not a `structure` in the AWS SDK that already has all of these defined?

modules/aws/config.go
```diff
@@ -0,0 +1,82 @@
...
+	DisableEndpointHostPrefix *bool
+
+	STSRegionalEndpoint endpoints.STSRegionalEndpoint
+
+	S3UsEast1RegionalEndpoint endpoints.S3UsEast1RegionalEndpoint
+}
+
+var once sync.Once
+
+var defaultConfig *Config
```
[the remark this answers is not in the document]
brikis98: I assume you're doing it this way to avoid having to add extra params to all AWS methods... 

At the very least, the `NewCustomConfig` method should probably log a warning if it's called with a config value different than what was already sett...

brikis98 (commented):
Thanks for the PR!

modules/aws/config.go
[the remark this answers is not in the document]
ffernandezcast: Hi @brikis98 the main reason is this one:

> add extra params to all AWS methods


But as you mentioned I have to take care about the concurrency. I will check it out

modules/aws/config.go
ffernandezcast: Naming sometimes is hard ahahaha but I fully agree as much clear semantically guide you easily across the code

modules/aws/config.go
ffernandezcast: Yeah it is in the SDK but you know trying to avoid the new depency inside the test I think could be a good Idea

modules/aws/config.go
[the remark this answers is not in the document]
ffernandezcast: Base on this comments:
Rethinked the approach

Now I'm using a custom endpoint resolver. In the test folder I include a new local file with the endpoints I like to configure locally:

1. Let me use it also in a CI pipeline
2. Avoid dealing with global configuration and possible concurrency issues, due the main goal is just be able to use it locally or in an snapshot environment


```
var LocalEndpoints = map[string]string{
	"apigateway":     "http://localhost:4567",
	"cloudformation": "http://localhost:4581",
	"cloudwatch":     "http://localhost:4582",
	"dynamodb":       "http://localhost:4569",
	"es":             "http://localhost:4578",
	"firehose":       "http://localhost:4573",
	"iam":            "http://localhost:4593",
	"kinesis":        "http://localhost:4568",
	"lambda":         "http://localhost:4574",
	"route53":        "http://localhost:4580",
	"redshift":       "http://localhost:4577",
	"s3":             "http://localhost:4572",
	"secretsmanager": "http://localhost:4584",
	"ses":            "http://localhost:4579",
	"sns":            "http://localhost:4575",
	"sqs":            "http://localhost:4576",
	"ssm":            "http://localhost:4583",
	"stepfunctions":  "http://localhost:4585",
	"sts":            "http://localhost:4592",
}

In My Test file:

	aws.SetAwsEndpointsOverrides(&LocalEndpoints)
```

.circleci/config.yml
```diff
@@ -1,256 +1,27 @@
-defaults: &defaults
```
yorinasub17, line 1: Reminder: You'll need to revert this before we can merge this in.

I assume you were doing this to try to get the builds going in your fork on CircleCI. If that is the case, try the following config:

```
defaults: &defaults
  docker:
    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.13

version: 2.0
jobs:
  test:
    <<: *defaults
    steps:
      - checkout

      - run:
          name: run tests
          command: |
            mkdir -p /tmp/logs
            run-go-tests --path test --timeout 2h | tee /tmp/logs/all.log
          no_output_timeout: 3600s

      - run:
          command: terratest_log_parser --testlog /tmp/logs/all.log --outputdir /tmp/logs
          when: always
      - store_artifacts:
          path: /tmp/logs
      - store_test_results:
          path: /tmp/logs

workflows:
  version: 2
  build-and-test:
    jobs:
      - test:
          filters:
            tags:
              only: /^v.*/
```

modules/aws/auth.go
```diff
@@ -153,3 +154,24 @@ type CredentialsError struct {
 func (err CredentialsError) Error() string {
 	return fmt.Sprintf("Error finding AWS credentials. Did you set the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables or configure an AWS profile? Underlying error: %v", err.UnderlyingErr)
 }
+
+func getNewSession(config *aws.Config) (*session.Session, error) {
```
brikis98, line 158: Please add comments to all new functions specifying what those functions do

modules/aws/auth.go
```diff
@@ -153,3 +154,24 @@ type CredentialsError struct {
...
+
+	if GlobalCustomEndpoints != nil {
```
brikis98, line 160: Instead of making it a pointer, which could lead to panics due to `nil` pointer dereferences, perhaps this could check that the `map` isn't empty?

modules/aws/auth.go
```diff
@@ -153,3 +154,24 @@ type CredentialsError struct {
...
+		config.WithS3ForcePathStyle(true).WithEndpointResolver(endpoints.ResolverFunc(endPointsCustomResolver))
+	}
+
+	return session.NewSession(config)
+}
+
+func endPointsCustomResolver(service, region string, optionalFunctions ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
+	customServiceEndpoint := (*GlobalCustomEndpoints)[service]
+	if customServiceEndpoint != "" {
```
brikis98, line 169: Continuing the thought above, if `GlobalCustomEndpoints` is just a `map` (rather than a pointer to a map), this becomes a bit safer:

```go
customServiceEndpoint, hasEndpoint := GlobalCustomEndpoints[service]
if hasEndpoint  {
  // ...
}
```

modules/aws/config.go
```diff
@@ -0,0 +1,16 @@
+package aws
+
+import (
+	"sync"
+)
+
+var once sync.Once
+var GlobalCustomEndpoints *map[string]string
+
+func SetAwsEndpointsOverrides(endpoints *map[string]string) *map[string]string {
```
brikis98, line 10: Why return the value?

modules/aws/config.go
```diff
@@ -0,0 +1,16 @@
+package aws
+
+import (
+	"sync"
+)
+
+var once sync.Once
+var GlobalCustomEndpoints *map[string]string
```
brikis98, line 8: So, this still has the same issue of concurrency. That is, if multiple tests, running concurrently, want different endpoints, it's not really possible. That said, I think I can live with that, as _probably_ the common use case is to set it once for the entire test suite. However, this all needs to be explained clearly via comments on the function below.

brikis98 (commented):
Thanks for the updates! I suggested a few more changes. One other important thing: we need to add an automated test of some kind for this behavior.

modules/aws/config.go
ffernandezcast: In this case is not a issue at all. As you mentioned the goal is just change the endpoint to local for instance the local endpoint to S3 will not change between tests. I don't know if GO have a mechanics to do a global setup for the tests suite

.circleci/config.yml
ffernandezcast: Yeah that's the goal test the fork circleci compilation

modules/aws/config.go
ffernandezcast: Not it's an issue

modules/aws/auth.go
ffernandezcast: Sure I have to verify this one, Sorry my lack of expertise in Go. This is my first try with the language

.circleci/config.yml
ffernandezcast: image
@yorinasub17 I'm still trying to test the build but now I receiving this error I could use circleci/golang:1.13 but maybe I need configuration for **run-go-tests**

ffernandezcast: Error using circleci/golang:1.13

image

yorinasub17: Ah I thought our image was public, but it is not :-/

In any case, a lot of the scripts we use are in our closed source library. I recommend running just the tests you added locally, and once you are confident, we can kick off the build ourselves (we have a script that pulls down the current snapshot of forks into our repo so circle builds it). We can only do this if you back out the circleci config, so as soon as you roll that back, I can kick off a build.

modules/aws/config.go
ffernandezcast: I tested TestMain and work fine as expected

.circleci/config.yml
```diff
@@ -9,10 +9,13 @@ defaults: &defaults
     TERRAGRUNT_VERSION: v0.21.11
     PACKER_VERSION: 1.5.1
     GOLANG_VERSION: 1.13
-    K8S_VERSION: v1.10.0  # Same as EKS
+    K8S_VERSION: v1.15.0  # Same as EKS
+    MINIKUBE_VERSION: v1.9.2
     HELM_VERSION: v3.1.1
     KUBECONFIG: /home/circleci/.kube/config
 
+setup_minikube: &setup_minikube
```
brikis98, line 17: These changes are already in master... Do you need to rebase?

modules/aws/config.go
brikis98: > As you mentioned the goal is just change the endpoints to local, for instance the local endpoint to S3 will not change between tests

In your specific usage of this, sure. But Terratest is open source, and once we expose this as a public API, people will use it in a thousand different ways. So if this is not goroutine-safe, we need to mark that vary clearly in the API docs.

brikis98: Also, why is `GlobalCustomEndpoints` a pointer to a map rather than just a map? I guess it's because you're allowing the user to override it, but it seems safer to copy the keys/values...

brikis98: Actually, here's what we can do:

Instead of using `once.Do` and silently ignoring future calls to `SetAwsEndpointsOverrides`, why not set it the first time and then return an error if it's called more than once? That way, a user would least get notified if invalid usage (including concurrent usage).

brikis98 (commented):
Sorry for the delay; I was on vacation. Left a few more thoughts.

.circleci/config.yml
ffernandezcast: Yes

modules/aws/config.go
ffernandezcast: I will remove the pointe and once.Do.

I have additional comments about it. Maybe Imaging the scenario he have  n one test or call just a few custom endpoints:

var customLocalEndpointsOne = map[string]string{
	"apigateway":     "http://localhost:4567",
	"cloudformation": "http://localhost:4581"
}

aws.SetAwsEndpointsOverrides(customLocalEndpointsOne) 
and in a second call:

var customLocalEndpointsTwo = map[string]string{
	"route53":        "http://localhost:4580",
	"redshift":       "http://localhost:4577",
	"s3":             "http://localhost:4572",
}
aws.SetAwsEndpointsOverrides(customLocalEndpointsOne) 


what do you think about evaluate if the service already exist in the global GlobalCustomEndpoints and during the SetAwsEndpointsOverrides evaluate merge them or if already exist as you mentioned throw an exception? 

@brikis98 

Let me know your opinion about

brikis98: Well, if you have the `once.Do`, then in your example, the call to `aws.SetAwsEndpointsOverrides(customLocalEndpointsTwo)` gets silently ignored, which would be very confusing. My point is that if we only want `SetAwsEndpointsOverrides` to be called once, we should be vocal about it and return an error when it's called a second time.

pinmarva:
Hi,
i am interested on this PR. Any update?

ffernandezcast:
> Hi,
> i am interested on this PR. Any update?

You should pay for it ;-) As to your crazy micromanager :-D Bandido!

MrZablah:
Hi,

Any updates on this?

synfinatic:
This PR has been open for nearly 9mo.  Any chance of this getting merged soon-ish? @yorinasub17 & @brikis98 any chance one of you can review?

rdelpret:
Also particularly interested in this feature!

rdelpret:
@yorinasub17 @brikis98  anything I can do to help shepard this PR through? Seem's like it is more or less done?

brikis98:
I think the missing bit is some sort of automated testing for this new functionality. If someone wants to open up a new PR and wrap this up, please feel free!

ffernandezcast:
> I think the missing bit is some sort of automated testing for this new functionality. If someone wants to open up a new PR and wrap this up, please feel free!

I will try to do it(I'm facing some times constraints). But any contributions are welcome!!!

modules/aws/config.go
```diff
@@ -0,0 +1,30 @@
...
+	"log"
+	"os"
+)
+
+var customEndpoints map[string]string
+
+// SetAwsEndpointsOverrides function to override the aws endpoints globall one time
+func SetAwsEndpointsOverrides(endpoints map[string]string) {
+
+	if HasAwsCustomEndPoints() {
```
synfinatic, line 13: Is this check actually required?  Since this variable is per-process and all tests run in a single process, it means that if there is a bug in localstack or moto and you want to use native AWS for a single test, there is no good way to do that.

ffernandezcast: Yes. It is. The function just validate your are setting one time the custom endpoint. If you have a issue with a specific localstack endpoint you always could remove it from the custom endpoints array. I added some test checking this scenario

ffernandezcast:
Hi peers @brikis98 @yorinasub17 @synfinatic I added some tests. Waiting for your feedback Thanks!

modules/aws/config.go
synfinatic: Could you explain why it is required?  I'm actually confused on this point.

I understand you don't have to list a given service as a custom endpoint, but for example if I'm writing unit tests for an S3 module, and there is a bug in localstack for a single S3 API call, then all my tests would have to use AWS which defeats the whole purpose of this PR.

Alternatively, if it was re-modifiable, then it would be rather trivial to use localstack with one test suite and AWS with another suite.

ffernandezcast: @synfinatic some historical info maybe could help to clarify https://github.com/gruntwork-io/terratest/pull/495#discussion_r415615658

synfinatic: Fair enough.   As long as everyone understands that this global is for _all tests_ not just for a specific _test suite_.  The global is per-process and not per-suite (at least in the context of using `stretchr/testify/suite`)

ffernandezcast:
@brikis98 @yorinasub17 @synfinatic any additional feedback or observation?

synfinatic:
LGTM.  I would definitely squash the 94 commits though :)  Unfortunately, I'm just some rando on the internet and don't actually have any approval permissions

modules/aws/config.go
```diff
@@ -0,0 +1,30 @@
...
+		log.Println("The aws custom endpoints have been already setup, check your configuration. The global endpoint override is just allow one time")
+		os.Exit(1)
```
synfinatic, line 15: after giving this more thought, do we have to exit?  why not just return an error?  calling exit() should only be done for un-recoverable failures and this is has a simple recovery: do nothing.

synfinatic: oh, just realized people should call HasAwsCustomEndpoints() first... sure that works.

synfinatic: Since this is still open, I will point out that `os.Exit()` skips `defer` calls which is definitely not helpful.   Probably should use `log.Panicln()` instead.

rdelpret:
@brikis98 @yorinasub17 any feedback on this one?

brikis98:
Sorry folks, we are unusually busy right now, and doing our best to work through the backlog of issues/PRs. We'll get to this as soon as we can.

synfinatic:
So one thing that this PR does not support is running multiple localstack containers for testing in parallel.  This is because the variables storing the AWS Endpoints are global.  Ideally this would not be the case, but TBH this is better than nothing.  That said, I'm not sure adding this is supported without massive changes to terratest.

whummer:
This looks awesome - would also love to see this getting merged. 👍   Anything we can do to help push this forward? @ffernandezcast @brikis98 Thanks! 🙏

oblogic7:
Would be nice to get this merged.  Any reason it is sitting for so long?

tata2000:
Are there any updates on this PR? Will this PR be merged at all?

segator:
why is this not yet merged? how people test with localstack?

brikis98:
Hey folks, my apologies for the continued delay here. There are two things at play here:

1. **To properly support LocalStack, there are two areas to consider, and there are open questions with both:**
    1. Updating Terratest's own helpers to be able to talk to LocalStack. This is what this PR does, which is handy. However, it's only part of the story. Moreover, there are some open questions here to think through: e.g., the list of `localEndpoints` is already out of date, so how do we do this in a maintainable way? 
    2. Updating the code you're deploying with Terratest to talk to LocalStack. For example, if you're using Terratest's `terraform.Apply(t testing.T, opts *terraform.Options)` helper to deploy a Terraform module, it would make sense to update the `terraform.Options` struct with a `UseLocalStack` parameter, that if set, overrides the Terraform module's `provider` configuration to use LocalStack (similar to the [tflocal script](https://github.com/localstack/terraform-local)). I don't think this PR does this yet. Moreover, there are open questions here to think through as well: e.g., do we just call `tflocal` directly or re-implement its logic in Go (and how maintainable will that be)?
1. **We are very overloaded right now:** 
    1. We're a small, bootstrapped company, so we have to be very picky in where we spend our time & resources. We do our best to maintain all aspects of Terratest that we depend on (e.g., the Terraform helpers, Docker helpers, K8S helpers, etc), but we don't currently use LocalStack ourselves. So it has been hard to justify carving out our time to think through those open questions. That said, we should've done a better job of updating you all on this!
    2. Note that we are currently working on our roadmap for the next few months and using LocalStack is one of the items we're considering. I don't know if it'll make the cut in the near term, but when it does, then we'll come back and give more thought to this PR. If not, others are welcome to think through the open questions above, and we'll try to respond if we can, but please understand we're doing our best to maintain this (fairly popular) open source project with limited resources, so we have to pick our battles.

synfinatic:
@brikis98 Well, it's been 18mo since I started following this PR and honestly I no longer care at this point.  I've killed my project which was dependent on this PR and moved on to other projects.

whummer:
Hi @brikis98 , thanks for the detailed response!

> the list of localEndpoints is already out of date, so how do we do this in a maintainable way?

That's a great point - in fact, the list of services would need to be adjusted to cover the latest additions to LocalStack. The endpoint URL is the same for all services (port 4566 by default), so we can parameterize the list of services, then generate the `localEndpoints` map dynamically. 👍 

> it would make sense to update the terraform.Options struct with a UseLocalStack parameter, that if set, overrides the Terraform module's provider configuration to use LocalStack

Thanks for the pointer - that would be quite useful indeed!

> LocalStack is one of the items we're considering [...] so we have to pick our battles

That's great, glad to hear that! And yes, very much understandable that you need to pick your battles. I'll see if I or someone from the LocalStack team can help address the points you've raised, so we can jointly work on a solution for this. Thanks for your help!

---
@ffernandezcast Would you be ok if we work on this PR collaboratively, could you please give me push access to your fork/branch? Thanks!

blairham:
@ ffernandezcast Thank you for this, I've had to fork this repo and apply your patch. Works perfectly, too bad GruntWork hasn't addressed this in the 2+ years now

ffernandezcast:
> 



> 

👏 thanks @blairham .  It's  a pity after we the community members invest our time for free and also promote and evangelise  about framework use. Also a need to answer to @brikis98 because maybe there is some misunderstanding in his comments

coding-yogi:
@brikis98 , there are two different things we are talking about
1. To support terratest terraform APIs to talk to localstack 
2. To support terratest aws APIs to talk to localstack

First point is not even an issue currently as this can be easily handled from the `.tf` files with absolutely no changes to terratest APIs. I am not sure why this has to be even considered for this PR

PR is targeting point 2 where we are unable to use the `aws` module just because we cannot set an endpoint to do so.

Regarding your comment on time crunch, I can fully understand and that is why there is a community to help. When @ffernandezcast has spent considerable time on PR and even you have spent time reviewing it, I don't understand where its stuck? Reviewer's team has to spend time reviewing it, which is already being done.

This PR is fulfilling the purpose it was open for. It is not meant to do a full-fledge support for LocalStack as such. Cos this implementation can support `moto` as well which has nothing to do with `localstack`. I request you to re-look into this PR as its almost ready. This can help many of us :pray:

blairham:
> 👏 thanks @blairham . It's a pity after we the community members invest our time for free and also promote and evangelise about framework use. Also I should answer to @brikis98 comments because I feel there is some misunderstanding in his comments about the proposal

I have it working flawlessly in my own fork, it is a pitty as I opened an MR but they wanted way too much from me to merge it. I'll just continue to use my fork and maybe open it up to everyone else where the community can provide input

steve0hh:
As this is still not merged, here's a workaround if you're like me looking for a solution to use localstack with terratest.
You can override the AWS SDK endpoint URLs using environment variables since it's using the aws go sdkv2 - https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html.

```
$ export AWS_ENDPOINT_URL=http://localhost:4566
$ export AWS_ENDPOINT_URL_S3=http://s3.localhost.localstack.cloud:4566 
$ export TERRATEST_BINARY=~/localstack/bin/tflocal  
$ go test -timeout 30m
```

Note that we need to override `AWS_ENDPOINT_URL_S3` again because localstack does not support virtualhost addressing style - [source](https://github.com/localstack/localstack/issues/7652#issuecomment-1424510263)

james00012:

PR #496, opened 2020-04-07
Use large machine class

yorinasub17:
This enables the large machine class for our circleci builds, but only on the main test suite. The helm and kubernetes test suites did not benefit from using the larger instance classes, running at roughly the same speed.

Speed wise, the large instance class didn't help with the main test suite either, but it looks like it helps make the tests stable as I got two tests to pass in a row: something that we struggled to do with this repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,5 @@
 defaults: &defaults
+  resource_class: large
   machine:
     enabled: true
     image: "ubuntu-1604:201903-01"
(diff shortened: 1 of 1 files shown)


yorinasub17:
Gah! Of course when I claim stability the test destabilizes 😓 . The test failure is from the known issue with GCP.

yorinasub17:
Thanks for the reviews. Will go ahead and merge this in.

> Would we benefit from DLC? I can't remember if we build any Docker images as part of this repo's test suite.

I don't think it will make a difference in this repo as the test times are completely dwarfed by kubernetes boot time and AWS.

Issue #493, opened 2020-04-07
Underlying: bufio.Scanner: token too long when reading terraform plan as json with Azure ARM template resource


nitinmuteja:
I am using terraform azure arm templates as a resource in one module https://www.terraform.io/docs/providers/azurerm/r/template_deployment.html
When I create a plan and run the show command as json using the plan file, I get the following error:
Underlying: bufio.Scanner: token too long

TestUT_VerifyResourceNamingConventions 2020-04-07T19:41:19+05:30 command.go:158: 
2020/04/07 19:41:20 Ran terraform plan command
TestUT_VerifyResourceNamingConventions 2020-04-07T19:41:20+05:30 retry.go:72: terraform [show -json terraform.tfplan]
TestUT_VerifyResourceNamingConventions 2020-04-07T19:41:20+05:30 command.go:87: Running command terraform with args [show -json terraform.tfplan]
TestUT_VerifyResourceNamingConventions 2020-04-07T19:41:22+05:30 retry.go:80: Returning due to fatal error: FatalError{Underlying: bufio.Scanner: token too long}

V 0.26.2

nitinmuteja:
shell.command outputmaxlinesize is not set from terraform.options.outputmaxlinesize

brikis98:
Would you be up for a quick PR to fix this?

nitinmuteja:
Yes I can.

yorinasub17:
Fixed in https://github.com/gruntwork-io/terratest/releases/tag/v0.26.4

PR #497, opened 2020-04-08
#493 passed the parameter OutputMaxLineSize received from options. Re…

nitinmuteja:
…factored code to a common function

### DIFF
--- a/modules/terraform/cmd.go
+++ b/modules/terraform/cmd.go
@@ -59,7 +59,7 @@ func RunTerraformCommand(t testing.TestingT, additionalOptions *Options, args ..
 func RunTerraformCommandE(t testing.TestingT, additionalOptions *Options, additionalArgs ...string) (string, error) {
 	options, args := GetCommonOptions(additionalOptions, additionalArgs...)
 
-	cmd := generateCommand(additionalOptions, args...)
+	cmd := generateCommand(options, args...)
 	description := fmt.Sprintf("%s %v", options.TerraformBinary, args)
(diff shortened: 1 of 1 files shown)


modules/terraform/cmd.go
```diff
@@ -48,14 +59,7 @@ func RunTerraformCommand(t testing.TestingT, additionalOptions *Options, args ..
...
 
-	cmd := shell.Command{
-		Command:           options.TerraformBinary,
-		Args:              args,
-		WorkingDir:        options.TerraformDir,
-		Env:               options.EnvVars,
-		OutputMaxLineSize: options.OutputMaxLineSize,
-	}
-
+	cmd := generateCommand(additionalOptions, additionalArgs...)
```
yorinasub17, line 62: Build failed because this should actually be:
```suggestion
	cmd := generateCommand(options, args...)
```
This needs to be fixed with the other functions as well.

nitinmuteja:
corrected the function call. passed args rather than additionalargs

yorinasub17:
You need to also modify it to pass in `options` instead of `additionalOptions`, since `GetCommonOptions` modifies the options with the defaults for anything unset (e.g. `TerraformBinary`.

yorinasub17:
Thanks for making the changes! Just kicked off a build and if it passes, we can merge this in.

yorinasub17:
Thanks for the contribution! The build failed, but for a known transient failure unrelated to this PR so will go ahead and merge + release.

nitinmuteja:
thanks

Issue #487, opened 2020-03-30
Kubernetes Load Balancer type of service is not being recognized by Terratest tests


aamirganihar:
I have a test that creates a Kubernetes deployment and a service of LoadBalancer type. The test then looks for this service to execute an HTTP request. The test keeps waiting for the service to be provisioned, even though the service is up and running and then eventually fails. I executed a curl request on the service and it returned a successful response. On changing the service type from LoadBalancer to ClusterIP, leads to a success result. I had logged the same bug a couple of months ago and it had worked perfectly for version 0.23.5. I pulled the newest Terratest version (0.26.1) yesterday and suddenly this issue cropped up. Am I missing any step here? I am new to both Terratest and Go, so I am not sure if I am following the right steps. I pulled the latest Terratest version using the command `go get -d -v github.com/gruntwork-io/terratest@v0.26.1`. The content of the go.mod file is as shown below:

```
module Terratest

go 1.13

require (
        github.com/go-sql-driver/mysql v1.4.1
        github.com/gruntwork-io/terratest v0.26.1
)
```
The last time, all I had done was to run the go get command with version 0.23.5 and it had started to work.

yorinasub17:
What Kubernetes cluster are you targeting? We recently updated the logic for getting the host name of `LoadBalancer` service types in https://github.com/gruntwork-io/terratest/releases/tag/v0.24.2, and it is possible the new logic broke it for you. But the fix is unclear to me without knowledge of the Kubernetes cluster since the behavior of `LoadBalancer` is highly dependent on the platform.

aamirganihar:
I am using a single cluster setup on Minkube

yorinasub17:
Hmm that is strange, as we use minikube internally for testing and it works perfectly fine. Can you share:

- Which OS?
- What version of Minikube?
- The return value of `k8s.GetServiceEndpoint`?

aamirganihar:
OS: MacOS Catalina 10.5.1
Minikube version: 1.8.2

I have the following lines in the code:

```
service := k8s.GetService(t, options, "conduit-frontend-lb")
endpoint := k8s.GetServiceEndpoint(t, options, service, 80)
fmt.Sprintf("http://%s", endpoint)
```

Since the test case fails to find the service and the endpoint function needs this service in order to run the function, I am not being able to find the return value of the service endpoint. Below is the error message that I receive:

```
TestConduitFrontEndServiceExample: service.go:106: 
        	Error Trace:	service.go:106
[...]

PR #498, opened 2020-04-08
Address different flavors of minikube

yorinasub17:
This attempts to address https://github.com/gruntwork-io/terratest/issues/487 by switching the `IsMinikubeE` check to be based on the node labels, which is more likely to be consistent than the nodename.

### DIFF
--- a/modules/k8s/minikube.go
+++ b/modules/k8s/minikube.go
@@ -4,33 +4,35 @@ import (
 	"strings"
 
 	"github.com/gruntwork-io/terratest/modules/testing"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // IsMinikubeE returns true if the underlying kubernetes cluster is Minikube. This is determined by getting the
-// associated nodes and checking if:
-// - there is only one node
-// - the node has at least one label namespaced with "minikube.k8s.io"
+// associated nodes and checking if all nodes has at least one label namespaced with "minikube.k8s.io".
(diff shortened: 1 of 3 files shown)


modules/k8s/minikube.go
```diff
@@ -1,15 +1,36 @@
...
-// - the node is named "minikube"
+// - the node has at least one label namespaced with "minikube.k8s.io"
 func IsMinikubeE(t testing.TestingT, options *KubectlOptions) (bool, error) {
 	nodes, err := GetNodesE(t, options)
 	if err != nil {
 		return false, err
 	}
-	return len(nodes) == 1 && nodes[0].GetName() == "minikube", nil
+
+	// ASSUMPTION: Minikube always only has one node.
```
brikis98, line 19: Is that a necessary assumption to make? Or is it safer to loop over all nodes and all labels and see if any of them has the minikube label?

yorinasub17: > Or is it safer to loop over all nodes and all labels and see if any of them has the minikube label?

The safest would be to validate that all nodes have the minikube label. If any node does not have the minikube label, it will most likely be in a mixed cluster mode with minikube and non-minikube nodes, so we should assume it is not minikube.

yorinasub17: Updated: d05f5a1

yorinasub17:
Thanks for review! Going to merge and release this now.

PR #499, opened 2020-04-10
Fix for error [flag used but not declared -auto-approve]

sharma1612harshit:
Instead of using -auto-approve using -force to destroy, as I was facing an issue while running the tests in a Docker with go1.13.4 and terraform0.11.3

### DIFF
--- a/modules/terraform/destroy.go
+++ b/modules/terraform/destroy.go
@@ -1,7 +1,6 @@
 package terraform
 
 import (
-	"flag"
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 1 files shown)


yorinasub17:
Thanks for your contribution! However, `-force` for destroy is deprecated since [v0.11.4 when `-auto-approve` was introduced](https://github.com/hashicorp/terraform/blob/v0.11/CHANGELOG.md#0114-march-15-2018). This change shouldn't be needed if you bump your terraform version one patch version.

With that said, we dropped support for terraform 11 in terratest beginning with [v0.24.0](https://github.com/gruntwork-io/terratest/releases/tag/v0.24.0). Please consider doing one of the following:

- Upgrading terraform to 0.12
- Downgrading terratest to v0.23.5
- Forking terratest for tf11 compatibility

I'll close this given the reasons above.

PR #501, opened 2020-04-13
Check in config for no-response experiment

zackproser:
This config will be read by our installed [No Response bot](https://github.com/probot/no-response) so that we can evaluate its effectiveness for our workflow. 

Once merged, any issue that: 
1. has been given the label: `more-information-needed` 
2. hasn't received an update within 30 days 

will be automatically closed out with an explanation as configured in this PR. 

Note that, should the **original issue author** later respond with additional info, the issue will be re-opened automatically.

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,13 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 30
+# Label requiring a response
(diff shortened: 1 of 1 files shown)


zackproser:
Thanks for reviewing! Test failures look to be unrelated to this - so I'm going to merge now. 

If this ends up being useful - my thinking is we could slowly roll it out across more repos - and possibly expand the scope of the automation as well. Right now, I only granted no-response bot access to the terratest repo.

PR #502, opened 2020-04-14
Use ReadLine to get stable output scanning

tommyknows:
This commit switches from bufio/Scanner to bufio/Reader to read command execution's stdout and stderr. This removes the need for a `MaxLineSize` in the buffer.

The reason why this was replaced is because I was able to make a command's execution hang forever. I'm not quite sure what caused it, but the last thing I saw is:

```
TestModule/FreshInstallSubtest 2020-04-14T08:41:37Z command.go:87: Running command terraform with args [show -no-color -json /tmp/620357425/p.tfplan]
```

The call never returned, and when I had a look at the code I decided to remove the Scanner and replace it with a Reader instead. Though it may be a bit slower (https://stackoverflow.com/a/48598005), it fixes the bug and makes the API easier to use, as no MaxLineSize needs to be passed through anymore.

The output in question is coming from a `terraform show -json` call, which resulted in an output that would be one line and ~500KB. I saved that json-output to a file and tried reproducing the bug with a cut-down version of the shell-module (basically using a os.File in call to NewScanner), 
but it works fine like that. I am able to reproduce the bug with this minimal .go file here:
https://gist.github.com/tommyknows/0e0e995c03f75ac474aa135af1870e53

Sadly, I cannot share the terraform plan output. The terraform plan is ~50KB, the json version is ~500KB, but because the terraform workspace needs to be initialised and it's a plan from internal infrastructure, I cannot share it.

Signed-off-by: Ramon Rüttimann <ramon@nine.ch>


tommyknows:
Oh, right, missed it because it's behind a build tag.

Question:
The build tag is there because of this: https://github.com/gruntwork-io/terratest/pull/317
I can fix the log parser while I'm on it, should I open a separate PR for that one?

I'm wondering because if there's a runtime-dependency (meaning the test will use the go code directly from the same, checked out version) on that parser, I can remove the build tag & fix the parser in this PR. 
If the fixed parser needs to be merged before I can move the test, I'll open up a separate PR.

WDYT?

tommyknows:
opened #505, which I guess would need to be merged first in order for the build here to pass.

yorinasub17:
#505 is now merged so you can rebase on `master` and I can kick off the build.

tommyknows:
done, thank you 👍

tommyknows:
@yorinasub17 any feedback on this? Thanks 🙂

yorinasub17:
Sorry for the delay here! I just reviewed the code again and kicked off the build and it passed, so will go ahead to merge and release!

PR #504, opened 2020-04-15
Fix #500: rollback should use namespace in args

### LINKED ISSUE #500: helm rollback command is broken when specific to a namespace: Error: release: not found
terratest v0.26.3
Helm v3.1.2+gd878d4d

Doesnt take the namespace into account therefore :
Error: release: not found

I suspect you need something similar to the install/upgrade functions where the namespace is added to the args for the helm command

yorinasub17:


brikis98:
Oh, quick sanity check: should there be a test for this?

yorinasub17:
Yup there should be. Will add test at the next earliest convenience.

yorinasub17:
Finally got a chance to add a test! Test failure is in main terratest library and not helm, so going to treat that as a pass and merge.

Going to merge this in now.

Issue #223, opened 2019-01-16
New pattern to consider for testing Terraform code


brikis98:
I've been contemplating a simple wrapper method we can create to simplify the typical pattern for testing Terraform code. We add a new method to the `terraform` package with a signature similar to the following:

```go
func TestModule(t *testing.T, options *terraform.Options, validate func(outputs *terraform.Outputs))
```

When you call this method, you pass it an `options` object with the path to the Terraform module to test and the variables to set for it, and `TestModule` does the following:

1. Use `test_structure.CopyTerraformFolderToTemp` to copy your code to a temp folder so that concurrent tests against the same Terraform code don't clash.
1. Run `terraform init`.
1. Run `terraform apply`, forwarding your vars along.
1. Call your `validate` function, passing it all the outputs from your module.
1. Call `terraform destroy` at the end of the test.
1. Delete the temp folder at the end of the test.
1. Note that the `apply` and `destroy` steps would be wrapped in `test_structure.RunTestStage` with well-known names, so you can skip one or the other or both while iterating in dev mode.

This reduces a typical Terraform test to:

```go
func MyTest(t *testing.T) {
  options := &terraform.Options {
    Path: "../my-module",
    Vars: map[string]interface{
      "foo": "bar"
    },
  }

  terraform.TestModule(t, options, func(outputs *terraform.Outputs) {
    publicIp := outputs.get("public_ip") as string
    http_helper.HttpGetWithRetry(t, publicIp)
  })
}
```

yorinasub17:
Should this also call `test_structure.CopyTerraformFolderToTemp` by default? It makes debugging harder, but at the same time a lot of test isolation issues go away when we do this and I almost feel like this should be a standard pattern.

brikis98:
Yes! I meant to include that but totally forgot. Thx for the reminder!

It might also automatically create some "stages" for you so you can run only `apply` or only `destroy` if you need to. I'll update the original description.

brikis98:
BTW, now that I think about it, I wonder if we could avoid the need to copy the Terraform folder around if we: 

1. Add a `StateFileName` field to `terraform.Options`.
1. Set it to a unique name using the `UniqueId` method for each test.
1. Set `-state=<StateFileName>` whenever calling `terraform apply`, `terraform destroy`, etc.

That is, by ensuring each test gets its own state file, they can all run in the same folder without clashing.

floresj:
[...]

PR #506, opened 2020-04-16
Proposal to simplify common test cases

jfharden:
Please find documentation updates to describe a proposed solution to https://github.com/gruntwork-io/terratest/issues/223

I've tried to make it relatively flexible so we can cover most use cases without adding much complexity.

## Changes

1. I've updated the quick start guide to show a simple "real" example
2. I've updated the page on Unit and integration testing. In all honesty I wasn't sure where to put this proposal, and given it's focussed around providing a solution for both unit and integration testing (and it follows nicely on from the talk in the video on that page) this seemed like an ok place. Possibly better in it's own page with a link out to it from here instead though.

### DIFF
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -112,7 +112,7 @@ func TestWebServer(t *testing.T) {
     TerraformDir: "../web-server",
   },
 
-  terrform.UnitTest(&terraform.UnitTestPlan{
+  terrform.UnitTest(t, &terraform.UnitTestPlan{
     TerraformOptions: terraformOptions,
     Validators: []terraform.Validator{
       // A list of validation functions to test your module
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -107,10 +107,10 @@ Using Terratest, you can write an automated test that performs the exact same st
 
 ```go
 func TestWebServer(t *testing.T) {
-  terraformOptions :=  &terraform.Options {
+  terraformOptions := &terraform.Options {
     // The path to where your Terraform code is located
     TerraformDir: "../web-server",
-  },
+  }
 
   terrform.UnitTest(t, &terraform.UnitTestPlan{
     TerraformOptions: terraformOptions,
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -107,19 +107,23 @@ Using Terratest, you can write an automated test that performs the exact same st
 
 ```go
 func TestWebServer(t *testing.T) {
-  terraformOptions := &terraform.Options {
+  terraformOptions :=  &terraform.Options {
     // The path to where your Terraform code is located
     TerraformDir: "../web-server",
-  }
+  },
 
-  // At the end of the test, run `terraform destroy`
-  defer terraform.Destroy(t, terraformOptions)
+  terrform.UnitTest(&terraform.UnitTestPlan{
+    TerraformOptions: terraformOptions,
+    Validators: []terraform.Validator{
+      // A list of validation functions to test your module
+      validateWebServer,
+    },
+  })
+}
 
-  // Run `terraform init` and `terraform apply`
-  terraform.InitAndApply(t, terraformOptions)
-
-  // Run `terraform output` to get the value of an output variable
-  url := terraform.Output(t, terraformOptions, "url")
+func validateWebServer(t *testing.T, outputs *terraform.ModuleOutputs, workingDir string) {
+  // Retrieve the URL from the terraform apply output
+  url := outputs.get("url")
 
   // Verify that we get back a 200 OK with the expected text. It
   // takes ~1 min for the Instance to boot, so retry a few times.
@@ -131,8 +135,11 @@ func TestWebServer(t *testing.T) {
 }
 ```
 
-The code above does all the steps we mentioned above, including running `terraform init`, `terraform apply`, making HTTP requests to the web server (retrying up to 15 times with 5 seconds between retries), and running `terraform destroy` (using [`defer`](https://blog.golang.org/defer-panic-and-recover) to run it at the end of the test, whether the test succeeds or fails). If you put this code in a file called `web_server_test.go`, you can run it by executing `go test`, and you’ll see output that looks like this (truncated for readability):
+The UnitTest function above does all the steps we mentioned above, including running `terraform init`, `terraform apply`, calling your validator to make HTTP requests to the web server (retrying up to 15 times with 5 seconds between retries), and running `terraform destroy` (using [`defer`](https://blog.golang.org/defer-panic-and-recover) to run it at the end of the test, whether the test succeeds or fails). You can easily add more validations by providing extra validation functions to the Validators list.
 
+If you put this code in a file called `web_server_test.go`, you can run it by executing `go test`, and you’ll see output that looks like this (truncated for readability):
+
+**_TODO: The following will have to be updated with the correct output later once the UnitTest framework is done_**
 ```
 $ go test -v
 === RUN   TestWebServer
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -112,8 +112,7 @@ func TestWebServer(t *testing.T) {
     TerraformDir: "../web-server",
   }
 
-  terraform.UnitTest(t, &terraform.UnitTestPlan{
-
+  terraform.UnitTest(t, &terraform.UnitTestConfig{
     TerraformOptions: terraformOptions,
     Validators: []terraform.Validator{
       // A list of validation functions to test your module
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -112,7 +112,8 @@ func TestWebServer(t *testing.T) {
     TerraformDir: "../web-server",
   }
 
-  terrform.UnitTest(t, &terraform.UnitTestPlan{
+  terraform.UnitTest(t, &terraform.UnitTestPlan{
+
     TerraformOptions: terraformOptions,
     Validators: []terraform.Validator{
       // A list of validation functions to test your module
--- a/docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
+++ b/docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
@@ -57,14 +57,14 @@ Signature of UnitTest:
 type ModuleOutputs map[string]interface{}
 
 // A validation function to test your terraform code
-type Validator func(t *testing.T, moduleOutputs *terraform.ModuleOutputs, workingDir string)
+type Validator func(t *testing.T, moduleOutputs terraform.ModuleOutputs, workingDir string)
 
 type UnitTestPlan struct {
   TerraformOptions *terraform.Options
   // If you do not provide a working directory it will default to ./.terratest-unit-test/TEST_FUNCTION_NAME/
   WorkingDirectory string
   Setup func(t *testing.T, workingDir string)
-  TearDown func(t *testing.T, moduleOutputs *terraform.ModuleOutputs, workingDir string)
+  TearDown func(t *testing.T, moduleOutputs terraform.ModuleOutputs, workingDir string)
   Validators []Validator
 }
 
@@ -92,7 +92,7 @@ func TestWebServer(t *testing.T) {
   })
 }
 
-func validateWebServer(t *testing.T, outputs *terraform.ModuleOutputs, workingDir string) {
+func validateWebServer(t *testing.T, outputs terraform.ModuleOutputs, workingDir string) {
 }
 ```
 
@@ -117,7 +117,7 @@ func TestWebServer(t *testing.T) {
   })
 }
 
-func validateWebServer(t *testing.T, outputs *terraform.ModuleOutputs, workingDir string) {
+func validateWebServer(t *testing.T, outputs terraform.ModuleOutputs, workingDir string) {
   ...
 }
 
@@ -147,7 +147,7 @@ Signature of IntegrationTest:
 
 ```go
 // Output of every module applied, by name
-type AllModuleOutputs map[string]*ModuleOutputs
+type AllModuleOutputs map[string]ModuleOutputs
 
 type ModuleUnderTest struct {
   Name  string
@@ -157,14 +157,14 @@ type ModuleUnderTest struct {
 IntegrationTest(t *testing.T, testPlan *terraform.IntegrationTestPlan
 
 // A validation function to test your terraform code
-type IntegrationTestValidator func(t *testing.T, allModuleOutputs *terraform.AllModuleOutputs, workingDir string)
+type IntegrationTestValidator func(t *testing.T, allModuleOutputs terraform.AllModuleOutputs, workingDir string)
 
 type IntegrationTestPlan struct {
   Modules []*ModuleUnderTest
   // If you do not provide a working directory it will default to ./.terratest-integration-test/TEST_FUNCTION_NAME/
   WorkingDirectory string
   Setup func(t *testing.T, workingDir string)
-  TearDown func(t *testing.T, allModuleOutputs *terraform.AllModuleOutputs, workingDir string)
+  TearDown func(t *testing.T, allModuleOutputs terraform.AllModuleOutputs, workingDir string)
   Validators []IntegrationTestValidator
 }
 ```
--- a/docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
+++ b/docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
@@ -86,7 +86,7 @@ func TestWebServer(t *testing.T) {
     Validators: []terraform.Validator{
       // A list of validation functions to test your module
       validateWebServer,
-      valudateImageHosting,
+      validateImageHosting,
       validateSSLApplied,
     },
   })
@@ -111,7 +111,7 @@ func TestWebServer(t *testing.T) {
     Validators: []terraform.Validator{
       // A list of validation functions to test your module
       validateWebServer,
-      valudateImageHosting,
+      validateImageHosting,
       validateSSLApplied,
     },
   })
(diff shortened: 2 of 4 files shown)


docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+  TerraformOptions *terraform.Options
+}
+
+IntegrationTest(t *testing.T, testPlan *terraform.IntegrationTestPlan
+
+// A validation function to test your terraform code
+type IntegrationTestValidator func(t *testing.T, allModuleOutputs *terraform.AllModuleOutputs, workingDir string)
+
+type IntegrationTestPlan struct {
+  Modules []*ModuleUnderTest
```
jfharden, line 163: Hmm, I was thinking a bit more about this, often you want outputs from one module to setup another module, so maybe on integration tests there should be callback functions to generate the terraform options, they could take the outputs of all previously applied modules, so like:

```
ModuleUnderTest struct {
    Name string
    OptionsGenerator func(t *testing.T, allModuleOutputs *terraform.AllModuleOutputs, workingDir string) *terraform.Options
}
```

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+  // If you do not provide a working directory it will default to ./.terratest-unit-test/TEST_FUNCTION_NAME/
```
jfharden, line 164:
```suggestion
  // If you do not provide a working directory it will default to ./.terratest-integration-test/TEST_FUNCTION_NAME/
```

docs/_docs/01_getting-started/quick-start.md
```diff
@@ -112,14 +112,18 @@ func TestWebServer(t *testing.T) {
     TerraformDir: "../web-server",
   }
 
-  // At the end of the test, run `terraform destroy`
-  defer terraform.Destroy(t, terraformOptions)
-
-  // Run `terraform init` and `terraform apply`
-  terraform.InitAndApply(t, terraformOptions)
+  terrform.UnitTest(t, &terraform.UnitTestPlan{
```
brikis98, line 115:
```suggestion
  terraform.UnitTest(t, &terraform.UnitTestPlan{

```

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+```go
+// The output of the module after it has been applied, as returned by terraform.OutputAll
+type ModuleOutputs map[string]interface{}
+
+// A validation function to test your terraform code
+type Validator func(t *testing.T, moduleOutputs *terraform.ModuleOutputs, workingDir string)
+
+type UnitTestPlan struct {
+  TerraformOptions *terraform.Options
+  // If you do not provide a working directory it will default to ./.terratest-unit-test/TEST_FUNCTION_NAME/
```
brikis98, line 64: Is this more of a scratch directory?

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+Signature of UnitTest:
+
+```go
+// The output of the module after it has been applied, as returned by terraform.OutputAll
+type ModuleOutputs map[string]interface{}
+
+// A validation function to test your terraform code
+type Validator func(t *testing.T, moduleOutputs *terraform.ModuleOutputs, workingDir string)
+
+type UnitTestPlan struct {
```
brikis98, line 62: The word `Plan` already has meaning in Terraform, so this might be slightly confusing.

docs/_docs/01_getting-started/quick-start.md
```diff
@@ -131,8 +135,11 @@ func TestWebServer(t *testing.T) {
 }
 ```
 
-The code above does all the steps we mentioned above, including running `terraform init`, `terraform apply`, making HTTP requests to the web server (retrying up to 15 times with 5 seconds between retries), and running `terraform destroy` (using [`defer`](https://blog.golang.org/defer-panic-and-recover) to run it at the end of the test, whether the test succeeds or fails). If you put this code in a file called `web_server_test.go`, you can run it by executing `go test`, and you’ll see output that looks like this (truncated for readability):
+The UnitTest function above does all the steps we mentioned above, including running `terraform init`, `terraform apply`, calling your validator to make HTTP requests to the web server (retrying up to 15 times with 5 seconds between retries), and running `terraform destroy` (using [`defer`](https://blog.golang.org/defer-panic-and-recover) to run it at the end of the test, whether the test succeeds or fails). You can easily add more validations by providing extra validation functions to the Validators list.
```
brikis98, line 138: Does the unit test run `apply` or `plan`? If it runs `apply`, what's the difference with integration test then?

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+
+This example shows the simplest example with multiple Validators, with the default working directory
+```go
+func TestWebServer(t *testing.T) {
+  terraformOptions := &terraform.Options {
+    // The path to where your Terraform code is located
+    TerraformDir: "../web-server",
+  }
+
+  terrform.UnitTest(&terraform.UnitTestPlan{
```
brikis98, line 84: If you need to run `apply` on several modules, how would you compose multiple `UnitTest` calls together? Or is that where integration test comes in?

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+
+
+## Integration test helpers
+
+**_TODO: As this is a one day project I won't attempt to complete this today _**
+
+Integration tests are very similar to unit tests, only you are testing multiple modules together. To facilitate this
+terraform provides an IntegrationTest function, the main difference to the UnitTest is the IntegrationTestPlan takes a
+list of modules, each with a unique name. It passes a map of ModuleOutputs which are keyed on the unique name you gave
+the module.
```
brikis98, line 137: Ah, OK, I see how unit and integration tests compare!

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
brikis98: Was just about to bring this very thing up!

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+    // The path to where your Terraform code is located
+    TerraformDir: "../web-server",
+  }
+
+  terrform.UnitTest(&terraform.UnitTestPlan{
+    TerraformOptions: terraformOptions,
+    Validators: []terraform.Validator{
+      // A list of validation functions to test your module
+      validateWebServer,
+      valudateImageHosting,
```
marinalimeira, line 89:
```suggestion
      validateImageHosting,
```

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,149 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+  }
+
+  terrform.UnitTest(&terraform.UnitTestPlan{
+    TerraformOptions: terraformOptions,
+    WorkingDirectory "../web-server",
+    Setup: setup,
+    Validators: []terraform.Validator{
+      // A list of validation functions to test your module
+      validateWebServer,
+      valudateImageHosting,
```
marinalimeira, line 114:
```suggestion
      validateImageHosting,
```

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
marinalimeira: What should be the difference of making a integration test of a list with one module? I have the idea that unit tests are for cases where there is no side effect (mostly I/O) and in the case of terraform, we don't do since it's always calling the provider API.

brikis98: @marinalimeira Make sure to watch https://www.infoq.com/presentations/automated-testing-terraform-docker-packer/. With infra code, you're right, all testing is really integration testing... But rather than just throw the term "unit testing" away entirely, it can be useful in this context to think of testing a single module as a "unit test."

jfharden: It's a fine distinction, and in this proposal you could write any unit test as an integration test with a single module, but it requires deeper knowledge and more code, so I think there is real value in providing the more simplified interface of UnitTest since that will cover many cases and allows someone to get started with much less knowledge about the internals of terratest, and maybe less go knowledge too.

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
jfharden: Good point, I've renamed this (hopefully everywhere) to UnitTestConfig.

jfharden:
I've pushed some simple integration tests (it needs more, but this is a start)

modules/terraform/test_helpers_test.go
```diff
@@ -0,0 +1,66 @@
...
+		TerraformOptions: &Options{
+			TerraformDir: "../../examples/terraform-hello-world-example/",
+		},
+		Validators: []Validator{
+			validateOutput,
+		},
+	}
+
+	UnitTest(t, testConfig)
+}
```
brikis98, line 20: Sanity check: how does it compare with the "hello, world" example we have today (e.g., [on this page](https://terratest.gruntwork.io/examples/))? E.g.,: 

- Is it shorter? 
- Easier to understand for newbies? 
- Less error prone?

jfharden: It's definitely not shorter, nor easier to understand for a newbie to terratest, it also wouldn't teach them the very basics of what's going on under the hood.

I might argue it's less error prone since they can't forget to destroy the infrastructure they setup, and they are freely provided the outputs without having to call for them.

However although it's slightly longer and harder than the very basic hello world example the strength comes when you have more tests, more complex modules, and especially when compared with the simplest examples of using test stages (especially since it will deal with saving the outputs and reloading them if you are skipping the init_apply step.

I used the hello world example here since it was the simplest way to flex the few things I wanted to do, although the example probably shouldn't be in the quick start guide, since it doesn't teach the newbie what's going on.

I _think_ (but I haven't yet written much in the way of integration tests) that the integration test version should be especially powerful, and much simpler than orchestrating all of the interactions yourself (especially when combined with the test stages).

jfharden: I think also someone reading later reasoning about the tests will find it easier to understand what tests are running if the boilerplate includes a nice long list of tests, like

```
TerraformOptions: &Options{
  TerraformDir: "../../examples/terraform-hello-world-example/",
},
Validators: []Validator{
  validateOutput,
  validateRds,
  validateAlb,
  validateRedis,
  validateEcsTask
  validateEcsService,
  validateCloudMap
},
```

It also sets a patter for function signatures which might encourage reuse of validators (although that is 100% conjecture on my part)

brikis98: > I used the hello world example here since it was the simplest way to flex the few things I wanted to do, although the example probably shouldn't be in the quick start guide, since it doesn't teach the newbie what's going on.

There are a large number of unit and integration test examples in the `test` folder... And in other open source `gruntwork-io` repos (e.g., https://github.com/hashicorp/terraform-aws-vault, https://github.com/hashicorp/terraform-aws-consul)... Do you want to try out the pattern on something more real-world to see if it scales better to more complicated use cases?

brikis98: (And by "try out the pattern", I mean, write up a few more examples just to see what they would look like with the new API)

jfharden: Sure thing

jfharden: A real problem created here is scope.....

....with the current scheme it's easy to set up expected values, use them to initialise the vars of the module, and then reuse them later.

If we break out the validations into external functions that scoping is lost and we need to communicate those things around (test_structure.Load/Save* can help, but it's just adding more boiler plate), since it's going to automatically save the terraformOptions you could load them again with test_structure.Load/Save (which save you doing the save yourself, but still adds a lot of boiler plate of loading everything you expect in every test).

You can define the validators as anonymous functions inside the TestConfig struct and keep those expectations in scope, but doing so makes a long struct, which might be hard to read, ends up with multiple levels of extra indentation and definitely breaks the advantage I mentioned earlier with being able to easily see a nice list of validators.

There could be an extra item in the Config (something like `TestData map[string]string` which could be passed around for you, but now the function signatures are getting quite long needed 4 args for every validator, and in the integration test this would become _another_ map of module names to TestData.

brikis98: Ah, that's a good point. And without generics, you also lose type safety with those maps, so you'd have to cast everything. 

Perhaps this is a terrible idea, but I wonder if you could declare a `struct` that contains all the values you care about, and that single struct is passed to each function instead of a `map`?

jfharden: Do you mean to allow the user to define a struct to pass around, like (in their client code, not terratest code):

```
type MyStruct struct {
   Region string
   Foo string
   Bar []int
}
```

Then pass it back to them on every test so they only have to:

```
func validateFoo(t *testing.T, outputs ModuleOutputs, testData interface{}, workingDir string) {
  myStruct := testData.(*MyStruct)
  assert.Equal(t, myStruct.Foo, outputs["foo"])
}
```

Even doing this you lost the compile time type safety, the type error only surfaces when the execution occurs :S

brikis98: Yes, and, yes, we lose type safety... But at least we reduce it to one cast per method. But yea, still not great.

jfharden: I've added a dynamodb test (which is a lot of copy+paste from https://github.com/gruntwork-io/terratest/blob/master/test/terraform_aws_dynamodb_example_test.go) and refactored to use the new interface (excluding passing the struct around we just discussed, with a Save/LoadString workaround.

I'll update the docs now to reflect the struct discussion we just had.

jfharden: I've also updated the docs (for UnitTest, not Integration test yet)

jfharden: Updated the Dynamo test now to include the new test data struct

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
jfharden: No, since we are going to be automating saving and loading the terraform options (and maybe other things, like the testData) we have to use some working directory or other, we can allow the user to specify it, but so they don't have to (and so it could be easily git ignored) I chose this path. The test name comes from testing.T and means we can have a unique directory per test, so they wont collide even when run in parallel.

See the simple implementation here: https://github.com/jfharden/terratest/pull/1/files#diff-abf55a1c7dcad9df4f582c66ee96b4d8R32

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,178 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+      // A list of validation functions to test your module
+      validateWebServer,
+      validateImageHosting,
+      validateSSLApplied,
+    },
+  })
+}
+
+func validateWebServer(t *testing.T, outputs terraform.ModuleOutputs, testDataStruct interface{}, workingDir string) {
+}
```
yorinasub17, line 101: NIT: Should have a comment indicating that `validateImageHosting` and `validateSSLApplied` are additional functions defined here that will have the same signature.

docs/_docs/02_testing-best-practices/unit-integration-end-to-end-test.md
```diff
@@ -22,3 +22,178 @@ Link to the video at [infoq.com](https://www.infoq.com/presentations/automated-t
...
+    EnvVars: map[string]string{
+      "AWS_DEFAULT_REGION": testData.AwsRegion,
+    },
+  }
+
+  terrform.UnitTest(&terraform.UnitTestConfig{
+    TerraformOptions: terraformOptions,
+    WorkingDirectory "../web-server",
+    Setup: setup,
+    TestData: testData
```
yorinasub17, line 134:
```suggestion
    TestData: testData,

```

docs/_docs/01_getting-started/quick-start.md
```diff
@@ -112,14 +112,18 @@ func TestWebServer(t *testing.T) {
     TerraformDir: "../web-server",
   }
 
-  // At the end of the test, run `terraform destroy`
-  defer terraform.Destroy(t, terraformOptions)
-
-  // Run `terraform init` and `terraform apply`
-  terraform.InitAndApply(t, terraformOptions)
+  terraform.UnitTest(t, &terraform.UnitTestConfig{
+    TerraformOptions: terraformOptions,
```
yorinasub17, line 116: I think this should be a function to generate the options, that takes in the test data.

Oftentimes the terraform options will depend on random data. E.g., you might use `random.UniqueId` to seed a name variable. When iterating locally, you want to ensure that you can repeatedly run apply using the same random values. In this workflow, I would want to:

- Use `setup` to define the random values and pack it into `TestData`
- In `init_apply`, load the test data and generate the options based on that

This way, whenever I want to iterate, I would skip the setup step after the first run so that I preserve all the randomly generated data across the different test runs.

Hope I'm making sense here!

Issue #358, opened 2019-08-30
how to disable terraform's output when running test


owulveryck:
Hello and thank you for you tool.

I am using it to write some tests against my infrastructure.
I would like to inhibit the output of terraform while running `go test` and use the `output` of the functions to display what happened **only** if something fails:

```go
output, err := terraform.InitAndPlanE(&testing.T{}, terraformOptions)
if err != nil {
    t.Log(output)
    t.Fail()
}
```

If the test succeeds, a `go test` will only return 
`PASS`

Is there a way to easily achieve that?

owulveryck:
I managed to do it by tweaking the `logger` package.

The patch is:

```diff
--- a/modules/logger/logger.go
+++ b/modules/logger/logger.go
@@ -11,6 +11,13 @@ import (
        "time"
 )

+var output io.Writer = os.Stdout
+
+// Use a new writer to send the logs to
+func Use(w io.Writer) {
+       output = w
+}
+
 // Logf logs the given format and arguments, formatted using fmt.Sprintf, to stdout, along with a timestamp and information
 // about what test and file is doing the logging. This is an alternative to t.Logf that logs to stdout immediately,
 // rather than buffering all log output and only displaying it at the very end of the test. This is useful because:
@@ -29,14 +36,14 @@ import (
 // Note that there is a proposal to improve t.Logf (https://github.com/golang/go/issues/24929), but until that's
 // implemented, this method is our best bet.
 func Logf(t *testing.T, format string, args ...interface{}) {
-       DoLog(t, 2, os.Stdout, fmt.Sprintf(format, args...))
+       DoLog(t, 2, output, fmt.Sprintf(format, args...))
 }

 // Log logs the given arguments to stdout, along with a timestamp and information about what test and file is doing the
 // logging. This is an alternative to t.Logf that logs to stdout immediately, rather than buffering all log output and
 // only displaying it at the very end of the test. See the Logf method for more info.
 func Log(t *testing.T, args ...interface{}) {
-       DoLog(t, 2, os.Stdout, args...)
+       DoLog(t, 2, output, args...)
 }

 // DoLog logs the given arguments to the given writer, along with a timestamp and information about what test and file is
```

In my tests, I simply do a 

```go
logger.Use(ioutil.Discard)
```
 if I do not want the logs.

WDYT? Do you want me to submit a PR?

brikis98:
[...]

PR #510, opened 2020-04-19
Logging

tommyknows:
This commit introduces a better logging facility. It adds some types to
allow the caller to customise logging, while not introducing
API-breaking changes.

It enables the user to disable logging for terraform commands and shell
commands. The user may also supply his own Logf function (or even
multiple Logf functions) to these packages to further customise the
output.

The default Logf function is kept and has changed to use testing.T.Logf
if the Go version is at least 1.14. This is because with Go 1.14,
streaming output for Logf has been introduced. The inability to print
the output during testing was why the logger package was created in the
first place. This is fixed now, so if the logger package can determine
the Go version and that version supports streaming Logf, it will be used
instead.

Furthermore, some default loggers have been created:

- Discard:  discards all logging. Useful for functions where the output
            is too large or or unnecessary.
- TestingT: uses testing.T, if supplied (falls back to the "old" logger
            if not). This means that users with pre-1.14 versions can
            use testing.T.Logf anyway, if they want to.
- Logger:   This is the "old" logger that has been used so far. Can be
            used if, for example, the go version is >=1.14, but the
            "default" testing.T.Logf should not be used for whatvever
            reason.

This is all implemented while maintaining the old API. For users of that
API, nothing changes.

Signed-off-by: Ramon Rüttimann <me@ramonr.ch>

### DIFF
--- a/modules/docker/run.go
+++ b/modules/docker/run.go
@@ -48,8 +48,8 @@ type RunOptions struct {
 	// solely focus on the most important ones.
 	OtherOptions []string
 
-	// Set one or more loggers that should be used. See the logger package for more info.
-	Log *logger.Loggers
+	// Set a logger that should be used. See the logger package for more info.
+	Logger *logger.Logger
 }
 
 // Run runs the 'docker run' command on the given image with the given options and return stdout/stderr. This method
@@ -62,7 +62,7 @@ func Run(t testing.TestingT, image string, options *RunOptions) string {
 
 // RunE runs the 'docker run' command on the given image with the given options and return stdout/stderr, or any error.
 func RunE(t testing.TestingT, image string, options *RunOptions) (string, error) {
-	options.Log.Logf(t, "Running 'docker run' on image '%s'", image)
+	options.Logger.Logf(t, "Running 'docker run' on image '%s'", image)
 
 	args, err := formatDockerRunArgs(image, options)
 	if err != nil {
@@ -72,7 +72,7 @@ func RunE(t testing.TestingT, image string, options *RunOptions) (string, error)
 	cmd := shell.Command{
 		Command: "docker",
 		Args:    args,
-		Log:     options.Log,
+		Logger:  options.Logger,
 	}
 
 	return shell.RunCommandAndGetOutputE(t, cmd)
@@ -89,7 +89,7 @@ func RunAndGetID(t testing.TestingT, image string, options *RunOptions) string {
 // RunAndGetIDE runs the 'docker run' command on the given image with the given options and returns the container ID
 // that is returned in stdout, or any error.
 func RunAndGetIDE(t testing.TestingT, image string, options *RunOptions) (string, error) {
-	options.Log.Logf(t, "Running 'docker run' on image '%s', returning stdout", image)
+	options.Logger.Logf(t, "Running 'docker run' on image '%s', returning stdout", image)
 
 	args, err := formatDockerRunArgs(image, options)
 	if err != nil {
@@ -99,7 +99,7 @@ func RunAndGetIDE(t testing.TestingT, image string, options *RunOptions) (string
 	cmd := shell.Command{
 		Command: "docker",
 		Args:    args,
-		Log:     options.Log,
+		Logger:  options.Logger,
 	}
 
 	return shell.RunCommandAndGetStdOutE(t, cmd)
--- a/modules/docker/run.go
+++ b/modules/docker/run.go
@@ -47,6 +47,9 @@ type RunOptions struct {
 	// Terratest to not have to support every single command-line option offered by the 'docker run' command, and
 	// solely focus on the most important ones.
 	OtherOptions []string
+
+	// Set one or more loggers that should be used. See the logger package for more info.
+	Log *logger.Loggers
 }
 
 // Run runs the 'docker run' command on the given image with the given options and return stdout/stderr. This method
@@ -59,7 +62,7 @@ func Run(t testing.TestingT, image string, options *RunOptions) string {
 
 // RunE runs the 'docker run' command on the given image with the given options and return stdout/stderr, or any error.
 func RunE(t testing.TestingT, image string, options *RunOptions) (string, error) {
-	logger.Logf(t, "Running 'docker run' on image '%s'", image)
+	options.Log.Logf(t, "Running 'docker run' on image '%s'", image)
 
 	args, err := formatDockerRunArgs(image, options)
 	if err != nil {
@@ -69,6 +72,7 @@ func RunE(t testing.TestingT, image string, options *RunOptions) (string, error)
 	cmd := shell.Command{
 		Command: "docker",
 		Args:    args,
+		Log:     options.Log,
 	}
 
 	return shell.RunCommandAndGetOutputE(t, cmd)
@@ -85,7 +89,7 @@ func RunAndGetID(t testing.TestingT, image string, options *RunOptions) string {
 // RunAndGetIDE runs the 'docker run' command on the given image with the given options and returns the container ID
 // that is returned in stdout, or any error.
 func RunAndGetIDE(t testing.TestingT, image string, options *RunOptions) (string, error) {
-	logger.Logf(t, "Running 'docker run' on image '%s', returning stdout", image)
+	options.Log.Logf(t, "Running 'docker run' on image '%s', returning stdout", image)
 
 	args, err := formatDockerRunArgs(image, options)
 	if err != nil {
@@ -95,6 +99,7 @@ func RunAndGetIDE(t testing.TestingT, image string, options *RunOptions) (string
 	cmd := shell.Command{
 		Command: "docker",
 		Args:    args,
+		Log:     options.Log,
 	}
 
 	return shell.RunCommandAndGetStdOutE(t, cmd)
@@ -144,15 +149,11 @@ func formatDockerRunArgs(image string, options *RunOptions) ([]string, error) {
 		args = append(args, "--volume", volume)
 	}
 
-	for _, opt := range options.OtherOptions {
-		args = append(args, opt)
-	}
+	args = append(args, options.OtherOptions...)
 
 	args = append(args, image)
 
-	for _, arg := range options.Command {
-		args = append(args, arg)
-	}
+	args = append(args, options.Command...)
 
 	return args, nil
 }
--- a/modules/docker/stop.go
+++ b/modules/docker/stop.go
@@ -13,6 +13,9 @@ import (
 type StopOptions struct {
 	// Seconds to wait for stop before killing the container (default 10)
 	Time int
+
+	// Set one or more loggers that should be used. See the logger package for more info.
+	Log *logger.Loggers
 }
 
 // Stop runs the 'docker stop' command for the given containers and return the stdout/stderr. This method fails
@@ -25,7 +28,7 @@ func Stop(t testing.TestingT, containers []string, options *StopOptions) string 
 
 // StopE runs the 'docker stop' command for the given containers and returns any errors.
 func StopE(t testing.TestingT, containers []string, options *StopOptions) (string, error) {
-	logger.Logf(t, "Running 'docker stop' on containers '%s'", containers)
+	options.Log.Logf(t, "Running 'docker stop' on containers '%s'", containers)
 
 	args, err := formatDockerStopArgs(containers, options)
 	if err != nil {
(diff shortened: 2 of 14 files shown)


tommyknows:
PR is already based on commits from #502.

If you're not happy about the default behaviour to use Testing.T in Go 1.14, let me know.

modules/logger/logger.go
```diff
@@ -6,31 +6,125 @@ import (
...
-//
-// 3. If you have a test that takes a long time to complete, some CI systems will kill the test suite prematurely
-//    because there is no log output with t.Logf (e.g., CircleCI kills tests after 10 minutes of no log output). With
-//    this log method, you get log output continuously.
+// defaultLogf will be used if the caller uses the function
+// Logf, where on the first call to that function, a sane
+// default logging function will be set. We keep that in a
+// global variable so that we don't need to define it on
+// every call to Logf.
+var defaultLogf LogFunc
```
tommyknows, line 22: Technically, we could make this variable public so that users can override it at run time.

This would mean that even if an "intermediate" package does not support setting a logger (for example, the k8s package currently), the user could set it himself.

However, it also means that packages could override these defaults and make a mess out of logging - that's why I kept it private.

What do you think?

tommyknows:
Would also close #358

tommyknows:
One thing that might need some fine-tuning is the shell command output.

i have the following usage / issue:
I'm executing `terraform plan -out <file>` and then `terraform show <file> -json`.

The `terraform show` spoils the output completely, so I used the `logger.Discard` to discard the logging. 
However, now when that command fails, I just get a `FatalError: exit code 1` info, I don't have any clue _why_ it failed.

What I did to mitigate that is:
```diff
diff --git a/modules/shell/command.go b/modules/shell/command.go
index b4ac85c..82ee6d3 100644
--- a/modules/shell/command.go
+++ b/modules/shell/command.go
@@ -131,6 +131,11 @@ func readStdoutAndStderr(t testing.TestingT, logf logger.LogFunc, stdout, stderr
        wg := &sync.WaitGroup{}
        mutex := &sync.Mutex{}

+       stdErrLog := func(t testing.TestingT, format string, args ...interface{}) {
+               logger.TestingT(t, format, args...)
+               logf(t, format, args...)
+       }
+
        wg.Add(2)
        var stdoutErr, stderrErr error
        go func() {
@@ -139,7 +144,7 @@ func readStdoutAndStderr(t testing.TestingT, logf logger.LogFunc, stdout, stderr
        }()
        go func() {
                defer wg.Done()
-               stderrErr = readData(t, logf, stderrReader, mutex, storedStderr)
+               stderrErr = readData(t, stdErrLog, stderrReader, mutex, storedStderr)
        }()
        wg.Wait()
```

With this, StdErr is always printed, no matter which logger, at the end of the test.

This is okay for my usecase, but maybe defeats the point of providing `logger.Discard`?

I just want to discard stdout, but not stderr. 
I _could_ add an argument to RunShellCommand that's then passed through somehow, but I'm not sure if that's the cleanest solution.

tommyknows:
Alright @yorinasub17 thanks for merging the other PR! This just has been rebased and is ready for review.

modules/logger/logger.go
```diff
@@ -6,31 +6,125 @@ import (
...
+
+type Loggers []LogFunc
+type LogFunc func(t testing.TestingT, format string, args ...interface{})
```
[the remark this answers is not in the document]
tommyknows: Should the interface contain both `Log` and `Logf` functions?

tommyknows:
Thanks for your inputs @yorinasub17 !
I haven't seen the style from the base64 package before, but I really like it. I adjusted the code to use interfaces instead. 
I had to wrap that interface in a concrete type to get the "no logger is given, do sane thing" behaviour, but it is quite similar now to the base64 package. 

I also removed the last two commits from the command-output, sorry for that. Definitely should go in a separate PR.

tommyknows:
any progress on this, @yorinasub17? Or is there anything else I can do?

yorinasub17:
Apologies! I was on vacation and then have been buried in catch up on various tasks and had put this off. I should be able to review this prior to the end of the week. Thanks for your patience!

tommyknows:
No worries, thank you!

modules/docker/run.go
```diff
@@ -59,7 +62,7 @@ func Run(t testing.TestingT, image string, options *RunOptions) string {
 
 // RunE runs the 'docker run' command on the given image with the given options and return stdout/stderr, or any error.
 func RunE(t testing.TestingT, image string, options *RunOptions) (string, error) {
-	logger.Logf(t, "Running 'docker run' on image '%s'", image)
+	options.Logger.Logf(t, "Running 'docker run' on image '%s'", image)
```
yorinasub17, line 65: You need to check for `nil` here, as it will default to `nil` unless the user explicitly passes it in. This means that this code, as written, will cause a panic for users updating to this version as opposed to a compile error.

modules/docker/run.go
```diff
@@ -85,7 +89,7 @@ func RunAndGetID(t testing.TestingT, image string, options *RunOptions) string {
 // RunAndGetIDE runs the 'docker run' command on the given image with the given options and returns the container ID
 // that is returned in stdout, or any error.
 func RunAndGetIDE(t testing.TestingT, image string, options *RunOptions) (string, error) {
-	logger.Logf(t, "Running 'docker run' on image '%s', returning stdout", image)
+	options.Logger.Logf(t, "Running 'docker run' on image '%s', returning stdout", image)
```
yorinasub17, line 92: Check for nil.

modules/docker/stop.go
```diff
@@ -25,7 +28,7 @@ func Stop(t testing.TestingT, containers []string, options *StopOptions) string
 
 // StopE runs the 'docker stop' command for the given containers and returns any errors.
 func StopE(t testing.TestingT, containers []string, options *StopOptions) (string, error) {
-	logger.Logf(t, "Running 'docker stop' on containers '%s'", containers)
+	options.Logger.Logf(t, "Running 'docker stop' on containers '%s'", containers)
```
yorinasub17, line 31: Check for nil.

modules/logger/logger.go
```diff
@@ -7,36 +7,135 @@ import (
...
 
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
+var (
+	// DefaultLogf will be used if the caller uses the function Logf, where on the first call to that function, a sane
+	// default logging function will be set. We keep that in a global variable so that we don't need to define it on
+	// every call to Logf.
+	// DefaultLogf is the default logger that is used if no other logger should be provided. It uses the TerratestLogger
+	// to log messages. This can be overwritten to change the logging globally.
```
yorinasub17, line 21: I'm confused by the docs here. Perhaps reword to say something like: `Defaul`t is the default logger that is used for the `Logf` function?

modules/logger/logger.go
```diff
@@ -7,36 +7,135 @@ import (
...
+		l,
+	}
+}
+
+func (l *Logger) Logf(t testing.TestingT, format string, args ...interface{}) {
+	if tt, ok := t.(helper); ok {
+		tt.Helper()
+	}
+
+	if l == nil || l.l == nil {
```
yorinasub17, line 66: I don't think `l` will ever be `nil`, given that you can't do `nil.Logf`.
```suggestion
	if l.l == nil {
```

modules/packer/packer.go
```diff
@@ -88,7 +89,7 @@ func BuildArtifact(t testing.TestingT, options *Options) string {
 
 // BuildArtifactE builds the given Packer template and return the generated Artifact ID.
 func BuildArtifactE(t testing.TestingT, options *Options) (string, error) {
-	logger.Logf(t, "Running Packer to generate a custom artifact for template %s", options.Template)
+	options.Logger.Logf(t, "Running Packer to generate a custom artifact for template %s", options.Template)
```
yorinasub17, line 92: Check for nil

modules/shell/command.go
```diff
@@ -79,11 +82,12 @@ func RunCommandAndGetStdOutE(t testing.TestingT, command Command) (string, error
...
-// runCommandAndStoreOutputE runs a shell command and stores each line from stdout and stderr in the given
-// storedStdout and storedStderr variables, respectively. The stdout and stderr of that command will also
-// be printed to the stdout and stderr of this Go program to make debugging easier.
+// runCommandAndStoreOutputE runs a shell command and stores each line from stdout
+// and stderr in the given storedStdout and storedStderr variables, respectively.
+// Depending on the logger, the stdout and stderr of that command will also be
+// printed to the stdout and stderr of this Go program to make debugging easier.
 func runCommandAndStoreOutputE(t testing.TestingT, command Command, storedStdout *[]string, storedStderr *[]string) error {
-	logger.Logf(t, "Running command %s with args %s", command.Command, command.Args)
+	command.Logger.Logf(t, "Running command %s with args %s", command.Command, command.Args)
```
yorinasub17, line 90: Check for nil

modules/shell/command.go
```diff
@@ -165,7 +169,7 @@ func readData(t testing.TestingT, reader *bufio.Reader, mutex *sync.Mutex, allOu
 			break
 		}
 
-		logger.Log(t, line)
+		log.Logf(t, line)
```
yorinasub17, line 172: Check for nil

modules/logger/logger.go
tommyknows: Typing this on my phone, will have a look at everything tomorrow.

However, you can do this; which is also why I don‘t check for nil anywhere.
Basically, what we‘re doing is:

<pre>
var l *Logger // is now a nil pointer of type *Logger
l.Logf // actually translates to Logf(nil, ...)
</pre>

That‘s why I‘m checking for nil here, as an nil pointer can then be made useful as the default.

tommyknows: https://play.golang.org/p/isXuh90a3E1

yorinasub17: Ahhh that is neat. I didn't know that. Can you add a comment to reference this, since it's not something that is common that we do?

modules/docker/run.go
yorinasub17: As mentioned in https://github.com/gruntwork-io/terratest/pull/510#discussion_r425459142, this is being handled in the Logger type so will mark as resolved.

modules/logger/logger.go
tommyknows: Yup, creates a clean and non-dangerous API 😄 
Added a comment.

modules/logger/logger.go
tommyknows: there was also the comment still there, removed that and implemented your suggestion.

modules/logger/logger.go
```diff
@@ -7,36 +7,134 @@ import (
...
 	"runtime"
 	"strings"
+	gotesting "testing"
 	"time"
 
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
+var (
+	// DefaultLogf is the default logger that is used for the Logf function, if no one is provided. It uses the
```
yorinasub17, line 17:
```suggestion
	// Default is the default logger that is used for the Logf function, if no one is provided. It uses the
```

modules/shell/command_test.go
```diff
@@ -89,6 +90,7 @@ wait
 	cmd := Command{
 		Command: "bash",
 		Args:    []string{"-c", bashCode},
+		Logger:     logger.With(logger.Discard),
```
yorinasub17, line 93:
```suggestion
		Logger:     logger.New(logger.Discard),
```

modules/shell/command_test.go
```diff
@@ -113,6 +115,7 @@ echo
 	cmd := Command{
 		Command: "bash",
 		Args:    []string{"-c", bashCode},
+		Logger:     logger.With(logger.Discard), // don't print that line to stdout
```
yorinasub17, line 118:
```suggestion
		Logger:     logger.New(logger.Discard), // don't print that line to stdout
```

yorinasub17:
Ok I think this is ready for testing. Just a few more minor bugs, and then I'll kick off the build.

modules/shell/command_test.go
tommyknows: not quite, as `logger.Discard` is of type `*Logger, we can just use it directly even 🙂

tommyknows:
Thanks, missed those!
Fixed & ready 🙂

yorinasub17:
Ok build failed, but for a reason unrelated to this PR. Given that, I'll go forward with merging this in. Thanks for your contribution!

blame19:
Isn't modules/k8s/kubectl_options.go missing the capability to use a custom logger as well?

Issue #300, opened 2019-05-15
OutputList and OutputMap only support single level data structures


yorinasub17:
Both OutputList and OutputMap coerce nested structures into a list. E.g A terraform list of the form `[["hello"]]` is converted to `[]string{"[\"hello\"]"}` when calling `OutputList`.

solarmosaic-kflorence:
Seems like it would be nice to support parsing the output as arbitrary JSON.

PR #513, opened 2020-04-21
Add support for ListOfObjects and MapOfObjects outputs from terraform.

deberon:
I needed support for a more complex output structure than only Lists or only Maps. So I added an output function for lists of maps. I believe this could close https://github.com/gruntwork-io/terratest/issues/300 but am not sure what the requirements were for that issue.

I read the instructions at https://terratest.gruntwork.io/docs/community/contributing/ but couldn't find information on running the tests so here are the outputs from the test cases I created:

https://gist.github.com/deberon/4e32f0bea2a26c5be66d4fc47a479bad


modules/terraform/output.go
```diff
@@ -49,6 +49,42 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
 	return out, nil
 }
 
+// OutputListOfMaps calls terraform output for the given variable and returns its value as a list of maps.
+// If the output value is not a list of maps, then it fails the test.
+func OutputListOfMaps(t testing.TestingT, options *Options, key string) []map[string]string {
```
brikis98, line 54: Thanks for the PR! Once concern is that this only handles a list of maps where all the values are strings. What about maps where the values could be numbers, booleans, or other nested structures? 

I'd like to avoid having to add a never-ending list of `OutputListOfIntMaps`, `OutputListOfBoolMaps`, `OutputListOfMapMaps`, etc. functions. 

Perhaps this one should return `[]map[string]interface{}` instead?

deberon:
I modeled this function after `OutputMap` and `OutputList` which both return strings. Taking some time to think about things, I could write 2 new functions along the lines of `OutputListOfObjects` and `OutputMapOfObjects`.

`OutputMapOfObjects` would be mostly similar to `OutputMap` but with `interface` objects instead.

`OutputListOfObjects` would be my new `OutputListOfMaps` function but with the type changes you suggested.

Does this sound good? I apologize if my naming isn't appropriate, but I'm not really sure if we should refer to them as the golang types or the terraform types.

brikis98:
That sounds good, thanks!

deberon:
@brikis98 I made those proposed changes. The implementation I went with even supports recursive objects, which I also built rudimentary tests for.

modules/terraform/output.go
```diff
@@ -49,6 +50,131 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+		}
+		result = append(result, m)
+	}
+
+	return result, nil
+
+}
+
+// parseMap takes a map of interfaces and parses the types. Is also recursive for nested objects.
+func parseMap(m map[string]interface{}) (map[string]interface{}, error) {
```
brikis98, line 69: I'm a bit confused... If this is already a `map[string]interface{}` and you're returning a `map[string]interface{}`, what is changing?

modules/terraform/output.go
```diff
@@ -49,6 +50,131 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+			result[k] = nestedMap
+		case []interface{}:
+			nestedList, err := parseListOfMaps(v.([]interface{}))
+			if err != nil {
+				return nil, err
+			}
+			result[k] = nestedList
+		default:
+			// Test for int conversion
+			testInt, err := strconv.ParseInt((fmt.Sprintf("%v", v)), 10, 0)
```
brikis98, line 90: What about float? Bool?

modules/terraform/output.go
deberon: I think you're asking why I have a second `parseMap` function? If so, I have it to support recursive functions. See https://github.com/gruntwork-io/terratest/pull/513/files#diff-40b865b3d1190588582bbeaba6461217R76-R81

modules/terraform/output.go
deberon: I thought about adding support for the other data types as well, but the `OutputMap` and `OutputList` functions also only return string versions of their values. I will work on a new commit to support this right away.

deberon:
@brikis98 I pushed a new commit with support for floats and bools as well. Thanks!

modules/terraform/output.go
```diff
@@ -49,6 +50,135 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+// parseMap takes a map of interfaces and parses the types. Is also recursive for nested objects.
+func parseMap(m map[string]interface{}) (map[string]interface{}, error) {
+
+	var result map[string]interface{}
+	result = make(map[string]interface{})
+
+	for k, v := range m {
+		switch v.(type) {
+		case map[string]interface{}:
+			nestedMap, err := parseMap(v.(map[string]interface{}))
```
brikis98, line 77: NIT: you don't need to manually  cast `v` here. Instead, you can use a [type switch](https://tour.golang.org/methods/16) to have Go do it for you automatically (and more safely).

modules/terraform/output.go
```diff
@@ -49,6 +50,135 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+		case []interface{}:
+			nestedList, err := parseListOfMaps(v.([]interface{}))
+			if err != nil {
+				return nil, err
+			}
+			result[k] = nestedList
+		case bool:
+			result[k] = v
+		case float64:
+			// Test for int conversion as well. If it can be an int it will be.
```
brikis98, line 91: Will all numbers end up with type `float64`? Or will some already be of type `int`? That is, do you need another `case` statement here?

modules/terraform/output.go
brikis98: No, what I meant was this: you have this `parseMap` function that takes as input somemthing that is _already_ a `map[string]interface{}`. So what parsing, exactly, are you doing? 

I can see that within `parseMap`, you loop over the keys and values of the passed-in `map`, look at the type of each value, and if the value is:

- Already a map or list, you call `parseMap` recursively... But the type stays the same.
- Boolean: the type stays the same.
- Float: the type stays the same, unless it's an int.
- Everything else: you convert it to a string.

My question is... Do we need to do this "parsing" at all? Or could we just leave it as-is, based on what you get back from `json.Unmarshall`?

deberon: Ahh so sorry. I completely misunderstood the question. So the reason I had to go through all this is due to how [Unmarshal()](https://golang.org/pkg/encoding/json/#Unmarshal) converts all numbers to `float`s, including `int`s.

This was causing my tests to fail and I figured it wouldn't be intuitive for developers to write their tests using floats instead of ints. Although, thanks to your review I do see that I needlessly set up case statements for `bool` and `float`. I will update my PR to reflect this.

Sorry again for the misunderstanding, I'm still pretty new to go.

FYI, this is what the test looks like when I don't do the number parsing:

```
TestOutputMapOfObjects 2020-04-30T10:02:42-04:00 command.go:158: {"nest":{"five":"five","four":4},"nest_list":[{"seven":"seven","six":6}],"one":1,"somebool":true,"somefloat":1.1,"three":"three","two":"two"}
    TestOutputMapOfObjects: output_test.go:130:
                Error Trace:    output_test.go:130
                Error:          Not equal:
                                expected: map[string]interface {}{"nest":map[string]interface {}{"five":"five", "four":4}, "nest_list":[]map[string]interface {}{map[string]interface {}{"seven":"seven", "six":6}}, "one":1, "somebool":true, "somefloat":1.1, "three":"three", "two":"two"}
                                actual  : map[string]interface {}{"nest":map[string]interface {}{"five":"five", "four":4}, "nest_list":[]map[string]interface {}{map[string]interface {}{"seven":"seven", "six":6}}, "one":1, "somebool":true, "somefloat":1.1, "three":"three", "two":"two"}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -3,3 +3,3 @@
                                   (string) (len=4) "five": (string) (len=4) "five",
                                -  (string) (len=4) "four": (int) 4
                                +  (string) (len=4) "four": (float64) 4
                                  },
                                @@ -8,6 +8,6 @@
                                    (string) (len=5) "seven": (string) (len=5) "seven",
                                -   (string) (len=3) "six": (int) 6
                                +   (string) (len=3) "six": (float64) 6
                                   }
                                  },
                                - (string) (len=3) "one": (int) 1,
                                + (string) (len=3) "one": (float64) 1,
                                  (string) (len=8) "somebool": (bool) true,
                Test:           TestOutputMapOfObjects
```

modules/terraform/output.go
deberon: The `Unmarshal()` process converts all numbers to floats. So I had to do that `ParseInt` with some error handling to convert ints when I find them.

deberon:
@brikis98 sorry for the lengthy PR, but I have submitted the changes you suggested.

modules/terraform/output.go
deberon: I will definitely add comments to the new functions, thanks! I actually had an iteration that used `useNumber` along with `Decoder`. But this resulted in more code and the same end result we have here (in terms of how `int`s are handled).

In my mind, if somebody passes:

```
{
  "foo": 1,
  "bar": 1.3
}
```
I would expect `foo` to always return an int. However, with my changes, if they instead had:
```
{
  "foo": 1.0,
  "bar": 1.3
}
```
Then `foo` will still be an int (because of using ParseInt). I think this could be resolved by documentation stating a test will need to manually convert from the `int` to a `float` if that functionality is needed. Think that would be ok?

modules/terraform/output.go
```diff
@@ -49,6 +50,131 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+				return nil, err
+			}
+			result[k] = nestedMap
+		case []interface{}:
+			nestedList, err := parseListOfMaps(vt)
+			if err != nil {
+				return nil, err
+			}
+			result[k] = nestedList
+		case int, float64:
```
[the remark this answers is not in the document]
deberon: Good catch. `int` will never actually be a case here because `Unmarshal` only returns floats. I will update my code accordingly.

deberon:
@brikis98 I added commits addressing your most recent comments. Thanks!

modules/terraform/output.go
```diff
@@ -49,6 +50,143 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+
+	var result map[string]interface{}
+
+	result, err = parseMap(output)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
```
brikis98, line 148:
```suggestion
	return parseMap(output)
```

modules/terraform/output.go
```diff
@@ -49,6 +50,143 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+		newMap, err := parseMap(m)
+
+		if err != nil {
+			return nil, err
+		}
+
+		result = append(result, newMap)
+	}
+
+	return result, nil
```
brikis98, line 187:
```suggestion
	return parseListOfMaps(output)
```

modules/terraform/output.go
```diff
@@ -49,6 +50,143 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
...
+// the return value to a float.
+//
+// This function exists to map return values of the terraform outputs to intuitive
+// types. ie, if you are expecting a value of "1" you are implicitly expecting an int.
+//
+// This also allows the work to be executed recursively to support complex data types.
+func parseMap(m map[string]interface{}) (map[string]interface{}, error) {
+
+	var result map[string]interface{}
+	result = make(map[string]interface{})
```
brikis98, line 84:
```suggestion
	result := make(map[string]interface{})
```

modules/terraform/output.go
deberon: This part is actually needed to change the type from `[]interface` to `[]map[string]interface`.

brikis98: Isn't that exactly what `parseListOfMaps` does?

deberon: no

```$ go test . -timeout 30m -run TestOutputListOfObjects
# github.com/gruntwork-io/terratest/modules/terraform [github.com/gruntwork-io/terratest/modules/terraform.test]
./output.go:179:24: cannot use output (type []map[string]interface {}) as type []interface {} in argument to parseListOfMaps
FAIL    github.com/gruntwork-io/terratest/modules/terraform [build failed]
FAIL```

**Comment by solarmosaic-kflorence** [comment]
@brikis98 would it be better to work with the output as JSON values? that is actually what the output is afterall.

File: modules/terraform/output.go (hunk 8 above)
Comment (brikis98) [reply modules/terraform/output.go:178]: Gah. Go and it's lack of generics.

**Comment by brikis98** [comment]
> would it be better to work with the output as JSON values? that is actually what the output is afterall.

What do you mean?

File: modules/terraform/output.go [hunk 10]
```diff
@@ -49,6 +50,134 @@ func OutputRequiredE(t testing.TestingT, options *Options, key string) (string,
 	return out, nil
 }
 
+// parseListOfMaps takes a list of maps and parses the types.
+// It is mainly a wrapper for parseMap to support lists.
+func parseListOfMaps(l []interface{}) ([]map[string]interface{}, error) {
+	var result []map[string]interface{}
+
+	for _, v := range l {
+		m, err := parseMap(v.(map[string]interface{}))
```
brikis98, line 59: This will panic if `v` is not a `map[string]interface{}`. We do NOT want our tests to panic, ever, if we can avoid it, so please break the inner part out into a separate line:

```go
asMap, isMap = v.(map[string]interface{})
if !isMap {
  // return error
}

m, err := parseMap(asMap)
```

solarmosaic-kflorence:
@brikis98 I am new to Go, so pardon my ignorance, but it seems like it would be better for terratest to just [unmarshall](https://golang.org/pkg/encoding/json/) the terraform output into a struct which can then be asserted upon

brikis98:
Ah, so  you mean an API like:

```go
func OutputStruct(t *testing.T, options *Options, key string, out interface{})
```

Where `out` would be a pointer. So the usage of it would be something like:

```go
type Example struct {
  foo string
  bar int
}

var example Example
terraform.OutputStruct(t *testing.T, options, "example", &example)
```

If so, I think that makes sense for more complicated use cases, and would certainly welcome a separate PR to add that! That said, supporting maps and lists better out of the box, as in this PR, is useful too.

solarmosaic-kflorence:
yes, exactly like that @brikis98 -- I can open up a separate feature request issue for that.

brikis98:
Thanks!

@deberon What's the latest with this PR? Are you waiting on a review from me?

deberon:
@brikis98 I am not actually sure. I think I need that approval.

modules/terraform/output.go
brikis98: This issue still needs to be resolved.

brikis98:
@deberon OK, took one more pass. There's just one issue left, and I can kick off tests. Thanks!

deberon:
@brikis98 I made your requested changes

brikis98:
OK, all tests passed, except one, which has a known issue that causes an intermittent failure. I'm going to merge this now, thanks!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.28.2

PR #515, opened 2020-04-27
Add ApiVersions helm Option.

gregsidelinger:
Enables the use of Capabilities.APIVersions.Has in charts

Signed-off-by: Greg Sidelinger <gate@ilive4code.net>

### DIFF
--- a/modules/helm/options.go
+++ b/modules/helm/options.go
@@ -13,4 +13,5 @@ type Options struct {
 	HomePath       string              // The path to the helm home to use when calling out to helm. Empty string means use default ($HOME/.helm).
 	EnvVars        map[string]string   // Environment variables to set when running helm
 	Version        string              // Version of chart
+	ApiVersions    []string            // List of api-versions to use with the helm template command
 }
--- a/modules/helm/options.go
+++ b/modules/helm/options.go
@@ -13,5 +13,4 @@ type Options struct {
 	HomePath       string              // The path to the helm home to use when calling out to helm. Empty string means use default ($HOME/.helm).
 	EnvVars        map[string]string   // Environment variables to set when running helm
 	Version        string              // Version of chart
-	ApiVersions    []string            // List of api-versions to use with the helm template command
 }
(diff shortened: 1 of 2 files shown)


brikis98:
@yorinasub17 Could you take a look when you get a chance?

modules/helm/options.go
```diff
@@ -13,4 +13,5 @@ type Options struct {
 	HomePath       string              // The path to the helm home to use when calling out to helm. Empty string means use default ($HOME/.helm).
 	EnvVars        map[string]string   // Environment variables to set when running helm
 	Version        string              // Version of chart
+	ApiVersions    []string            // List of api-versions to use with the helm template command
```
yorinasub17, line 16: `helm.Options` is intended to capture common options across all the helm commands and I am slightly disinclined to have the precedence of packing it with command specific args. Is it possible to instead have `RenderTemplate` take in variadic string arguments at the end of the function to pass to the command as extra flags? 

E.g., here is what I had in mind:

```go
helm.RenderTemplate(
    // Existing args
    t, options, "chart", "release", []string{},
    // Additional custom args to pass to the template command. This can be anything that template supports, not just --api-versions
    "--api-versions", "helm.k8s.io/test", "--skip-crds",
)
```

gregsidelinger: Makes sense.  My only question is do you want want to use variadic string arguments or a simple array of extra args for RenderTemplate?  Here is my quick mockup.

```
func RenderTemplate(t testing.TestingT, options *Options, chartDir string, releaseName string, templateFiles []string, extraArgs ...string) string {
...
}
```

And then in RenderTemplateE  join the array?
`args = append(args, extraArgs...)`

I can update the PR shortly after I do a test.

yorinasub17: Yup how you had it is what I had in mind, so that we maintain backwards compatibility.

yorinasub17:
The test failed for an unrelated reason, so going to merge this in! Thanks for your contribution!

Issue #517, opened 2020-04-30
Terraform options does not allow to specify backend file


ArseniiPetrovich:
I want to give my users an option to select either local or remote backend. But it seems Terratest does not support setting path to the config file via `-backend-config=PATH` parameter, only key-value parameters.

yorinasub17:
Seems like a useful thing to add. Would you be up for contributing a PR to implement this? I think a good way to do this is to output just the `key` and not `key=value` when `value` is `null` in this function https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/format.go#L83

PR #518, opened 2020-05-02
Add possibility to pass backend config file

ArseniiPetrovich:
Hi, @yorinasub17, I'm super new to the GoLang, so I'm sorry in advance if this is completely wrong, but I think this should work to solve #517 issue.

Sincerely,
Arsenii Petrovich

### DIFF
--- a/modules/terraform/init_backend_file_test.go
+++ b/modules/terraform/init_backend_file_test.go
@@ -1,7 +1,6 @@
 package terraform
 
 import (
-	"io/ioutil"
 	"path/filepath"
 	"testing"
 
@@ -9,20 +8,15 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestInitBackendConfig(t *testing.T) {
+func TestInitBackendConfigFile(t *testing.T) {
 	t.Parallel()
-
-	stateDirectory, err := ioutil.TempDir("", t.Name())
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	customBackendFile := filepath.Join(stateDirectory, "terraform.backend")
 
 	testFolder, err := files.CopyTerraformFolderToTemp("../../test/fixtures/terraform-backend-file", t.Name())
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	customBackendFile := filepath.Join(testFolder, "terraform.backend")
 
 	options := &Options{
 		TerraformDir: testFolder,
@@ -33,5 +27,5 @@ func TestInitBackendConfig(t *testing.T) {
 
 	InitAndApply(t, options)
 
-	assert.FileExists(t, remoteStateFile)
+	assert.FileExists(t, customBackendFile)
 }
--- a/modules/terraform/init_backend_file_test.go
+++ b/modules/terraform/init_backend_file_test.go
@@ -17,9 +17,9 @@ func TestInitBackendConfig(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	remoteStateFile := filepath.Join(stateDirectory, "backend.tfstate")
+	customBackendFile := filepath.Join(stateDirectory, "terraform.backend")
 
-	testFolder, err := files.CopyTerraformFolderToTemp("../../test/fixtures/terraform-backend", t.Name())
+	testFolder, err := files.CopyTerraformFolderToTemp("../../test/fixtures/terraform-backend-file", t.Name())
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -27,7 +27,7 @@ func TestInitBackendConfig(t *testing.T) {
 	options := &Options{
 		TerraformDir: testFolder,
 		BackendConfig: map[string]interface{}{
-			"path": remoteStateFile,
+			customBackendFile: nil,
 		},
 	}
 
--- a/modules/terraform/init_test.go
+++ b/modules/terraform/init_test.go
@@ -27,7 +27,7 @@ func TestInitBackendConfig(t *testing.T) {
 	options := &Options{
 		TerraformDir: testFolder,
 		BackendConfig: map[string]interface{}{
-			"path": remoteStateFile,
+			remoteStateFile: nil,
 		},
 	}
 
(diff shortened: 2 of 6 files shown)


modules/terraform/init_test.go
```diff
@@ -27,7 +27,7 @@ func TestInitBackendConfig(t *testing.T) {
 	options := &Options{
 		TerraformDir: testFolder,
 		BackendConfig: map[string]interface{}{
-			"path": remoteStateFile,
+			remoteStateFile: nil,
```
yorinasub17, line 30: I think this still needs to be what it was since this is updating the `local` backend's `path` key. Can you actually create a new test that is doing something similar, but with the backend config stored in a file?

ArseniiPetrovich:
@yorinasub17 How can I test on my side that the solution is working? I get the following:
```
ubuntu@selfportal:~/terratest/modules/terraform$ go test init_backend_file_test.go 
go: downloading github.com/stretchr/testify v1.4.0
# command-line-arguments [command-line-arguments.test]
./init_backend_file_test.go:27:14: undefined: Options
./init_backend_file_test.go:34:2: undefined: InitAndApply
./init_backend_file_test.go:36:23: undefined: remoteStateFile
FAIL	command-line-arguments [build failed]
FAIL
```

ArseniiPetrovich:
However I think I have added a new test that should work... Potentially... 😅

yorinasub17:
> How can I test on my side that the solution is working? I get the following:

This is happening because you are only compiling your test file and not the entire package. Here is the command to compile the package and run just your test (in the terraform folder):

```
go test -v -run TestInitBackendConfig$ .
```

ArseniiPetrovich:
Thanks, @yorinasub17!
I've run my tests, found some bugs and fixed it.
However there is one strange bug I can't explain. Here are the logs:
```
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 retry.go:72: terraform [init -upgrade=false -backend-config=/tmp/TestInitBackendConfigFile206760874/terraform-backend-file/terraform.backend]
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:86: Running command terraform with args [init -upgrade=false -backend-config=/tmp/TestInitBackendConfigFile206760874/terraform-backend-file/terraform.backend]
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: 
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: Initializing the backend...
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: 
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: 
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: Warning: Missing backend configuration
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: 
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: -backend-config was used without a "backend" block in the configuration.
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: 
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: If you intended to override the default local backend configuration,
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: no action is required, but you may add an explicit backend block to your
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: configuration to clear this warning:
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: 
TestInitBackendConfigFile 2020-05-04T19:59:51+03:00 command.go:168: terraform {
```
The thing is that when I run `terraform init -upgrade=false -backend-config=/tmp/TestInitBackendConfigFile206760874/terraform-backend-file/terraform.backend`  from CLI it shows no warning to me. What might be the reason for that?

ArseniiPetrovich:
Don't work for some reason. I'll create an issue at Terraform repo.

test/fixtures/terraform-backend-file/terraform.backend
```diff
@@ -0,0 +1,4 @@
+terraform {
+  backend "local" {}
+}
```
yorinasub17, line 3: Reading the docs, I think this is supposed to be the contents of the `config` block. See the comment about backend files in https://www.terraform.io/docs/backends/config.html#partial-configuration, specifically

> A backend configuration file has the contents of the backend block as top-level attributes, without the need to wrap it in another terraform or backend block:

modules/terraform/init_backend_file_test.go
```diff
@@ -0,0 +1,31 @@
...
+	options := &Options{
+		TerraformDir: testFolder,
+		BackendConfig: map[string]interface{}{
+			customBackendFile: nil,
+		},
+	}
+
+	InitAndApply(t, options)
+
+	assert.FileExists(t, customBackendFile)
```
yorinasub17, line 30: This is actually testing if `CopyTerraformFolderToTemp` works, instead of the functionality of the backend configuration file. Can you:

- Update the configuration file to specify a different path for the local terraform state
- Verify that the configuration was successfully overridden by checking for the existence of the custom terraform state file

ArseniiPetrovich: Yeaah... Okay, I got it. Yes, will do.

test/fixtures/terraform-backend-file/terraform.backend
yorinasub17: Ahhh yes you can only configure the backend specified in the module using `--backend-config`. You can't dynamically pick and choose the actual backend. To do that, you will need something like [terragrunt](https://terragrunt.gruntwork.io/) which can generate the remote state config dynamically before `terraform` is called. You can do this with a `generate` block that adjusts content based on an environment variable (controlled with `get_env`).

Note that `terratest` supports calling `terragrunt` by setting the `TerraformBinary` field in `terraform.Options`.

---

I still think this specific PR would be a useful addition to have.

tonyszhang:
@ArseniiPetrovich why was this close?

PR #520, opened 2020-05-14
Add function for lowercase uniqueId

tomasbackman:
First wanted to just add arguments to existing `UniqueId` function, but since go does not support default values I decided to create a new function instead.


yorinasub17:
Thanks for your contribution! I'm not sure we would accept this as is though given that it is just as easy and straightforward to get lowercase unique id with `strings.ToLower(random.UniqueId())` (which is what we do at Gruntwork).

We also have https://github.com/gruntwork-io/gruntwork-cli/blob/master/random/random.go#L18 in `gruntwork-io/gruntwork-cli` for more complex unique id generation.

I think we are sufficiently covered between these approaches, so will be closing this PR. Please let us know if they don't work for you!

tomasbackman:
Ah of course, both options makes much more sense. =)
Thanks/sorry for taking your time.

PR #521, opened 2020-05-16
Improve shell command output handling

tommyknows:
Changes in `shell` package:
While there are many public functions for grabbing a command's mixed
output or stdout output only, there is no way to specifically get Stderr
instead. Stderr is needed to debug issues, which are (almost) always
accompanied by an error (non-0 exit status). This commit changes the
handling of stdout and stderr to make it possible to return a "merged"
string (stdout+stderr) or stdout only. As we can now grab stderr
separately, we can also add it to the returned error, if any.

This means that every error that is returned by public functions in this
package is now of type `ErrWithCmdOutput`, which makes it possible for the
callee to explicitely get the stderr stream, if needed, or just print
the whole error including stderr.
It also means that per default, an error will not just be
"FatalError{Underlying: exit status 1}", but instead contain the whole
Stderr stream. The underlying error is encapsulated in the new error
type.

Signed-off-by: Ramon Rüttimann <me@ramonr.ch>

### DIFF
--- a/modules/shell/command.go
+++ b/modules/shell/command.go
@@ -11,10 +11,9 @@ import (
 	"sync"
 	"syscall"
 
-	"github.com/stretchr/testify/require"
-
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
 )
 
 // Command is a simpler struct for defining commands than Go's built-in Cmd.
@@ -27,44 +26,46 @@ type Command struct {
 	Logger *logger.Logger
 }
 
-// RunCommand runs a shell command and redirects its stdout and stderr to the stdout of the atomic script itself.
+// RunCommand runs a shell command and redirects its stdout and stderr to the stdout of the atomic script itself. If
+// there are any errors, fail the test.
 func RunCommand(t testing.TestingT, command Command) {
 	err := RunCommandE(t, command)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 }
 
-// RunCommandE runs a shell command and redirects its stdout and stderr to the stdout of the atomic script itself.
+// RunCommandE runs a shell command and redirects its stdout and stderr to the stdout of the atomic script itself. Any
+// returned error will be of type ErrWithCmdOutput, containing the output streams and the underlying error.
 func RunCommandE(t testing.TestingT, command Command) error {
-	_, err := RunCommandAndGetOutputE(t, command)
-	return err
+	output, err := runCommand(t, command)
+	if err != nil {
+		return &ErrWithCmdOutput{err, output}
+	}
+	return nil
 }
 
-// RunCommandAndGetOutput runs a shell command and returns its stdout and stderr as a string. The stdout and stderr of that command will also
-// be printed to the stdout and stderr of this Go program to make debugging easier.
+// RunCommandAndGetOutput runs a shell command and returns its stdout and stderr as a string. The stdout and stderr of
+// that command will also be logged with Command.Log to make debugging easier. If there are any errors, fail the test.
 func RunCommandAndGetOutput(t testing.TestingT, command Command) string {
 	out, err := RunCommandAndGetOutputE(t, command)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	return out
 }
 
-// RunCommandAndGetOutputE runs a shell command and returns its stdout and stderr as a string. The stdout and stderr of that command will also
-// be printed to the stdout and stderr of this Go program to make debugging easier.
+// RunCommandAndGetOutputE runs a shell command and returns its stdout and stderr as a string. The stdout and stderr of
+// that command will also be logged with Command.Log to make debugging easier. Any returned error will be of type
+// ErrWithCmdOutput, containing the output streams and the underlying error.
 func RunCommandAndGetOutputE(t testing.TestingT, command Command) (string, error) {
-	allOutput := []string{}
-	err := runCommandAndStoreOutputE(t, command, &allOutput, &allOutput)
+	output, err := runCommand(t, command)
+	if err != nil {
+		return output.Combined(), &ErrWithCmdOutput{err, output}
+	}
 
-	output := strings.Join(allOutput, "\n")
-	return output, err
+	return output.Combined(), nil
 }
 
-// RunCommandAndGetStdOut runs a shell command and returns solely its stdout (but
-// not stderr) as a string. The stdout and stderr of that command will also be
-// printed to the stdout and stderr of this Go program to make debugging easier.
-// If there are any errors, fail the test.
+// RunCommandAndGetStdOut runs a shell command and returns solely its stdout (but not stderr) as a string. The stdout and
+// stderr of that command will also be logged with Command.Log to make debugging easier. If there are any errors, fail
+// the test.
 func RunCommandAndGetStdOut(t testing.TestingT, command Command) string {
 	output, err := RunCommandAndGetStdOutE(t, command)
 	require.NoError(t, err)
@@ -73,20 +74,29 @@ func RunCommandAndGetStdOut(t testing.TestingT, command Command) string {
 
 // RunCommandAndGetStdOutE runs a shell command and returns solely its stdout (but not stderr) as a string. The stdout
 // and stderr of that command will also be printed to the stdout and stderr of this Go program to make debugging easier.
+// Any returned error will be of type ErrWithCmdOutput, containing the output streams and the underlying error.
 func RunCommandAndGetStdOutE(t testing.TestingT, command Command) (string, error) {
-	stdout := []string{}
-	stderr := []string{}
-	err := runCommandAndStoreOutputE(t, command, &stdout, &stderr)
+	output, err := runCommand(t, command)
+	if err != nil {
+		return output.Stdout(), &ErrWithCmdOutput{err, output}
+	}
 
-	output := strings.Join(stdout, "\n")
-	return output, err
+	return output.Stdout(), nil
 }
 
-// runCommandAndStoreOutputE runs a shell command and stores each line from stdout
-// and stderr in the given storedStdout and storedStderr variables, respectively.
-// Depending on the logger, the stdout and stderr of that command will also be
-// printed to the stdout and stderr of this Go program to make debugging easier.
-func runCommandAndStoreOutputE(t testing.TestingT, command Command, storedStdout *[]string, storedStderr *[]string) error {
+type ErrWithCmdOutput struct {
+	Underlying error
+	Output     *output
+}
+
+func (e *ErrWithCmdOutput) Error() string {
+	return fmt.Sprintf("error while running command: %v; %s", e.Underlying, e.Output.Stderr())
+}
+
+// runCommand runs a shell command and stores each line from stdout and stderr in Output. Depending on the logger, the
+// stdout and stderr of that command will also be printed to the stdout and stderr of this Go program to make debugging
+// easier.
+func runCommand(t testing.TestingT, command Command) (*output, error) {
 	command.Logger.Logf(t, "Running command %s with args %s", command.Command, command.Args)
 
 	cmd := exec.Command(command.Command, command.Args...)
@@ -96,66 +106,63 @@ func runCommandAndStoreOutputE(t testing.TestingT, command Command, storedStdout
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	stderr, err := cmd.StderrPipe()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	err = cmd.Start()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	if err := readStdoutAndStderr(t, command.Logger, stdout, stderr, storedStdout, storedStderr); err != nil {
-		return err
+	output, err := readStdoutAndStderr(t, command.Logger, stdout, stderr)
+	if err != nil {
+		return output, err
 	}
 
-	if err := cmd.Wait(); err != nil {
-		return err
-	}
-
-	return nil
+	return output, cmd.Wait()
 }
 
 // This function captures stdout and stderr into the given variables while still printing it to the stdout and stderr
 // of this Go program
-func readStdoutAndStderr(t testing.TestingT, log *logger.Logger, stdout, stderr io.ReadCloser, storedStdout, storedStderr *[]string) error {
+func readStdoutAndStderr(t testing.TestingT, log *logger.Logger, stdout, stderr io.ReadCloser) (*output, error) {
+	out := newOutput()
 	stdoutReader := bufio.NewReader(stdout)
 	stderrReader := bufio.NewReader(stderr)
 
 	wg := &sync.WaitGroup{}
-	mutex := &sync.Mutex{}
 
 	wg.Add(2)
 	var stdoutErr, stderrErr error
 	go func() {
 		defer wg.Done()
-		stdoutErr = readData(t, log, stdoutReader, mutex, storedStdout)
+		stdoutErr = readData(t, log, stdoutReader, out.stdout)
 	}()
 	go func() {
 		defer wg.Done()
-		stderrErr = readData(t, log, stderrReader, mutex, storedStderr)
+		stderrErr = readData(t, log, stderrReader, out.stderr)
 	}()
 	wg.Wait()
 
 	if stdoutErr != nil {
-		return stdoutErr
+		return out, stdoutErr
 	}
 	if stderrErr != nil {
-		return stderrErr
+		return out, stderrErr
 	}
 
-	return nil
+	return out, nil
 }
 
-func readData(t testing.TestingT, log *logger.Logger, reader *bufio.Reader, mutex *sync.Mutex, allOutput *[]string) error {
+func readData(t testing.TestingT, log *logger.Logger, reader *bufio.Reader, writer io.StringWriter) error {
 	var line string
-	var err error
+	var readErr error
 	for {
-		line, err = reader.ReadString('\n')
+		line, readErr = reader.ReadString('\n')
 
 		// remove newline, our output is in a slice,
 		// one element per line.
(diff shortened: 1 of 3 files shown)


modules/shell/command.go
```diff
@@ -73,20 +74,29 @@ func RunCommandAndGetStdOut(t testing.TestingT, command Command) string {
...
 // RunCommandAndGetStdOutE runs a shell command and returns solely its stdout (but not stderr) as a string. The stdout
 // and stderr of that command will also be printed to the stdout and stderr of this Go program to make debugging easier.
+// Any returned error will be of type ErrWithCmdOutput, containing the output streams and the underlying error.
 func RunCommandAndGetStdOutE(t testing.TestingT, command Command) (string, error) {
-	stdout := []string{}
-	stderr := []string{}
-	err := runCommandAndStoreOutputE(t, command, &stdout, &stderr)
+	output, err := runCommand(t, command)
+	if err != nil {
+		return output.stdout.String(), &ErrWithCmdOutput{err, output}
```
brikis98, line 81: Should this not do the `strings.Join` thing as a couple lines below?

modules/shell/command.go
```diff
@@ -96,84 +106,85 @@ func runCommandAndStoreOutputE(t testing.TestingT, command Command, storedStdout
...
-		line = strings.TrimSuffix(line, "\n")
-
-		// only return early if the line does not have
-		// any contents. We could have a line that does
-		// not not have a newline before io.EOF, we still
-		// need to add it to the output.
-		if len(line) == 0 && err == io.EOF {
-			break
+		// separate block so that we do not overwrite the err variable
+		// that we need afterwards to check if we're done
```
brikis98, line 168: NIT: Could use a different var name instead, right?

modules/shell/command.go
```diff
@@ -96,84 +106,85 @@ func runCommandAndStoreOutputE(t testing.TestingT, command Command, storedStdout
...
+			// only return early if the line does not have
+			// any contents. We could have a line that does
+			// not not have a newline before io.EOF, we still
+			// need to add it to the output.
+			if len(line) == 0 && err == io.EOF {
+				break
+			}
+
+			log.Logf(t, line)
+			if _, err := writer.WriteString(line); err != nil {
```
brikis98, line 183: Since there's no more mutex, we're assuming this writer is goroutine-safe?

modules/shell/output.go
```diff
@@ -0,0 +1,75 @@
...
+
+type merged struct {
+	// ensure that there are no parallel writes
+	sync.Mutex
+	Lines []string
+}
+
+func (m *merged) String() string {
+	m.Lock()
+	defer m.Unlock()
```
brikis98, line 63: Why do you need a lock on read?

modules/shell/command.go
```diff
@@ -27,44 +26,46 @@ type Command struct {
...
-// be printed to the stdout and stderr of this Go program to make debugging easier.
+// RunCommandAndGetOutputE runs a shell command and returns its stdout and stderr as a string. The stdout and stderr of
+// that command will also be logged with Command.Log to make debugging easier. Any returned error will be of type
+// ErrWithCmdOutput, containing the output streams and the underlying error.
 func RunCommandAndGetOutputE(t testing.TestingT, command Command) (string, error) {
-	allOutput := []string{}
-	err := runCommandAndStoreOutputE(t, command, &allOutput, &allOutput)
+	output, err := runCommand(t, command)
+	if err != nil {
+		return output.merged.String(), &ErrWithCmdOutput{err, output}
```
brikis98, line 60: `output` is often returned as `nil` by `runCommand`, and yet, many of the methods reach into the various `output` fields and call methods on them. Are all those methods well behaved even if `output` is `nil`? We may want to add unit tests for that specifically, as we do NOT want a panic at test time.

modules/shell/command.go
tommyknows: They are equal, but instead of doing `strings.Join` here, I'll replace the one below with `output.stdout.String()`, so that we use the same function everywhere.

modules/shell/command.go
tommyknows: No; as they don't need to be (unless they write to the same underlying store).
As long as the caller does not call this method in parallel with the same writer, we do not need a goroutine-safe one.

In this case, `output.merged` is protected by a mutex. However, only because of the structure of it, there's no relation to this function here.

modules/shell/output.go
tommyknows: You're right, I don't 🙂 Although this may result in the callee getting "old" data; if the command would still be running. But I don't think that's relevant here.

modules/shell/command.go
tommyknows: Good catch, that would (in most cases) panic!

tommyknows:
The outputs do not really differ actually, and I would argue that this change is backwards compatible as there are no public API changes.

Comparison:
<details><summary>Successful</summary><br>
Old:

```
TestThis 2020-05-17T14:06:48+02:00 logger.go:66: Running command echo with args [Hello world]
TestThis 2020-05-17T14:06:48+02:00 logger.go:66: Hello world
    TestThis: main_test.go:14: Success.
                Output: Hello world
                Error: <nil>
```

New:

```
TestThis 2020-05-17T14:12:31+02:00 logger.go:66: Running command echo with args [Hello world]
TestThis 2020-05-17T14:12:31+02:00 logger.go:66: Hello world
    TestThis: main_test.go:14: Success.
                Output: Hello world
                Error: <nil>
```
</details>

<details><summary>Failing</summary><br>
Old: 

```
TestThis 2020-05-17T14:06:48+02:00 logger.go:66: Running command doesnotexist with args [Hello world]
    TestThis: main_test.go:20: Failure.
                Output:
                Error: exec: "doesnotexist": executable file not found in $PATH
```

New:

```
TestThis 2020-05-17T14:12:31+02:00 logger.go:66: Running command doesnotexist with args [Hello world]
    TestThis: main_test.go:20: Failure.
                Output:
                Error: error while running command: exec: "doesnotexist": executable file not found in $PATH;
```
</details>

If you want the exact same outputs as before, I could remove the `error while running command` string from the error.

However, what is now possible is to access the output of a failed command:
```go

```

modules/shell/output.go
```diff
@@ -0,0 +1,88 @@
...
+		},
+	}
+}
+
+func (o *output) Stdout() string {
+	if o == nil {
+		return ""
+	}
+
+	return o.stdout.String()
```
brikis98, line 34: Do you need to check if `o.stdout == nil` as well?

modules/shell/output.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (o *output) Stderr() string {
+	if o == nil {
+		return ""
+	}
+
+	return o.stderr.String()
```
brikis98, line 42: Do you need to check if `o.stderr == nil` as well?

modules/shell/output.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (o *output) Combined() string {
+	if o == nil {
+		return ""
+	}
+
+	return o.merged.String()
```
brikis98, line 50: Do you need to check if `o.merged == nil` as well?

modules/shell/command.go
brikis98: > As long as the caller does not call this method in parallel with the same writer, we do not need a goroutine-safe one.

And I guess since `readData` is a private method, you've gone through all the existing callers to make sure they aren't using the same writer anywhere? 

Either way, it's worth updating the comment on `readData` that the method is NOT go-routine safe, so you should not re-use writers.

brikis98:
> The outputs do not really differ actually, and I would argue that this change is backwards compatible as there are no public API changes.

Hm, I was expecting / referring to the inverse of this!

That is, I thought the log output _did_ change to make the error more obvious (rather than always saying "exit status 1"). This is a change I'd very much welcome.

The backwards incompatibility I was referring to is in the API, but in a subtle way: the error type we return is now different, with an extra layer of wrapping. If anyone was expecting that error to be of a certain type (e.g., doing some sort of casting or type comparison), that will no longer work. I'm guessing it's a very rare case, so I'm OK with this breaking change, but I do think we have to call it out.

modules/shell/output.go
tommyknows: No, as even if `o.stdout` is nil, we can call `*outputStream.String()` with a nil pointer.
The actual `String()` methods check if the pointer is nil and just return an empty string in this case

(I actually noticed that I missed `*outputStream.String`, but I did this with `*merged.String` already. Now fixed and implemented with outputStream too.)

tommyknows:
> > The outputs do not really differ actually, and I would argue that this change is backwards compatible as there are no public API changes.
> 
> Hm, I was expecting / referring to the inverse of this!
> 
> That is, I thought the log output _did_ change to make the error more obvious (rather than always saying "exit status 1"). This is a change I'd very much welcome.

Actually, that is the case - it just really depends on the failure.
So what this change does is that it returns stderr within the error. 
That _should_ make the error more obvious. It is also possible to grab all the output from the error instead (by doing the type assertion).

For example:

```go
out, err = shell.RunCommandAndGetOutputE(t, shell.Command{
         Command: "sh",
         Args:    []string{"-c", "echo \"hello world\" >&2 && exit 1"},
})
t.Logf("Failure.\n\tOutput: %v\n\tError: %v", out, err)
```

Before:
```
    TestThis: main_test.go:20: Failure.
                Output: hello world
                Error: exit status 1
```

After:
```
    TestThis: main_test.go:20: Failure.
                Output: hello world
                Error: error while running command: exit status 1; hello world
```
Previously, the `Error` would have only printed `exit status 1`. 

This was not in my example above as I actually tested the "wrong" thing. In the example, I ran a command that did not exist, instead of running a command that exits with a non-0 exit code.


> The backwards incompatibility I was referring to is in the API, but in a subtle way: the error type we return is now different, with an extra layer of wrapping. If anyone was expecting that error to be of a certain type (e.g., doing some sort of casting or type comparison), that will no longer work. I'm guessing it's a very rare case, so I'm OK with this breaking change, but I do think we have to call it out.

Oh yeah, that makes sense, that is a breaking change.

I added an example to the GoDoc to show how it can be used.

modules/shell/example_test.go
```diff
@@ -0,0 +1,35 @@
+package shell_test
+
+import (
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/shell"
+)
+
+func ExampleCommand() {
```
brikis98, line 9: `t.Parallel()`

modules/shell/example_test.go
```diff
@@ -0,0 +1,35 @@
...
+	var t *testing.T
+
+	out, err := shell.RunCommandAndGetOutputE(t, shell.Command{
+		Command: "sh",
+		Args:    []string{"-c", `echo "hello world" && echo "this command has failed" >&2 && exit 1`},
+	})
+	t.Logf("Failure.\n\tOutput: %v\n\tError: %v", out, err)
```
brikis98, line 16: Will this also print `stderr`? Even without the casting below?

modules/shell/example_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	if err != nil {
+		o, ok := err.(*shell.ErrWithCmdOutput)
+		if !ok {
+			t.Errorf("did not get correct type. got=%T", err)
+			return
```
brikis98, line 22: NIT: can combine these two lines into `t.Fatalf`

brikis98: Or even reduce the whole if-block to a `require.IsType` one-liner.

modules/shell/example_test.go
brikis98: Oh, just realized this isn't a valid test function...

brikis98:
> Previously, the Error would have only printed exit status 1.
>
> This was not in my example above as I actually tested the "wrong" thing. In the example, I ran a command that did not exist, instead of running a command that exits with a non-0 exit code.

OK, great!

modules/shell/example_test.go
tommyknows: Yes; the error.String now always contains `stderr`. Output too, depending on the method that is called (`shell.RunCommandAndGetOutput` gets both stdout & stderr)

modules/shell/example_test.go
tommyknows: I usually try to keep code as "obvious" as possible; only importing stdlibs if possible.
But I don't have a strong opinion on this, you can decide on what the example should use.

Would the require usage be:
```go
var errOut *shell.ErrWithCmdOutput
require.IsType(t, errOut, err)
```

brikis98: If this is an automated test (`example_test.go`), then importing that library is fine, as downstream users won't be affected.

Note also that, in general, we put example infra code (e.g., Terraform code) in the top-level `examples` folder and example test code into the top-level `test` folder. So this example probably makes more sense in `test`, which no one imports anyway.

modules/shell/example_test.go
brikis98: Nice!

modules/shell/example_test.go
brikis98: It should be named `TestXXX` and take a `t *testing.T` as a param.

tommyknows: It's not a test, it's an example:
https://blog.golang.org/examples

modules/shell/example_test.go
tommyknows: As noted above, it's not a test, it's an example.
The advantage of having an example like that is that is shows up in the GoDoc. 
For example (no pun intended) https://godoc.org/github.com/pkg/errors#Cause / https://github.com/pkg/errors/blob/master/example_test.go#L106

That's why I hesitate from importing the `require` package; although it would work, the standard type assertion is more obvious to the regular dev.

Moving it to a different package would not make sense, as then it doesn't show up in the GoDoc anymore.
However, the `test` package you have is "different" anyway, as it is actual integration-like tests. I don't want to add a test here, but documentation.

brikis98:
Apologies for the delay! I'm super booked this week, but will try to respond in the next few days.

modules/shell/example_test.go
brikis98: OK, sorry for the long delay, just coming back to this now.

Understood on this being an example. So far, in Terratest, we've only used actual tests as examples, and not official Go-native examples. 

###  Advantages of using tests

1. Terratest is _meant_ to be used as part of an automated test suite with the `t *testing.T` arg passed in. So using real test functions ends up being a more accurate example and encourages following the right practices.
1. Automated tests can run in parallel. Examples cannot. The Terratest test suite is already getting quite slow, and I worry about it getting 10x slower if we have lots of examples.
1. Many Terratest functions are not for pure unit testing, and can only be used with lots of context from the outside world: e.g., Terraform code, Packer templates, AWS credentials, running K8S cluster, etc. This sort of thing doesn't work well in the world of Go examples, which I believe are meant for pure unit tests only.

### Advantages of examples

1. They show up in GoDoc, and can even be executed in Go playground right from the browser.

### Overall

I'm leaning towards stick with using real tests as our "examples," but am open to other opinions and trade-offs we should consider. @yorinasub17 Would be curious to hear your thoughts too!

tommyknows: Well, Go examples actually never get executed (unless in the playground from the browser).
So (2.) does not really apply here.  

The function itself is already tested, and Examples are purely for documentation purposes, demonstrating, in this case, how to extract the output from the returned error. As this also doesn't really need any kind of setup / terraform files, I thought that an example may be a perfect fit.

(my opinion, I'm okay with doing it according to your standards 🙂)

brikis98: > Well, Go examples actually never get executed (unless in the playground from the browser).
> So (2.) does not really apply here.

Actually, they do! If you specify the `// Output:` comment block, Go runs examples as part of your automated test suite and compares `stdout` to that comment block.

tommyknows: Oh you're right, and it actually fails because of the nil `*testing.T` 😅 

Means I could either move it to the example-dir, or remove the `// Output:` block.

brikis98: Yea, I worry any example code we have that isn't executed will eventually be broken... So another vote for sticking with real automated tests for our examples.

tommyknows: True, although they still get compiled at least. 
I could rewrite the example to be:
```go
func ExampleCommand() {
        t := new(testing.T)

        out, err := shell.RunCommandAndGetOutputE(t, shell.Command{
                Command: "sh",
                Args:    []string{"-c", `echo "hello world" && echo "this command has failed" >&2 && exit 1`},
                Logger:  logger.Discard,
        })
        t.Logf("Failure.\n\tOutput: %v\n\tError: %v", out, err)

        if err != nil {
                o, ok := err.(*shell.ErrWithCmdOutput)
                if !ok {
                        t.Fatalf("did not get correct type. got=%T", err)
                }
                fmt.Printf("Outputs:\n\tStdout: %s\n\tStderr: %s\n", o.Output.Stdout(), o.Output.Stderr())
        }

        // Output:
        // Outputs:
        //      Stdout: hello world
        //      Stderr: this command has failed
}
```

The test works this way, and the documentation is correct too.

brikis98: That's certainly an improvement, but I think in net, using a proper automated test is still going to be the better fit for Terratest.

yorinasub17: I agree with @brikis98 that we should stick to our integration test suite for the example. While the example may be a fit here, the problem is that this sets a precedent in our code base. Going from 0 examples to 1 example is actually a big change, as it implies that we want examples.

This leads to questions everytime someone comes into the code base:

- Why does `RunCommandAndGetOutputE` have an example and the other functions do not?
- Would you accept contributions that add examples for all the other functions, like `aws` package? (this is where the execution time becomes a big problem)
- I'm adding new functionality. Should I add an example?

Unless we have the intention to go through and add examples for most or all of our functions, I feel that it is safer from a maintainability standpoint to avoid them altogether.

tommyknows: Alright, thanks for the feedback!

I moved the example to an actual test and ensured that the output is correct.
I did not, however, move it to the `test` folder. The way it currently is, it's more of a unit test than an example. 

Should I also add a separate example? or move that unit test to the example folder?

brikis98: I think this is fine as-is, thanks!

brikis98:
There was an unrelated test failure that looked like a transient issue. All other tests passed, so I'm going to merge. Thanks for the PR!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.28.0

tommyknows:
Thank you for taking your time and looking at this!

PR #526, opened 2020-05-24
Update dependencies

edumco:
### Several dependencies are outdated

Using @dependabot I have selected 5 out of 15 outdated dependencies.

The select are the less outdated ones. This makes the change less prone to problems.

Do you have any policy about updates?


go.mod
```diff
@@ -3,10 +3,10 @@ module github.com/gruntwork-io/terratest
 go 1.13
 
 require (
-	cloud.google.com/go v0.51.0
-	cloud.google.com/go/storage v1.0.0
+	cloud.google.com/go v0.56.0
+	cloud.google.com/go/storage v1.6.0
 	github.com/Azure/azure-sdk-for-go v38.1.0+incompatible
-	github.com/Azure/go-autorest/autorest v0.9.3
+	github.com/Azure/go-autorest/autorest v0.10.2
```
brikis98, line 9: Seems like a potentially backwards incompatible change?

PR #528, opened 2020-05-25
Default logger needs to mark itself as helper to avoid the line numbers

yorinasub17:
Right now, the default logger does not call the helper function, so it ends up logging an extraneous line number `logger.go:66:`.

### DIFF
--- a/modules/logger/logger.go
+++ b/modules/logger/logger.go
@@ -99,6 +99,10 @@ func (_ testingT) Logf(t testing.TestingT, format string, args ...interface{}) {
 type terratestLogger struct{}
 
 func (_ terratestLogger) Logf(t testing.TestingT, format string, args ...interface{}) {
+	if tt, ok := t.(helper); ok {
+		tt.Helper()
+	}
+
 	DoLog(t, 3, os.Stdout, fmt.Sprintf(format, args...))
 }
 


modules/logger/logger.go
```diff
@@ -99,6 +99,10 @@ func (_ testingT) Logf(t testing.TestingT, format string, args ...interface{}) {
 type terratestLogger struct{}
 
 func (_ terratestLogger) Logf(t testing.TestingT, format string, args ...interface{}) {
+	if tt, ok := t.(helper); ok {
+		tt.Helper()
```
brikis98, line 103: Sorry, I'm totally missing some context here. What is this helper thing?

yorinasub17:
Hmm this didn't have the expected effect so will close.

modules/logger/logger.go
yorinasub17: See https://golang.org/pkg/testing/#B.Helper. Specifically:

> When printing file and line information, that function will be skipped.

Right now all the logs are outputting `logger.go:66` as the file + linenumber information, which is this function.

brikis98: Ohhhh, neat, didn't know about that! Shame it didn't seem to work :-\

PR #530, opened 2020-05-26
Bump Kubernetes 1.18.3

### LINKED ISSUE #529: Bump the K8s version to latest 1.18.3
In order to use Terratest with projects that have already migrated their k8s client code to 1.18.x, Terratest needs to bump its dependency from 1.17.0. This incurs slight changes in some of the API calls to K8s in `modules/k8s`, mostly adding `context.Background()` as the first argument to the calls.

nickolaev:
In order to use Terratest with projects that have already migrated their k8s client code to 1.18.x, Terratest needs to bump its dependency from 1.17.0. This incurs slight changes in some of the API calls to K8s in modules/k8s, mostly adding context.Background() as the first argument to the calls.

Fixes #529


yorinasub17:
Build passed, so will merge and release. Thanks!

PR #532, opened 2020-05-26
Handle nil in input vars

yorinasub17:
Make sure to convert `nil` to `null` in `-var` so that terraform knows how to handle it correctly and we can pass in `null` as input vars.

### DIFF
--- a/modules/terraform/format.go
+++ b/modules/terraform/format.go
@@ -182,6 +182,10 @@ func mapToHclString(m map[string]interface{}) string {
 // Convert a primitive, such as a bool, int, or string, to an HCL string. If this isn't a primitive, force its value
 // using Sprintf. See ToHclString for details.
 func primitiveToHclString(value interface{}, isNested bool) string {
+	if value == nil {
+		return "null"
+	}
+
 	switch v := value.(type) {
 
 	case bool:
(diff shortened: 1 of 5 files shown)


yorinasub17:
> ... but I have some memory of having tried this a while back, and had to revert it. IIRC, Terraform didn't parse null correctly when it was passed in via TF_VAR_xxx env vars or CLI -var xxx=yyy args.

This still doesn't work. However, this change improves over the status quo:

- Currently, `nil` gets translated to `<nil>` in the use case above, and is still passed in as a string. At least it is a bit more sensible when it gets translated to `null`.
- Translating to the string literal `null` in a complex data type now works. E.g., `null` in map and lists will work with this change.

The latter is what we need right now.

One thing I can do is refactor this routine to support error handling, and update the function to return an error if it attempts to translate a `nil` at the top level, as opposed to within a complex type.

brikis98:
> One thing I can do is refactor this routine to support error handling, and update the function to return an error if it attempts to translate a nil at the top level, as opposed to within a complex type.

That

brikis98:
_(Sorry, hit enter too soon on previous comment)_

> This still doesn't work. However, this change improves over the status quo:

Roger, thx for confirming. The improvement in the PR is still an improvement!

> One thing I can do is refactor this routine to support error handling, and update the function to return an error if it attempts to translate a nil at the top level, as opposed to within a complex type.

Either that, or at least document this as a known issue.

innovia:
need that too - following

yorinasub17:
> Either that, or at least document this as a known issue.

Decided to document the behavior for now (8de6386) in the interest of getting this shipped sooner.

---

Thanks for the review! Going to merge and release this now.

PR #537, opened 2020-05-31
Update quick start examples to use tested example code

brikis98:
This PR updates the Terratest quick start guide to use our tested example code (that is, the code from the `examples` folder that we run through automated tests) instead of code copy/pasted into the Markdown files. The copy/pasted code is _already_ broken (e.g., bad AMI ID, not enough retries), and this will only get worse over time, so switching to example code that's actually tested makes our documentation more up to date and maintainable. We were already using real examples on the home page and examples page of Terratest, so this update is mainly about:

1. Refactoring the example code so that it's possible to embed single files rather than a list of examples.
1. Updating the UI for the example code so that certain aspects can be hidden, such as the nav, "learn more" links, etc.
1. Using these new super powers in the quick start guide to replace the hard-coded, out-of-date examples with the tested examples from the home page.

### DIFF
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -7,6 +7,9 @@ tags: ["quick-start"]
 order: 101
 nav_title: Documentation
 nav_title_link: /docs/
+custom_js:
+  - examples
+  - prism
 ---
 
 ## Requirements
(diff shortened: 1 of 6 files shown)


docs/_docs/01_getting-started/quick-start.md
```diff
@@ -59,91 +62,91 @@ The basic usage pattern for writing automated tests with Terratest is to:
...
-output "url" {
-  value = "http://${aws_instance.web_server.public_ip}:8080"
-}
+## Example #1: Terraform "Hello, World"
+
+Let's start with the simplest possible [Terraform](https://www.terraform.io/) code, which just outputs the text, 
+"Hello, World" (if you’re new to Terraform, check out our [Comprehensive Guide to 
+Terraform](https://blog.gruntwork.io/a-comprehensive-guide-to-terraform-b3d32832baca)): 
+ 
+{% include examples/explorer.html example_id='terraform-hello-world' file_id='terraform_code' id='quick_start_page' class='wide quick-start-examples' skip_learn_more=true skip_view_on_github=true skip_tags=true %}
```
brikis98, line 71: This `include` is how we can now include real example code, including one file at a time, in our docs.

brikis98:

Issue #536, opened 2020-05-31
TestImportSSHKeyOSLogin failing


brikis98:
[CircleCi Build](https://circleci.com/gh/gruntwork-io/terratest/3286)
[Build log](https://3286-53174289-gh.circle-artifacts.com/0/tmp/logs/TestImportSSHKeyOSLogin.log)

```
=== RUN   TestImportSSHKeyOSLogin
=== PAUSE TestImportSSHKeyOSLogin
=== CONT  TestImportSSHKeyOSLogin
TestImportSSHKeyOSLogin 2020-05-29T14:21:09Z key_pair.go:31: Generating new public/private key of size 2048
TestImportSSHKeyOSLogin 2020-05-29T14:21:10Z oslogin.go:27: Importing SSH key for user circleci@terratest-214610.iam.gserviceaccount.com
TestImportSSHKeyOSLogin 2020-05-29T14:21:10Z oslogin.go:61: Deleting SSH key for user circleci@terratest-214610.iam.gserviceaccount.com
TestImportSSHKeyOSLogin 2020-05-29T14:21:10Z oslogin.go:101: Getting login profile for user circleci@terratest-214610.iam.gserviceaccount.com
--- FAIL: TestImportSSHKeyOSLogin (1.01s)
    oslogin.go:20: 
        	Error Trace:	oslogin.go:20
        	            				oslogin_test.go:18
        	Error:      	Received unexpected error:
        	            	googleapi: Error 400: Login profile size exceeds 32 KiB. Delete profile values to make additional space., badRequest
        	Test:       	TestImportSSHKeyOSLogin
        	Messages:   	Could not add SSH Key to user circleci@terratest-214610.iam.gserviceaccount.com
```

brikis98:
@robmorgan Do we need to clean something up in one of our GCP accounts?

robmorgan:
@brikis98 yes we do, it looks like our test IAM user has too many public keys associated with this. I'll look into this, this week.

brikis98:
Roger, thx Rob!

robmorgan:
Yori and I looked at the imported SSH keys for the Terratest CircleCI user and found they were at the 70 limit. See https://github.com/gruntwork-io/terratest/pull/298 for more context.

[...]

PR #539, opened 2020-06-01
Switch magiconair/properties with stretchr/testify

### LINKED ISSUE #538: Possible wrong test library
I was looking into the `test` files and realize that terratest is using `github.com/magiconair/properties/assert` and `github.com/stretchr/testify` assert. This is the correct behavior?

Related:
https://github.com/gruntwork-io/terratest/blob/master/test/docker_hello_world_example_test.go#L7
https://github.com/gruntwork-io/terratest/blob/master/test/packer_hello_world_example_test.go
https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/apply_test.go#L6
https://github.com/gruntwork-io/terratest/blob/master/modules/gcp/compute_test.go#L12

yorinasub17:
Fixes #538


robmorgan:
@cairesvs nice find 👍  it looks like we slipped that in years ago.

yorinasub17:
Thanks for review. Merging this in: the test failures are from https://github.com/gruntwork-io/terratest/issues/536

PR #540, opened 2020-06-03
Feature/adding AWS secret manager

ismferd:
Hi guys,

I need to include SecretManager as part of terratest AWS module. 

I share the gist file with test passed:
https://gist.github.com/ismferd/466cb3eb14331de0760181681c2f73da

Thanks for this awesome project!

### DIFF
--- a/modules/aws/secretmanager.go
+++ b/modules/aws/secretmanager.go
@@ -1,7 +1,6 @@
 package aws
 
 import (
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/secretsmanager"
 	"github.com/gruntwork-io/terratest/modules/testing"
@@ -62,7 +61,7 @@ func NewSecretManagerClient(t testing.TestingT, region string) *secretsmanager.S
 }
 
 //NewSecretManagerClientE creates a new SQS client.
-func NewSecretManagerClientE(t testing.TestingT, region string) ( *secretsmanager.SecretsManager, error) {
+func NewSecretManagerClientE(t testing.TestingT, region string) (*secretsmanager.SecretsManager, error) {
 	sess, err := NewAuthenticatedSession(region)
 	if err != nil {
 		return nil, err
--- a/modules/aws/secretmanager.go
+++ b/modules/aws/secretmanager.go
@@ -1,0 +1,71 @@
+package aws
+
+import (
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+//PutSecret creates new secret of SecretManager.
+func PutSecret(t testing.TestingT, awsRegion string, secretID string, secretDescription string, keyValue string) string {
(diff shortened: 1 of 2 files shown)


yorinasub17:
Thanks for your contribution! However, given the relative simplicity of the functions, I am not sure if we will include this into our API. Could use a second opinion from @brikis98 on this.

To explain, as Terratest has grown (we're approaching 4,000 stars!), we've had to think more about what types of helper functions we want in the library to make sure it doesn't become so big as to be unusable. Please check out the [contribution guidelines](https://terratest.gruntwork.io/docs/community/contributing/), especially the part about [infrastructure and validation helpers](https://terratest.gruntwork.io/docs/community/contributing/#infrastructure-and-validation-helpers) for more details. These secretsmanager helpers meet requirements for "platforms," "popularity," and "creating infrastructure," but fall short on "complexity" (i.e., these are direct wrappers around the AWS SDK as opposed to something more complicated).

ismferd:
@yorinasub17 thanks for your answer, basically one of the thing that I want to achieve with that is the possibility of add the retry.DowithRetry in order to get my resource as I see in other parts of your project.

ismferd:
That's probably a special case but I could extend the SecretManager API in a close future.

So, I'm looking forward your answer.

yorinasub17:
> add the retry.DowithRetry in order to get my resource as I see in other parts of your project.

Can you elaborate a bit further? E.g., would something like the following work for you:

```go
sess, err := terraws.NewAuthenticatedSession(region)
require.NoError(t, err)
secretsMgr := secretsmanager.New(sess)
secret := retry.DoWithRetry(... omitted for brevity ..., func() (string, error) {
  resp, err := secretsMgr.GetSecretValue(&secretsmanager.GetSecretValueInput{SecretId: aws.String(secretID)})
  if err != nil {
    return "", err
  }
  return aws.StringValue(resp.SecretString), nil
})
```

> I could extend the SecretManager API in a close future.

If you can describe some of your use cases here that may be difficult to do with the SDK directly or is not obvious, that can help in deciding if we should include this.

ismferd:
Currently, the case is basically that: get a secret previously added into secretmanager.
Trying to abstract me of AWS API (things such as sessions) and just making sense to my test code. 

Thanks a lot @yorinasub17

brikis98:
A few thoughts:

1. As-written, this does seem like a very thin wrapper over the AWS SDK. Such thin wrappers require effort to document, find, learn, maintain, extend to handle other use cases, etc, so as Yori said, I'm not sure it's worth adding to Terratest.
1. AWS Secrets Manager is eventually consistent, so after you write a secret, you do have to use retries to wait until you can read that secret back. Adding that to this PR would make it _slightly_ more compelling, but (a) we already have a `retry` helper, so as Yori showed, you can do that in your own code and (b) not all use cases would need retries.

So I think I'm in agreement with Yori that this probably doesn't quite meet the bar of being complicated enough to merit creating an abstraction for it in Terratest.

ismferd:
@brikis98 thanks for your answer,
I only kept in mind to continue with the abstraction of the aws module that you have. 
If you don't see appropriate to have an abstraction of secret manager I'm going to use @yorinasub17 tip.

Thank both for your quick answer.

james00012:

Issue #534, opened 2020-05-27
Terratest fails for azure modules with "features": required field is not set"


papineni87:
I am trying to test terraform azure module and facing below issue

TestTerraformAzureExample 2020-05-27T21:22:08+05:30 logger.go:66: 2020/05/27 21:22:08 [ERROR] <root>: eval: *terraform.EvalSequence, err: "features": required field is not set
TestTerraformAzureExample 2020-05-27T21:22:08+05:30 logger.go:66: 
TestTerraformAzureExample 2020-05-27T21:22:08+05:30 logger.go:66: Error: "features": required field is not set

I am not sure why features field is required for modules as it is required only in the provider. 
I am not using the azurerm provider in my modules.

brikis98:
Hm, looks like the Azurerm 2.0 provider now requires the `features` block to be set: https://github.com/terraform-providers/terraform-provider-azurerm/issues/5867.

Would you be up for a PR to add that to the Azure examples?

tsalright:
The example I am seeing in the code is pointed to azurerm 1.31.0. It shouldn't require the feature field.

Locally, I ran the test straight from master and it works. I updated it to version 2.0.0 and ran it against my azure account without feature and was able to replicate the issue. I then added `features {}` on the next line and the tests are working for provider 2.0.0. 

I threw a pull request together while i was testing it out.
https://github.com/gruntwork-io/terratest/pull/547

brikis98:
Please see https://github.com/gruntwork-io/terratest/issues/89 for the status on Terratest with Azure. In particular, we're [starting some work](https://github.com/gruntwork-io/terratest/issues/89#issuecomment-650052374) around it, and when that work is done, we'll be able to come back to this question!

davesee:
Hi @tsalright , 

Are you still having the problem? I see features{} is now included in the providers.

There have been several updates to the Azure tests since this issue was created and I would like to help if it is still needed.

tsalright:
@davesee I am not having the issue anymore. I think we updated most of these when we restructured the modules and tests. Should be good now.

davesee:
Please close @yorinasub17

HadwaAbdelhalem:
@brikis98  would you please close this issue. the recent contributions fixed that

yorinasub17:
Thanks for the updates! Just closed as requested!

PR #547, opened 2020-06-05
test: update basic azure example to provider 2.0.0

tsalright:
Updated the **provider azurerm** version to 2.0.0 and added the missing feature {} field. This addresses issue https://github.com/gruntwork-io/terratest/issues/534.


brikis98:
Please see https://github.com/gruntwork-io/terratest/issues/89 for the status on Terratest with Azure. In particular, we're [starting some work](https://github.com/gruntwork-io/terratest/issues/89#issuecomment-650052374) around it, and when that work is done, we'll be able to come back to this PR!

yorinasub17:
@tsalright Is this PR still necessary or should we close this out? It looks like the version locking was removed when we started testing: https://github.com/gruntwork-io/terratest/blob/master/examples/azure/terraform-azure-example/main.tf

rguthriemsft:
we should close this PR out.  I don't think it is needed anymore.

yorinasub17:
Closing as obsolete. Thanks!

PR #554, opened 2020-06-11
Idempotent apply

### LINKED ISSUE #551: Does terratest support checking that no changes happen a second apply?
I search but didn't find anything relevant.

shadycuz:
Closes #551 

This PR adds a "Apply" function to `apply.go`. This function will apply the terraform config and then immediately afterward run the plan function to verify that the configuration doesn't produce any plan changes. 

I have updated the Docs with an example. I have added two tests in `apply_test.go`. One test works as expected but the other I can't get to work correctly. 

In `ApplyAndIdempotent()` function I use this to error if there is changes on the second plan. 

```
exitCode := Plan(t, options)
require.Equal(t, DefaultSuccessExitCode, exitCode)
```

I don't know how to "catch" the test failure caused by `require.Equal`. I can't seem to find a way to mark my test as "FailureExpected". I tried catching like this:

```
require.Panicsf(t, func() { ApplyAndIdempotent(t, options) }, "Should Fail")
```

But that doesn't work. I decided it's best to get some guidance. I don't write in Golang often.

Edit: It looks like maybe I should create a "wrapper" function that returns an error and test that for failures. That seems to be the pattern like, `Plan()` and `PlanE()`

### DIFF
--- a/docs/_docs/04_community/contributing.md
+++ b/docs/_docs/04_community/contributing.md
@@ -115,8 +115,8 @@ We also recommend updating the automated tests *before* updating any code (see [
 Development](https://en.wikipedia.org/wiki/Test-driven_development)). That means you add or update a test case,
 verify that it's failing with a clear error message, and *then* make the code changes to get that test to pass. This
 ensures the tests stay up to date and verify all the functionality in this Module, including whatever new
-functionality you're adding in your contribution. Check out the [test](https://github.com/gruntwork-io/terratest/tree/master/test) folder for instructions on running the
-automated tests.
+functionality you're adding in your contribution. The instructions for running the automated tests can be
+found [here](https://terratest.gruntwork.io/docs/community/contributing/#developing-terratest).
 
 ### Update the code
 
--- a/modules/terraform/apply.go
+++ b/modules/terraform/apply.go
@@ -63,7 +63,7 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
 	return RunTerraformCommandE(t, options, FormatArgs(options, "apply-all", "-input=false", "-lock=false", "-auto-approve")...)
 }
 
-// ApplyAndIdempotent runs terraform init and apply with the given options and return stdout/stderr from the apply command. It then runs
+// ApplyAndIdempotent runs terraform apply with the given options and return stdout/stderr from the apply command. It then runs
 // plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
 // the caller is responsible for cleaning up any resources created by running apply.
 func ApplyAndIdempotent(t testing.TestingT, options *Options) string {
@@ -74,11 +74,11 @@ func ApplyAndIdempotent(t testing.TestingT, options *Options) string {
 	return out
 }
 
-// ApplyAndIdempotentE runs terraform init and apply with the given options and return stdout/stderr from the apply command. It then runs
+// ApplyAndIdempotentE runs terraform apply with the given options and return stdout/stderr from the apply command. It then runs
 // plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
 // the caller is responsible for cleaning up any resources created by running apply.
 func ApplyAndIdempotentE(t testing.TestingT, options *Options) (string, error) {
-	out, err := InitAndApplyE(t, options)
+	out, err := ApplyE(t, options)
 
 	if err != nil {
 		return "", err
--- a/modules/terraform/apply.go
+++ b/modules/terraform/apply.go
@@ -80,17 +80,17 @@ func ApplyAndIdempotentE(t testing.TestingT, options *Options) (string, error) {
 	out, err := ApplyE(t, options)
 
 	if err != nil {
-		return "", err
+		return out, err
 	}
 
 	exitCode, err := PlanExitCodeE(t, options)
 
 	if err != nil {
-		return "", err
+		return out, err
 	}
 
 	if exitCode != 0 {
-		return "", errors.New("terraform configuration not idempotent")
+		return out, errors.New("terraform configuration not idempotent")
 	}
 
 	return out, nil
--- a/modules/terraform/apply.go
+++ b/modules/terraform/apply.go
@@ -84,11 +84,39 @@ func ApplyAndIdempotentE(t testing.TestingT, options *Options) (string, error) {
 		return "", err
 	}
 
-	exitCode := PlanExitCode(t, options)
+	exitCode, err := PlanExitCodeE(t, options)
+
+	if err != nil {
+		return "", err
+	}
 
 	if exitCode != 0 {
 		return "", errors.New("terraform configuration not idempotent")
 	}
 
 	return out, nil
+}
+
+// InitAndApplyAndIdempotent runs terraform init and apply with the given options and return stdout/stderr from the apply command. It then runs
+// plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
+// the caller is responsible for cleaning up any resources created by running apply.
+func InitAndApplyAndIdempotent(t testing.TestingT, options *Options) string {
+
+	out, err := InitAndApplyAndIdempotentE(t, options)
+	require.NoError(t, err)
+
+	return out
+}
+
+// InitAndApplyAndIdempotentE runs terraform init and apply with the given options and return stdout/stderr from the apply command. It then runs
+// plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
+// the caller is responsible for cleaning up any resources created by running apply.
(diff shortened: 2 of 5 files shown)


shadycuz:
Looks good now.

Note I didn't run the full test suite. The instructions say `Check out the test folder for instructions on running the automated tests.` but I didn't find any instructions. 

I'm guessing:

```bash
cd test
go mod init tests
go test
```
or something like that would work?

modules/terraform/apply.go
```diff
@@ -60,3 +62,33 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
...
+	require.NoError(t, err)
+
+	return out
+}
+
+// ApplyAndIdempotentE runs terraform init and apply with the given options and return stdout/stderr from the apply command. It then runs
+// plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
+// the caller is responsible for cleaning up any resources created by running apply.
+func ApplyAndIdempotentE(t testing.TestingT, options *Options) (string, error) {
+	out, err := InitAndApplyE(t, options)
```
brikis98, line 81: Should only call `Apply`. If you want a version that calls `init` and `apply`, it would have to be called something like `InitAndApplyAndIdempotent`.

modules/terraform/apply.go
```diff
@@ -60,3 +62,33 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
...
+
+	if err != nil {
+		return "", err
+	}
+
+	exitCode := PlanExitCode(t, options)
```
brikis98, line 87: This should call `PlanExitCodeE` and return the error if it hits one.

modules/terraform/apply_test.go
```diff
@@ -119,3 +119,36 @@ func TestTgApplyOutput(t *testing.T) {
...
+	require.NoError(t, err)
+
+	options := &Options{
+		TerraformDir: testFolder,
+		NoColor:      true,
+	}
+
+	out, err := ApplyAndIdempotentE(t, options)
+
+	_ = out
```
brikis98, line 150: What is this for?

modules/terraform/apply_test.go
```diff
@@ -119,3 +119,36 @@ func TestTgApplyOutput(t *testing.T) {
...
+
+	require.Error(t, err)
```
brikis98, line 152: Might want to check it's the error you expect.

brikis98:
> The instructions say Check out the test folder for instructions on running the automated tests. but I didn't find any instructions.

Oh, hm, where does it say that? It should be updated to point to: https://terratest.gruntwork.io/docs/community/contributing/#developing-terratest

modules/terraform/apply.go
shadycuz: @brikis98 I changed the call to be `Apply` only. Do you think we should offer one that init's as well?

shadycuz: I went ahead and created one as that follows the previous code style.

modules/terraform/apply.go
shadycuz: yip, makes sense.

modules/terraform/apply_test.go
shadycuz: The compiler was complaining about an unused variable. I googled and it said to do this 🤣. 

I changed it to `require.Equal(t, out, "")`

modules/terraform/apply_test.go
shadycuz: Thats a good catch, up to this point I had not ran the tests with -v, it was failing with an error from terraform because I hadn't ran init. Which really surprised me because the other test worked without init. Turns our null resource has a provider. So the test was passing but not working as expected.

Fixed with this `require.EqualError(t, err, "terraform configuration not idempotent")`

shadycuz:
> > The instructions say Check out the test folder for instructions on running the automated tests. but I didn't find any instructions.
> 
> Oh, hm, where does it say that? It should be updated to point to: https://terratest.gruntwork.io/docs/community/contributing/#developing-terratest

It was in the contributing.md, its fixed now. https://github.com/gruntwork-io/terratest/pull/554/commits/da7b476a8cfba2401bbc3366b4bcea69ecacdac8

docs/_docs/04_community/contributing.md
```diff
@@ -115,8 +115,8 @@ We also recommend updating the automated tests *before* updating any code (see [
 Development](https://en.wikipedia.org/wiki/Test-driven_development)). That means you add or update a test case,
 verify that it's failing with a clear error message, and *then* make the code changes to get that test to pass. This
 ensures the tests stay up to date and verify all the functionality in this Module, including whatever new
-functionality you're adding in your contribution. Check out the [test](https://github.com/gruntwork-io/terratest/tree/master/test) folder for instructions on running the
-automated tests.
+functionality you're adding in your contribution. The instructions for running the automated tests can be
+found [here](https://terratest.gruntwork.io/docs/community/contributing/#developing-terratest).
```
brikis98, line 119: Thx for updating this 👍

modules/terraform/apply.go
```diff
@@ -60,3 +62,61 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
 
 	return RunTerraformCommandE(t, options, FormatArgs(options, "apply-all", "-input=false", "-lock=false", "-auto-approve")...)
 }
+
+// ApplyAndIdempotent runs terraform apply with the given options and return stdout/stderr from the apply command. It then runs
+// plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
+// the caller is responsible for cleaning up any resources created by running apply.
+func ApplyAndIdempotent(t testing.TestingT, options *Options) string {
+
```
brikis98, line 70: NIT: extra blank line.

modules/terraform/apply.go
```diff
@@ -60,3 +62,61 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
...
+}
+
+// ApplyAndIdempotentE runs terraform apply with the given options and return stdout/stderr from the apply command. It then runs
+// plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
+// the caller is responsible for cleaning up any resources created by running apply.
+func ApplyAndIdempotentE(t testing.TestingT, options *Options) (string, error) {
+	out, err := ApplyE(t, options)
+
+	if err != nil {
+		return "", err
```
brikis98, line 84: All the errors here should probably `return out, err`

modules/terraform/apply.go
```diff
@@ -60,3 +62,61 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
...
+	}
+
+	return out, nil
+}
+
+// InitAndApplyAndIdempotent runs terraform init and apply with the given options and return stdout/stderr from the apply command. It then runs
+// plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
+// the caller is responsible for cleaning up any resources created by running apply.
+func InitAndApplyAndIdempotent(t testing.TestingT, options *Options) string {
+
```
brikis98, line 104: NIT: extra blank line

modules/terraform/apply.go
```diff
@@ -60,3 +62,61 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
...
+	require.NoError(t, err)
+
+	return out
+}
+
+// InitAndApplyAndIdempotentE runs terraform init and apply with the given options and return stdout/stderr from the apply command. It then runs
+// plan again and will fail the test if plan requires additional changes. Note that this method does NOT call destroy and assumes
+// the caller is responsible for cleaning up any resources created by running apply.
+func InitAndApplyAndIdempotentE(t testing.TestingT, options *Options) (string, error) {
+
```
brikis98, line 115: NIT: extra blank line here and at the end of the method.

shadycuz:
@brikis98 all fixed up. 

I didn't run any of the tests in /test because none of them will call the new idempotent functions in this PR. I supposed I could have written one but since we are just recomposing existing well tested logic I felt the 2 unit tests would work fine.

brikis98:
Great, thank you! Kicking off tests now.

brikis98:
One test failed:

```
TestPackerMultipleConcurrentAmis 2020-06-14T13:32:10Z logger.go:66: 1592141530,,ui,error,Build 'amazon-ebs' errored: Error waiting for AMI. Reason: {\n  Code: "Server.InternalError"%!!(MISSING)(PACKER_COMMA)\n  Message: "Server.InternalError"\n}
```

This looks like an intermittent issue in AWS and has nothing to do with this PR. Since all the other tests passed, I'm going to merge. Thanks!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.28.5

PR #567, opened 2020-06-28
Add ability for parseMap() to parse non-map list values

clockworksoul:
When working with outputs from Terraform `aws_security_group` resources I found that the `OutputListOfObjectsE` function couldn't handle lists nested inside the maps elements (such as the `cidr_blocks` values in the `aws_security_group.foo.ingress` list). I tracked the issue to the inability of `parseMap` to parse maps that contained values of a list type other than `[]map[string]interface{}`. 

To test, I extended the maps `test/fixtures/terraform-output-listofobjects/output.tf` to include lists, as seen in the diff. 

Running `go test -run TestOutputListOfObjects` against the test condition returned the same error:

```
$ go test -run TestOutputListOfObjects
...
--- FAIL: TestOutputListOfObjects (0.27s)
    output.go:155: 
        	Error Trace:	output.go:155
        	            				output_test.go:160
        	Error:      	Received unexpected error:
        	            	Type switching to map[string]interface{} failed.
        	Test:       	TestOutputListOfObjects
FAIL
```

I'll admit that this diff isn't pretty, but it adds the desired functionality.

Test results, post change:

```
$ go test -run TestOutputListOfObjects
...
TestOutputListOfObjects 2020-06-28T17:26:52-04:00 logger.go:66: Running command terraform with args [output -no-color -json list_of_maps]
TestOutputListOfObjects 2020-06-28T17:26:52-04:00 logger.go:66: [{"even_more":["six"],"more":{"five":"five","four":4},"one":1,"three":"three","two":"two"},{"even_more":[6],"more":[{"five":"five","four":4}],"one":"one","three":3,"two":2}]
PASS
```

### DIFF
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -100,6 +100,52 @@ func parseMap(m map[string]interface{}) (map[string]interface{}, error) {
 			}
 			result[k] = nestedMap
 		case []interface{}:
+			if len(vt) == 0 {
+				result[k] = []interface{}{}
+			} else {
+				switch vt[0].(type) {
+				case map[string]interface{}:
+					nestedList, err := parseListOfMaps(vt)
+					if err != nil {
+						return nil, err
+					}
+					result[k] = nestedList
+				default:
+					nestedList, err := parseListOutputTerraform(vt, k)
+					if err != nil {
+						return nil, err
+					}
+					result[k] = nestedList
+				}
+			}
+		case float64:
+			testInt, err := strconv.ParseInt((fmt.Sprintf("%v", vt)), 10, 0)
+			if err == nil {
+				result[k] = int(testInt)
+			} else {
+				result[k] = vt
+			}
+		default:
+			result[k] = vt
+		}
+	}
+
+	return result, nil
+}
+
+func parseMapOriginal(m map[string]interface{}) (map[string]interface{}, error) {
+
+	result := make(map[string]interface{})
+
+	for k, v := range m {
+		switch vt := v.(type) {
+		case map[string]interface{}:
+			nestedMap, err := parseMap(vt)
+			if err != nil {
+				return nil, err
+			}
+			result[k] = nestedMap
+		case []interface{}:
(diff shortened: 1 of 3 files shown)


modules/terraform/output.go
```diff
@@ -100,11 +100,24 @@ func parseMap(m map[string]interface{}) (map[string]interface{}, error) {
...
+			} else {
+				switch vt[0].(type) {
+				case map[string]interface{}:
+					nestedList, err := parseListOfMaps(vt)
+					if err != nil {
+						return nil, err
+					}
+					result[k] = nestedList
+				default:
+					nestedList, err := parseListOutputTerraform(vt, k)
```
clockworksoul, line 114: Is `parseListOutputTerraform` the correction function to be using here? Should there perhaps be a `parseList` instead?

brikis98:
I think for more complicated structures than the basics supported by the existing `OutputXXX` methods, we'd need to add the following: https://github.com/gruntwork-io/terratest/pull/513#issuecomment-627833280.

clockworksoul:
Ah, that's very interesting, and I can see the utility for complex cases! I can do that.

So that I understand, though, do we not want to support `OutputListOfObjects` in the case where the objects themselves contain lists that aren't type `[]map[string]interface{}`?  Also, how would this new function affect existing functionality?

brikis98:
> So that I understand, though, do we not want to support the current `OutputListOfObjects` function in the case where the objects themselves contain lists that aren't type `[]map[string]interface{}`? 

I don't think these "simpler" methods are well-suited to deeply-nested types. For that, I think defining a struct is a better way to go.

> Also, how would this new function affect existing functionality?

I think we'd leave existing functionality as-is for simple, non-nested use cases.

clockworksoul:
Makes sense, and seems straight-forward enough. I'll close this PR and start a new one for `OutputStructE` and `OutputStruct`.

clockworksoul:
For closure: followed up with https://github.com/gruntwork-io/terratest/pull/569

Issue #338, opened 2019-07-22
Wrong log message for GCP zonal instance groups


Etiene:
This confused the hell out of me when working on a nomad module that was still using zonal instance groups. 

The method `getRandomInstanceE` can be called by both types of instance groups (Regional and Zonal) but the log has a hardcoded mention to Regional 

* https://github.com/gruntwork-io/terratest/blob/71a1fe923037323023f8b59368eca5a3bb3423d1/modules/gcp/compute.go#L560

* https://github.com/gruntwork-io/terratest/blob/71a1fe923037323023f8b59368eca5a3bb3423d1/modules/gcp/compute.go#L565

j4ng5y:
I have a PR ready for this, but I need to clean up my local repo before I issue it :smile:

robmorgan:
@j4ng5y sure 👍

PR #584, opened 2020-08-21
Update error message for zonal instance group

toshi0607:
Fix this issue 
close https://github.com/gruntwork-io/terratest/issues/338

`getRandomInstanceE` is also used for zonal instance group as the issue says
https://github.com/gruntwork-io/terratest/blob/master/modules/gcp/compute.go#L534

I was really impressed by terratest while reading Terraform Up & Running. Thank you so much. Let me contribute especially for GCP and Kubernetes 😄


yorinasub17:
One test failed, but it is unrelated to this change so will go ahead and merge + release this!

PR #588, opened 2020-08-26
GetCurrentBranchName update and GetCurrentGitRef creation

### LINKED ISSUE #574: git.GetCurrentBranchName returns HEAD when on a release tag
When you are running a build from a release tag, `git.GetCurrentBranchName` will return `HEAD` because you are not on a branch (detached state). This makes it difficult to run certain operations that depend on checking out the specific `ref`.

As a proposed remedy:

- `git.GetCurrentBranchName` should return `""` when we are not on a specific branch.
- Implement `git.GetCurrentGitRef` which will return the branch if on a branch (using `git.GetCurrentBranchName`), the underlying tag is not on a branch (`git.GetCurrentBranchName(t) == ""`), or the commit SHA if not on a tag.

colens3:
Fixes #574 

- Updated GetCurrentBranchNameE function so that it returns empty string for detached state. GetCurrentBranchName function behaviour is updated too since it uses GetCurrentBranchNameE.
- Added GetCurrentGitRef and GetCurrentGitRefE functions which will return:
    - Branch name - in case of non-detached state
    - Light (non-annotated) tag - in case of detached state and non existing tag on commit
    - Tag - in case of detached state and existing tag on commit

Additional notes:
- Similar git command mocking approach used as within https://github.com/golang/go/blob/master/src/os/exec/exec_test.go

### DIFF
--- a/modules/git/git.go
+++ b/modules/git/git.go
@@ -14,12 +14,57 @@ func GetCurrentBranchName(t testing.TestingT) string {
 	if err != nil {
 		t.Fatal(err)
 	}
+	if out == "HEAD" {
+		return ""
+	}
 	return out
 }
 
-// GetCurrentBranchNameE retrieves the current branch name.
-func GetCurrentBranchNameE(t testing.TestingT) (string, error) {
+// GetCurrentBranchNameE retrieves the current branch name. Created as variable
+// to enable mocking within the tests.
+var GetCurrentBranchNameE = func(t testing.TestingT) (string, error) {
 	cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD")
+	bytes, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(bytes)), nil
+}
+
+// GetCurrentGitRef retrieves current branch name or most recent tag from a commit. If the tag
+// points to the commit, then only tag is returned.
+func GetCurrentGitRef(t testing.TestingT) string {
+	out, err := GetCurrentGitRefE(t)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return out
+}
+
+// GetCurrentGitRefE retrieves current branch name or most recent tag from a commit. If the tag
+// points to the commit, then only tag is returned.
+func GetCurrentGitRefE(t testing.TestingT) (string, error) {
+	out, err := GetCurrentBranchNameE(t)
+
+	if err != nil {
+		return "", err
+	}
+
+	if out != "" {
+		return out, nil
+	}
+
+	out, err = GetMostRecentTagE(t)
+	if err != nil {
+		return "", err
+	}
+	return out, nil
+}
+
+// GetMostRecentTagE retrieves most recent tag that is reachable from a commit. If the tag points
+// to the commit, then only the tag is returned. Created as variable to enable mocking within tests.
+var GetMostRecentTagE = func(t testing.TestingT) (string, error) {
+	cmd := exec.Command("git", "describe", "--tags")
 	bytes, err := cmd.Output()
 	if err != nil {
 		return "", err
--- a/modules/git/git.go
+++ b/modules/git/git.go
@@ -49,7 +49,7 @@ func GetCurrentGitRef(t testing.TestingT) string {
 
 // GetCurrentGitRefE retrieves current branch name, lightweight (non-annotated) tag or
 // if tag points to the commit exact tag value.
-var GetCurrentGitRefE = func(t testing.TestingT) (string, error) {
+func GetCurrentGitRefE(t testing.TestingT) (string, error) {
 	out, err := GetCurrentBranchNameE(t)
 
 	if err != nil {
--- a/modules/git/git.go
+++ b/modules/git/git.go
@@ -12,7 +12,8 @@ import (
 var branchNameExecCommand = exec.Command
 var tagExecCommand = exec.Command
 
-// GetCurrentBranchName retrieves the current branch name.
+// GetCurrentBranchName retrieves the current branch name or
+// empty string in case of detached state.
 func GetCurrentBranchName(t testing.TestingT) string {
 	out, err := GetCurrentBranchNameE(t)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


colens3:
Not sure if this is the format of the required test output, also please let me know if anything else needs to be changed :)
https://gist.github.com/colens3/593a9b74d672ed59b807e34b5156a9b7

modules/git/git.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
+// Separate variables created to be able to mock commands in tests
+var branchNameExecCommand = exec.Command
+var tagExecCommand = exec.Command
```
yorinasub17, line 13: I think we should run the actual command during testing. You can still test this accordingly by cloning a copy of terratest to a temporary folder and checking out various branches and tags.

modules/git/git.go
```diff
@@ -19,7 +23,54 @@ func GetCurrentBranchName(t testing.TestingT) string {
 
 // GetCurrentBranchNameE retrieves the current branch name.
```
yorinasub17, line 24: Can you update this comment to indicate that it will return empty string when we are not on a branch?

modules/git/git.go
```diff
@@ -19,7 +23,54 @@ func GetCurrentBranchName(t testing.TestingT) string {
...
+	out, err := GetCurrentGitRefE(t)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return out
+}
+
+// GetCurrentGitRefE retrieves current branch name, lightweight (non-annotated) tag or
+// if tag points to the commit exact tag value.
+var GetCurrentGitRefE = func(t testing.TestingT) (string, error) {
```
yorinasub17, line 52:
```suggestion
func GetCurrentGitRefE(t testing.TestingT) (string, error) {
```

modules/git/git_test.go
```diff
@@ -1,14 +1,85 @@
...
+	return cmd
+}
+
+func TestExecCommandHelper(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+
+	fmt.Fprintf(os.Stdout, os.Getenv("STDOUT"))
+	os.Exit(0)
```
yorinasub17, line 72: It's generally not a good idea to call `os.Exit` in test, as it can cause the test process to exit. `t.Fatal` is generally better.

modules/git/git_test.go
```diff
@@ -1,14 +1,85 @@
...
+	"os/exec"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetCurrentBranchName(t *testing.T) {
-	t.Parallel()
+var branchNameMock string
+var tagMock string
```
yorinasub17, line 13: As mentioned above, we will want to test if the `git` command is correct. We are more likely to get the `git` command wrong than the surrounding logic, so it's important that we run the actual command instead of mocking it out.

modules/git/git.go
colens3: Sure, done.

modules/git/git.go
colens3: Sure, makes sense. Will give it a try with proposed approach

modules/git/git_test.go
colens3: Got it. This part will be removed since we will not use command mocks.

modules/git/git.go
colens3: @yorinasub17 Performance results are not so good for terratest temporary repo clone (depends on the connection but on my end git module test execution lasts for around 20s). Is that acceptable or should we change the approach?

modules/git/git_test.go
colens3: Removed mock, switched to temp repo cloning strategy

colens3:
Hi @yorinasub17,
PR is ready for another review (whenever you find time)

modules/git/git_test.go
```diff
@@ -1,14 +1,89 @@
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetCurrentBranchName(t *testing.T) {
-	t.Parallel()
+func TestGetCurrentBranchNameReturnsBranchName(t *testing.T) {
+	err := exec.Command("git", "checkout", "master").Run()
+	assert.Nil(t, err)
```
yorinasub17, line 14: Can you actually clone the repo to a new temp dir and then run all of these tests in a subtest? That way, it doesn't have the surprising effect of changing what branch you are working on which can cause problems in CI when running all the tests.

Something like:

```go
import (
    "testing"
    "io/ioutil"
    "path/filepath"

    "github.com/gruntwork-io/terratest/modules/shell"
)

func TestGitRefChecks(t *testing.T) {
    t.Parallel()

    tmpdir, err := ioutil.TempDir("", "")
    require.NoError(t, err)
    defer os.RemoveAll(tmpdir)
    gitWorkDir := filepath.Join(tmpdir, "terratest")

    cmd := shell.Command{
        Command: "git",
        Args: []string{"clone", "https://github.com/gruntwork-io/terratest.git", gitWorkDir},
    }
    shell.RunCommand(t, cmd)

    t.Run("GetCurrentBranchNameReturnsBranchName", testGetCurrentBranchNameReturnsBranchName)
    t.Run("GetCurrentBranchNameReturnsEmptyForDetachedState", testetCurrentBranchNameReturnsEmptyForDetachedState)
    // ... and so on ...
}
```

colens3: I think I've done exactly that (sorry if i misunderstood).
Temp repo is cloned within the setup function which is run before whole test module (cleanup is added also).

Please check:
https://github.com/gruntwork-io/terratest/pull/588/files#diff-1b9f8aa5feac10022b44c29ce84f78e1R57-R70

yorinasub17: Ahhhh sorry I missed that when quickly glancing.

With that said, we prefer the pattern of using test functions and subtests as opposed to test main. Can you switch the pattern so you are using subtests instead as shown above?

colens3: Sure thing. Switched to subtests, ready for another review :)

yorinasub17:
Thanks for working through that with me! The updates look good! Just kicked off a build and if that passes, we can merge this in.

colens3:
Great! Thank you for the reviews and patience

yorinasub17:
The build failed, but it is because of a flaky packer issue. So this is good to go! Thanks again for your contribution!

PR #591, opened 2020-08-31
feat: add Azure FunctionApp Module

polatengin:
This PR is for adding Azure Function App module to the Terratest.

With this module, it can be possible to test Azure Function App resources on Terraform scripts.


HadwaAbdelhalem:
Hi @yorinasub17, PR is updated and ready for your review. Here is the related [CI](https://github.com/polatengin/terratest/runs/3081781881?check_suite_focus=true). the failure in it is unrelated to the change, but it is due to Azure [failure](https://github.com/polatengin/terratest/runs/3081781881?check_suite_focus=true#step:10:2072)  to create cosmos DB due to unavailability

modules/azure/appService.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/web/mgmt/2019-08-01/web"
+	"github.com/stretchr/testify/require"
+)
+
+// AppExistsE indicates whether the specified application exists.
```
yorinasub17, line 11:
```suggestion
// AppExists indicates whether the specified application exists.
```

yorinasub17 (commented):
LGTM! I just found one typo in the docs. I'll kick off a regression build now to make sure our precommit hooks run.

modules/azure/appService.go
HadwaAbdelhalem: Fixed

yorinasub17 (approved):
LGTM and sanity check build passed, so we can merge this in!

PR #594, opened 2020-09-02
feat: ci pipeline for azure

polatengin:
ci pipeline with manual trigger, for azure tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,28 +68,29 @@ jobs:
         env:
           AZURE_CREDENTIALS: ${{ secrets.AZURE_CREDENTIALS }}
         run: |
-          az account show
           cd test
-          TEST_VERSION=`date +"%Y%m%d%H%M"`
-          export ARM_SUBSCRIPTION_ID=`az account show --query "id" --output tsv`
 
           APP_ID=`echo $AZURE_CREDENTIALS | jq -r -c ".clientId"`
-
           APP_PASSWORD=`echo $AZURE_CREDENTIALS | jq -r -c ".clientSecret"`
-
           TENANT_ID=`echo $AZURE_CREDENTIALS | jq -r -c ".tenantId"`
 
+          # if clientId, subscriptionId, tenantId doesn't provide to the go tests
+          # by default, terratest reads them from below environment variables
           export ARM_CLIENT_ID="$APP_ID"
           export ARM_CLIENT_SECRET="$APP_PASSWORD"
-          export ARM_SUBSCRIPTION_ID="$ARM_SUBSCRIPTION_ID"
+          export ARM_SUBSCRIPTION_ID=`az account show --query "id" --output tsv`
           export ARM_TENANT_ID="$TENANT_ID"
 
+          # some resources may require ssh keys (e.g. Kubernetes, VMs, etc.)
+          # terraform will read below environment variables
+          # if those values didn't provide to the terraform explicitly
           rm -rf ssh_key*
           ssh-keygen -m PEM -t rsa -b 4096 -f ./ssh_key -q -N ""
           export TF_VAR_ssh_public_key="$PWD/ssh_key.pub"
           export TF_VAR_client_id="$APP_ID"
           export TF_VAR_client_secret="$APP_PASSWORD"
 
+          # run the actual tests under the `azure` subfolder
           go test ./azure/* -v -timeout 90m
       - name: report back the result
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,96 @@
+name: ci-workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      repo:
+        description: 'Repository info'
+        required: true
+      branch:
+        description: 'Name of the branch'
+        required: true
+      target_repository:
+        description: 'Repository name of the terratest'
+        required: true
+      target_pr:
+        description: 'PR number on the terratest repo'
+        required: true
+
+jobs:
+  ci-job:
+    runs-on: [ubuntu-latest]
+    steps:
+      - name: checkout to repo
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.event.inputs.repo }}
+          ref: ${{ github.event.inputs.branch }}
+      - name: list module files to lint
+        run: |
+          [ -d "modules/azure" ] && ls -li "modules/azure"
+      - name: lint modules/azure folder
+        id: azure_module_lint
+        uses: golangci/golangci-lint-action@v2.2.0
+        with:
+          version: v1.30
+          working-directory: modules/azure
+      - name: list test files to lint
+        run: |
+          [ -d "test/azure" ] && ls -li "test/azure"
+      - name: lint test/azure folder
+        id: azure_test_lint
+        uses: golangci/golangci-lint-action@v2.2.0
+        with:
+          version: v1.30
+          working-directory: test/azure
+      - name: login to azure cli
+        uses: azure/login@v1.1
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+      - name: run go test for azure
+        id: azure_test
+        env:
+          AZURE_CREDENTIALS: ${{ secrets.AZURE_CREDENTIALS }}
+        run: |
+          az account show
+          cd test
+          TEST_VERSION=`date +"%Y%m%d%H%M"`
+          export ARM_SUBSCRIPTION_ID=`az account show --query "id" --output tsv`
+
+          APP_ID=`echo $AZURE_CREDENTIALS | jq -r -c ".clientId"`
+
+          APP_PASSWORD=`echo $AZURE_CREDENTIALS | jq -r -c ".clientSecret"`
+
+          TENANT_ID=`echo $AZURE_CREDENTIALS | jq -r -c ".tenantId"`
+
+          export ARM_CLIENT_ID="$APP_ID"
+          export ARM_CLIENT_SECRET="$APP_PASSWORD"
+          export ARM_SUBSCRIPTION_ID="$ARM_SUBSCRIPTION_ID"
+          export ARM_TENANT_ID="$TENANT_ID"
+
+          rm -rf ssh_key*
+          ssh-keygen -m PEM -t rsa -b 4096 -f ./ssh_key -q -N ""
+          export TF_VAR_ssh_public_key="$PWD/ssh_key.pub"
+          export TF_VAR_client_id="$APP_ID"
+          export TF_VAR_client_secret="$APP_PASSWORD"
+
+          go test ./azure/* -timeout 90m
+      - name: report back the result
+        if: always()
+        env:
+          CURRENT_REPOSITORY: ${{ github.repository }}
+          RUN_ID: ${{ github.run_id }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TARGET_REPOSITORY: ${{ github.event.inputs.target_repository }}
+          TARGET_PR: ${{ github.event.inputs.target_pr }}
+          TEST_RESULT: ${{ steps.azure_test.conclusion }}
+          TEST_LINT_RESULT: ${{ steps.azure_test_lint.conclusion }}
+          MODULE_LINT_RESULT: ${{ steps.azure_module_lint.conclusion }}
+        run: |
+          if [ "$TEST_RESULT" == "success" ] && [ "$TEST_LINT_RESULT" == "success" ] && [ "$MODULE_LINT_RESULT" == "success" ]; then
+            BODY_PAYLOAD="[Microsoft CI Bot] TL;DR; success :thumbsup:\n\nYou can check the status of the CI Pipeline logs here ; https://github.com/${CURRENT_REPOSITORY}/actions/runs/$RUN_ID"
+            curl --request POST --header "Authorization: token ${GITHUB_TOKEN}" --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/${TARGET_REPOSITORY}/issues/${TARGET_PR}/comments --data "{\"body\":\"${BODY_PAYLOAD}\"}"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+        description: 'Name of the branch'
+        required: true
+      target_repository:
+        description: 'Repository name of the terratest'
+        required: true
+      target_pr:
+        description: 'PR number on the terratest repo'
+        required: true
+
+jobs:
```
yorinasub17, line 19: It looks like the job is running the linter for `modules/azure`, but not the tests. Is that intentional?

yorinasub17 (commented):
Had a few questions, but overall this makes sense to me!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+  workflow_dispatch:
+    inputs:
+      repo:
+        description: 'Repository info'
+        required: true
+      branch:
+        description: 'Name of the branch'
+        required: true
+      target_repository:
+        description: 'Repository name of the terratest'
```
brikis98, line 13: Repository name of the terratest repo, perhaps?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+        required: true
```
brikis98, line 14: Should these all have defaults? i.e., The official fork repo and official Terratest repo?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: login to azure cli
+        uses: azure/login@v1.1
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+      - name: run go test for azure
+        id: azure_test
+        env:
+          AZURE_CREDENTIALS: ${{ secrets.AZURE_CREDENTIALS }}
+        run: |
+          az account show
```
brikis98, line 55: Sanity check: this doesn't show any secrets or confidential info, does it?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          TARGET_REPOSITORY: ${{ github.event.inputs.target_repository }}
+          TARGET_PR: ${{ github.event.inputs.target_pr }}
+          TEST_RESULT: ${{ steps.azure_test.conclusion }}
+          TEST_LINT_RESULT: ${{ steps.azure_test_lint.conclusion }}
+          MODULE_LINT_RESULT: ${{ steps.azure_module_lint.conclusion }}
+        run: |
+          if [ "$TEST_RESULT" == "success" ] && [ "$TEST_LINT_RESULT" == "success" ] && [ "$MODULE_LINT_RESULT" == "success" ]; then
+            BODY_PAYLOAD="[Microsoft CI Bot] TL;DR; success :thumbsup:\n\nYou can check the status of the CI Pipeline logs here ; https://github.com/${CURRENT_REPOSITORY}/actions/runs/$RUN_ID"
+            curl --request POST --header "Authorization: token ${GITHUB_TOKEN}" --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/${TARGET_REPOSITORY}/issues/${TARGET_PR}/comments --data "{\"body\":\"${BODY_PAYLOAD}\"}"
+          elif [ "$TEST_RESULT" != "failure" ] || [ "$TEST_LINT_RESULT" != "failure" ] || [ "$MODULE_LINT_RESULT" != "failure" ]; then
```
brikis98, line 93: Why an `elif` here rather than just `else`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TARGET_REPOSITORY: ${{ github.event.inputs.target_repository }}
+          TARGET_PR: ${{ github.event.inputs.target_pr }}
+          TEST_RESULT: ${{ steps.azure_test.conclusion }}
+          TEST_LINT_RESULT: ${{ steps.azure_test_lint.conclusion }}
+          MODULE_LINT_RESULT: ${{ steps.azure_module_lint.conclusion }}
+        run: |
+          if [ "$TEST_RESULT" == "success" ] && [ "$TEST_LINT_RESULT" == "success" ] && [ "$MODULE_LINT_RESULT" == "success" ]; then
+            BODY_PAYLOAD="[Microsoft CI Bot] TL;DR; success :thumbsup:\n\nYou can check the status of the CI Pipeline logs here ; https://github.com/${CURRENT_REPOSITORY}/actions/runs/$RUN_ID"
+            curl --request POST --header "Authorization: token ${GITHUB_TOKEN}" --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/${TARGET_REPOSITORY}/issues/${TARGET_PR}/comments --data "{\"body\":\"${BODY_PAYLOAD}\"}"
```
brikis98, line 92: So these `curl` calls would update the test status on the official Terratest repo?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+
+jobs:
+  ci-job:
+    runs-on: [ubuntu-latest]
+    steps:
+      - name: checkout to repo
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.event.inputs.repo }}
+          ref: ${{ github.event.inputs.branch }}
```
brikis98, line 27: How does the code from the official Terratest repo end up in the fork?

.github/workflows/ci.yml
rguthriemsft: This will not show secrets.  The secrets for subscription access are stored in github secrets so the log should mask out the values.

.github/workflows/ci.yml
rguthriemsft: correct, @polatengin can confirm but I am pretty sure that is how it works.

.github/workflows/ci.yml
rguthriemsft: This should be normal process for fork development where once the fork is created, you would setup official repo as upstream and pull/merge changes from upstream master.  This will be a required step for PRs from forks to pull latest before submitting/approving PR.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          export ARM_SUBSCRIPTION_ID="$ARM_SUBSCRIPTION_ID"
+          export ARM_TENANT_ID="$TENANT_ID"
+
+          rm -rf ssh_key*
+          ssh-keygen -m PEM -t rsa -b 4096 -f ./ssh_key -q -N ""
+          export TF_VAR_ssh_public_key="$PWD/ssh_key.pub"
+          export TF_VAR_client_id="$APP_ID"
+          export TF_VAR_client_secret="$APP_PASSWORD"
+
+          go test ./azure/* -timeout 90m
```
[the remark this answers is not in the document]
polatengin: good point, I had that in mind. thanks for the feedback. I just updated the pr.

.github/workflows/ci.yml
polatengin: actually, I separated steps to run linter for `modules/azure` and `test/azure` folders.

[lint modules/azure folder](https://github.com/gruntwork-io/terratest/pull/594/files#diff-e9f950f17198d3d5e3122a44230a09b9R31)

[lint test/azure folder](https://github.com/gruntwork-io/terratest/pull/594/files#diff-e9f950f17198d3d5e3122a44230a09b9R40)

.github/workflows/ci.yml
polatengin: 🤦‍♂️ you're right, fixed..

.github/workflows/ci.yml
yorinasub17: Oh sorry, what I meant was that you weren't running the tests in `modules/azure` folder, and not that you weren't running the linter in tests folder.

If you look in [modules/azure](https://github.com/gruntwork-io/terratest/tree/master/modules/azure), you will find a few unit test files in there and I was wondering if we were intentionally going to remove those or not.

.github/workflows/ci.yml
polatengin: at the end of the run, CI pipeline creates a comment on the PR, saying either "success" or "failure" and the link to the run logs.

curl call creates that comment.

I just noticed, I should add some comments to make it more clear.

polatengin: added couple of comments to make the intent clear

.github/workflows/ci.yml
polatengin: we're checking the previous steps statuses, they can be "success", "failure", "skipped" ,etc.

I want to check if all the steps are success or at least 1 step failed.

I need to check 2 specific condition, that's why I can't use `else`.

.github/workflows/ci.yml
polatengin: that's correct, in the logs and Settings -> Secrets page, you can't see the secrets, it's always masked out.

.github/workflows/ci.yml
rguthriemsft: @polatengin we should run go test for unit test also.  Can we run those test first?

.github/workflows/ci.yml
polatengin: only target_repository can have a default value (which is gruntwork-io/terratest), I just pushed a commit to bring that default value

.github/workflows/ci.yml
polatengin: also, I added a comment block to clarify the fork->develop->pr flow

https://github.com/gruntwork-io/terratest/pull/594/files#diff-e9f950f17198d3d5e3122a44230a09b9R3-R16

.github/workflows/ci.yml
rguthriemsft: I spoke to @polatengin  and we are moving the test in modules to the test folder.  They are actually integration test.  We can discuss unit test approach in future PR.  He will add a commit that moves those test files. @yorinasub17  does that work?

yorinasub17: Yup that is fine with me! Just wanted to make sure how we wanted to do that going forward.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          export ARM_CLIENT_ID="$APP_ID"
+          export ARM_CLIENT_SECRET="$APP_PASSWORD"
+          export ARM_SUBSCRIPTION_ID="$ARM_SUBSCRIPTION_ID"
+          export ARM_TENANT_ID="$TENANT_ID"
+
+          rm -rf ssh_key*
+          ssh-keygen -m PEM -t rsa -b 4096 -f ./ssh_key -q -N ""
+          export TF_VAR_ssh_public_key="$PWD/ssh_key.pub"
+          export TF_VAR_client_id="$APP_ID"
+          export TF_VAR_client_secret="$APP_PASSWORD"
```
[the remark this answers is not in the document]
polatengin: I just pushed a commit, which adds comments to make it more clear.

TF_VAR variables are safe nets for the terraform scripts, terraform will use these variables as a fallback, if the values didn't provide to the script explicitly

.github/workflows/ci.yml
polatengin: I created another PR for that, it seemed not related to CI pipeline to me.

here is the PR : https://github.com/gruntwork-io/terratest/pull/596

.github/workflows/ci.yml
```diff
@@ -0,0 +1,115 @@
...
+# actors
+  # source repo: official terratest repo (gruntwork-io/terratest)
+  # forked repo: (e.g., xyz/terratest, abc/terratest)
+  # pr: created from forked repo, against source repo (gruntwork-io/terratest/pull/{PR NUMBER})
+# flow
+  # developer fork the source repo
+  # developer creates a branch in the forked repo
+  # developer does the development 😎
+  # developer creates a pr
+  # this pipeline will be triggered automatically, with the following input values
```
brikis98, line 12: Is it triggered automatically or manually? I thought `workflow_dispatch` was for manually triggered workflows?

.github/workflows/ci.yml
brikis98: Apologies if I'm missing something obvious, but it seems like the logic here is:

1. If all the checks were successful, update the PR with a success message.
1. Else if at least one of the checks was NOT failure, then update the PR with a failure message. 

I'm not sure I understand the second branch? Shouldn't we update the PR with failure if any of the checks were failures? Or even simpler, shouldn't it just be an "else"? As in, if all the checks weren't successful (the if-statement), then there must have been a failure, so update the PR accordingly?

.github/workflows/ci.yml
polatengin: yes, `workflow_dispatch` trigger type is for manually triggered workflows. I added a couple of lines to the comment section to make the flow more clear. webhook on source repo is triggering a webapp, webapp triggers this pipeline automatically. I hope this clarifies the flow. let me know if it needs to be improved.

.github/workflows/ci.yml
polatengin: what was I thinking when I wrote that line? 🤦‍♂️ I even thought I tested that clause (https://github.com/polatengin/terratest/pull/14#issuecomment-685994863)

it should be "if test step failed OR lint test folder step failed OR lint module folder step failed"

pushed the fix and tested with cases, it should be ok now, sorry for the confusion...

.github/workflows/ci.yml
brikis98: Ah, OK, that's very helpful, thanks!

.github/workflows/ci.yml
brikis98: OK, that new version makes much more sense 😄 

One more sanity check on these if-statements: what are the other possible statuses? Right now, it updates the PR with (a) a success message if everything succeeded or (b) a failure message if there were any failures. Should it update the PR on other statuses? E.g., If tests were skipped, would that be useful to know?

polatengin: conclusion of a step can be, `failure`, `success`, `cancelled` or `skipped`

as you already mentioned, we're creating a comment on either "all success" or "any failure" conditions.

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#steps-context

brikis98 (approved):
OK, I think this is good to go!

PR #595, opened 2020-09-02
fix: some dependencies bumped to latest versions

polatengin:
azure-sdk-for-go, go-autorest and gruntwork-cli dependencies bumped to the latest versions


rguthriemsft:
@polatengin, we should probably regenerate go.sum file with this PR similar to PR #609.

yorinasub17:
Closing as requested, so that it can be moved to the Microsoft fork.

PR #597, opened 2020-09-04
fix: existing azure examples are moved to azure subfolders

polatengin:
`terraform_azure_example_test.go` , `terraform_azure_aks_example_test.go` and related `terraform script files` are moved into `azure` subfolders.


brikis98 (commented):
Thanks for the PR!

We should define and document the folder structure we want. 

- E.g., If all Azure examples/tests go into `azure` subfolders, should all AWS examples/tests go into `aws` subfolders? Same for GCP? 
- What about examples that use technologies that work across clouds? E.g., K8S, Terraform, Packer, Docker?
- Would someone learning Terratest prefer to browse examples by _cloud_ or by _technology_? That is, what would be more expected for a newbie: [`examples/aws`, `examples/azure`, `examples/gcp`] or [`examples/terraform`, `examples/k8s`, `examples/docker`], etc?
- My guess is that the technology-focused examples would _probably_ be more natural to use... If so, could we adopt a convention where examples/tests that use cloud `xxx` include `xxx` in the example and test name? E.g., `examples/terraform/hello-world-aws` and `examples/terraform/hello-world-azure`. Could we use that to figure out which tests to run where? Or is that too complex?

@yorinasub17 Would welcome your thoughts here!

polatengin:
> We should define and document the folder structure we want.
> 
> * E.g., If all Azure examples/tests go into `azure` subfolders, should all AWS examples/tests go into `aws` subfolders? Same for GCP?

where should we document it? I can document for `azure`

brikis98:
> > We should define and document the folder structure we want.
> > 
> > * E.g., If all Azure examples/tests go into `azure` subfolders, should all AWS examples/tests go into `aws` subfolders? Same for GCP?
> 
> where should we document it? I can document for `azure`

Well, there are two ways we'd want to document this:

1. By updating the existing code into the folder structure we want. That way, by following the existing structure of the repo, most people will get it right automatically.
1. By documenting what structure we want and _why_ we want it in the [Terratest docs](https://terratest.gruntwork.io/), perhaps in the [contribution section](https://terratest.gruntwork.io/docs/community/contributing/)?

That said, the first step is to decide on the folder structure we want. I suggested two alternatives above:

1. Cloud at the top level, technology at the second level: e.g., `examples/aws/terraform-hello-world`, `examples/azure/terraform-hello-world`, `examples/azure/packer-hello-world`, etc.
1. Technology at the top level, cloud at the second level: e.g., `examples/terraform/hello-world-aws`, `examples/terraform/hello-world-azure`, `examples/packer/hello-world-azure`. 

I think I have a slight preference for the latter, but I'd love to hear what others think, or if there are other alternatives worth considering?

yorinasub17:
> Well, there are two ways we'd want to document this:

brikis98:
> Given that this is one of the key cornerstones for getting the azure workflow complete (including testing), I want to propose that we:
> 
> * Merge this PR as is.
> * File a issue with the desire to reorg the repo examples/test.
> * Iterate through the workflow to ensure the CI and PR flow works.
> * Write docs for Azure specific contributions. Mention the significance of the `azure` folder for `tests` and `examples` there.

OK, I'm on board with this! Could you take the lead on the first couple steps? I'm buried in calls today and want to unblock the Azure work. Thanks!

yorinasub17:
> Could you take the lead on the first couple steps?

Roger!

---

I'll kick off the build as a sanity check to make sure this doesn't break the other tests, and if that passes, I can go ahead and merge this in.

Also filed https://github.com/gruntwork-io/terratest/issues/600

brikis98:
Thank you @yorinasub17!

yorinasub17:
The build initially failed because this didn't include the context file, but it is now passing once I rebased: https://app.circleci.com/pipelines/github/gruntwork-io/terratest/115/workflows/51683ae9-3190-4c19-bf88-9dee5c8ec85f

However, I just realized that the docs may be incorrect. I'm standing up a local server now to double check.

yorinasub17:
Ok the docs do indeed need to be updated, so I did that here: https://github.com/gruntwork-io/terratest/pull/602

I believe this PR is linked to that one so this should autoclose when we merge that one.

yorinasub17:
This was merged as #602

PR #601, opened 2020-09-09
fix: linter steps on the CI Pipeline and not require couple of input variables

polatengin:
linter steps are merged with the related "list files" steps

also, we're using official golangci-lint binaries to run the linters, so, it'll be the same experience on local runs and pipeline runs

pr number input variable is not required anymore, if it's provided, then the result will be report back to the pr, if it's not provided, report back step simply exits


yorinasub17 (approved):
LGTM!

rguthriemsft:
Look good to me

Issue #603, opened 2020-09-09
Add negative test case to test TestGetTargetAzureSubscription in common_test.go


rguthriemsft:
Adding an issue to add a negative test case that test no subscription id passed to TestGetTargetAzureSubscription.

james03160927:
No activity for the past 5 years. Closing this issue.

PR #604, opened 2020-09-09
Fix failing test in common_test.go

rguthriemsft:
When we move to CI for Azure, ARM_SUBSCRIPTION_ID is required to be set for all test to run.  This requirement breaks the test TestGetTargetAzureSubscription case subIDNotProvided.  This test cannot pass in CI because if we don't set env all other test fail.  We will try to find a way to add this test case back in but we have modified the test case to test lookup of env value when no subscription id is passed to getTargetAzureSubscription in common.go.  I have added an issue to track adding back the negative test case.  https://github.com/gruntwork-io/terratest/issues/603

### DIFF
--- a/test/azure/common_test.go
+++ b/test/azure/common_test.go
@@ -23,6 +23,7 @@ func TestGetTargetAzureSubscription(t *testing.T) {
 	//Lookup ARM_SUBSCRIPTION_ID env variable, if not set, fail test.
 	if id, exists = os.LookupEnv(azure.AzureSubscriptionID); !exists {
 		fmt.Printf("ARM_SUBSCRIPTION_ID environment variable not set")
+		id = ""
 	}
 
 	type args struct {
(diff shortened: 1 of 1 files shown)


test/azure/common_test.go
```diff
@@ -15,6 +17,15 @@ import (
...
 	t.Parallel()
 
+	var id string
+	var exists bool
+
+	//Lookup ARM_SUBSCRIPTION_ID env variable, CI requires this value to run all test.
+	if id, exists = os.LookupEnv(azure.AzureSubscriptionID); !exists {
+		fmt.Printf("ARM_SUBSCRIPTION_ID environment variable not set")
+		id = ""
+	}
```
yorinasub17, line 27: Can this be simplified to:

```
require.NotEmpty(t, os.Getenv(azure.AzureSubscriptionID), "ARM_SUBSCRIPTION_ID environment variable not set")
```

This will fail the test if the environment variable is not set.

yorinasub17 (commented):
I think I understand what is going on here, but it took me a bit of staring. I have a suggestion to simplify this, but it is based on my understanding that what we really want to test here is that the specific env var is set.

test/azure/common_test.go
rguthriemsft: This is a better solutoin.  I updated the test to include this check.  We still need to add a test to test throwing the exception.  I have someone on my team looking into that now and will submit in a seperate PR.

yorinasub17 (approved):
LGTM!

PR #605, opened 2020-09-10
fix: ci pipeline report back step

polatengin:
instead of calling github rest api, ci pipeline creates a graphql payload and call the github graphql api.

because, it seems github action is running as a user and that user is degraded to readonly (https://github.com/actions/first-interaction/issues/10#issuecomment-546628432)

so, pipeline couldn't create a comment on the pr, changing the step to use the graphql endpoints fixing that problem.


polatengin:
test comment to check if PAT is working or not

PR #608, opened 2020-09-10
report back step is fixed

polatengin:
instead of calling github rest api, ci pipeline creates a graphql payload and call the github graphql api.

because, it seems github action is running as a user and that user is degraded to readonly (actions/first-interaction#10 (comment))

so, pipeline couldn't create a comment on the pr, changing the step to use the graphql endpoints fixing that problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
 
           if [[ -z "$BODY_PAYLOAD" ]]
           then
-                echo "BODY_PAYLOAD is empty"
+            echo "BODY_PAYLOAD is empty"
           else
             echo "Here is the target repository: ${TARGET_REPOSITORY}"
             # TARGET_REPOSITORY is in {owner}/{repo format}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,7 @@ jobs:
 
           # if pipeline has something to report back to the PR
 
-          if [ -z "$BODY_PAYLOAD" ]
+          if [[ -z "$BODY_PAYLOAD" ]]
           then
                 echo "BODY_PAYLOAD is empty"
           else
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,12 +109,49 @@ jobs:
           # if no PR number provided, simply exit...
           [[ -z "$TARGET_PR" ]] && { echo "No PR Number provided, exiting..." ; exit 0; }
 
+          echo "PR Number provided... ${TARGET_REPOSITORY}:#${TARGET_PR}"
+
+          BODY_PAYLOAD=""
           # if all the previous steps finished successfully, create a comment on the PR with the "success" information
           if [ "$TEST_RESULT" == "success" ] && [ "$TEST_LINT_RESULT" == "success" ] && [ "$MODULE_LINT_RESULT" == "success" ]; then
             BODY_PAYLOAD="[Microsoft CI Bot] TL;DR; success :thumbsup:\n\nYou can check the status of the CI Pipeline logs here ; https://github.com/${CURRENT_REPOSITORY}/actions/runs/$RUN_ID"
-            curl --request POST --header "Authorization: token ${GITHUB_TOKEN}" --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/${TARGET_REPOSITORY}/issues/${TARGET_PR}/comments --data "{\"body\":\"${BODY_PAYLOAD}\"}"
           # if at least one of the previous steps failed, create a comment on the PR with the "failure" information
           elif [ "$TEST_RESULT" == "failure" ] || [ "$TEST_LINT_RESULT" == "failure" ] || [ "$MODULE_LINT_RESULT" == "failure" ]; then
             BODY_PAYLOAD="[Microsoft CI Bot] TL;DR; failure :facepalm:\n\nYou can check the status of the CI Pipeline logs here ; https://github.com/${CURRENT_REPOSITORY}/actions/runs/$RUN_ID"
-            curl --request POST --header "Authorization: token ${GITHUB_TOKEN}" --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/${TARGET_REPOSITORY}/issues/${TARGET_PR}/comments --data "{\"body\":\"${BODY_PAYLOAD}\"}"
+          fi
+
+          echo "Comment message is ready..."
+          echo "${BODY_PAYLOAD}"
+
+          # if pipeline has something to report back to the PR
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -109,12 +109,49 @@ jobs:
...
             BODY_PAYLOAD="[Microsoft CI Bot] TL;DR; failure :facepalm:\n\nYou can check the status of the CI Pipeline logs here ; https://github.com/${CURRENT_REPOSITORY}/actions/runs/$RUN_ID"
-            curl --request POST --header "Authorization: token ${GITHUB_TOKEN}" --header "Accept: application/vnd.github.v3+json" https://api.github.com/repos/${TARGET_REPOSITORY}/issues/${TARGET_PR}/comments --data "{\"body\":\"${BODY_PAYLOAD}\"}"
+          fi
+
+          echo "Comment message is ready..."
+          echo "${BODY_PAYLOAD}"
+
+          # if pipeline has something to report back to the PR
+
+          if [ -z "$BODY_PAYLOAD" ]
```
yorinasub17, line 128: We follow [the google style guide for shell](https://google.github.io/styleguide/shellguide.html), and one of the recommendations is to prefer `[[]]` over `[]` (see [here](https://google.github.io/styleguide/shellguide.html#s6.3-tests) for explanation).

```suggestion
          if [[ -z "$BODY_PAYLOAD" ]]
```

.github/workflows/ci.yml
```diff
@@ -109,12 +109,49 @@ jobs:
...
+          then
+                echo "BODY_PAYLOAD is empty"
```
yorinasub17, line 130: NIT: Looks like this indentation is inconsistent with the else block.

yorinasub17 (commented):
> changing the step to use the graphql endpoints fixing that problem.

Does this actually work? After reading through the linked issue in its entirety, I am pretty sure this is a security vulnerability if it is working and we should report to Github as their intent of the read only token is clear 😅 This restriction also makes perfect sense: we only want to allow write access to workflows specified in a canonical base branch in the base repo as opposed to the fork so that a malicious fork can't suddenly go and delete all the metadata in a repo. So I wouldn't expect you to be able to write comments to the PR from the fork.

It looks like [the official solution](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run) here is to trigger a new workflow using [workflow_run](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run). I believe the idea is:

- Configure a workflow yaml file to run upon completion of a workflow (in our case, it will be this current workflow we have)
- Configure the jobs to check the results of the previous workflow that triggered the run (I had to dig around, but found [the reference to do this in the community forum](https://github.community/t/workflow-run-completed-event-triggered-by-failed-workflow/128001)).
- Based on the result, do actions (in this case, it will be the code you have right here).

Does that make sense? Could definitely use a second opinion here @rguthriemsft and @brikis98 

NOTE: based on how it works, we'd probably need to merge the workflow for commenting into `master` first before we can test this.

.github/workflows/ci.yml
polatengin: fixed it with a commit, thanks

polatengin:
> Does that make sense? Could definitely use a second opinion here @rguthriemsft and @brikis98

GitHub Actions are using pre-populated GITHUB_TOKEN secret to impersonate a user.

I was testing the pipeline on a few different repos. Now I'm figuring out, I was missing a critical point; all the repos I'm testing was in my account or in an org I'm owning.

so, GITHUB_TOKEN, I was using, had write permissions on PRs.

But running the pipeline on a completely different repo gives you a GITHUB_TOKEN which has write permission to that repo and read only permission to other repos.

We can split the workflow into 2 (1 for running linters and tests, 1 for creating the comment) but I think this approach will improve the complexity and it'll be relatively hard to debug and troubleshoot. 'cause first pipeline will run on the _special repo_ and the second pipeline will run on the _source repo_

I have another approach in mind, we can run the linters, tests AND creating comment in the same pipeline, by giving the pipeline a PAT

Because you can create and update comments in other repositories by using a PAT instead of GITHUB_TOKEN. The user associated with the PAT must have write access to the repository.

I just checked if I could create a comment on a PR on gruntwork-io/terratest repo , and I could (https://github.com/gruntwork-io/terratest/pull/605#issuecomment-690890565)

Only ran the commands in the report back step with a PAT I created on my account

> NOTE: based on how it works, we'd probably need to merge the workflow for commenting into `master` first before we can test this.

As you said, it's not easy to troubleshoot and debug a pipeline if it's not merged into master.

I think this is a valid and a good engineering feedback for GitHub, I'll create an engineering feedback internally.

yorinasub17:
Gotcha. So just to make sure I understand:

- We are setting up a PAT with write access as a secret in Microsoft fork, so that any actions run on Microsoft fork can write to PRs in Gruntwork terratest repo.
- Forks from Gruntwork terratest repo **does not** run the actions. Instead, they are run on the Microsoft fork (when approved to run).
- When a fork is approved to be run on Microsoft, a webhook triggers the github action on the Microsoft fork, which will run the CI. This will use the token configured on the Microsoft fork.
- Forks from the Microsoft repo **do not** get access to the token, due to the protection settings that prevent secrets from forwarding to fork actions.

Is that correct?

polatengin:
> * We are setting up a PAT with write access as a secret in Microsoft fork, so that any actions run on Microsoft fork can write to PRs in Gruntwork terratest repo.

Actually, people who are forking your repository will create the PAT and will set it as a secret, you don't have to create a PAT, especially your GITHUB_TOKEN already has write access to your repo

> * Forks from Gruntwork terratest repo **does not** run the actions. Instead, they are run on the Microsoft fork (when approved to run).

yep, that's correct, Azure Function _listens_ rest calls from the webhook you created (btw, we'll need to adjust it when we --as Microsoft-- has a new place to run it) and trigger the CI Pipeline on the Microsoft fork, which has an active Azure Subscription

> * When a fork is approved to be run on Microsoft, a webhook triggers the github action on the Microsoft fork, which will run the CI. This will use the token configured on the Microsoft fork.

Actually, when a PR is created against gruntwork-io/terratest and has "Azure" label, Azure Function will be invoked by the webhook and Azure Function will trigger the CI Pipeline on the fork on Microsoft account (microsoft/terratest or something like that)

> * Forks from the Microsoft repo **do not** get access to the token, due to the protection settings that prevent secrets from forwarding to fork actions.

Microsoft repo will be read-only and will not accept PRs, it's just for to have a place that has an active Azure Subscription

> Is that correct?

Sorry for the confusion, is it clarified now?

yorinasub17:
Yup it is clarified thanks! I do have a question about the workflow though. Will discuss in slack.

yorinasub17 (approved):

Issue #607, opened 2020-09-10
Package name incorrect for common_test.go, compute_test.go and region_test.go


rguthriemsft:
We found a bug after moving the test files into test/azure folder.  These three files have package azure (their previous package location) and need to be changed to package test.

rguthriemsft:
Please assign to @rguthriemsft

rguthriemsft:
fixed in #609

PR #609, opened 2020-09-10
Fix test packages - Issue #607

rguthriemsft:
This PR is to fix issue https://github.com/gruntwork-io/terratest/issues/607.  In fixing the package name we hit an issue with the golangci linter where the go.sum file was out of date.  We regenerated the go.sum file by executing:

rm -rf go.sum
go clean -modcache
go mod tidy

and committing the updated go.sum file.  We also increased the timeout for the linter from 1m0s to 5m0s.  I ran the full CI on my fork here https://github.com/rguthriemsft/terratest/actions/runs/248696494

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
           [ -d "modules/azure" ] && ls -li "modules/azure"
 
           # run the linter
-          ./bin/golangci-lint run ./modules/azure/ --build-tags=azure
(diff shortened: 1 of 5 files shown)


yorinasub17 (approved):
LGTM! Let me kick off a regression build to make sure the changes to go.sum doesn't break any of our non-azure tests.

yorinasub17:
Tests passed so merging this in!

Issue #622, opened 2020-09-14
Add support to Azure CI to skip provider registration.


rguthriemsft:
Currently terraform provider will try to register resource managers it needs to operate against azure by default.  If the user/service principal does not have owner rights to the subscription this will fail.  Terraform supports adding a flag to provider block to skip registration to support contributor access.  CI needs to add ARM_SKIP_PROVIDER_REGISTRATION to the env variables so a user can enable this for their CI to function correctly.

tsalright:
That comment looks good Richard. Thank you for logging this and getting a PR in place for it it. It makes much more sense to use the Environment Variable for this than adding it to the provider on each of the scripts.

brikis98:
Was this resolved by #624?

james03160927:
Closing this as this was resolved in other issue.

PR #624, opened 2020-09-14
add skip provider registration

rguthriemsft:
Fix to resolve issue #622

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,10 @@ on:
       target_pr:
         description: 'PR number on the official terratest repo'
         required: false
+      skip_provider_registration:
+        description: 'When set to true, terraform will skip provider registration (see: https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration for more information)'
+        required: true
+        default: 'false'
 
 jobs:
   ci-job:
@@ -82,10 +86,7 @@ jobs:
           export ARM_CLIENT_SECRET="$APP_PASSWORD"
           export ARM_SUBSCRIPTION_ID=`az account show --query "id" --output tsv`
           export ARM_TENANT_ID="$TENANT_ID"
-          
-          # This environment variable can be uncommented if the service principal only has contributor access and cannot register resource providers.  
-          # This will fail if the resource providers have never been registered in this subscription (see: https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration)
-          #export ARM_SKIP_PROVIDER_REGISTRATION=true
+          export ARM_SKIP_PROVIDER_REGISTRATION=${{ github.event.inputs.skip_provider_registration }}
 
           # some resources may require ssh keys (e.g. Kubernetes, VMs, etc.)
           # terraform will read below environment variables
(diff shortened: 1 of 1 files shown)


tsalright:
This looks great to me Richard. This will help me run individual validations against my employer's subscriptions where I don't have the owner rights to any of the subscriptions I would use and have ran into this many times trying to use Terraform to provision against Azure.

yorinasub17 (dismissed):
This LGTM, but I need some context on how this would be used. In particular, this seems to only apply if one is running the github action in their fork against their subscription. But is that a common model that we need to support?

And if someone is running the github action in their fork, will that conflict with the one run in the microsoft fork?

I guess I am confused as to when one would edit this action to uncomment this line of code, and how that workflow works for the contributor.

rguthriemsft:
@tsalright and @yorinasub17, I made a change to have this as an input with a default value of false.  This should elminate need to create a commit on user fork but still achieve desired behavior.  This is needed to support a user/service principal that has contributor access to subscription but not owner.  Terraform by default wants to register resource providers in Azure before it starts provisioning resources.  Contributors do not have permissions to register resource providers but do have permissions to provision resources which is common scenario.

rguthriemsft:
@tsalright, can you run test with this input fix and confirm it is working for you as expected?

tsalright:
@rguthriemsft, I will run it and get back to you shortly. I like the use of the variable to pass in on execution. Removes the need for me to do an extra commits to test the pipeline in my fork.

yorinasub17 (approved):
Thanks for the clarification. I understand the reason this needs to be set to run against certain subscriptions, but my question was around when would one be running against their subscription?

Sorry, I must be missing something obvious. In my mind, any test runs during development would be done locally and not through the CI workflow, and one would configure their access in the shell and not through the github action. When would one be running the actions on their fork?

tsalright:
@rguthriemsft that change worked for me. https://github.com/tsalright/terratest/runs/1113774135

rguthriemsft:
@yorinasub17, there is still the scenario where someone wanting to contribute to terratest, forks main repo, makes their change and then runs the full CI to ensure they didn't break any other code in terratest.  They may only be a contributor to the subscription they have access to do the validation in.  We had envisioned that with this CI job, a terratest contributor could run it against their subscription for their PR and expose the results as part of their PR.  as long as the CI shows it ran against same commit at head of their PR that should be valid assuming all the test passed.  Hope that makes sense?

yorinasub17:
Thanks for the explanation, but I think I am just struggling with the why. In any case, that point doesn't really matter for merging this in so will go ahead to merge this in.

If you want to know my thought process...

I am struggling to see what is the advantage of doing that from their fork in Github Actions in their subscription, as opposed to running it locally or waiting for the integration with the Microsoft fork. E.g., if we are going to run the full test anyway on the Microsoft fork, why would contributors want to setup their fork to run the github action?

If they want to test before they commit and push I am more inclined to recommend and support running it locally, outside of CI. This workflow actually gives a lot of benefits including:

- Being able to use `TestStages`.
- Being able to rerun `destroy` if the test fails with a panic or some other catastrophic way and you need to clean up the resources.
- Being able to run a subset of tests using `-run`.

I guess this comes from our experience with AWS and GCP contributions, where in almost all cases contributors would run a subset of the tests from their local machine against their sandbox account for verification testing, and rely on our build process in our account for the full test build.

PR #626, opened 2020-09-14
feat: adding Azure Availability Set module

### LINKED ISSUE #623: Add Azure Availability Set module
This module is intended to help test Azure Availability Set deployed with a set Fault Domain Count and a related Virtual Machine. This will provide sample Terraform deployment as well as testing examples.

davesee:
Implements basic exists and prop get example with an advanced VM contained check

Fixes #623


examples/azure/terraform-azure-availabilityset-example/README.md
```diff
@@ -0,0 +1,109 @@
+# Terraform Azure Availability Set Example
+
+This folder contains a simple Terraform module that deploys resources in [Azure](https://azure.microsoft.com/) to demonstrate
+how you can use Terratest to write automated tests for your Azure Terraform code. This module deploys an Availability Set with two attched Virtual Machines.
+
+* A [Availability Set](https://docs.microsoft.com/en-us/azure/virtual-machines/availability) that gives the module the following:
```
davidkpiano, line 6:
```suggestion
* An [Availability Set](https://docs.microsoft.com/en-us/azure/virtual-machines/availability) that gives the module the following:
```

rguthriemsft:
CI result here https://github.com/rguthriemsft/terratest/runs/1114813177?check_suite_focus=true

examples/azure/terraform-azure-availabilityset-example/README.md
```diff
@@ -0,0 +1,109 @@
+# Terraform Azure Availability Set Example
+
+This folder contains a simple Terraform module that deploys resources in [Azure](https://azure.microsoft.com/) to demonstrate
+how you can use Terratest to write automated tests for your Azure Terraform code. This module deploys an Availability Set with two attched Virtual Machines.
+
+* An [Availability Set](https://docs.microsoft.com/en-us/azure/virtual-machines/availability) that gives the module the following:
+    * `Availability Set` with the name specified in the `availability_set_name` output variable.
+    * `Fault Domain Count` with the value specified in the `availability_set_fdc` output variable.
+* A [Virtual Machines](https://azure.microsoft.com/en-us/services/virtual-machines/) that gives the Availability Set the following:
```
brikis98, line 9:
```suggestion
* [Virtual Machines](https://azure.microsoft.com/en-us/services/virtual-machines/) that gives the Availability Set the following:
```

examples/azure/terraform-azure-availabilityset-example/README.md
```diff
@@ -0,0 +1,109 @@
...
+## Running automated tests against this module
+
+1. Sign up for [Azure](https://azure.microsoft.com/).
+1. Configure your Azure credentials using one of the [supported methods for Azure CLI
+   tools](https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration?view=azure-cli-latest).
+1. Install [Terraform](https://www.terraform.io/) and make sure it's on your `PATH`.
+1. [Review environment variables](#review-environment-variables).
+1. Install [Golang](https://golang.org/) and make sure this code is checked out into your `GOPATH`.
+1. `cd test`
+1. Make sure [the azure-sdk-for-go versions match](#check-go-dependencies) in [/go.mod](/go.mod) and in [test/azure/terraform_azure_availabilityset_example_test.go](/test/azure/terraform_azure_availabilityset_example_test.go).
```
brikis98, line 41: Bit confused... Since dependencies are managed in `go.mod` and `go` now makes use of that automatically, how would the dependencies not match?

examples/azure/terraform-azure-availabilityset-example/README.md
```diff
@@ -0,0 +1,109 @@
...
+1. `go build terraform_azure_availabilityset_example_test.go`
+1. `go test -v -run TestTerraformAzureAvailabilitySetExample`
+
+## Module test APIs
```
brikis98, line 45: Does this belong in the example?

examples/azure/terraform-azure-availabilityset-example/README.md
```diff
@@ -0,0 +1,109 @@
...
+)
+```
+
+If we make changes to either the **go.mod** or the **go test file**, we should make sure that the go build command works still.
+
+```powershell
+go build terraform_azure_availabilityset_test.go
+```
+
+## Review Environment Variables
```
brikis98, line 91: I think we were discussing this in Slack, but it may make sense to extract this into, say, `examples/azure/README.md` and have all the individual examples cross-link to it rather than copy/pasting it, as this is a good amount of content.

examples/azure/terraform-azure-availabilityset-example/README.md
```diff
@@ -0,0 +1,109 @@
...
+- `GetAvailabilitySetVMs` gets a list of VM names in the specified Azure Availability Set\
+    func GetAvailabilitySetVMs(t testing.TestingT, avsName string, resGroupName string, subscriptionID string) []string
+    
+- `GetAvailabilitySetE` gets an Availability Set in the specified Azure Resource Group\
+    func GetAvailabilitySetE(t testing.TestingT, avsName string, resGroupName string, subscriptionID string) (*compute.AvailabilitySet, error)
+    
+- `GetAvailabilitySetClientE` gets a new Availability Set client in the specified Azure Subscription\
+    func GetAvailabilitySetClientE(subscriptionID string) (*compute.AvailabilitySetsClient, error)
+
+## Check Go Dependencies
```
brikis98, line 67: I'm a bit confused why this is necessary?

examples/azure/terraform-azure-availabilityset-example/main.tf
```diff
@@ -0,0 +1,104 @@
...
+resource "azurerm_availability_set" "main" {
+  name                        = "${var.prefix}-example"
+  location                    = azurerm_resource_group.main.location
+  resource_group_name         = azurerm_resource_group.main.name
+  platform_fault_domain_count = var.avs_fault_domain_count
+  managed                     = true
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY VIRTUAL NETWORK RESOURCES
```
brikis98, line 38: Is this necessary? Or could a default network be used to reduce the complexity of this example? For all our AWS examples, we always use the default VPC.

examples/azure/terraform-azure-availabilityset-example/main.tf
```diff
@@ -0,0 +1,104 @@
...
+
+  ip_configuration {
+    name                          = "terratestconfiguration1"
+    subnet_id                     = azurerm_subnet.internal.id
+    private_ip_address_allocation = "Dynamic"
+  }
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY VIRTUAL MACHINE TO THE AVAILABILITY SET
```
brikis98, line 68: NIT: add a comment that this VM doesn't really do anything; it's just an empty Ubuntu VM.

examples/azure/terraform-azure-availabilityset-example/main.tf
```diff
@@ -0,0 +1,104 @@
...
+    name              = "${var.prefix}-osdisk"
+    caching           = "ReadWrite"
+    create_option     = "FromImage"
+    managed_disk_type = "Standard_LRS"
+  }
+
+  os_profile {
+    computer_name  = "${var.prefix}-vm"
+    admin_username = var.username
+    admin_password = var.password
```
brikis98, line 98: What's the password for? Is this an SSH password? This brings up all sorts of security implications, so I'm wondering if we really need it for this example?

examples/azure/terraform-azure-availabilityset-example/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+variable "username" {
+  description = "The username to be provisioned into your VM"
+  type        = string
+  default     = "testadmin"
+}
+
+variable "vm_size" {
+  description = "The username to be provisioned into your VM"
+  type        = string
+  default ="Standard_B1s"
```
brikis98, line 54: Please make sure to run `terraform fmt` on this code.

brikis98: Also, are these the cheapest VMs available?

modules/azure/availabilityset.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// CheckAvailabilitySetContainsVM checks if the Virtual Machine is contained in the Availability Set VMs
+func CheckAvailabilitySetContainsVM(t testing.TestingT, vmName string, avsName string, resGroupName string, subscriptionID string) bool {
+	avsVMs, err := GetAvailabilitySetVMsE(t, avsName, resGroupName, subscriptionID)
+	if err != nil {
+		return false
+	}
+
+	return collections.ListContains(avsVMs, strings.ToLower(vmName))
```
brikis98, line 36: Why `strings.ToLower`? Is it all case insensitive?

modules/azure/availabilityset.go
```diff
@@ -0,0 +1,132 @@
...
+func AvailabilitySetExistsE(t testing.TestingT, avsName string, resGroupName string, subscriptionID string) (bool, error) {
+	_, err := GetAvailabilitySetE(t, avsName, resGroupName, subscriptionID)
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// CheckAvailabilitySetContainsVM checks if the Virtual Machine is contained in the Availability Set VMs
+func CheckAvailabilitySetContainsVM(t testing.TestingT, vmName string, avsName string, resGroupName string, subscriptionID string) bool {
```
brikis98, line 30: Should there be a `CheckAvailabilitySetContainsVME`?

modules/azure/availabilityset.go
```diff
@@ -0,0 +1,132 @@
...
+
+	avs, err := client.Get(context.Background(), resGroupName, avsName)
+	if err != nil {
+		return nil, err
+	}
+
+	vms := []string{}
+
+	for _, vm := range *avs.VirtualMachines {
+		vms = append(vms, strings.ToLower(GetNameFromResourceID(*vm.ID)))
```
brikis98, line 61: Same question here on `strings.ToLower`.

Also, is a `nil` check necessary on `*vm.ID`?

modules/azure/availabilityset.go
```diff
@@ -0,0 +1,132 @@
...
+// GetAvailabilitySetFaultDomainCount gets the Fault Domain Count for the specified Azure Availability Set
+func GetAvailabilitySetFaultDomainCount(t testing.TestingT, avsName string, resGroupName string, subscriptionID string) int32 {
+	avsFaultDomainCount, err := GetAvailabilitySetFaultDomainCountE(t, avsName, resGroupName, subscriptionID)
+	require.NoError(t, err)
+
+	return avsFaultDomainCount
+}
+
+// GetAvailabilitySetFaultDomainCountE gets the Fault Domain Count for the specified Azure Availability Set
+func GetAvailabilitySetFaultDomainCountE(t testing.TestingT, avsName string, resGroupName string, subscriptionID string) (int32, error) {
```
brikis98, line 76: Sanity check: this is a very thing wrapper around the underlying Azure SDK. In our [contribution guidelines](https://terratest.gruntwork.io/docs/community/contributing/#infrastructure-and-validation-helpers), to avoid Terratest becoming too bloated, we typically prefer to only include helpers that wrap some degree of complexity. Of course, this is balanced by the need to keep code short/DRY for very common operations, so if you expect a lot of test code to be looking up the fault domain count, then a short method for doing that may be worth maintaining.

modules/azure/resourceid.go
```diff
@@ -0,0 +1,36 @@
...
+	return lastValue
+}
+
+// GetSliceLastValueE will take a source string and returns the last value when split by the seperaror char
+func GetSliceLastValueE(source string, seperator string) (string, error) {
+	if !(len(source) == 0 || len(seperator) == 0 || !strings.Contains(source, seperator)) {
+		tmp := strings.Split(source, seperator)
+		return tmp[len(tmp)-1], nil
+	}
+	return "", errors.New("invalid input or no slice available")
```
brikis98, line 23: NIT: reading "not" logic is always harder mentally. Consider flipping the structure:

```go
if (len(source) == 0 || ...) {
  return "", errors.New()
}

tmp := strings...
```

brikis98: Also, we declare explicit types for each error so that in tests, you can check for those exact types. Could you replace `errors.New` with a custom type?

modules/azure/resourceid.go
```diff
@@ -0,0 +1,36 @@
...
+// GetSliceIndexValueE will take a source string and returns the value at the given index when split by the seperaror char
+func GetSliceIndexValueE(source string, seperator string, index int) (string, error) {
+	if !(len(source) == 0 || len(seperator) == 0 || !strings.Contains(source, seperator) || index < 0) {
+		tmp := strings.Split(source, seperator)
+		if !(len(tmp) >= index) {
+			return "", errors.New("index out of slice range")
+		}
+		return tmp[index], nil
+	}
+	return "", errors.New("invalid input or no slice available")
```
brikis98, line 35: Same note here on (a) inverting the structure and (b) custom error type.

modules/azure/resourceid.go
```diff
@@ -0,0 +1,36 @@
...
+func GetNameFromResourceID(resourceID string) string {
+	lastValue, err := GetSliceLastValueE(resourceID, "/")
+	if err != nil {
+		return ""
+	}
+	return lastValue
+}
+
+// GetSliceLastValueE will take a source string and returns the last value when split by the seperaror char
+func GetSliceLastValueE(source string, seperator string) (string, error) {
```
brikis98, line 18: This doesn't seem specific to Azure. Perhaps move to `collections` package?

brikis98: Also, this should have dedicated unit tests.

test/azure/availabilityset_test.go
```diff
@@ -0,0 +1,74 @@
+// +build azure
```
brikis98, line 1: Typically, the tests we put in the `test` package are tests for the example code in the `examples` folder (e.g., Terraform examples). We see these as "integration" tests.

If you have "unit" test that calls some method directly, without any other example code in between (e.g., no Terraform examples), then that should go into the same package as the method you're testing: e.g., if you're unit testing a method in `modules/azure/foo.go`, the tests go into `modules/azure/foo_test.go`.

test/azure/availabilityset_test.go
```diff
@@ -0,0 +1,74 @@
...
+	subscriptionID = ""
+	rgName         = "terratest-rg"
+)
+
+/*
+The below tests are currently stubbed out, with the expectation that they will throw errors.
+If/when methods to create and delete network resources are added, these tests can be extended.
+*/
+
+func TestGetAvailabilitySetClientE(t *testing.T) {
```
brikis98, line 27: It seems like the way to test these is to deploy the example you created above using Terratest's `terraform` package and then use these methods to validate that what got deployed is working...

modules/azure/resourceid.go
brikis98: Ah, I see some tests below. Those should be moved to `modules/azure/resourceid_test.go`.

test/azure/resourceid_test.go
```diff
@@ -0,0 +1,96 @@
...
+	// verify success
+	resultSuccess := azure.GetNameFromResourceID(sliceSource)
+	assert.Equal(t, sliceResult, resultSuccess)
+
+	// verify error when seperator not found
+	resultBadSeperator := azure.GetNameFromResourceID(sliceNotFound)
+	assert.Equal(t, "", resultBadSeperator)
+}
+
+func TestGetSliceLastValue(t *testing.T) {
```
brikis98, line 31: Please make use of [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests).

test/azure/terraform_azure_availabilityset_example_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTerraformAzureAvailabilitySetExample(t *testing.T) {
+	t.Parallel()
+
+	// subscriptionID is overridden by the environment variable "ARM_SUBSCRIPTION_ID"
+	subscriptionID := ""
+	prefix := "terratest-avs"
```
brikis98, line 22: This prefix should use `random.UniqueId()`. This ensures that the test resources all get unique names that won't conflict with other resources in the same Azure account, including other resources from tests running in parallel.

test/azure/terraform_azure_availabilityset_example_test.go
```diff
@@ -0,0 +1,58 @@
...
+	// At the end of the test, run `terraform destroy` to clean up any resources that were created
+	defer terraform.Destroy(t, terraformOptions)
+
+	// Run `terraform init` and `terraform apply`. Fail the test if there are any errors.
+	terraform.InitAndApply(t, terraformOptions)
+
+	// Run `terraform output` to get the values of output variables
+	resourceGroupName := terraform.Output(t, terraformOptions, "resource_group_name")
+	availabilitySetName := terraform.Output(t, terraformOptions, "availability_set_name")
+	fdc, _ := strconv.ParseInt(terraform.Output(t, terraformOptions, "availability_set_fdc"), 10, 32)
```
brikis98, line 44: Don't ignore errors that are returned! Call `require.NoError` on it instead.

brikis98 (commented):
Thanks for the PR!

test/azure/availabilityset_test.go
yorinasub17: Note that this was a decision we made to move all azure tests to the `test` package.

See https://github.com/gruntwork-io/terratest/pull/594#discussion_r483192207 for the relevant discussion.

I was on the fence, but decided that it was probably worth experimenting this style in Azure modules given that I personally always found it confusing to have these "unit" tests since it is almost always better to launch infrastructure with terraform than go.

brikis98: Hm, I think we want to keep the `test` folder solely for "integration tests" that have copy/pasteable examples for users. We don't want all our unit tests mixed up in here, as those are not something users would take advantage of...

rguthriemsft: This was an area we were not sure what the right approach was.  In summary, we should move unit test back under modules folder. as @yorinasub17 mentioned.  Will work with team on that.

yorinasub17: Fair enough and I am onboard with porting that back.

Side note:

> as those are not something users would take advantage of...

I think I keep coming back to the desire to have `examples-for-humans` and `examples-for-machines`, and `tests-for-humans` and `tests-for-machines`. Right now we are overloading the `examples` and `tests` folder to serve both purposes: for learning, and for testing the repo. I think we should call out which examples and tests are specifically for learning, and map out a learning path with increasingly complex examples.

I just find it awkward to create infrastructure in golang like we are doing in the `aws` module to test a bunch of things in unit, because we want to avoid polluting the `examples` folder which we view as copy pastable by humans. As mentioned above, it feels much cleaner to rely on `terraform` to create infrastructure than awkward SDK calls in golang...

examples/azure/terraform-azure-availabilityset-example/main.tf
rguthriemsft: In Azure there is no default network, hence the need to either provision one in the module or reference an already deployed virtual network from another deployment.  We wanted to make the example self contained and hence the vnet is in the example.

examples/azure/terraform-azure-availabilityset-example/main.tf
rguthriemsft: @davesee, can you look at dropping the os_profile from the example?  https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html#os_profile.  Not sure if that affects the test or not.

examples/azure/terraform-azure-availabilityset-example/variables.tf
rguthriemsft: Yes, this is a really small VM size. https://docs.microsoft.com/en-us/azure/virtual-machines/sizes-b-series-burstable

examples/azure/terraform-azure-availabilityset-example/README.md
davesee: Was originally 2 VMs but reduced to one so changed the plural to singular.

examples/azure/terraform-azure-availabilityset-example/README.md
davesee: Was copied from a previous example, removed erroneous instruction.

examples/azure/terraform-azure-availabilityset-example/README.md
davesee: Moved general instructions to examples/azure/README.md and linked to it in step 4.

examples/azure/terraform-azure-availabilityset-example/README.md
davesee: Was copied from a previous example, removed erroneous instruction.

examples/azure/terraform-azure-availabilityset-example/README.md
davesee: Removed.

examples/azure/terraform-azure-availabilityset-example/main.tf
davesee: It is necessary for the VM creation unfortunately. I put a note in the variable.tf file to highlight the implications.

variable.tf > password > description:
"Password for the VM OS, for demonstration ONLY! This should not be used with production resources"

Test error without OS Terraform section:
Error: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="Required parameter 'osProfile' is missing (null)." Target="osProfile"
...
--- FAIL: TestTerraformAzureAvailabilitySetExample (287.60s)

davesee: Since the VM is not being accessed, I can use the Terraform random_password generator to ensure stagnant resources are not compromised.

https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password

modules/azure/resourceid.go
davesee: Moved string/separator/slice/select functionality to collections, not sure what to name the file though.

test/azure/resourceid_test.go
davesee: Done, please see: modules/collections/stringslicevalue_test.go

modules/azure/resourceid.go
davesee: Please see: modules/collections/stringslicevalue.go

modules/azure/availabilityset.go
davesee: This is not necessary due to the for loop, if there are no VMs present, there will be no vm to nil panic on.

test/azure/availabilityset_test.go
davesee: Unit tests reverted to modules/azure

test/azure/availabilityset_test.go
davesee: Agreed, just stubbed out the methods to confirm consistency and signatures as noted in the file comments. I can remove if preferred.

New location: modules/azure/availabilityset_test.go

modules/azure/availabilityset.go
davesee: The resource ID in these calls come all CAPS.. I changed the method name to reflect that so there is no ambiguity and removed the lowercase conversion.

modules/azure/availabilityset.go
davesee: Added.

test/azure/terraform_azure_availabilityset_example_test.go
davesee: Used the native Terraform random_strings resource to generate unique names. The names are collected in outputs, removing the need for the prefix. See main.tf for the implementation. 

https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/string

test/azure/terraform_azure_availabilityset_example_test.go
davesee: Error properly handled.

examples/azure/terraform-azure-availabilityset-example/variables.tf
davesee: It is very small, updated to the absolutely smallest Linux size, Standard_B1ls

modules/azure/resourceid.go
davesee: Removed not logic and created custom error, NewSliceValueNotFoundError

modules/azure/resourceid.go
davesee: Removed not logic and used custom error.

examples/azure/terraform-azure-availabilityset-example/main.tf
davesee: Comment added.

modules/azure/availabilityset.go
davesee: Kept since is helpful. Reduced complexity.

davesee: The method was shortened, also kept as it provides a useful property return and a sample for similar getters.

test/azure/terraform_azure_availabilityset_example_test.go
davesee: Please provide feedback on the random_string implementation and which strategy is best:
1. random_string and name concatenation in terraform, values are red from output.
2.  terraform.Options with one prefix using the random.UniqueId() and concatenation using terraform vars "${var.prefix}-network"
3. terraform.Options with one random.UniqueId() per resource name

Of these, the team saw #3 implemented in the azure and is considered the current standard.

examples/azure/terraform-azure-availabilityset-example/README.md
```diff
@@ -0,0 +1,41 @@
...
+automated tests for this module.
+
+Note that the Availability Set and VM in this module don't actually do anything; it just runs the resources for
+demonstration purposes.
+
+**WARNING**: This module and the automated tests for it deploy real resources into your Azure account which can cost you
+money. The resources are all part of the [Azure Free Account](https://azure.microsoft.com/en-us/free/), so if you haven't used that up,
+it should be free, but you are completely responsible for all Azure charges.
+
+## Running this module manually
```
yorinasub17, line 22: These instructions should link to the root README `Review Environment Variables` section.

examples/azure/terraform-azure-availabilityset-example/main.tf
```diff
@@ -0,0 +1,124 @@
...
+# The examples have been upgraded to 0.12 syntax
+# ---------------------------------------------------------------------------------------------------------------------
+
+terraform {
+  required_version = ">= 0.12"
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# See test/azure/terraform_azure_availabilityset_example_test.go for how to write automated tests for this code.
+# ---------------------------------------------------------------------------------------------------------------------
```
yorinasub17, line 17: NIT: This block should be at the top of the file.

modules/azure/availabilityset.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"context"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// AvailabilitySetExists indicates whether the speficied Azure Availability Set exists
```
yorinasub17, line 12: NIT: All the functions that are not `E` should have the following text:

```suggestion
// AvailabilitySetExists indicates whether the speficied Azure Availability Set exists. This function would fail the test if there is an error.
```

modules/collections/stringslicevalue.go
```diff
@@ -0,0 +1,26 @@
+package collections
+
+import (
+	"strings"
+)
+
+// GetSliceLastValueE will take a source string and returns the last value when split by the seperaror char
```
yorinasub17, line 7:
```suggestion
// GetSliceLastValueE will take a source string and returns the last value when split by the seperator char
```

modules/collections/stringslicevalue.go
```diff
@@ -0,0 +1,26 @@
...
+func GetSliceLastValueE(source string, seperator string) (string, error) {
+	if len(source) > 0 && len(seperator) > 0 && strings.Contains(source, seperator) {
+		tmp := strings.Split(source, seperator)
+		return tmp[len(tmp)-1], nil
+	}
+	return "", NewSliceValueNotFoundError(source)
+}
+
+// GetSliceIndexValueE will take a source string and returns the value at the given index when split by the seperaror char
```
yorinasub17, line 16:
```suggestion
// GetSliceIndexValueE will take a source string and returns the value at the given index when split by the seperator char
```

yorinasub17 (dismissed):
The updates LGTM! I had a bunch of final nits for the documentation: none of them major. Once those are addressed I think this one is good to go!

modules/azure/aks.go
```diff
@@ -7,7 +7,8 @@ import (
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
-// GetManagedClustersClientE is a helper function that will setup an Azure ManagedClusters client on your behalf
+// GetManagedClustersClientE is a helper function that will setup an Azure ManagedClusters client on your behalf.
+// This function would fail the test if there is an error.
```
yorinasub17, line 11: Oh sorry if I wasn't clear, but I meant all the functions that don't have the `E`. The convention is:

- `xxx` function should fail the test instead of returning errors when there is an error, and should include that comment.
- `xxxE` functions should return errors instead of failing the tests.

davesee: Yes, that makes perfect sense. I missed the "not" and just updated it previously without considering the implication.

modules/collections/stringslicevalue.go
davesee: Fixed all occurrences.

modules/collections/stringslicevalue.go
davesee: Fixed all occurrences.

modules/azure/availabilityset.go
davesee: Fixed all occurrences.

examples/azure/terraform-azure-availabilityset-example/main.tf
davesee: Fixed all occurrences. I updated the other Azure main.tf files as well with a header format that combines the recently added description and the test instructions for consistency.

examples/azure/terraform-azure-availabilityset-example/README.md
davesee: Added step root doc ref.

modules/azure/aks.go
davesee: Made changes, running CI for success and will update when complete.

davesee: https://github.com/rguthriemsft/terratest/runs/1150630104?check_suite_focus=true

Completed CI successfully.

davesee:
https://github.com/rguthriemsft/terratest/runs/1150630104?check_suite_focus=true

yorinasub17 (commented):
Updates LGTM! I am going to kick off a build in our CI as well so that we run the tests for the `collections` library.

@rguthriemsft @polatengin 
One thing I noticed is that since we restored the tests in to `modules/azure`, I think we need to update the github action so that it runs those tests. I think we will need the updated the flow + test results to make sure the tests in that folder are passing before we merge this in.

So to summarize the remaining TODOs here are:

- [x] ~Wait for Gruntwork terratest build of core library~ Build failed, but it is unrelated to this PR so we can consider this a pass!
- [x] ~Update Azure Github actions to run the unit tests in `modules/azure`~ Done in https://github.com/gruntwork-io/terratest/pull/651
- [ ] Wait for report from updated test run.

Does that make sense?

rguthriemsft:
@yorinasub17 I would like to submit an update to the CI to enable unit test as a seperate PR to keep the history clean if that is ok.  Will try to add it shortly.

rguthriemsft:
CI ran with unit test execution here https://github.com/rguthriemsft/terratest/actions/runs/269456385

yorinasub17:
Thanks! Was that on the latest commit (99191cc)?

modules/azure/availabilityset.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"context"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// AvailabilitySetExists indicates whether the speficied Azure Availability Set exists.
```
yorinasub17, line 12:
```suggestion
// AvailabilitySetExists indicates whether the specified Azure Availability Set exists.
```

modules/azure/availabilityset.go
```diff
@@ -0,0 +1,153 @@
...
+// This function would fail the test if there is an error.
+func AvailabilitySetExists(t testing.TestingT, avsName string, resGroupName string, subscriptionID string) bool {
+	exists, err := AvailabilitySetExistsE(t, avsName, resGroupName, subscriptionID)
+	require.NoError(t, err)
+	return exists
+}
+
+// AvailabilitySetExistsE indicates whether the speficied Azure Availability Set exists
```
yorinasub17, line 20:
```suggestion
// AvailabilitySetExistsE indicates whether the specified Azure Availability Set exists
```

yorinasub17 (commented):
Took one more pass and this LGTM! Spotted two more typos in comments that I missed in past go arounds, but from our side I think this is good to go.

I noticed there was still some work being done on this though: is that all done?

yorinasub17 (approved):
This is good to go! Merging in now!

Issue #621, opened 2020-09-14
Need Azure Resource Group module 


wmattlong:
Need to Azure resource group module for checking the existence of a resource group.

HadwaAbdelhalem:
@brikis98 , would you please close it as PR merged

brikis98:
Thx for the heads up. Closing!

PR #637, opened 2020-09-16
pull-request-621

skunklab:
Adds the following Azure resourcegroup module for terratest
(i) Terraform example in examples/azure folder
(ii) resouregroup.go module in modules/azure folder
(iii) Unit test in test/azure folder
(iv) Integration test that runs example and makes assertions in test/azure folder.

### DIFF
--- a/examples/azure/terraform-azure-resourcegroup-example/main.tf
+++ b/examples/azure/terraform-azure-resourcegroup-example/main.tf
@@ -1,0 +1,4 @@
+resource "azurerm_resource_group" "main" {
+  name     =  "${var.prefix}-${var.resourceGroupName}"
+  location = var.location
+}
--- a/modules/azure/resourcegroup.go
+++ b/modules/azure/resourcegroup.go
@@ -1,0 +1,48 @@
+package azure
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
+	"github.com/stretchr/testify/require"
+)
+
+// ResourceGroupExists indicates whether a resource group exists within a subscription; otherwise false
+func ResourceGroupExists(t *testing.T, resourceGroupName string, subscriptionID string) bool {
+	result, err := ResourceGroupExistsE(resourceGroupName, subscriptionID)
+	require.NoError(t, err)
+	return result
+}
+
+// ResourceGroupExistsE indicates whether a resource group exists within a subscription
+func ResourceGroupExistsE(resourceGroupName, subscriptionID string) (bool, error) {
+	subscriptionID, err := getTargetAzureSubscription(subscriptionID)
+	if err != nil {
+		return false, err
+	}
+	client, err := GetResourceGroupClientE(subscriptionID)
+	if err != nil {
+		return false, err
+	}
+	rg, err := client.Get(context.Background(), resourceGroupName)
+	if err != nil {
+		return false, err
+	}
+	return (resourceGroupName == *rg.Name), nil
+}
+
+//GetResourceGroupClientE gets a resource group client in a subscription
+func GetResourceGroupClientE(subscriptionID string) (*resources.GroupsClient, error) {
+	subscriptionID, err := getTargetAzureSubscription(subscriptionID)
+	if err != nil {
+		return nil, err
+	}
+	resourceGroupClient := resources.NewGroupsClient(subscriptionID)
+	authorizer, err := NewAuthorizer()
+	if err != nil {
+		return nil, err
+	}
+	resourceGroupClient.Authorizer = *authorizer
+	return &resourceGroupClient, nil
+}
--- a/examples/azure/terraform-azure-resourcegroup-example/README.md
+++ b/examples/azure/terraform-azure-resourcegroup-example/README.md
@@ -1,0 +1,105 @@
+# Terraform Azure Resource Group Example
+
+
+This folder contains a simple Terraform module that deploys resources in [Azure](https://azure.microsoft.com/) to demonstrate how you can use Terratest to write automated tests for your Azure Terraform code. This module deploys the following:
+
+
+* A [Resource Group](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/overview) that gives the module the following:
+    * [Resource Group](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/overview)  with the value specified in the `resource_group_name` output variable.
+
+Check out [test/azure/terraform_azure_@module@_test.go](/test/azure/terraform_azure_@module@_test.go) to see how you can write
+automated tests for this module.
+
+Note that the resources deployed in this module don't actually do anything; it just runs the resources for
+demonstration purposes.
+
+**WARNING**: This module and the automated tests for it deploy real resources into your Azure account which can cost you
+money. The resources are all part of the [Azure Free Account](https://azure.microsoft.com/en-us/free/), so if you haven't used that up,
+it should be free, but you are completely responsible for all Azure charges.
+
+## Running this module manually
+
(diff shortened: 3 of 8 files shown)


examples/azure/terraform-azure-resourcegroup-example/main.tf
```diff
@@ -0,0 +1,32 @@
...
+  # 0.12.26 as the minimum version, as that version added support for required_providers with source URLs, making it
+  # forwards compatible with 0.13.x code.
+  required_version = ">= 0.12.26"
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY A RESOURCE GROUP
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "azurerm_resource_group" "resourcegroup" {
```
rguthriemsft, line 29: Gruntwork style is to use underscore_case for labels.  label should be resource_group

modules/azure/resourcegroup_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+/*
+The below tests are currently stubbed out, with the expectation that they will throw errors.
+If/when CRUD methods are introduced for Azure Virtual Machines, these tests can be extended
+(see AWS S3 tests for reference).
```
rguthriemsft, line 12: refactor comment.

examples/azure/terraform-azure-resourcegroup-example/main.tf
```diff
@@ -0,0 +1,32 @@
...
+  name     = "terratest-rg-${var.postfix}"
```
rguthriemsft, line 30: Move base part of name to variable.

test/azure/terraform_azure_resourcegroup_example_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTerraformAzureResourceGroupExample(t *testing.T) {
+	t.Parallel()
+
+	// subscriptionID is overridden by the environment variable "ARM_SUBSCRIPTION_ID"
+	subscriptionID := ""
```
rguthriemsft, line 21: Set base part of name.  Test that name = base + postifx and exist.

examples/azure/terraform-azure-resourcegroup-example/main.tf
skunklab: done

modules/azure/resourcegroup_test.go
skunklab: done

examples/azure/terraform-azure-resourcegroup-example/main.tf
skunklab: done

test/azure/terraform_azure_resourcegroup_example_test.go
skunklab: done

rguthriemsft:
@yorinasub17 @skunklab This PR is ready for final review and merge. 
 Here is passing CI run.  https://github.com/rguthriemsft/terratest/actions/runs/280816954

rguthriemsft:
Closes #621

rguthriemsft (dismissed):

modules/azure/resourcegroup.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return false, err
+	}
+	client, err := GetResourceGroupClientE(subscriptionID)
+	if err != nil {
+		return false, err
+	}
+	rg, err := client.Get(context.Background(), resourceGroupName)
+	if err != nil {
+		return false, err
```
yorinasub17, line 31: This should check the error for the "not exists" case and return `nil` for the error. Otherwise, `ResourceGroupExists` will never return `false`. See https://github.com/gruntwork-io/terratest/blob/master/modules/azure/networkinterface.go#L20 where we do this.

modules/azure/resourcegroup_test.go
```diff
@@ -0,0 +1,20 @@
...
+The below tests are currently stubbed out, with the expectation that they will throw errors.
+If/when methods to create and delete resource groups are added, these tests can be extended.
+*/
+
+func TestResourceGroupExists(t *testing.T) {
+	t.Parallel()
+
+	resourceGroupName := "fakeResourceGroupName"
+	_, err := ResourceGroupExistsE(resourceGroupName, "")
+	require.Error(t, err)
```
yorinasub17, line 19: Should this actually check if it returns `false` without error? Since we are calling it on a resource group that we don't expect to exist?

yorinasub17 (commented):
Implementation LGTM! I had a few style nits, and one potential bug about the way we are returning `false`.

modules/azure/resourcegroup.go
skunklab: The E suffix function will attempt to get required resources, i.e., subscription, resource group client, etc. and if successfully compare the resource group name to the expected parameter input (T|F).  This is same pattern used in previous approved Exists functions.

modules/azure/resourcegroup_test.go
skunklab: This pattern currently used in previously approved Exists unit test functions, i.e., getting error when subscription and/or resource group no available.

modules/azure/resourcegroup.go
```diff
@@ -0,0 +1,50 @@
...
+
+// ResourceGroupExists indicates whether a resource group exists within a subscription; otherwise false
+func ResourceGroupExists(t *testing.T, resourceGroupName string, subscriptionID string) bool {
+	result, err := ResourceGroupExistsE(resourceGroupName, subscriptionID)
+	require.NoError(t, err)
+	return result
+}
+
+// ResourceGroupExistsE indicates whether a resource group exists within a subscription
+// This function would fail the test if there is an error.
```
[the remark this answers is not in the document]
skunklab: done

modules/azure/resourcegroup.go
yorinasub17: The expectation of a `ResourceGroupExists` function that returns `bool` is to return `false` if the resource doesn't exist, not to fail the test if it doesn't exist (if we want it to fail, then it should be called `AssertResourceGroupExists`).

Right now, as implemented, `ResourceGroupExists` will never return `false` because all the cases that return `false` in `ResourceGroupExistsE` returns it with a non-nil error (so the `require.NoError` call will fail the test). This breaks the expectation of the function.

So we should either:

- Add in the check to the error to only bubble it up if it is an error that doesn't indicate the resource doesn't exist.
- Rename the function to `AssertResourceGroupExists` and remove the `bool` return value.

> This is same pattern used in previous approved Exists functions.

We may have slipped through this on previous functions. I caught this in the PR referenced above. We should agree on what the right pattern is here, and fix it if it isn't the case in the merged code.

rguthriemsft: We should have exist. @skunklab see https://github.com/gruntwork-io/terratest/blob/48e24df893c925adef331c342b7d3979da08f51c/modules/azure/availabilityset.go#L22.  This pattern should replace the nil check for line 30-32.

skunklab: modified

rguthriemsft (approved):
Look good to me, Passing CI https://github.com/rguthriemsft/terratest/actions/runs/298025567

rguthriemsft:

You can check the status of the CI Pipeline logs here ; https://github.com/rguthriemsft/terratest/actions/runs/298025567

yorinasub17 (approved):
LGTM! Pardon the delay in getting this reviewed.

PR #618, opened 2020-09-14
feat: add Azure Disk module

davidkpiano:
This PR adds the module for [Azure managed disks](https://azure.microsoft.com/en-us/services/storage/disks).

Closes issue #614

PR #641, opened 2020-09-16
feat: adding Virtual Machine module

### LINKED ISSUE #635: Add Azure Virtual Machine module
This module advances the compute module and provide information its integrated resources.

Examples are divided into sub-tests that include examples that test for the following.
- Strategies: VM properties are retrieved using a single value get, by the object reference or by an instance getter
- MultiipleVMs: Check several VMs at once in a given resource group
- Information: Key VM properties are inspected for deployment confirmation
- Availability Set: Availability Set client and associated properties
- Disk: Disk client is leveraged since offline VMs do not report complete disk properties
- Network Interface: Networking and IP related VM properties
- Virtual Network & Subnet: Networking related VM properties

davesee:
Adding more comprehensive VM and dependency resource tests.

Fixes #635


davesee:
CI Pass: https://github.com/rguthriemsft/terratest/runs/1176719381?check_suite_focus=true

examples/azure/terraform-azure-vm-example/main.tf
```diff
@@ -0,0 +1,159 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY A VIRTUAL MACHINE
+# This is an advanced example of how to deploy an Azure Virtual Machine in an availability set, managed disk 
+# and Networking with a Public IP.
+# ---------------------------------------------------------------------------------------------------------------------
+# See test/azure/terraform_azure_vm_example_test.go for how to write automated tests for this code.
+# ---------------------------------------------------------------------------------------------------------------------
+
+provider "azurerm" {
+  version = "~> 2.20"
```
rguthriemsft, line 10: bump to latest azurerm provider 2.29 or greater

examples/azure/terraform-azure-vm-example/main.tf
```diff
@@ -0,0 +1,159 @@
...
+  # 0.12.26 as the minimum version, as that version added support for required_providers with source URLs, making it
+  # forwards compatible with 0.13.x code.
+  required_version = ">= 0.12.26"
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY A RESOURCE GROUP
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "azurerm_resource_group" "vm" {
```
rguthriemsft, line 30: name vm_rg

modules/azure/compute.go
```diff
@@ -8,97 +8,444 @@ import (
...
-		return nil, err
+		return nics, err
 	}
 
-	// Attach authorizer to the client
-	vmClient.Authorizer = *authorizer
+	vmNICs := *vm.NetworkProfile.NetworkInterfaces
+	if len(vmNICs) == 0 {
+		// If the NIC is not present
+		return nics, NewNotFoundError("Network Interface", "Any", vmName)
```
rguthriemsft, line 54: Is this really an error if no nics are present on the VM?

modules/azure/compute.go
```diff
@@ -8,97 +8,444 @@ import (
...
+// GetVirtualMachineAvailabilitySetIDE gets the Availability Set ID of the specified Azure Virtual Machine.
+func GetVirtualMachineAvailabilitySetIDE(vmName string, resGroupName string, subscriptionID string) (string, error) {
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
+	if err != nil {
+		return "", err
+	}
+
+	if vm.AvailabilitySet == nil {
+		return "", NewNotFoundError("Availability Set", "", vmName)
```
rguthriemsft, line 191: If the VM is not in an availability set, that is not necessarily an error but maybe you are thinking that if someone called this function and no availability set was found, this should be an error.  Am I understanding that correctly?

modules/azure/compute.go
```diff
@@ -8,97 +8,444 @@ import (
...
+	var vmImage VMImage
+
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
+	if err != nil {
+		return vmImage, err
+	}
+
+	if vm.StorageProfile == nil {
+		return vmImage, NewNotFoundError("Image Reference", "Any", vmName)
```
rguthriemsft, line 225: I think the string in this function in errors. go is wrong, your third param is not a region as it should not be.  Should we refactor in this PR or seperate?

func (err NotFoundError) Error() string {
	return fmt.Sprintf("Object of type %s with id %s not found in region %s", err.objectType, err.objectID, err.region)
}

modules/azure/compute.go
```diff
@@ -8,97 +8,444 @@ import (
...
+	return tags, nil
+}
+
+// ***************************************************** //
+// Get multiple Virtual Machines from a Resource Group
+// ***************************************************** //
+
+// GetResourceGroupVirtualMachines gets a list of all Virtual Machine names in the specified Resource Group.
+// This function would fail the test if there is an error.
+func GetResourceGroupVirtualMachines(t testing.TestingT, resGroupName string, subscriptionID string) []string {
```
rguthriemsft, line 310: should function name be ListVirtualMachinesForResourceGroup?

modules/azure/compute.go
```diff
@@ -8,97 +8,444 @@ import (
...
+
+	for _, v := range vms.Values() {
+		vmDetails = append(vmDetails, *v.Name)
+	}
+	return vmDetails, nil
+}
+
+// GetResourceGroupVirtualMachinesObjects gets all Virtual Machine objects in the specified Resource Group.
+// This function would fail the test if there is an error.
+func GetResourceGroupVirtualMachinesObjects(t testing.TestingT, resGroupName string, subscriptionID string) *map[string]compute.VirtualMachineProperties {
```
rguthriemsft, line 338: should name be GetVirtualMachinesForResourceGroup or ListVirtualMachinesForResourceGroup?

rguthriemsft (changes requested):

modules/azure/compute.go
davesee: Correct again, removed error.

modules/azure/compute.go
davesee: There is a NewNotFoundError now also in modules/azure/errors.go, this one is more generic.

modules/azure/compute.go
davesee: Updated

modules/azure/compute.go
davesee: Updated

davesee:
Updates complete.
CI Pass: https://github.com/rguthriemsft/terratest/runs/1194826162?check_suite_focus=true

rguthriemsft:
@yorinasub17 @brikis98 this PR is ready for final Gruntworks review.

rguthriemsft (dismissed):

davesee:
https://github.com/rguthriemsft/terratest/runs/1196414380?check_suite_focus=true

rguthriemsft (dismissed):

examples/azure/terraform-azure-example/main.tf
```diff
@@ -1,14 +1,16 @@
 # ---------------------------------------------------------------------------------------------------------------------
-# DEPLOY AN AZURE VIRTUAL MACHINE
-# This is a basic example of how to deploy an Azure Virtual Machine with the minimum network resources.
+# DEPLOY A SIMPLE AZURE VIRTUAL MACHINE
+# This is a simple example of how to deploy an Azure Virtual Machine with the minimum network resources.
```
yorinasub17, line 3: NIT: We've been trying to avoid using words like `simple`, `just`, or `easy` in docs as it doesn't really add anything, and is not very newbie friendly. The main reason is that it can make users feel bad if they actually don't understand what is going on: "oh if this is the simple example, and I don't understand what is going on, there is no way I can learn this..."

See [this talk](https://www.writethedocs.org/videos/prague/2018/don-t-say-simply-jim-fisher/) for more info on this.

I think this is perfectly fine as `DEPLOY AN AZURE VIRTUAL MACHINE`. In fact, I would probably drop `basic` from the original text too.

modules/azure/compute.go
```diff
@@ -8,97 +8,451 @@ import (
...
+
+	for i, nic := range vmNICs {
+		nics[i] = GetNameFromResourceID(*nic.ID)
+	}
+	return nics, nil
+}
+
+// GetVirtualMachineNicCount gets the Network Interface count of the specified Azure Virtual Machine.
+// This function would fail the test if there is an error.
+func GetVirtualMachineNicCount(t testing.TestingT, vmName string, resGroupName string, subscriptionID string) int {
```
yorinasub17, line 67: NIT: Is this function necessary? Or can one do `len(GetVirtualMachineNics(t, vmName, resGroupName, subscriptionID))` to get the count?

modules/azure/compute.go
```diff
@@ -8,97 +8,451 @@ import (
...
-	return &vmClient, nil
+	return diskNames, nil
 }
 
-// GetSizeOfVirtualMachine gets the size type of the given Azure Virtual Machine.
+// GetVirtualMachineManagedDiskCount gets the Managed Disk count of the specified Azure Virtual Machine.
 // This function would fail the test if there is an error.
-func GetSizeOfVirtualMachine(t testing.TestingT, vmName string, resGroupName string, subscriptionID string) compute.VirtualMachineSizeTypes {
-	size, err := GetSizeOfVirtualMachineE(t, vmName, resGroupName, subscriptionID)
+func GetVirtualMachineManagedDiskCount(t testing.TestingT, vmName string, resGroupName string, subscriptionID string) int {
```
yorinasub17, line 122: NIT: Same question as above re: reusing the other function.

modules/azure/compute.go
```diff
@@ -8,97 +8,451 @@ import (
...
+	if err != nil {
+		return -1, err
+	}
+
+	return len(*vm.StorageProfile.DataDisks), nil
+}
+
+// GetVirtualMachineOsDiskName gets the OS Disk name of the specified Azure Virtual Machine.
+// This function would fail the test if there is an error.
+func GetVirtualMachineOsDiskName(t testing.TestingT, vmName string, resGroupName string, subscriptionID string) string {
```
yorinasub17, line 142: NIT: Go convention is to all caps on acronyms, so this should be `GetVirtualMachineOSDiskName`

modules/azure/compute.go
```diff
@@ -8,97 +8,451 @@ import (
...
+// GetVirtualMachineStateE gets the State of the specified Azure Virtual Machine.
+func GetVirtualMachineStateE(vmName string, resGroupName string, subscriptionID string) (string, error) {
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
+	if err != nil {
+		return "", err
+	}
+
+	return vm.Status, nil
+}
```
yorinasub17, line 177: NIT: I think this function is a bit extraneous, given that one can get this with `GetVirtualMachine(vmName, resGroupName, subscriptionID).Status`

As a rule of thumb, a single attribute probably doesn't need a getter, while nested attributes would be useful (e.g., like the `GetVirtualMachineOsDiskName` function)

modules/azure/compute.go
```diff
@@ -8,97 +8,451 @@ import (
...
+		return nil, err
+	}
+
+	for _, v := range vms.Values() {
+		vmDetails = append(vmDetails, *v.Name)
+	}
+	return vmDetails, nil
+}
+
+// GetVirtualMachinesForResourceGroup gets all Virtual Machine objects in the specified Resource Group.
```
yorinasub17, line 343: NIT: it helps to explain the map keys when returning `map`. What do the keys represent in the return type?

modules/azure/compute.go
```diff
@@ -8,97 +8,451 @@ import (
...
+// This function would fail the test if there is an error.
+func GetVirtualMachinesForResourceGroup(t testing.TestingT, resGroupName string, subscriptionID string) *map[string]compute.VirtualMachineProperties {
```
yorinasub17, line 345: Don't use pointers with map in golang, because it's already a pointer: https://blog.golang.org/maps

> Map types are reference types, like pointers or slices

modules/azure/disk.go
```diff
@@ -0,0 +1,103 @@
...
+		if ResourceNotFoundErrorExists(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// GetDiskType returns the Disk Storage Account Type of the Azure Managed Disk
+// This property also accessible from the VM client disk storage object but only works
```
yorinasub17, line 33:
```suggestion
// This property is also accessible from the VM client disk storage object but only works
```

modules/azure/disk.go
```diff
@@ -0,0 +1,103 @@
...
+// when the VM is online, while this direct call to GetDiskType always works.
+// This function would fail the test if there is an error.
+func GetDiskType(t testing.TestingT, diskName string, resGroupName string, subscriptionID string) compute.DiskStorageAccountTypes {
+	diskType, err := GetDiskTypeE(t, diskName, resGroupName, subscriptionID)
+	require.NoError(t, err)
+	return diskType
+}
+
+// GetDiskTypeE returns the Disk Storage Account Type of the Azure Managed Disk
```
yorinasub17, line 42: NIT: The extra comments in `GetDiskType` applies here as well, so it should be replicated.

modules/azure/disk.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		if ResourceNotFoundErrorExists(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// GetDiskType returns the Disk Storage Account Type of the Azure Managed Disk
```
yorinasub17, line 32:
```suggestion
// GetDiskType returns the Disk Storage Account Type of the Azure Managed Disk.
```

modules/azure/disk.go
```diff
@@ -0,0 +1,103 @@
...
+	// Get the Disk object
+	disk, err := GetDiskE(t, diskName, resGroupName, subscriptionID)
+	if err != nil {
+		return "", err
+	}
+
+	return disk.Sku.Name, nil
+}
+
+// GetDisk returns a Disk in the specified Azure Resource Group
```
yorinasub17, line 53:
```suggestion
// GetDisk returns a Disk in the specified Azure Resource Group.
```

modules/random/random.go
```diff
@@ -24,7 +24,7 @@ func RandomString(elements []string) string {
 	return elements[index]
 }
 
-const base62chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+const base62chars = "0123456789012345678901234567890123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```
yorinasub17, line 27: This seems like an unneeded change, since it is replicating the digits.

modules/azure/compute.go
```diff
@@ -8,97 +8,451 @@ import (
...
+	if err != nil {
+		return "", err
+	}
+
+	return string(*vm.OsProfile.AdminUsername), nil
+}
+
+// GetVirtualMachineSize gets the Size Type of the specified Azure Virtual Machine.
+// This function would fail the test if there is an error.
+func GetVirtualMachineSize(t testing.TestingT, vmName string, resGroupName string, subscriptionID string) compute.VirtualMachineSizeTypes {
```
yorinasub17, line 265: Is this rename necessary? The reason I ask is that this would be a backwards incompatible change, and I am not sure the name change is worth the backwards incompatibility.

test/azure/terraform_azure_vm_example_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTerraformAzureVmExample(t *testing.T) {
+	t.Parallel()
+
+	// subscriptionID is overridden by the environment variable "ARM_SUBSCRIPTION_ID"
+	subscriptionID := ""
+	uniquePostfix := random.UniqueId() // "resource" //
```
yorinasub17, line 25: NIT: This comment doesn't quite make sense to me. Either expand, or remove.

test/azure/terraform_azure_vm_example_test.go
```diff
@@ -0,0 +1,191 @@
...
+	expectedPublicAddressName := terraform.Output(t, terraformOptions, "public_ip_name")
+	expectedNicName := terraform.Output(t, terraformOptions, "network_interface_name")
+	expectedAvsName := terraform.Output(t, terraformOptions, "availability_set_name")
+	expectedOSDiskName := terraform.Output(t, terraformOptions, "os_disk_name")
+	expectedDiskName := terraform.Output(t, terraformOptions, "managed_disk_name")
+
+	// Comment for PR, to be removed:
+	// Please let me know if there are too many tests or alternate examples, happy to reduce the
+	// complexity and amount of code to be maintained. I tried to illustrate different approaches
+	// we have used in various scenarios to illustrate the flexability of Terratest.
```
yorinasub17, line 74: This is actually fine, but consider breaking up each of the following subtests into a separate function to make it easier to read/follow.

test/azure/terraform_azure_vm_example_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	t.Run("Strategies", func(t *testing.T) {
+		// Check the VM Size directly.
+		actualVMSize := azure.GetVirtualMachineSize(t, virtualMachineName, resourceGroupName, subscriptionID)
+		assert.Equal(t, expectedVMSize, actualVMSize)
+
+		// Check the VM Size by reference alternate example.
```
yorinasub17, line 81: NIT: What do you mean by "alternate example"?

test/azure/terraform_azure_vm_example_test.go
```diff
@@ -0,0 +1,191 @@
...
+		vmList := azure.ListVirtualMachinesForResourceGroup(t, resourceGroupName, subscriptionID)
+		assert.True(t, len(vmList) > 0)
+		assert.Contains(t, vmList, virtualMachineName)
+
+		// Get all VMs in a Resource Group by reference alternate example.
+		// This strategy is beneficial when checking multiple VMs & their properties by avoiding
+		// multiple SDK calls. The penalty for this approach is introducing direct references
+		// which need to be checked for nil for optional configurations.
+		vmsByRef := azure.GetVirtualMachinesForResourceGroup(t, resourceGroupName, subscriptionID)
+		assert.True(t, len(*vmsByRef) > 0)
```
yorinasub17, line 109: NIT: use `assert.Greater`, as it will show you both numbers when the assertion fails.

modules/azure/disk.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// DiskExists indicates whether the specified Azure Managed Disk exists
```
yorinasub17, line 11:
```suggestion
// DiskExists indicates whether the specified Azure Managed Disk exists.
```

yorinasub17 (commented):
Mostly LGTM! Almost all my comments are style nits, but I did have a few questions.

test/azure/terraform_azure_vm_example_test.go
davesee: I labeled it as alternate since the three asserts in this section "Strategies" test for the same VM property "VM Size" using different patterns:
- getPropertyOfVM
- getVM then vm.Property
-getVMInstance then vmInstance.GetProperty()

I added the 2 alternates to show the how properties can be accessed on this complex resource while reducing the underlying calls to the sdk by reusing the same VM reference or instance. I did not create any other tests to support the alternate examples since these are meant to give the audience flexibility if they need to create their own tests.

If that does not align with the module and causes overhead I can remove these and stick to one pattern, getPropertyOfVM.

yorinasub17: Ah gotcha! The examples are fine, but I was confused by the comment. I suggest rewording to:

> An alternate example for checking the VM size by reference

As written, I grouped "reference alternate example" together which made it confusing, instead of reading it as "Check the VM Size by reference - alternate example".

test/azure/terraform_azure_vm_example_test.go
rguthriemsft: Fixed.

test/azure/terraform_azure_vm_example_test.go
rguthriemsft: Fixed.

davesee:
CI Pass: https://github.com/rguthriemsft/terratest/runs/1203133483?check_suite_focus=true

modules/azure/disk.go
davesee: Updated.

test/azure/terraform_azure_vm_example_test.go
davesee: Updated along with the sub test breakout.

test/azure/terraform_azure_vm_example_test.go
davesee: Refactored.

test/azure/terraform_azure_vm_example_test.go
davesee: Removed.. it was just there to replace the UniqueId when using Terraform manual deployment since it took longer to spin up and destroy these resources.

modules/azure/compute.go
davesee: Not necessary, reverted.

modules/random/random.go
davesee: Reverted. The frequency of numbers returned was just low since there is a 5:26 ratio of numeric to alpha chars, tried to even the odds.

modules/azure/disk.go
davesee: Updated

modules/azure/disk.go
davesee: Updated

modules/azure/disk.go
davesee: Updated

modules/azure/disk.go
davesee: Updated

modules/azure/compute.go
davesee: Updated, thanks for the link, helpful to know!

modules/azure/compute.go
davesee: Updated

modules/azure/compute.go
davesee: Removed! Good point.

modules/azure/compute.go
davesee: Updated

modules/azure/compute.go
davesee: Reused, good point!

modules/azure/compute.go
davesee: Reused, good point!

examples/azure/terraform-azure-example/main.tf
davesee: Updated, makes sense. Thanks for the link :)

davesee:
https://github.com/rguthriemsft/terratest/runs/1203353211?check_suite_focus=true

rguthriemsft (approved):
Look good to me

rguthriemsft (approved):

examples/azure/terraform-azure-vm-example/main.tf
```diff
@@ -0,0 +1,161 @@
...
+# ---------------------------------------------------------------------------------------------------------------------
+
+provider "azurerm" {
+  version = "~> 2.29"
+  features {}
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# PIN TERRAFORM VERSION TO >= 0.12
+# The examples have been upgraded to 0.12 syntax
```
brikis98, line 16: NIT: given that the comment block below already explains the TF version, fix this out-of-date one. 

```suggestion
# PIN TERRAFORM VERSION
```

examples/azure/terraform-azure-vm-example/main.tf
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  os_profile_windows_config {
+    provision_vm_agent = true
+  }
+
+  depends_on = [random_password.rand]
+}
+
+resource "random_password" "rand" {
```
brikis98, line 134: Same question about password management

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
 	if err != nil {
 		return nil, err
 	}
 
-	// Create a VM client
-	vmClient := compute.NewVirtualMachinesClient(subscriptionID)
+	vmNICs := *vm.NetworkProfile.NetworkInterfaces
```
brikis98, line 50: Do you need to check this isn't `nil`? E.g., If there are no NICs attached? It's really important to avoid panics in Terratest.

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
 	if err != nil {
 		return nil, err
 	}
 
-	// Attach authorizer to the client
-	vmClient.Authorizer = *authorizer
+	// Get VM attached Disks
+	vmDisks := *vm.StorageProfile.DataDisks
```
brikis98, line 84: Same question on `nil`

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
-	return &vmClient, nil
+	// No Attached Disks present
+	if len(vmDisks) == 0 {
+		return nil, nil
+	}
+
+	// Get the Names of the attached Managed Disks
+	diskNames := make([]string, len(vmDisks))
+	for i, v := range vmDisks {
+		diskNames[i] = *v.Name
```
brikis98, line 94: Yet another sanity check on `nil`. Any chance the disk has no name set?

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
 	}
 
-	// Get the details of the target virtual machine
-	vm, err := vmClient.Get(context.Background(), resGroupName, vmName, compute.InstanceView)
+	// Virtual Machine has no associated Availability Set
+	if vm.AvailabilitySet == nil {
+		return "", nil
+	}
+
+	return GetNameFromResourceID(*vm.AvailabilitySet.ID), nil
```
brikis98, line 142: `nil` check again

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+	var vmImage VMImage
+
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
+	if err != nil {
+		return vmImage, err
+	}
+
+	if vm.StorageProfile == nil {
+		return vmImage, NewNotFoundError("Image Reference", "Any", vmName)
```
brikis98, line 173: Why does a `nil` storage profile mean a `NewNotFoundError`?

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+	}
+
+	// Populate VM Image; values always present, no nil checks needed
```
brikis98, line 176: 👍

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+
+// GetVirtualMachineAdminUserE gets the Admin Username of the specified Azure Virtual Machine.
+func GetVirtualMachineAdminUserE(vmName string, resGroupName string, subscriptionID string) (string, error) {
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
+	if err != nil {
+		return "", err
+	}
+
+	return string(*vm.OsProfile.AdminUsername), nil
```
brikis98, line 201: This method feels like a very thin wrapper on `GetVirtualMachineE`. Is it worth the extra code to maintain/document/test? Or would it make sense for users to call `GetVirtualMachineE` directly?

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+		return nil, err
+	}
+
+	vms, err := vmClient.List(context.Background(), resourceGroupName)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, v := range vms.Values() {
+		vmDetails = append(vmDetails, *v.Name)
```
brikis98, line 279: Can name be `nil`?

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+	}
+
+	vms, err := vmClient.List(context.Background(), resourceGroupName)
+	if err != nil {
+		return nil, err
+	}
+
+	vmDetails := make(map[string]compute.VirtualMachineProperties, len(vms.Values()))
+	for _, v := range vms.Values() {
+		vmDetails[string(*v.Name)] = *v.VirtualMachineProperties
```
brikis98, line 308: You can guess what I'm going to ask 😁 

Also, why does `*v.Name` need to be cast to a `string`?

modules/azure/compute.go
```diff
@@ -8,97 +8,392 @@ import (
...
+
+// GetVirtualMachineInstanceE gets a local Virtual Machine instance in the specified Resource Group.
+func GetVirtualMachineInstanceE(vmName string, resGroupName string, subscriptionID string) (*Instance, error) {
+	// Get VM Object
+	vm, err := GetVirtualMachineE(vmName, resGroupName, subscriptionID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Instance{vm}, nil
```
brikis98, line 338: I'm a bit confused... This just wraps `GetVirtualMachineE` in a different return type? If so, does this really need two more methods?

modules/azure/compute_test.go
```diff
@@ -13,18 +13,63 @@ import (
...
+	vmName := ""
+	rgName := ""
+	subID := ""
+
+	_, err := GetVirtualMachineInstanceE(vmName, rgName, subID)
+
+	require.Error(t, err)
+}
+
+func TestListVirtualMachinesForResourceGroupE(t *testing.T) {
```
brikis98, line 43: Are these all unit tests just checking the methods handle invalid params?

modules/azure/disk.go
```diff
@@ -0,0 +1,101 @@
...
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// DiskExists indicates whether the specified Azure Managed Disk exists.
+// This function would fail the test if there is an error.
+func DiskExists(t testing.TestingT, diskName string, resGroupName string, subscriptionID string) bool {
```
brikis98, line 13: Was everything in this file supposed to be part of https://github.com/gruntwork-io/terratest/pull/618?

test/azure/terraform_azure_example_test.go
```diff
@@ -1,12 +1,11 @@
-// +build azure
+// +build azure azureslim,compute
```
brikis98, line 1: What are the new tags?

test/azure/terraform_azure_vm_example_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// Check Availability Set for multiple VMs.
+	actualVMsInAvs := azure.GetAvailabilitySetVMNamesInCaps(t, expectedAvsName, resourceGroupName, subscriptionID)
+	assert.Contains(t, actualVMsInAvs, strings.ToUpper(expectedVMName))
+
+	// Get all VMs in a Resource Group, including their properties, therefore avoiding
+	// multiple SDK calls. The penalty for this approach is introducing direct references
+	// which need to be checked for nil for optional configurations.
+	vmsByRef := azure.GetVirtualMachinesForResourceGroup(t, resourceGroupName, subscriptionID)
+	thisVM := (vmsByRef)[expectedVMName]
```
brikis98, line 99: NIT: why parens around `vmsByRef`?

brikis98 (commented):
This is close. I left a bunch of comments, but they are almost almost all NITs or questions  about `nil`. We really have to be careful to avoid a nil pointer dereference or anything else that causes a panic in Terratest, as a panic will kill the test suite part way through, preventing any cleanup code from running.

examples/azure/terraform-azure-example/main.tf
```diff
@@ -102,13 +90,23 @@ resource "azurerm_virtual_machine" "main" {
...
   }
 
   os_profile_linux_config {
     disable_password_authentication = false
   }
+
+  depends_on = [random_password.main]
 }
 
+resource "random_password" "main" {
```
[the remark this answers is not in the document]
davesee: It was for simplicity and remote VM security, note added.

examples/azure/terraform-azure-vm-example/main.tf
davesee: Updated

examples/azure/terraform-azure-vm-example/main.tf
davesee: Updated to match.

modules/azure/compute.go
davesee: Similar reasoning, the OS disk is storage and will always have one. Added the extra check still. If you rather remove these nil checks I will do so, just exercising an abundance of caution to avoid future panics.

modules/azure/compute.go
davesee: Added *E and check.

modules/azure/compute.go
davesee: One NIC is required or you get the Terraform error below, so technically the length check is not necessary either. I also confirmed that once created, at leas one NIC is required and cannot be removed via the UI and the attached NIC cannot be deleted because it is a dependency.

Error: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="VirtualMachineMustHaveAtLeastOneNetworkInterface" Message="Virtual machine vm-resource must have at least one network interface." Details=[]

Added comment to support the absence of the `nil` check.

examples/azure/terraform-azure-vm-example/main.tf
davesee: Updated

modules/azure/compute.go
davesee: Disk names are required. Added comment to support the absence of the nil check.

modules/azure/compute.go
davesee: I checked and this is required as well, removed.

modules/azure/compute.go
davesee: Documentation preferred. Learning :)

modules/azure/compute.go
davesee: Can be called directly, removed func and unit test, updated terratest example.

modules/azure/compute.go
davesee: Every VM has a required name. Added comment to support the absence of the nil check.

modules/azure/compute.go
davesee: removed unnecessary cast!

modules/azure/compute.go
davesee: Removed the two functions and leveraged the existing VM in the test. I kept all the logic in the module at first to have a concrete type assertion in one space but yes, it was redundant at that point.

modules/azure/compute_test.go
davesee: It has been helpful as the library changed with number of params and names updated as well.

modules/azure/disk.go
davesee: Yes, they were introduced at different times but are valuable independently.

test/azure/terraform_azure_example_test.go
davesee: @rguthriemsft bought this up so we can have smaller CI builds under common groups of tests. Another example can be seen with azureslim,network on a few of those network tests.

test/azure/terraform_azure_vm_example_test.go
davesee: Removed extra brackets.

modules/azure/disk.go
davesee: Note this test does use the GetDiskType which was described as a thin wrapper in the Disk PR.  It adds more code to the test example but updated this VM test to check for the property via the disk object.

rguthriemsft:
[Microsoft CI Bot] TL;DR; failure :facepalm:

You can check the status of the CI Pipeline logs here ; https://github.com/rguthriemsft/terratest/actions/runs/305145049

rguthriemsft:
[Microsoft CI Bot] TL;DR; failure :facepalm:

You can check the status of the CI Pipeline logs here ; https://github.com/rguthriemsft/terratest/actions/runs/305145049

rguthriemsft:
[Microsoft CI Bot] TL;DR; failure :facepalm:

You can check the status of the CI Pipeline logs here ; https://github.com/rguthriemsft/terratest/actions/runs/305145049

rguthriemsft:

You can check the status of the CI Pipeline logs here ; https://github.com/rguthriemsft/terratest/actions/runs/305145049

rguthriemsft (approved):
@yorinasub17 @brikis98 Look good to me

yorinasub17 (approved):
LGTM!

bwhaley (approved):

yorinasub17:
Pardon the delay in reviewing this, but this is good to go now! Thanks for the contribution! Will merge and release.

PR #642, opened 2020-09-17
Update repo to work with Terraform 0.13.x

brikis98:
1. Update CircleCi to use Terraform 0.13.x.
1. Update version constraint on all examples to require at least 0.12.26.
1. Update pre-commit hooks to add `terraform-fmt`.
1. Fix some `terraform fmt` issues that were found.

In general, TF 0.13.x seems backwards compatible from a testing perspective: I did not need to change anything in `modules` in this PR. The only gotcha I've seen so far is with the way `plan` works: in 0.13.x, if you've never run `apply` at all, Terraform `plan` always believes there are changes it has to make, even if the Terraform module doesn't contain any resources. Therefore, I had to update some of our tests for `plan` to run `apply` first.


brikis98:
OK, tests are passing. This is ready for review!

yorinasub17 (approved):
LGTM!

brikis98:

rhoboat:
So you have to run `apply` before `plan` with TF 0.13?? Even if it's only if you've never run apply before, how does this make any sense?

brikis98:
> So you have to run `apply` before `plan` with TF 0.13?? Even if it's only if you've never run apply before, how does this make any sense?

Hm, perhaps I wasn't clear enough:

1. If you had `.tf` files with some resources in them and ran `plan`, either with TF 0.12 or 0.13, Terraform would tell you there are changes to apply, and `-detailed-exitcode` would exit with a code of 2. So nothing changed here.
1. However, if you had `.tf` files with no resources in them—perhaps solely data sources or just output variables as  part of a silly example or test, as we have in this repo—and ran `plan`, then TF 0.12 would tell you there's nothing to do and `-detailed-exitcode` would exit with a code  of 0, whereas in 0.13, if you had never run `apply` even once, it says there are changes to apply and `-detailed-exitcode` would exit with a code of 2. This is the change. 

So this doesn't really affect anything in the real world, I don't think...

PR #657, opened 2020-09-25
Adds helpers to create, get, and delete ECR repositories

bwhaley:
Adds new functions for interfacing with ECR.

### DIFF
--- a/modules/aws/ecr.go
+++ b/modules/aws/ecr.go
@@ -1,0 +1,93 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// CreateECRRepo creates a new ECR Repository
+func CreateECRRepo(t testing.TestingT, region string, name string) *ecr.Repository {
+	repo, err := CreateECRRepoE(t, region, name)
+	require.NoError(t, err)
+	return repo
+}
+
+// CreateECRRepoE creates a new ECR Repository
+func CreateECRRepoE(t testing.TestingT, region string, name string) (*ecr.Repository, error) {
+	client := NewECRClient(t, region)
+	resp, err := client.CreateRepository(&ecr.CreateRepositoryInput{RepositoryName: aws.String(name)})
+	if err != nil {
+		return nil, err
+	}
+	return resp.Repository, nil
+}
+
+// GetECRRepo gets an ECR repository by name
+func GetECRRepo(t testing.TestingT, region string, name string) *ecr.Repository {
+	repo, err := GetECRRepoE(t, region, name)
+	require.NoError(t, err)
+	return repo
+}
+
+// GetECRRepoE gets an ECR Repository by name
+func GetECRRepoE(t testing.TestingT, region string, name string) (*ecr.Repository, error) {
+	client := NewECRClient(t, region)
+	repositoryNames := []*string{aws.String(name)}
+	resp, err := client.DescribeRepositories(&ecr.DescribeRepositoriesInput{RepositoryNames: repositoryNames})
+	if err != nil {
+		return nil, err
+	}
+	if len(resp.Repositories) != 1 {
+		return nil, fmt.Errorf("There is no repository named %s", name)
+	}
+	return resp.Repositories[0], nil
+}
+
+// DeleteECRRepo will force delete the ECR repo by deleting all images prior to deleting the ECR repository.
+func DeleteECRRepo(t testing.TestingT, region string, repo *ecr.Repository) {
+	err := DeleteECRRepoE(t, region, repo)
+	require.NoError(t, err)
+}
+
+// DeleteECRRepoE will force delete the ECR repo by deleting all images prior to deleting the ECR repository.
+func DeleteECRRepoE(t testing.TestingT, region string, repo *ecr.Repository) error {
+	client := NewECRClient(t, region)
+	resp, err := client.ListImages(&ecr.ListImagesInput{RepositoryName: repo.RepositoryName})
+	if err != nil {
+		return err
+	}
+	if len(resp.ImageIds) > 0 {
+		_, err = client.BatchDeleteImage(&ecr.BatchDeleteImageInput{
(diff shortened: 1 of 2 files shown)


modules/aws/ecr.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// CreateECRRepo creates a new ECR Repository
```
yorinasub17, line 12: NIT: We've been adding a note on the non-E functions to mention that:
```suggestion
// CreateECRRepo creates a new ECR Repository. This will fail the test and stop execution if there is an error.
```

This applies to all the non-E functions in this file.

modules/aws/ecr.go
```diff
@@ -0,0 +1,93 @@
...
+// GetECRRepoE gets an ECR Repository by name
+func GetECRRepoE(t testing.TestingT, region string, name string) (*ecr.Repository, error) {
+	client := NewECRClient(t, region)
+	repositoryNames := []*string{aws.String(name)}
+	resp, err := client.DescribeRepositories(&ecr.DescribeRepositoriesInput{RepositoryNames: repositoryNames})
+	if err != nil {
+		return nil, err
+	}
+	if len(resp.Repositories) != 1 {
+		return nil, fmt.Errorf("There is no repository named %s", name)
```
yorinasub17, line 45: NIT: We should return a custom error type here so that you can check for it. E.g., if one wanted to distinguish between error because repo doesn't exist and error because of AWS issues.

modules/aws/ecr_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func TestEcrRepo(t *testing.T) {
+	t.Parallel()
+
+	region := GetRandomStableRegion(t, nil, nil)
+	repo1, err := CreateECRRepoE(t, region, "terratest")
+	defer DeleteECRRepo(t, region, repo1)
+
+	assert.Nil(t, err)
```
yorinasub17, line 16:
```suggestion
	require.NoError(t, err)
```

So we don't do the rest of the steps and panic if there is an error when we dereference the pointer in the next step when there is an error.

This applies to the similar checks done below.

modules/aws/ecr_test.go
```diff
@@ -0,0 +1,23 @@
...
+	assert.Equal(t, "terratest", *repo1.RepositoryName)
```
yorinasub17, line 17: NIT: use `aws.String`

modules/aws/ecr_test.go
```diff
@@ -0,0 +1,23 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
```
yorinasub17, line 8: NIT: If you decide to add a custom error for the non-existent ECR repo, consider adding a unit test that checks for that error type when you call `GetECRRepoE` with a random name.

yorinasub17 (commented):
Thanks for porting it over here! Had a bunch of nits I found now that I'm reviewing it with a fresh mind. I know most of this was copy pasted from our internal repo, so feel free to punt on the suggestions.

modules/aws/ecr.go
bwhaley: Done in 2840220

modules/aws/ecr_test.go
bwhaley: Done in 2840220

modules/aws/ecr_test.go
bwhaley: Done in 2840220

modules/aws/ecr.go
bwhaley: This error condition should never occur. If no repositories are found, the `DescribeRepositories` call itself will return an error. We can't have more than one response because we're only providing one name to `RepositoryNames`. However, it's never a good idea to index a slice without length checking first, hence this check.. but I'm not sure how to cause this condition, so we also wouldn't be able to check for a custom error. Thus I just stuck with the `fmt.Errorf` (although I did adjust the wording). 

Note that this follows the convention already established throughout `terratest` (e.g. in ecs.go, ssm.go, and elsewhere).

modules/aws/ecr_test.go
bwhaley: See above comment on the issue with testing for this condition.

bwhaley:
Thanks for the review. Note that I tried to follow the conventions already established in the package and used ecs.go as a model, but I do understand that the preferences and best practices are evolving. Let me know if you have any further feedback after the most recent updates!

modules/aws/ecr.go
yorinasub17: Ahhh ok. In that case, the error message should actually mention that this is an impossible case and if a user runs into it, they should file an issue.

modules/aws/ecr.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// GetECRRepo gets an ECR repository by name. This will fail the test and stop execution if there is an error.
+func GetECRRepo(t testing.TestingT, region string, name string) *ecr.Repository {
+	repo, err := GetECRRepoE(t, region, name)
+	require.NoError(t, err)
+	return repo
+}
+
+// GetECRRepoE gets an ECR Repository by name.
```
yorinasub17, line 37: NIT: It's probably worth mentioning that this will error if the ECS repo doesn't exist.

yorinasub17 (approved):
Thanks for addressing! Had two more nits, but super minor. Feel free to punt to future PR.

modules/aws/ecr.go
bwhaley: Good thinking, that sounds like the right thing to do. 2e7a21e

modules/aws/ecr.go
bwhaley: 2e7a21e

bwhaley:
Ok, I think I've incorporated all the feedback!

yorinasub17 (approved):
LGTM!

bwhaley:
Thanks for the reviews and feedback!

bwhaley:
Had to catch up to master before merging - can I get another approval please?

yorinasub17 (approved):

Issue #660, opened 2020-09-28
Add a dns-helper module


askainet:
I'm struggling to write reliable tests for k8s deployments with public ingresses, i.e. http retry requests to public DNS hostnames asynchronously created by external-dns for ingresses in those deployments:

1. Install helm chart for a web application, including an ingress resource with a custom public DNS name.
2. Wait until ingress is ready and get its public DNS name.
2.1. Eventually, some tool in k8s like external-dns creates a public DNS record pointing to the ingress. **Terratest doesn't provide a reliable way to wait for this yet.**
3. Start retrying http requests to public DNS name...

If the DNS record has not been created/updated yet when the first http request is sent, chances are the system where the test runs caches the negative/old reply for a very long time (some systems behave worse than others, not even honoring original TTL set by authoritative nameservers), and the test will most likely fail.

To avoid being affected by this local DNS cache issues, a dns-helper module could provide a way of waiting until the authoritative nameservers for the domain reply as expected to the DNS record.

yorinasub17:
This was released in https://github.com/gruntwork-io/terratest/releases/tag/v0.30.14

PR #661, opened 2020-09-28
feat: add dns-helper module

askainet:
I've been missing some DNS lookup helper functions, so as an excuse to improve my Go skills and have some fun in the process, I decided to contribute with this new dns-helper module. I hope some of you will find this useful and I'm looking forward to getting your feedback.

Tries to deal with https://github.com/gruntwork-io/terratest/issues/660

Motivation:

I was struggling to write reliable tests for k8s deployments with public ingresses, i.e. http retry requests to public DNS hostnames asynchronously created by external-dns for ingresses in those deployments:

1. Install helm chart for a web application, including an ingress resource with a custom public DNS name.
2. Wait until ingress is ready and get its public DNS name.
2.1. Eventually, some tool in k8s like external-dns creates a public DNS record pointing to the ingress. Terratest doesn't provide a reliable way to wait for this yet.
3. Start retrying http requests to public DNS name...

If the DNS record has not been created/updated yet when the first http request is sent, chances are the system where the test runs caches the negative/old reply for a very long time (some systems behave worse than others, not even honoring original TTL set by authoritative nameservers), and the test will most likely fail.

To avoid being affected by this local DNS cache, the dns-helper module provides several convenient ways of checking DNS records directly with their authoritative nameservers, like waiting until all the nameservers reply with the expected answers. It bypasses the local DNS resolver configuration, except to find the authoritative nameservers for the DNS name being checked.

Usage example:

```go
k8s.WaitUntilIngressAvailable(t, kubectlOptions, ingressName, maxRetries, sleepBetweenRetries)
ingress := k8s.GetIngress(t, kubectlOptions, ingressName)
hostname := ingress.Spec.Rules[0].Host
dnsQuery := dns_helper.DNSQuery{"A", hostname}
dns_helper.DNSLookupAuthoritativeAllWithRetry(t, dnsQuery, nil, maxRetries, sleepBetweenRetries)
response := http_helper.HTTPDoWithRetry(t, "GET", "https://"+hostname, nil, nil, 200, maxRetries, sleepBetweenRetries, nil)
require.Contains(t, response, expectedResponse)
```

- It supports DNS records types A, AAAA, CNAME, MX, NS, TXT.
- Provides `All` versions of the functions, to require consistency in all authoritative answers.
- Provides retry versions of some functions.
- Provides with-validation versions as well. 
- All the functions in fail-on-error and return-error flavors.
- Extensive testing. Sample tests results https://gist.github.com/askainet/93d334396fd2ded4f0148ddcacc6c680

As a bonus, and as suggested by @yorinasub17, add a new retry function DoWithRetryInterfaceE to return output in any format, updating the original DoWithRetryE to use it while keeping compatibility with string output.

### DIFF
--- a/modules/dns-helper/dns_helper.go
+++ b/modules/dns-helper/dns_helper.go
@@ -12,17 +12,14 @@ import (
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/miekg/dns"
+	"github.com/stretchr/testify/require"
 )
 
 // DNSFindNameservers tries to find the NS record for the given FQDN, iterating down the domain hierarchy
 // until it founds the NS records and returns it. Fails if there's any error or no NS record is found up to the apex domain.
 func DNSFindNameservers(t testing.TestingT, fqdn string, resolvers []string) []string {
 	nameservers, err := DNSFindNameserversE(t, fqdn, resolvers)
-
-	if err != nil {
-		t.Fatal(err)
-	}
-
+	require.NoError(t, err)
 	return nameservers
 }
 
@@ -87,11 +84,7 @@ func DNSFindNameserversE(t testing.TestingT, fqdn string, resolvers []string) ([
 // Fails on any error from DNSLookupAuthoritativeE.
 func DNSLookupAuthoritative(t testing.TestingT, query DNSQuery, resolvers []string) DNSAnswers {
 	res, err := DNSLookupAuthoritativeE(t, query, resolvers)
-
-	if err != nil {
-		t.Fatal(err)
-	}
-
+	require.NoError(t, err)
 	return res
 }
 
@@ -126,11 +119,7 @@ func DNSLookupAuthoritativeE(t testing.TestingT, query DNSQuery, resolvers []str
 // Fails on any error from DNSLookupAuthoritativeWithRetryE.
 func DNSLookupAuthoritativeWithRetry(t testing.TestingT, query DNSQuery, resolvers []string, maxRetries int, sleepBetweenRetries time.Duration) DNSAnswers {
 	res, err := DNSLookupAuthoritativeWithRetryE(t, query, resolvers, maxRetries, sleepBetweenRetries)
-
-	if err != nil {
-		t.Fatal(err)
-	}
-
+	require.NoError(t, err)
 	return res
 }
 
@@ -155,11 +144,7 @@ func DNSLookupAuthoritativeWithRetryE(t testing.TestingT, query DNSQuery, resolv
 // Fails on any error from DNSLookupAuthoritativeAllE.
 func DNSLookupAuthoritativeAll(t testing.TestingT, query DNSQuery, resolvers []string) DNSAnswers {
 	res, err := DNSLookupAuthoritativeAllE(t, query, resolvers)
-
-	if err != nil {
-		t.Fatal(err)
-	}
-
+	require.NoError(t, err)
 	return res
 }
 
@@ -203,10 +188,7 @@ func DNSLookupAuthoritativeAllE(t testing.TestingT, query DNSQuery, resolvers []
 // Fails when max retries has been exceeded.
 func DNSLookupAuthoritativeAllWithRetry(t testing.TestingT, query DNSQuery, resolvers []string, maxRetries int, sleepBetweenRetries time.Duration) {
 	_, err := DNSLookupAuthoritativeAllWithRetryE(t, query, resolvers, maxRetries, sleepBetweenRetries)
-
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 }
 
 // DNSLookupAuthoritativeAllWithRetryE repeatedly sends DNS requests for the specified record and type,
@@ -229,10 +211,7 @@ func DNSLookupAuthoritativeAllWithRetryE(t testing.TestingT, query DNSQuery, res
 // Fails on any underlying error from DNSLookupAuthoritativeAllWithValidationE.
 func DNSLookupAuthoritativeAllWithValidation(t testing.TestingT, query DNSQuery, resolvers []string, expectedAnswers DNSAnswers) {
 	err := DNSLookupAuthoritativeAllWithValidationE(t, query, resolvers, expectedAnswers)
-
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 }
 
 // DNSLookupAuthoritativeAllWithValidationE gets authoritative answers for the specified record and type.
@@ -264,10 +243,7 @@ func DNSLookupAuthoritativeAllWithValidationE(t testing.TestingT, query DNSQuery
 // Fails when max retries has been exceeded.
 func DNSLookupAuthoritativeAllWithValidationRetry(t testing.TestingT, query DNSQuery, resolvers []string, expectedAnswers DNSAnswers, maxRetries int, sleepBetweenRetries time.Duration) {
 	err := DNSLookupAuthoritativeAllWithValidationRetryE(t, query, resolvers, expectedAnswers, maxRetries, sleepBetweenRetries)
-
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 }
 
 // DNSLookupAuthoritativeAllWithValidationRetryE repeatedly gets authoritative answers for the specified record and type
@@ -313,11 +289,7 @@ func DNSDoWithRetryE(t testing.TestingT, actionDescription string, maxRetries in
 // Supported record types: A, AAAA, CNAME, MX, NS, TXT
 func DNSLookup(t testing.TestingT, query DNSQuery, resolvers []string) DNSAnswers {
 	res, err := DNSLookupE(t, query, resolvers)
-
-	if err != nil {
-		t.Fatal(err)
-	}
-
+	require.NoError(t, err)
 	return res
 }
 
--- a/modules/dns-helper/dns_helper.go
+++ b/modules/dns-helper/dns_helper.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/retry"
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/miekg/dns"
 	"github.com/stretchr/testify/require"
@@ -116,14 +117,14 @@ func DNSLookupAuthoritativeWithRetry(t testing.TestingT, query DNSQuery, resolve
 // or until max retries has been exceeded.
 // If resolvers are defined, uses them instead of the default system ones to find the authoritative nameservers.
 func DNSLookupAuthoritativeWithRetryE(t testing.TestingT, query DNSQuery, resolvers []string, maxRetries int, sleepBetweenRetries time.Duration) (DNSAnswers, error) {
-	res, err := DNSDoWithRetryE(
+	res, err := retry.DoWithRetryInterfaceE(
 		t, fmt.Sprintf("DNSLookupAuthoritativeE %s record for %s using authoritative nameservers", query.Type, query.Name),
 		maxRetries, sleepBetweenRetries,
-		func() (DNSAnswers, error) {
+		func() (interface{}, error) {
 			return DNSLookupAuthoritativeE(t, query, resolvers)
 		})
 
-	return res, err
+	return res.(DNSAnswers), err
 }
 
 // DNSLookupAuthoritativeAll gets authoritative answers for the specified record and type.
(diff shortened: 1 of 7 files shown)


modules/dns-helper/dns_local_server.go
```diff
@@ -0,0 +1,112 @@
...
+	ns2, mux2 := runTestDNSServer("0")
+
+	q := DNSQuery{"NS", testDomain}
+	dnsData1 = &DNSData{q: DNSAnswers{{"NS", ns1}, {"NS", ns2}}}
+	dnsData2 = &DNSData{q: DNSAnswers{{"NS", ns1}, {"NS", ns2}}}
+
+	mux1.HandleFunc(testDomain+".", func(w dns.ResponseWriter, r *dns.Msg) { stdDNSHandler(w, r, dnsData1, false) })
+	mux2.HandleFunc(testDomain+".", func(w dns.ResponseWriter, r *dns.Msg) { stdDNSHandler(w, r, dnsData2, true) })
+
+	return
```
yorinasub17, line 29: Style wise, we prefer explicit return statements vs implicit. Can you switch to explicitly returning each val here? This applies to all functions in this file.

modules/dns-helper/dns_helper_test.go
```diff
@@ -0,0 +1,289 @@
...
+	require.NoError(t, err)
+	require.ElementsMatch(t, res, expected)
+}
+
+func TestOkLocalDNSLookupAuthoritative(t *testing.T) {
+	t.Parallel()
+	ns1, ns2, dnsData1, dnsData2 := setupTestDNSServers()
+	dnsData1 = dnsData2
+	for dnsQuery, expected := range testDNSData {
+		(*dnsData1)[dnsQuery] = expected
```
yorinasub17, line 83: This is a bit confusing to me, especially since the name is not very descriptive. Two confusing elements:

- Why is it ok to update this on every loop? Does this pollute the next check?
- Why are we throwing away `dnsData1`? Consider doing the following, which is easier to follow:

```go
ns1, ns2, _, dnsData := setupTestDNSServers()
// use `dnsData` where using `dnsData1`
```

yorinasub17: Ohhh I think I get what's going on now. This is an internal database to handle the internal DNS data...

This is actually pretty confusing, so either:

- Update to a model where you pass in the database data to `setupTestDNSServers` as a function arg (or define a function `addEntryToDNSDatabase`).
- Add ample comments to explain what is going on.
- Rename `dnsData` to `dnsDatabase`.

modules/dns-helper/dns_local_server.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/miekg/dns"
+)
+
+var testDomain = "gruntwork.io"
+
+// DNSData type
+type DNSData map[DNSQuery][]DNSAnswer
+
+func setupTestDNSServers() (ns1, ns2 string, dnsData1, dnsData2 *DNSData) {
```
yorinasub17, line 18: You don't need to use pointers with maps, because they are passed by reference. [Source](https://blog.golang.org/maps):

> Map types are reference types, like pointers or slices

modules/dns-helper/dns_local_server.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	mux := dns.NewServeMux()
+	server := &dns.Server{PacketConn: listener, Net: "udp", Handler: mux}
+
+	go func() {
+		log.Fatal(server.ActivateAndServe())
```
yorinasub17, line 59: Don't use `log.Fatal` in test functions as it can kill the entire test run, which is problematic for terratest where it could skip the destroy call in other tests.

Also, how does this server shutdown during testing?

modules/dns-helper/dns_local_server.go
```diff
@@ -0,0 +1,112 @@
+package dns_helper
```
yorinasub17, line 1: The functions in here need comments.

modules/dns-helper/dns_local_server.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, r := range answers {
+		if seen[r] {
+			continue
+		}
+		seen[r] = true
+
+		rr, err := dns.NewRR(fmt.Sprintf("%s %s", q.Name, r.String()))
+
+		if err != nil {
+			log.Fatalf("err: %s", err)
```
yorinasub17, line 85: Same comment as above re:`log.Fatal`

yorinasub17 (commented):
Thanks for the contribution! This is definitely a great addition! I left a bunch of comments, but most of them are style nits.

askainet:
> Thanks for the contribution! This is definitely a great addition! I left a bunch of comments, but most of them are style nits.

Thank you for your comments! Coming from other languages, they are very helpful to learn the ways and style to write proper Golang. I'll try to address them very soon.

modules/dns-helper/dns_helper.go
```diff
@@ -0,0 +1,428 @@
...
+		}
+	}
+
+	parts := strings.Split(fqdn, ".")
+
+	for i := range parts[:len(parts)-1] {
+		domain = strings.Join(parts[i:], ".")
+		res, err = lookupFunc(domain)
+
+		if err == nil || len(res) > 0 {
```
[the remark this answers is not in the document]
askainet: Oops, weird logic indeed. No need to check for error here really, if there are responses we are good to go.

modules/dns-helper/dns_helper.go
```diff
@@ -0,0 +1,428 @@
...
+			var nameservers []string
+
+			for _, ns := range res {
+				nameservers = append(nameservers, strings.TrimSuffix(ns, "."))
+			}
+
+			logger.Logf(t, "FQDN %s belongs to domain %s, found NS record: %s", fqdn, domain, nameservers)
+			return nameservers, nil
+		}
```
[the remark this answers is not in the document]
askainet: Logging errors from the lookup function now for debugging purposes.

modules/dns-helper/dns_helper.go
```diff
@@ -0,0 +1,428 @@
...
+)
+
+// DNSFindNameservers tries to find the NS record for the given FQDN, iterating down the domain hierarchy
+// until it founds the NS records and returns it. Fails if there's any error or no NS record is found up to the apex domain.
+func DNSFindNameservers(t testing.TestingT, fqdn string, resolvers []string) []string {
+	nameservers, err := DNSFindNameserversE(t, fqdn, resolvers)
+
+	if err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
askainet: Got it, updated everywhere to require.NoError(t, err)

modules/dns-helper/dns_helper.go
```diff
@@ -0,0 +1,428 @@
...
+
+	return nameservers
+}
+
+// DNSFindNameserversE tries to find the NS record for the given FQDN, iterating down the domain hierarchy
+// until it founds the NS records and returns it. Returns the last error if the apex domain is reached with no result.
+func DNSFindNameserversE(t testing.TestingT, fqdn string, resolvers []string) ([]string, error) {
+	var res []string
+	var err error
+	var domain string
```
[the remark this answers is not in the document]
askainet: Removed unnecessary globals.

modules/dns-helper/dns_helper.go
```diff
@@ -0,0 +1,428 @@
...
+		return nil, err
+	}
+
+	res, err := DNSLookupE(t, query, nameservers)
+
+	if err != nil {
+		if _, ok := err.(*NotFoundError); !ok {
+			return nil, err
+		}
+		err = &NotFoundAuthoritativeError{query, nameservers}
```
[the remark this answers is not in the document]
askainet: Removed the NotFoundAuthoritativeError as it's in fact redundant.

modules/dns-helper/dns_local_server.go
askainet: Added comments to all functions.

modules/dns-helper/dns_local_server.go
askainet: Refactored the server setup to properly defer it's shutdown from the tests functions.

modules/dns-helper/dns_local_server.go
askainet: Refactored the whole DNSDatabase concept.

modules/dns-helper/dns_helper_test.go
askainet: That was a poor attempt to reduce lines of code when both test servers must use the same answers, that in the end made things confusing.
Refactored the whole DNSData concept to embed internal DNSDatabase and DNSDatabaseRetry in each dnsTestServer.
New entries can be added using the dnsTestServer.addEntryToDNSDatabase method.

askainet:
Updated tests results for dns-helper, as well as retry and http-helper as one of the affected modules by the DoWithRetry changes:

- dns-helper https://gist.github.com/askainet/93d334396fd2ded4f0148ddcacc6c680
- retry https://gist.github.com/askainet/e01966f2677d41d547a67782d35eef3b
- http-helper https://gist.github.com/askainet/a41ec027e3ec14b8ad54ad4447be6af0

modules/dns-helper/dns_local_server.go
askainet: Ensured all return statements are explicit.

modules/dns-helper/dns_helper.go
```diff
@@ -0,0 +1,428 @@
...
+			return nil, DNSLookupAuthoritativeAllWithValidationE(t, query, resolvers, expectedAnswers)
+		})
+
+	return err
+}
+
+// DNSDoWithRetryE sends a DNS query for the specified record and type to the given nameserver.
+// If it returns a response with records, return their values.
+// If it returns an error, sleep for sleepBetweenRetries and try again, up to a maximum of
+// maxRetries retries. If maxRetries is exceeded, return a MaxRetriesExceeded error.
```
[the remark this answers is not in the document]
askainet: Agreed, I updated it following your suggestion and tested with the http-helper as well.

askainet:
@yorinasub17 this is ready for review after addressing your comments

modules/dns-helper/dns_helper.go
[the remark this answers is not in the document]
yorinasub17: Is this function still necessary given the new DoWithRetryInterfaceE implementation?

modules/dns-helper/dns_helper_test.go
```diff
@@ -0,0 +1,383 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/gruntwork-io/terratest/modules/retry"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var publicDomainNameservers = []string{
```
yorinasub17, line 12: Add comment indicating that these are the public nameservers for gruntwork.io, and should be updated if that ever changes.

modules/dns-helper/dns_local_server.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return s1, s2
+}
+
+// runTestDNSServer starts and returns a new dnsTestServer listening in localhost and the given UDP port
+func runTestDNSServer(port string) *dnsTestServer {
+	listener, err := net.ListenPacket("udp", "127.0.0.1:"+port)
+
+	if err != nil {
+		log.Fatal(err)
```
yorinasub17, line 94: As mentioned in the other comment, we want to avoid `log.Fatal` in tests. You should switch this to `t.Fatal`, which will basically do the same thing but terminate the test goroutine instead of the whole process.

modules/dns-helper/dns_local_server.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, r := range answers {
+		if seen[r] {
+			continue
+		}
+		seen[r] = true
+
+		rr, err := dns.NewRR(fmt.Sprintf("%s %s", q.Name, r.String()))
+
+		if err != nil {
+			log.Fatalf("err: %s", err)
```
yorinasub17, line 130: Same as above about avoiding `log.Fatal`.

yorinasub17 (commented):
The changes LGTM! Had a few more nits left, with the most important one being the use of `log.Fatal`.

And really sorry about the delay here in responding. I was out on vacation two weeks ago, and then I was bogged down with other initiatives at work and didn't have a chance to come back to this until now.

modules/dns-helper/dns_helper_test.go
askainet: Removed unnecessary specific function DNSDoWithRetryE after implementing new generic DoWithRetryInterfaceE
Added comment about public nameservers for gruntwork.io

modules/dns-helper/dns_local_server.go
askainet: Switch from `log.Fatal` to `t.Fatal`

modules/dns-helper/dns_local_server.go
askainet: Switch from `log.Fatalf` to `t.Fatalf`

askainet:
> The changes LGTM! Had a few more nits left, with the most important one being the use of `log.Fatal`.
> 
> And really sorry about the delay here in responding. I was out on vacation two weeks ago, and then I was bogged down with other initiatives at work and didn't have a chance to come back to this until now.

Oops, I missed those two `log.Fatal` and the now unnecessary specific function `DNSDoWithRetryE`, fixed!

I'm happy to see this is moving forward, as I believe it will help when writing reliable tests depending on proper resolvability of DNS records. Also, it's been a great exercise to improve my Go skills.

And of course no worries at all about the delay, I hope you enjoyed your vacation! 🙂

yorinasub17:
Ok updates looks good to me! I just kicked off the test build, and if it passes, we can merge this in.

rhoboat (approved):
LGTM

yorinasub17:
Thanks @rhoboat ! Since all checks passed, this is good to go!

Thanks @askainet again for your contribution! Will merge and release this momentarily.

yorinasub17:
https://github.com/gruntwork-io/terratest/releases/tag/v0.30.14

askainet:
> Thanks @rhoboat ! Since all checks passed, this is good to go!
> 
> Thanks @askainet again for your contribution! Will merge and release this momentarily.

Great, thanks for merging! Very happy to contribute 😄

PR #663, opened 2020-10-01
feat: Sovereign Cloud Support for Azure (initial work)

### LINKED ISSUE #625: Add support for Azure sovereign regions - Client Factory
Need to modify modules to correctly query and pass base URI's for the sovereign clouds (Gov, Germany, China). This requires a few additional methods to help read correct base URI's, as well as CI support for testing this integration.  This will involve (at a minimum):

* Creation of a client factory to hand out correctly configured Azure SDK clients to module authors
* Enforcement of default behavior being public cloud (so there is no breakage if not configured)
* Providing unit tests where applicable
* Updating existing compute samples to use client factory.

Following this work, separate issues will be filed for updating in-progress modules.

mjyeaney:
This PR implements a client factory that manages correctly setting up the Azure SDK resource clients to interact with sovereign clouds (Government, Germany, China, Azure Stack).  This behavior is controlled by setting the `AZURE_ENVIRONMENT` environment variable to the appropriate environment name (as covered in `examples/azure/client_factory.md`).

Additionally, this PR _only_ has updated the call-site methods for the `compute.go` examples; other Azure modules will be updated in a separate PR once we're established/integrated.

Note this is an initial PR that will be part of a multi-step effort to enable sovereign cloud support for Azure with Terratest; tagging @rguthriemsft for initial review and/or comments.

Closes #625


modules/azure/client_factory.go
```diff
@@ -0,0 +1,141 @@
...
+// that are properly setup for use with Public or Sovereign clouds (depending on configuration)
+type multiEnvClientFactory struct{}
+
+// NewClientFactory returns a new multi-environment client factory
+func NewClientFactory() ClientFactory {
+	return &multiEnvClientFactory{}
+}
+
+// GetVirtualMachinesClientE returns a configured compute client, setup for proper cloud environment use.
+func (factory *multiEnvClientFactory) GetVirtualMachinesClientE(subscriptionID string) (compute.VirtualMachinesClient, error) {
```
rguthriemsft, line 53: @mjyeaney do you see a way that these api clients can live in the respective module vs. having them all implemented in client factory?

modules/azure/client_factory.go
```diff
@@ -0,0 +1,141 @@
...
+	// Validate Azure subscription ID
+	subscriptionID, err := getTargetAzureSubscription(subscriptionID)
+	if err != nil {
+		return compute.VirtualMachinesClient{}, err
+	}
+
+	// Lookup environment URI
+	baseURI, err := factory.getEnvironmentBaseURI()
```
rguthriemsft, line 61: can we generalize this pattern to get env base URI? maybe something like func getBaseURI(client interface{}) (string, err)

modules/azure/client_factory.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if !exists || envName == "" {
+		envName = autorest.PublicCloud.Name
+	}
+
+	return envName
+}
+
+// getEnvironmentBaseUri returns the ARM management URI for the configured Azure environment.
+func (factory *multiEnvClientFactory) getEnvironmentBaseURI() (string, error) {
```
rguthriemsft, line 114: should these methods have an E on end of name?

mjyeaney: Great catch - yes they should to follow conventions in the repo. Will adjust.

modules/azure/client_factory.go
mjyeaney: Refactored after our discussion and put more onus on modules to appropriately leverage client interface.

modules/azure/client_factory.go
mjyeaney: Generalized into single factory method as discussed - feedback welcome.

examples/azure/client_factory.md
```diff
@@ -0,0 +1,133 @@
+## Azure SDK Client Factory 
+
+This documentation provides and overview of the `client_factory.go` module, targeted use cases, and behaviors. 
+
+## Multi-cloud environment support
```
brikis98, line 5: These are solid docs, but does putting it in the `examples` folder make sense? Or should this be top-level documentation part of the [Terratest documentation site](https://terratest.gruntwork.io/docs/)? Note that the site is generated from Markdown repos in the same repo, so this would be mostly about moving these docs. E.g., There could be an "Azure authentication" page under "Testing best practices."

examples/azure/client_factory.md
```diff
@@ -0,0 +1,133 @@
...
+
+In order to better support Azure customers, we needed to support both Public and sovereign cloud environments (includes USGovernment, Germany, China, and Azure Stack).  To do this, we need to use the correct base URI's for the Azure management plane when using the REST API (either directly or via SDK). For the Go SDK, this can be accomplished by using the `WithBaseURI` suffixed calls when creating service clients.
```
brikis98, line 7: NIT: We should write this from the voice of Terratest, the open source project, and not Microsoft, the company. 

```suggestion
The Azure APIs need to support both Public and sovereign cloud environments (includes USGovernment, Germany, China, and Azure Stack).  To do this, we need to use the correct base URI's for the Azure management plane when using the REST API (either directly or via SDK). For the Go SDK, this can be accomplished by using the `WithBaseURI` suffixed calls when creating service clients.
```

examples/azure/client_factory.md
```diff
@@ -0,0 +1,133 @@
...
+
+Note that when using the "AzureStackCloud" setting, you must also set the `AZURE_ENVIRONMENT_FILEPATH` variable to point to a JSON file containing your Azure Stack URI deatils.
+
+>NOTE: The default behavior of the `client_factory` is to use the AzurePublicCloud environment. This requires no work from the developer to configure, and ensures consistent behavior with the current SDK code.       
+
+## Usage Patterns
+
+Modules authors will interact with the `client_factory` through the `NewClientFactory` method on the `azure` package as shown below:
+
+```golang
```
brikis98, line 100: One advantage of moving these docs into the Terratest docs site is that you can use `include` to pull in code snippets from real, tested code, rather than copy/pasting that code directly into Markdown. See the `{% include ... %}` usage here, for example: https://github.com/gruntwork-io/terratest/blob/master/docs/_docs/01_getting-started/quick-start.md#example-1-terraform-hello-world. That `include` gets rendered as follows: https://terratest.gruntwork.io/docs/getting-started/quick-start/#example-1-terraform-hello-world

examples/azure/client_factory.md
```diff
@@ -0,0 +1,133 @@
...
+	vmClient, ok := client.(compute.VirtualMachinesClient)
+	if !ok {
+		return nil, fmt.Errorf("Unable to convert client to type compute.VirtualMachinesClient")
+	}
+```
+
+The `ClientFactory` interface currently exposes the following methods - these will be built out more as more modules are developed and integrated:
+
+* `GetClientE(clientType ClientType, subscriptionID string)`: returns a client instance based on the ClientType passed, along with any error.
+* `GetClientBaseURIE(clientType ClientType, client interface{})`: returns the configured BaseURI for the given client, along with any error.
```
brikis98, line 118: Any reason not to expose the non-E versions of the methods (i.e., `GetClient` and `GetClientBaseURI`) that fail the test if there was any error?

modules/azure/aks.go
```diff
@@ -2,29 +2,35 @@ package azure
...
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
 // GetManagedClustersClientE is a helper function that will setup an Azure ManagedClusters client on your behalf
 func GetManagedClustersClientE(subscriptionID string) (*containerservice.ManagedClustersClient, error) {
-	// Validate Azure subscription ID
-	subscriptionID, err := getTargetAzureSubscription(subscriptionID)
+	// Create a cluster client
+	factory := NewClientFactory()
+	client, err := factory.GetClientE(ManagedClustersClientType, subscriptionID)
```
brikis98, line 15: To help my understanding: should _all_ Azure/Terratest code using this factory from now on? I know we're not updating it all in this particular PR, but is the expectation that we do update it this way in the future?

If that's the case, and apologies if this is a dumb question, but should this factory not be part of the AWS SDK itself then?

modules/azure/client_factory.go
```diff
@@ -0,0 +1,147 @@
+/*
+
+This file implements an Azure client factory that automatically handles setting up Base URI
+values for sovereign cloud support. Note the list of clients below is not initially exhaustive;
+rather, additional clients will me added as-needed.
```
brikis98, line 5:
```suggestion
rather, additional clients will be added as-needed.
```

modules/azure/client_factory.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+const (
+	// AzureEnvironmentEnvName is the name of the Azure environment to use. Set to one of the following:
+	//
+	// "AzureChinaCloud":        ChinaCloud
+	// "AzureGermanCloud":       GermanCloud
+	// "AzurePublicCloud":       PublicCloud
+	// "AzureUSGovernmentCloud": USGovernmentCloud
+	// "AzureStackCloud":		 Azure stack
```
brikis98, line 28: Should these be available in exported `const` vars?

modules/azure/client_factory.go
```diff
@@ -0,0 +1,147 @@
...
+	// GetClientE returns a client instance based on the ClientType passed, or optionally an error.
+	GetClientE(clientType ClientType, subscriptionID string) (interface{}, error)
+
+	// GetClientBaseURIE returns the configured BaseURI for the given client, based on the passed ClientType.
+	GetClientBaseURIE(clientType ClientType, client interface{}) (string, error)
+}
+
+// multiEnvClientFactory is used to coordinate handing out properly configured Azure SDK clients
+// that are properly setup for use with Public or Sovereign clouds (depending on configuration)
+type multiEnvClientFactory struct{}
```
brikis98, line 57: Why attach these methods to a `struct` rather than just have them as methods on the package? Using the `struct` requires an extra call to `NewClientFactory()` every time, and as the `struct` stores no data at all, it's not clear what the benefit of that is?

modules/azure/client_factory.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Create correct client based on type passed
+	switch clientType {
+	case SubscriptionsClientType:
+		return subscriptions.NewClientWithBaseURI(baseURI), nil
+	case VirtualMachinesClientType:
+		return compute.NewVirtualMachinesClientWithBaseURI(baseURI, subscriptionID), nil
+	case ManagedClustersClientType:
+		return containerservice.NewManagedClustersClientWithBaseURI(baseURI, subscriptionID), nil
+	}
```
brikis98, line 86: I found the need to cast these clients quite painful, adding several more error-prone lines to the API... I know Go doesn't have generics yet, so there are limits to what we can do here, but perhaps it would be cleaner to have a `GetClientXXX` method for each client type, and that method can return the exact type users ask for? E.g., `GetClientVirtualMachines`, `GetClientManagedClusters`, etc. This would result in a tiny bit of extra copy/paste within `client_factory.go`, but it would significantly reduce the boilerplate for all users, and remove the need for casting.

modules/azure/client_factory.go
```diff
@@ -0,0 +1,147 @@
...
+
+// getDefaultEnvironmentName returns either a configured Azure environment name, or the public default
+func (factory *multiEnvClientFactory) getDefaultEnvironmentName() string {
+	envName, exists := os.LookupEnv(AzureEnvironmentEnvName)
+
+	if !exists || envName == "" {
+		envName = autorestAzure.PublicCloud.Name
+	}
+
+	return envName
```
brikis98, line 116: NIT: slightly more functional programming / immutable approach, with simpler conditional:

```suggestion
	if exists && len(envName) > 0 {
		return envName
	}
	
	return autorestAzure.PublicCloud.Name
```

modules/azure/client_factory_test.go
```diff
@@ -0,0 +1,106 @@
...
+	autorest "github.com/Azure/go-autorest/autorest/azure"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Local consts for this file only
+var GovCloudEnvName = "AzureUSGovernmentCloud"
+var PublicCloudEnvName = "AzurePublicCloud"
+var ChinaCloudEnvName = "AzureChinaCloud"
+var GermanyCloudEnvName = "AzureGermanCloud"
```
brikis98, line 22: 1. These should be `const`, not `var`, as I don't think they ever have a reason to change?
1. The var names start with capitals, which actually makes them public. If you want them private, you need to make them lower case: e.g., `germanyCloudEnvName`. 
1. That said, as I mentioned in a comment above, shouldn't these be exported publicly so users can use them instead of copy/paste?

modules/azure/client_factory_test.go
```diff
@@ -0,0 +1,106 @@
+// /*
+
+// This file contains unit tests for the client factory implementation(s).
+
+// */
```
brikis98, line 5: NIT: what's with the strange comment syntax?

modules/azure/client_factory_test.go
```diff
@@ -0,0 +1,106 @@
...
+var ChinaCloudEnvName = "AzureChinaCloud"
+var GermanyCloudEnvName = "AzureGermanCloud"
+
+func TestDefaultEnvIsPublicWhenNotSet(t *testing.T) {
+	// Get a client factory
+	factory := multiEnvClientFactory{}
+
+	// save any current env value and restore on exit
+	currentEnv := os.Getenv(AzureEnvironmentEnvName)
+	defer os.Setenv(AzureEnvironmentEnvName, currentEnv)
```
brikis98, line 30: NIT:

```suggestion
	originalEnv := os.Getenv(AzureEnvironmentEnvName)
	defer os.Setenv(AzureEnvironmentEnvName, originalEnv)
```

modules/azure/client_factory_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// Set env var to gov
+	os.Setenv(AzureEnvironmentEnvName, "")
```
brikis98, line 33: Comment says "gov", but you're setting it to an empty string?

brikis98 (commented):
Thanks for the PR! Left some detailed comments within the PR. Here's a high level thought:

Could we simplify the API from:

```go
factory := NewClientFactory()
client, err := factory.GetClientE(SubscriptionsClientType, "")
if err != nil {
  return nil, err
}

// type cast and verify
subscriptionClient, ok := client.(subscriptions.Client)
if !ok {
  return nil, fmt.Errorf("Unable to convert client to subscriptions.Client")
}
```

To:

```go
client := GetClientSubscriptions(t, "")
```

To do that, you would:

1. Instead of one `GetClient` method, have one `GetClientXXX` method per client `XXX`.
1. Make `GetClient` a top-level package method, instead of a method on a `struct`.
1. Support `GetClient` in addition to `GetClientE`.

examples/azure/client_factory.md
mjyeaney: Agreed on docs...we picked an initial spot that didn't get us into messing with the docs jekyl site at the same time. @rguthriemsft - thoughts here?

examples/azure/client_factory.md
mjyeaney: Will revisit on outcome of moving this into Jekyll site (above).

examples/azure/client_factory.md
mjyeaney: Took a stab at making this in a more "active" voice...not sure how to describe without losing detail.

examples/azure/client_factory.md
mjyeaney: Removed these specific examples from the docs; however, the `client_factory` is not intended to be used by _test writers_; instead, it's in support of _module authors_ to correctly support multicloud environment details.

modules/azure/aks.go
mjyeaney: The Azure SDK does in fact have support for leveraging multi-environment clouds. However, it require exactly the steps this code is hiding, and is something we wanted to keep away from the concern of module writers. With that said, we do intend to deliver this feedback to the SDK team to help make this easier.

mjyeaney: Yes - all future modules should be using this factory (in subsequent PR's)

modules/azure/client_factory.go
mjyeaney: Done - thanks for the suggestion!

modules/azure/client_factory_test.go
mjyeaney: Made these lower-case, as they're only in support of the test itself. End users would configure these via environment variables (likely set outside of code).

modules/azure/client_factory_test.go
mjyeaney: Double-commenting resulting from VSCode. Fixed and removed multi-line comment markers.

modules/azure/client_factory_test.go
mjyeaney: Fixed - thanks!

modules/azure/client_factory_test.go
mjyeaney: Bad comment - resolved

modules/azure/client_factory.go
mjyeaney: The environment vars would be set outside any code - listed here for reference only.

modules/azure/client_factory.go
mjyeaney: The idea originally was to "hide" the inner workings of the client_factory and provide only the needed methods (e.g., "pit of success") as this wasn't intended to be used outside of modules. 

That said - we have removed the extra interface / struct and just made all methods top-level.

modules/azure/client_factory.go
mjyeaney: Added explicit methods for each client type and removed general case / casting need.

rguthriemsft:
Look good to me - Here is a passing CI run for this PR.

rguthriemsft (dismissed):

examples/azure/client_factory.md
brikis98: Discussion of docs is happening [in Slack](https://gruntwork-io.slack.com/archives/C015S5V8DE2/p1602110186026800?thread_ts=1601681297.020100&cid=C015S5V8DE2), so I'll mark this one as unresolved for now.

modules/azure/client_factory_test.go
```diff
@@ -0,0 +1,143 @@
...
+		{"PublicCloud/SubscriptionClient", publicCloudEnvName, autorest.PublicCloud.ResourceManagerEndpoint},
+		{"ChinaCloud/SubscriptionClient", chinaCloudEnvName, autorest.ChinaCloud.ResourceManagerEndpoint},
+		{"GermanCloud/SubscriptionClient", germanyCloudEnvName, autorest.GermanCloud.ResourceManagerEndpoint},
+	}
+
+	// save any current env value and restore on exit
+	currentEnv := os.Getenv(AzureEnvironmentEnvName)
+	defer os.Setenv(AzureEnvironmentEnvName, currentEnv)
+
+	for _, tt := range cases {
```
brikis98, line 68: I think you need to store a local copy:

```suggestion
	for _, tt := range cases {
		// The following is necessary to make sure testCase's values don't
		// get updated due to concurrency within the scope of t.Run(..) below	
		tt := tt
```

It might not be necessary as you aren't using `t.Parallel()`, but it's always a good sanity check to have in case that ever changes in the future.

modules/azure/client_factory_test.go
```diff
@@ -0,0 +1,143 @@
...
+		{"PublicCloud/VMClient", publicCloudEnvName, autorest.PublicCloud.ResourceManagerEndpoint},
+		{"ChinaCloud/VMClient", chinaCloudEnvName, autorest.ChinaCloud.ResourceManagerEndpoint},
+		{"GermanCloud/VMClient", germanyCloudEnvName, autorest.GermanCloud.ResourceManagerEndpoint},
+	}
+
+	// save any current env value and restore on exit
+	currentEnv := os.Getenv(AzureEnvironmentEnvName)
+	defer os.Setenv(AzureEnvironmentEnvName, currentEnv)
+
+	for _, tt := range cases {
```
brikis98, line 99:
```suggestion
	for _, tt := range cases {
		// The following is necessary to make sure testCase's values don't
		// get updated due to concurrency within the scope of t.Run(..) below	
		tt := tt
```

modules/azure/client_factory_test.go
```diff
@@ -0,0 +1,143 @@
...
+		{"PublicCloud/ManagedClustersClient", publicCloudEnvName, autorest.PublicCloud.ResourceManagerEndpoint},
+		{"ChinaCloud/ManagedClustersClient", chinaCloudEnvName, autorest.ChinaCloud.ResourceManagerEndpoint},
+		{"GermanCloud/ManagedClustersClient", germanyCloudEnvName, autorest.GermanCloud.ResourceManagerEndpoint},
+	}
+
+	// save any current env value and restore on exit
+	currentEnv := os.Getenv(AzureEnvironmentEnvName)
+	defer os.Setenv(AzureEnvironmentEnvName, currentEnv)
+
+	for _, tt := range cases {
```
brikis98, line 130:
```suggestion
	for _, tt := range cases {
		// The following is necessary to make sure testCase's values don't
		// get updated due to concurrency within the scope of t.Run(..) below	
		tt := tt
```

brikis98 (commented):
The latest updates are much cleaner and clearer, thank you! I had a few minor NITs, plus we need to resolve the docs question, after which, this is good to go!

rguthriemsft:
@yorinasub17 , @brikis98 we have update the documentation and will need to wait for #673 to merge so we can add a link to this document in the contribution guide.   Can you please review the client_factory.md and let us know if you have any feedback?

docs/_docs/04_community/client_factory.md
```diff
@@ -0,0 +1,176 @@
+# Azure SDK Client Factory
```
brikis98, line 1: In order for this to show up in the docs site, it needs a front matter. See the other `.md` files in the same folder for an example!

docs/_docs/04_community/client_factory.md
```diff
@@ -0,0 +1,176 @@
...
+
+This documentation provides and overview of the `client_factory.go` module, targeted use cases, and behaviors.  This module is intended to provide support for and simplify working with Azure's multiple cloud environments (Azure Public, Azure Government, Azure China, Azure Germany and Azure Stack).  Developers looking to contribute to additional support for Azure to TerraTest should leverage client_factory and use the patterns below to add a resource REST client from Azure Go SDK.  By doing so, it provides a consistent means for developers using TerraTest to test their Azure Infrastructure to connect to the correct cloud and its associated REST apis.
```
brikis98, line 3:
```suggestion
This documentation provides and overview of the `client_factory.go` module, targeted use cases, and behaviors.  This module is intended to provide support for and simplify working with Azure's multiple cloud environments (Azure Public, Azure Government, Azure China, Azure Germany and Azure Stack).  Developers looking to contribute to additional support for Azure to Terratest should leverage client_factory and use the patterns below to add a resource REST client from Azure Go SDK.  By doing so, it provides a consistent means for developers using Terratest to test their Azure Infrastructure to connect to the correct cloud and its associated REST apis.
```

docs/_docs/04_community/client_factory.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Background
+
+The Azure REST APIs support both Public and sovereign cloud environments (at the moment this includes Public, US Government, Germany, China, and Azure Stack environments).  If you are interacting with an environment other than public cloud, you need to set the base URI for the Azure REST API you are interacting with.
+
+### Base URI
+
+You must use the correct base URI's for the Azure REST API's (either directly or via Azure SDK for GO) to communicate with a cloud environment other than Azure Public. The Azure Go SDK supports this by using the `WithBaseURI` suffixed calls when creating service clients. For example, when using the `VirtualMachinesClient` with the public cloud, a developer would normally write code for the public cloud like so:
+
+```golang
```
brikis98, line 13: One of the benefits of being in the docs site is you can include snippets from real examples. You add them to https://github.com/gruntwork-io/terratest/blob/master/docs/_data/examples.yml and can then include them using an `include`. See https://github.com/gruntwork-io/terratest/blob/d0363f939d71a682726a72d2f82787e43e324b28/docs/_docs/01_getting-started/quick-start.md#example-1-terraform-hello-world for an example and https://terratest.gruntwork.io/docs/getting-started/quick-start/#example-1-terraform-hello-world for how it's rendered. 

This will make the code snippets more maintainable, as they'll actually go through our test suite.

docs/_docs/04_community/client_factory.md
```diff
@@ -0,0 +1,176 @@
...
+    subscriptionID := "your subscription ID"
+
+    // Create a VM client and return
+    vmClient, err := compute.NewVirtualMachinesClient(subscriptionID)
+
+    // Use client / etc
+}
+```
+
+However, this code will not work in non-Public cloud environments as the REST endpoints have different URIs depending on environment.  Instead, you need to use an alternative method (provided in the Azure REST SDK for Go) to get a properly configured client(*all REST API clients should support this alternate method*):
```
brikis98, line 28:
```suggestion
However, this code will not work in non-Public cloud environments as the REST endpoints have different URIs depending on environment.  Instead, you need to use an alternative method (provided in the Azure REST SDK for Go) to get a properly configured client (*all REST API clients should support this alternate method*):
```

docs/_docs/04_community/client_factory.md
```diff
@@ -0,0 +1,176 @@
...
+|"AzureGermanCloud"         |GermanCloud        |
+|"AzurePublicCloud"         |PublicCloud        |
+|"AzureUSGovernmentCloud"   |USGovernmentCloud  |
+|"AzureStackCloud"          |Azure stack        |
+
+When using the "AzureStackCloud" setting, you MUST also set the `AZURE_ENVIRONMENT_FILEPATH` variable to point to a JSON file containing your Azure Stack URI details.
+
+## Putting it all together
+
+ `client_factory` implements this pattern described above in order to instantiate and return properly configured *REST SDK for GO* clients so that test implementers don't have to consider REST API client implementation as long as they have the correct `AZURE_ENVIRONMENT` env setting.  If this environment variable is not set, the client will assume public cloud as the cloud environment to communicate with.  We strongly recommend developers creating TerraTest helper methods for Azure use this pattern with client factory to create REST API clients.  This will reduce effort for TerraTest users creating test for Azure resources.
```
brikis98, line 66:
```suggestion
 `client_factory` implements this pattern described above in order to instantiate and return properly configured *REST SDK for GO* clients so that test implementers don't have to consider REST API client implementation as long as they have the correct `AZURE_ENVIRONMENT` env setting.  If this environment variable is not set, the client will assume public cloud as the cloud environment to communicate with.  We strongly recommend developers creating Terratest helper methods for Azure use this pattern with client factory to create REST API clients.  This will reduce effort for Terratest users creating test for Azure resources.
```

brikis98 (commented):
Very close now! Just a couple grammar tweaks. The only other item is using `include` to pull in real, tested code snippets, instead of the copy/pasted ones, as the latter will go out of date very quickly.

docs/_docs/04_community/client_factory.md
```diff
@@ -0,0 +1,176 @@
+# Azure SDK Client Factory
+
+This documentation provides and overview of the `client_factory.go` module, targeted use cases, and behaviors.  This module is intended to provide support for and simplify working with Azure's multiple cloud environments (Azure Public, Azure Government, Azure China, Azure Germany and Azure Stack).  Developers looking to contribute to additional support for Azure to Terratest should leverage client_factory and use the patterns below to add a resource REST client from Azure Go SDK.  By doing so, it provides a consistent means for developers using Terratest to test their Azure Infrastructure to connect to the correct cloud and its associated REST apis.
```
polatengin, line 3: I suggest to add something like, "_future environments will be available automatically_" or "_, etc_" into the parantesis

rguthriemsft:

You can check the status of the CI Pipeline logs here ; https://github.com/rguthriemsft/terratest/actions/runs/331858216

rguthriemsft:
@brikis98, we put in changes to the docs site to support parsing code by tags into the examples in this commit.  Please take a look https://github.com/gruntwork-io/terratest/pull/663/commits/a5be1edf6ccf673c18beb8aa702c036151375f2b

docs/assets/js/examples.js
```diff
@@ -126,6 +149,13 @@ $(document).ready(function () {
     })
   }
 
+  function searchTagInLines (tagRegExp, lines) {
+    for (var i=0; i<lines.length; i++) {
+        if (lines[i].match(tagRegExp)) return i;
+    }
+    return -1;
```
brikis98, line 156:
```suggestion
    return lines.findIndex(line => line.match(tagRegExp))
```

docs/assets/js/examples.js
```diff
@@ -101,6 +101,7 @@ $(document).ready(function () {
       const $activeCodeSnippet = $(activeCodeSnippet)
       const exampleTarget = $(this).data('example')
       const fileId = $(this).data('target')
+      const rangeId = $(this).data('range-id')
```
brikis98, line 104: Instead of `range-id`, should this be called `snippet-id`, here and throughout the PR, given that you use `snippet-tag-xxx` everywhere?

brikis98 (commented):
The new snippet tag support is great, thank you for adding that!

I left a bunch of minor comments on code style and potential minor bugs, but I'm very excited to get this merged 👍

docs/assets/js/examples.js
mjyeaney: Renamed all include params and local var to match "snippet" concept.

docs/assets/js/examples.js
```diff
@@ -114,7 +115,29 @@ $(document).ready(function () {
...
+            // Search the array for "// snippet-tag-end::{id}" - save location
+            //endLine = lines.indexOf(`// snippet-tag-end::${rangeId}`)
+            endLine = searchTagInLines(`\/\/ snippet-tag-end::${rangeId}`, lines)
+
+            // If you have both a start and end, slice as below
+            if ((startLine != -1) && (endLine != -1)) {
+              range = lines.slice(startLine + 2, endLine)
+              $activeCodeSnippet.find('code').text(range.join('\n'))
+            } else {
+              console.error('Could not find specified range.')
```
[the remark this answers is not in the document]
mjyeaney: Modified code to dump entire code snippet if the event the proper range isn't found.

docs/assets/js/examples.js
```diff
@@ -114,7 +115,29 @@ $(document).ready(function () {
             // Remove the website::tag::xxx:: comment entirely from the code snippet
             content = content.replace(/^.*website::tag.*\n?/mg, '')
           }
-          $activeCodeSnippet.find('code').text(content)
+          // Find the range specified by range-id if specified
+          if (rangeId != "") {
```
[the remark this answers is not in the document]
mjyeaney: Extracted snippet searching to its own function.

docs/assets/js/examples.js
```diff
@@ -114,7 +115,29 @@ $(document).ready(function () {
...
+            //startLine = lines.indexOf(`// snippet-tag-start::${rangeId}`)
+            startLine = searchTagInLines(`\/\/ snippet-tag-start::${rangeId}`, lines)
+
+            // Search the array for "// snippet-tag-end::{id}" - save location
+            //endLine = lines.indexOf(`// snippet-tag-end::${rangeId}`)
+            endLine = searchTagInLines(`\/\/ snippet-tag-end::${rangeId}`, lines)
+
+            // If you have both a start and end, slice as below
+            if ((startLine != -1) && (endLine != -1)) {
+              range = lines.slice(startLine + 2, endLine)
```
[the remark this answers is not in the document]
mjyeaney: There is an extra empty line that is present with just  +1; this was a simple way to remove it without potentially modifying the other use-cases of the `explorer.html` and `example.html` includes.

docs/assets/js/examples.js
```diff
@@ -114,7 +115,29 @@ $(document).ready(function () {
...
             content = content.replace(/^.*website::tag.*\n?/mg, '')
           }
-          $activeCodeSnippet.find('code').text(content)
+          // Find the range specified by range-id if specified
+          if (rangeId != "") {
+            // Split the content into an array of lines
+            lines = content.split('\n')
+            // Search the array for "// snippet-tag-start::{id}" - save location
+            //startLine = lines.indexOf(`// snippet-tag-start::${rangeId}`)
+            startLine = searchTagInLines(`\/\/ snippet-tag-start::${rangeId}`, lines)
```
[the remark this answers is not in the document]
mjyeaney: Adjusted regex per comments above.

docs/assets/js/examples.js
```diff
@@ -114,7 +115,29 @@ $(document).ready(function () {
...
+
+            // Search the array for "// snippet-tag-end::{id}" - save location
+            //endLine = lines.indexOf(`// snippet-tag-end::${rangeId}`)
```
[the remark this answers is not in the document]
mjyeaney: Removed dead comments.

docs/assets/js/examples.js
```diff
@@ -114,7 +115,29 @@ $(document).ready(function () {
             // Remove the website::tag::xxx:: comment entirely from the code snippet
             content = content.replace(/^.*website::tag.*\n?/mg, '')
           }
-          $activeCodeSnippet.find('code').text(content)
+          // Find the range specified by range-id if specified
+          if (rangeId != "") {
+            // Split the content into an array of lines
+            lines = content.split('\n')
+            // Search the array for "// snippet-tag-start::{id}" - save location
+            //startLine = lines.indexOf(`// snippet-tag-start::${rangeId}`)
```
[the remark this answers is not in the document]
mjyeaney: Removed dead comments.

docs/_includes/examples/explorer.html
```diff
@@ -3,14 +3,24 @@
...
-            class="examples__nav-item nav-{{ example.id }}"
-            data-id="{{ example.id }}"
-            data-name="{{ example.name }}"
-            data-url="{{ site.github_api_url }}{{ example.url }}"
-            >
-            <img src="{{ site.baseurl }}{{ example.image }}" alt="{{ example.name }}" />
-          </div>
+          {% assign render_example = false %}
+          
+          {% if include.test_for_display == true and example.display_in_examples == nil %}
```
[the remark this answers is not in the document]
mjyeaney: This flag was originally there to distinguish the use case for the include. Today, this same include is used for the examples browser _as well as_ this new snippet functionality. However, we don't want want examples that are ONLY used as snippets to show up in the "Examples" browser.  Ultimately, we decided to NOT modify anything in `data/examples.yml` except the new example, which simplified some logic. To that end, this `test_for_display` flag could be removed - thanks for calling out.

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
...
+|"AzureStackCloud"          |Azure stack        |
+
+When using the "AzureStackCloud" setting, you MUST also set the `AZURE_ENVIRONMENT_FILEPATH` variable to point to a JSON file containing your Azure Stack URI details.
+
+## Putting it all together
+
+ `client_factory` implements this pattern described above in order to instantiate and return properly configured *REST SDK for GO* clients so that test implementers don't have to consider REST API client implementation as long as they have the correct `AZURE_ENVIRONMENT` env setting.  If this environment variable is not set, the client will assume public cloud as the cloud environment to communicate with.  We strongly recommend developers creating Terratest helper methods for Azure use this pattern with client factory to create REST API clients.  This will reduce effort for Terratest users creating test for Azure resources.
+
+>NOTE: TERRAFORM uses [ARM_ENVIRONMENT](https://www.terraform.io/docs/backends/types/azurerm.html#environment) environment variable to set the correct cloud environment.  
+<!-- -->
```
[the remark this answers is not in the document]
mjyeaney: This is a result of a linting rule failure on our end complaining that there should be a line break between blockquote regions. If this is not an issue, we can remove it. /cc @rguthriemsft

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Background
+
+The Azure REST APIs support both Public and sovereign cloud environments (at the moment this includes Public, US Government, Germany, China, and Azure Stack environments).  If you are interacting with an environment other than public cloud, you need to set the base URI for the Azure REST API you are interacting with.
+
+### Base URI
+
+You must use the correct base URI's for the Azure REST API's (either directly or via Azure SDK for GO) to communicate with a cloud environment other than Azure Public. The Azure Go SDK supports this by using the `WithBaseURI` suffixed calls when creating service clients. For example, when using the `VirtualMachinesClient` with the public cloud, a developer would normally write code for the public cloud like so:
+
+```golang
```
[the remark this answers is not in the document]
mjyeaney: Given that these refer to _concepts_, we weren't sure if it would apply.  Glad to adjust if you think it would be a better fit.

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
+---
+layout: collection-browser-doc
+title: Azure SDK Client Factory
```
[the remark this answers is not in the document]
mjyeaney: Moved to the "Contributions" section as noted above.

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
...
+category: community
```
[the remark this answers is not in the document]
mjyeaney: Moved to "Contributing"

docs/_config.yml
```diff
@@ -34,6 +34,7 @@ twitter_username: https://twitter.com/gruntwork_io
 github_username:  https://github.com/gruntwork-io
 
 github_api_url: https://raw.githubusercontent.com/gruntwork-io/terratest/master
+#github_api_url: https://raw.githubusercontent.com/rguthriemsft/terratest/pull-request-625
```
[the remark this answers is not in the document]
mjyeaney: Removed - note that certain snippets / files will load with a `404: Not found` error as they aren't in master until this PR is merged.

docs/_docs/04_community/client-factory.md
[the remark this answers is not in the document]
rguthriemsft: @brikis98 the markdown linter I was using complained of no line break.  Let us know if you want to remove it.

rguthriemsft (approved):
Awaiting feedback from @brikis98 for one comment he left in his review but Look good to me

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
+---
+layout: collection-browser-doc
+title: Azure SDK Client Factory
+category: contributing
```
brikis98, line 4: This ends up as a totally new top-level nav category, with solely this Azure stuff in it, which probably isn't what we want:

Screen Shot 2020-11-17 at 5 11 15 PM

Note that we have an existing contributing page already under the Community nav:

Screen Shot 2020-11-17 at 5 12 52 PM

Probably the easiest thing to do would be to move the content of this page into that existing Contributing page, just below the "Developing Terratest" section.

docs/_docs/04_community/client-factory.md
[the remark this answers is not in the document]
brikis98: In general, cross-linking to real, tested, maintained code tends age better than copy/pasting code into Markdown docs. That said, that assumes there's a reasonable example somewhere that demonstrates what you want that you can cross-link to. If the code snippets shown here do show up in examples elsewhere, then tagging and cross-linking them is worthwhile. If there is no appropriate example, and creating one would be too weird and actually make maintenance harder, then what you have is totally fine.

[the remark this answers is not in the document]
brikis98: BTW, the syntax highlighting isn't working on this page. You may need to add the prism CSS?

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
...
+
+Using code similar to above, you can communicate with any Azure cloud environment just by changing the base URI that is passed to the clients (Azure Public shown in above example).
+
+## Lookup Environment Metadata
+
+Developers MUST avoid hardcoding these base URI's.  Instead, they should be looked up from an authoritative source. The AutoRest-GO library (used by the Go SDK) provides such functionality. The `client_factory` module makes use of the AutoRest `EnvironmentFromName(envName string)` function to return the appropriate structure.  This method and Environment structure is documented on GoDoc [here](https://godoc.org/github.com/Azure/go-autorest/autorest/azure#EnvironmentFromName).
+
+To configure different cloud environments, we will use the same `AZURE_ENVIRONMENT` environment variable that the Go SDK uses. This can currently be set to one of the following values:
+
+|Value                      |Cloud Environment  |
```
brikis98, line 72: Doh. Looks like we don't have table styling in the CSS for this repo. Or maybe we do and it's just not being applied for some reason? Prob worth filing a bug and we can come back to it later.

Screen Shot 2020-11-17 at 5 18 34 PM

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Putting it all together
+
+ `client_factory` implements this pattern described above in order to instantiate and return properly configured *REST SDK for GO* clients so that test implementers don't have to consider REST API client implementation as long as they have the correct `AZURE_ENVIRONMENT` env setting.  If this environment variable is not set, the client will assume public cloud as the cloud environment to communicate with.  We strongly recommend developers creating Terratest helper methods for Azure use this pattern with client factory to create REST API clients.  This will reduce effort for Terratest users creating test for Azure resources.
+
+>NOTE: TERRAFORM uses [ARM_ENVIRONMENT](https://www.terraform.io/docs/backends/types/azurerm.html#environment) environment variable to set the correct cloud environment.  
+<!-- -->
+>NOTE: The default behavior of the `client_factory` is to use the AzurePublicCloud environment. This requires no work from the developer to configure, and ensures consistent behavior with the current SDK code.
```
brikis98, line 88: Interestingly, two quote blocks back to back look a bit weird:

Screen Shot 2020-11-17 at 5 19 21 PM

Consider: 

1. Putting a real new line between them to space them out vertically.
1. Using the ℹ️ emoji and making the word NOTE bold to make it more of a "callout".

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
...
+
+ If you require a client that is not already implemented in client factory for your helper method, you will need to create a corresponding method that instantiates the client and accepts base URI following the patterns discussed.  Below is a walkthrough for adding a client to client factory.
+
+## Walkthrough, adding a client to client_factory
+
+### Add your client namespace to client factory
+
+In the Azure SDK for GO, each service should have a module that implements that services client.  You can find the correct module [here](https://godoc.org/github.com/Azure/azure-sdk-for-go).  Add that module to the client factory imports.  Below is an example for client imports that shows clients for compute, container service and subscriptions.
+
+{% include examples/explorer.html example_id='client-factory' file_id='client_factory_code' class='wide quick-start-examples' skip_learn_more=true skip_view_on_github=true skip_tags=true snippet_id='client_factory_example.imports' %}
```
brikis98, line 100: A few of these seem broken:

Screen Shot 2020-11-17 at 5 21 17 PM

But some do work, so I don't think it's a general problem with the snippets, but perhaps just a couple broken IDs or links or typos?

Screen Shot 2020-11-17 at 5 21 22 PM

docs/_docs/04_community/client-factory.md
```diff
@@ -0,0 +1,124 @@
...
+
+Below is an example of the Virtual Machines client unit test:
+
+{% include examples/explorer.html example_id='client-factory' file_id='client_factory_test' class='wide quick-start-examples' skip_learn_more=true skip_view_on_github=true skip_tags=true snippet_id='client_factory_example.UnitTest' %}
+
+### Use your CreateClient method in your helper
+
+We now can use this client creation method in our helpers to create a Virtual Machines client.  Below is an example for how to call into this create method from `client_factory`:
+
+{% include examples/explorer.html example_id='client-factory' file_id='client_factory_helper' class='wide quick-start-examples' skip_learn_more=true skip_view_on_github=true skip_tags=true snippet_id='client_factory_example.helper' %}
```
brikis98, line 124: This one, however, seems to include the  ENTIRE file, rather than a snippet... So it takes up a lot of space on the page. I assume that's not intentional?

docs/assets/js/examples.js
```diff
@@ -126,6 +134,28 @@ $(document).ready(function () {
...
+      return range.join('\n')
+    } else {
+      console.error('Could not find specified range.')
+      return content
+    }
+  }
+
+  function searchTagInLines (tagRegExp, lines) {
+    return lines.findIndex(line => line.match(tagRegExp))
+  }
```
brikis98, line 157: NIT: please use semicolons after every line.

docs/_docs/04_community/client_factory.md
mjyeaney: Added and resolved.

docs/_docs/04_community/client_factory.md
mjyeaney: Leaving these 2 code samples as-is as they are intended to show an approach, not working code. Can revisit if needed.

docs/_docs/04_community/client-factory.md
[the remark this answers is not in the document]
mjyeaney: Syntax highlighting is fixed - added Prism CSS.

docs/_docs/04_community/client-factory.md
mjyeaney: Moved content to "Contributing" section as called out.

docs/_docs/04_community/client-factory.md
mjyeaney: Added bug #722

docs/_docs/04_community/client-factory.md
mjyeaney: Refactored blockquotes into bulleted list.

docs/_docs/04_community/client-factory.md
mjyeaney: The 404's are a result of the content being pulled from the "master" branch of the live repo, which doesn't yet contain the appropriate files (they're in the PR). Also, when an entire file is loaded, it means the tags/snippets weren't found and the `examples.js` logic falls back to rendering the entire file.

docs/_docs/04_community/client-factory.md
brikis98: Thanks!

brikis98 (approved):
Code changes LGTM! Merging now, thank you!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.31.1

PR #664, opened 2020-10-01
Add a new helper function to construct terratest option with default retryable errors

yorinasub17:
I've been going through all our repos and setting the retryable errors to these settings, but realized that it wasn't very DRY. Then I had the insight that this is probably something everyone would want, so decided to elevate it here.

### DIFF
--- a/modules/terraform/options.go
+++ b/modules/terraform/options.go
@@ -10,6 +10,28 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	DefaultRetryableTerraformErrors = map[string]string{
+		// Helm related terraform calls may fail when too many tests run in parallel. While the exact cause is unknown,
+		// this is presumably due to all the network contention involved. Usually a retry resolves the issue.
+		".*read: connection reset by peer.*": "Failed to reach helm charts repository.",
+		".*transport is closing.*":           "Failed to reach Kubernetes API.",
+
+		// `terraform init` frequently fails in CI due to network issues accessing plugins. The reason is unknown, but
+		// eventually these succeed after a few retries.
+		".*unable to verify signature.*":             "Failed to retrieve plugin due to transient network error.",
+		".*unable to verify checksum.*":              "Failed to retrieve plugin due to transient network error.",
+		".*no provider exists with the given name.*": "Failed to retrieve plugin due to transient network error.",
+		".*registry service is unreachable.*":        "Failed to retrieve plugin due to transient network error.",
+		".*Error installing provider.*":              "Failed to retrieve plugin due to transient network error.",
+
+		// Provider bugs where the data after apply is not propagated. This is usually an eventual consistency issue, so
+		// retrying should self resolve it.
+		// See https://github.com/terraform-providers/terraform-provider-aws/issues/12449 for an example.
+		".*Provider produced inconsistent result after apply.*": "Provider eventual consistency error.",
+	}
+)
+
 // Options for running Terraform commands
 type Options struct {
 	TerraformBinary string // Name of the binary that will be used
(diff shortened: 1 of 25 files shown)


modules/terraform/options.go
```diff
@@ -37,3 +37,38 @@ type Options struct {
 	Logger                   *logger.Logger         // Set a non-default logger that should be used. See the logger package for more info.
 	Parallelism              int                    // Set the parallelism setting for Terraform
 }
+
+// NewTerraformOptions returns a Options object with sensible defaults for retryable errors. The included retryable
+// errors are typical errors that most terraform modules encounter during testing, and are known to self resolve upon
+// retrying.
+func NewTerraformOptionsWithDefaultRetryableErrors(modulePath string, vars map[string]interface{}) *Options {
```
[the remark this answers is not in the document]
yorinasub17: Ah the `WithDefaultOptions` approach is much better. Will update to that.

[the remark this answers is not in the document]
yorinasub17: Ok overhauled to take the new approach: 8630111

modules/terraform/options.go
```diff
@@ -37,3 +40,54 @@ type Options struct {
 	Logger                   *logger.Logger         // Set a non-default logger that should be used. See the logger package for more info.
 	Parallelism              int                    // Set the parallelism setting for Terraform
 }
+
+// Clone makes a deep copy of the Options object and returns it.
+func (options *Options) Clone() (*Options, error) {
+	newOptions := Options{}
+	if err := copier.Copy(&newOptions, options); err != nil {
```
brikis98, line 47: Oh, neat. We should prob use this in Terragrunt in the future too!

modules/terraform/options.go
```diff
@@ -37,3 +40,54 @@ type Options struct {
...
+
+// WithDefaultRetryableErrors makes a copy of the Options object and returns an updated object with sensible defaults
+// for retryable errors. The included retryable errors are typical errors that most terraform modules encounter during
+// testing, and are known to self resolve upon retrying.
+// This will fail the test if there are any errors in the cloning process.
+func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options {
+	newOptions, err := originalOptions.Clone()
+	require.NoError(t, err)
+
+	retryableTerraformErrors := map[string]string{
```
brikis98, line 61: NIT: Should this be extracted into a variable that's shared, in case someone wants to use it outside of this method?

modules/terraform/options.go
```diff
@@ -37,3 +40,54 @@ type Options struct {
...
+		".*Provider produced inconsistent result after apply.*": "Provider eventual consistency error.",
+	}
+	if newOptions.RetryableTerraformErrors == nil {
+		newOptions.RetryableTerraformErrors = map[string]string{}
+	}
+	for k, v := range retryableTerraformErrors {
+		newOptions.RetryableTerraformErrors[k] = v
+	}
+
+	// These defaults for retry configuration are arbitrary, but has worked well in practice across Gruntwork
```
brikis98, line 87: NIT: 

```suggestion
	// These defaults for retry configuration are arbitrary, but have worked well in practice across Gruntwork
```

modules/terraform/options.go
```diff
@@ -37,3 +40,54 @@ type Options struct {
...
+		return nil, err
+	}
+	return &newOptions, nil
+}
+
+// WithDefaultRetryableErrors makes a copy of the Options object and returns an updated object with sensible defaults
+// for retryable errors. The included retryable errors are typical errors that most terraform modules encounter during
+// testing, and are known to self resolve upon retrying.
+// This will fail the test if there are any errors in the cloning process.
+func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options {
```
brikis98, line 57: I like this pattern 👍 

In the future, we should adopt it for `packer.Options`, `docker.Options`, etc.

brikis98 (commented):
Just a couple NITs and this is good to go.

modules/terraform/options.go
yorinasub17: Good idea. Done! e5c4287

modules/terraform/options.go
yorinasub17: This library has a few issues with nested structs, so it probably will be problematic in Terragrunt. I added some comments on which fields will not be a deep clone: 2b049ca

brikis98: Noted

brikis98: Actually, wait... If we don't deep copy those two fields, and perhaps other nested structs in the future, this may cause issues. For example, someone may be doing table-driven tests of a single Terraform module, where they test the module in 20 different configurations using `t.Run()` and `t.Parallel()` to create sub-tests. They might be tempted to configure the basic `terraform.Options` once, use `Clone()` to copy it for each sub-test, and within that subtest, assume they can override the `Vars` (for different settings) and `Logger` (to have different logging for each test)... But without a deep copy, the latter would be shared amongst all the tests...

Could we work around this by manually calling `copier.Copy` on `SshAgent` and `Logger`? Or do those have nested structs that would run into issues anyway? Alternatively, we can try to put a bigger warning on the `Clone` method docs about not sharing a cloned `Options` if they plan to override these specific fields...

yorinasub17: Pardon the delay, but addressed this in 2b1d1e5

brikis98 (approved):
LGTM!

yorinasub17:
Build failed, but for what looks like a transient issue. Since it is unrelated to this PR, will merge now.

Thanks for the review!

PR #665, opened 2020-10-01
Add function for docker compose to return stdout

rhoboat:
What else does this PR need? 

Purpose: I've just added a function that lets docker compose return only `stdout`, so that TLS Scripts followups can get just the Certificate ARN from `stdout` and clean up the resource during the test cleanup stage.

### DIFF
--- a/modules/docker/docker_compose.go
+++ b/modules/docker/docker_compose.go
@@ -16,30 +16,20 @@ type Options struct {
 
 // RunDockerCompose runs docker-compose with the given arguments and options and return stdout/stderr.
 func RunDockerCompose(t testing.TestingT, options *Options, args ...string) string {
-	out, err := RunDockerComposeE(t, options, args...)
+	out, err := RunDockerComposeE(t, false, options, args...)
 	if err != nil {
 		t.Fatal(err)
 	}
 	return out
 }
 
-// RunDockerComposeE runs docker-compose with the given arguments and options and return stdout/stderr.
-func RunDockerComposeE(t testing.TestingT, options *Options, args ...string) (string, error) {
-	cmd := shell.Command{
-		Command: "docker-compose",
-		// We append --project-name to ensure containers from multiple different tests using Docker Compose don't end
-		// up in the same project and end up conflicting with each other.
-		Args:       append([]string{"--project-name", t.Name()}, args...),
-		WorkingDir: options.WorkingDir,
-		Env:        options.EnvVars,
-		Logger:     options.Logger,
-	}
-
-	return shell.RunCommandAndGetOutputE(t, cmd)
-}
-
 // RunDockerComposeStdoutE runs docker-compose with the given arguments and options and return only stdout.
 func RunDockerComposeStdOutE(t testing.TestingT, options *Options, args ...string) (string, error) {
+	return RunDockerComposeE(t, true, options, args...)
+}
+
+// RunDockerComposeE runs docker-compose with the given arguments and options and return stdout/stderr.
+func RunDockerComposeE(t testing.TestingT, stdout bool, options *Options, args ...string) (string, error) {
 	cmd := shell.Command{
 		Command: "docker-compose",
 		// We append --project-name to ensure containers from multiple different tests using Docker Compose don't end
@@ -50,5 +40,8 @@ func RunDockerComposeStdOutE(t testing.TestingT, options *Options, args ...strin
 		Logger:     options.Logger,
 	}
 
-	return shell.RunCommandAndGetStdOutE(t, cmd)
+	if stdout {
+		return shell.RunCommandAndGetStdOutE(t, cmd)
+	}
+	return shell.RunCommandAndGetOutputE(t, cmd)
 }
--- a/modules/docker/docker_compose.go
+++ b/modules/docker/docker_compose.go
@@ -37,3 +37,18 @@ func RunDockerComposeE(t testing.TestingT, options *Options, args ...string) (st
 
 	return shell.RunCommandAndGetOutputE(t, cmd)
 }
+
+// RunDockerComposeE runs docker-compose with the given arguments and options and return only stdout.
+func RunDockerComposeStdOutE(t testing.TestingT, options *Options, args ...string) (string, error) {
+	cmd := shell.Command{
+		Command: "docker-compose",
+		// We append --project-name to ensure containers from multiple different tests using Docker Compose don't end
+		// up in the same project and end up conflicting with each other.
+		Args:       append([]string{"--project-name", t.Name()}, args...),
+		WorkingDir: options.WorkingDir,
+		Env:        options.EnvVars,
+		Logger:     options.Logger,
+	}
+
+	return shell.RunCommandAndGetStdOutE(t, cmd)
+}
--- a/modules/docker/docker_compose.go
+++ b/modules/docker/docker_compose.go
@@ -38,7 +38,7 @@ func RunDockerComposeE(t testing.TestingT, options *Options, args ...string) (st
 	return shell.RunCommandAndGetOutputE(t, cmd)
 }
 
-// RunDockerComposeE runs docker-compose with the given arguments and options and return only stdout.
+// RunDockerComposeStdoutE runs docker-compose with the given arguments and options and return only stdout.
 func RunDockerComposeStdOutE(t testing.TestingT, options *Options, args ...string) (string, error) {
 	cmd := shell.Command{
 		Command: "docker-compose",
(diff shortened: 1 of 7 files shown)


modules/docker/docker_compose.go
```diff
@@ -37,3 +37,18 @@ func RunDockerComposeE(t testing.TestingT, options *Options, args ...string) (st
 
 	return shell.RunCommandAndGetOutputE(t, cmd)
 }
+
+// RunDockerComposeE runs docker-compose with the given arguments and options and return only stdout.
```
yorinasub17, line 41:
```suggestion
// RunDockerComposeStdoutE runs docker-compose with the given arguments and options and return only stdout.
```

modules/docker/docker_compose.go
```diff
@@ -37,3 +37,18 @@ func RunDockerComposeE(t testing.TestingT, options *Options, args ...string) (st
...
+func RunDockerComposeStdOutE(t testing.TestingT, options *Options, args ...string) (string, error) {
+	cmd := shell.Command{
+		Command: "docker-compose",
+		// We append --project-name to ensure containers from multiple different tests using Docker Compose don't end
+		// up in the same project and end up conflicting with each other.
+		Args:       append([]string{"--project-name", t.Name()}, args...),
+		WorkingDir: options.WorkingDir,
+		Env:        options.EnvVars,
+		Logger:     options.Logger,
+	}
```
yorinasub17, line 51: Is there someway to DRY this command object between the two functions?

yorinasub17 (commented):
Had some comments for improving the implementation slightly.

The one other thing to add is a unit or integration test for this function. You should be able to reuse the test for `RunDockerCompose` for this purpose (https://github.com/gruntwork-io/terratest/blob/master/test/packer_docker_example_test.go)

modules/docker/docker_compose.go
rhoboat: Attempted in https://github.com/gruntwork-io/terratest/pull/665/commits/d7b4366a6aa82d136bac91992928bc46e2d7f0ce

yorinasub17: Nice!

modules/docker/docker_compose.go
```diff
@@ -16,15 +16,20 @@ type Options struct {
...
 }
 
+// RunDockerComposeStdoutE runs docker-compose with the given arguments and options and return only stdout.
+func RunDockerComposeStdOutE(t testing.TestingT, options *Options, args ...string) (string, error) {
+	return RunDockerComposeE(t, true, options, args...)
+}
+
 // RunDockerComposeE runs docker-compose with the given arguments and options and return stdout/stderr.
-func RunDockerComposeE(t testing.TestingT, options *Options, args ...string) (string, error) {
+func RunDockerComposeE(t testing.TestingT, stdout bool, options *Options, args ...string) (string, error) {
```
yorinasub17, line 32: I think this should be an internal helper function that `RunDockerComposeE` calls under the hood, for backwards compatibility and consistency with the `RunDockerCompose` function.

rhoboat: Like this? https://github.com/gruntwork-io/terratest/pull/665/commits/4fe3a995298e5d284de0f48b239b5a78791f831a

rhoboat:
@yorinasub17 This test is failing, added in [this commit](https://github.com/gruntwork-io/terratest/pull/665/commits/7604e4f362fef3c34ad0bd6ef3e31727a4d31e4d), because the stdout actually includes a lot more stuff than I expect. The `stdout` should only contain `"stdout: message"`, but it contains all of this:

```
Error:      	"Attaching to testdockercomposestdoutexample_bash_script_1
        	bash_script_1  | stderr: error
        	bash_script_1  | stdout: message
        	testdockercomposestdoutexample_bash_script_1 exited with code 0
        	" should not contain "stderr: error"
```

modules/docker/docker_compose.go
yorinasub17: Yup!

modules/docker/docker_compose.go
```diff
@@ -16,15 +16,25 @@ type Options struct {
...
+	out, err := runDockerComposeE(t, false, options, args...)
 	if err != nil {
 		t.Fatal(err)
 	}
 	return out
 }
 
+// RunDockerComposeAndGetStdout runs docker-compose with the given arguments and options and returns only stdout.
+func RunDockerComposeAndGetStdOut(t testing.TestingT, options *Options, args ...string) string {
+	out, _ := runDockerComposeE(t, true, options, args...)
```
brikis98, line 28: Don't ignore errors, as it will making debugging very hard. 

```suggestion
	out, err := runDockerComposeE(t, true, options, args...)
	require.NoError(t, err)
```

test/docker_stdout_example_test.go
```diff
@@ -0,0 +1,33 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/docker"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDockerComposeStdoutExample(t *testing.T) {
```
brikis98, line 10: `t.Parallel()`

test/docker_stdout_example_test.go
```diff
@@ -0,0 +1,33 @@
...
+	dockerComposeFile := "../examples/docker-compose-stdout-example/docker-compose.yml"
+
+	// Build the Docker image.
```
brikis98, line 13: NIT: might want to mention that we do this as a separate step as it writes lots of stuff to `stdout` that we don't want to pollute the text we try to read in the next step.

test/docker_stdout_example_test.go
```diff
@@ -0,0 +1,33 @@
...
+	docker.RunDockerCompose(
+		t,
+		&docker.Options{},
+		"-f",
+		dockerComposeFile,
+		"build",
+	)
+
+	// Run the Docker image, read the text file from it, and make sure it contains the expected output.
```
brikis98, line 22:
```suggestion
	// Run the Docker image, read the stdout and stderr from it, and make sure it contains the expected output.
```

yorinasub17:
> The stdout should only contain "stdout: message", but it contains all of this:

Ahhhh this might not be possible then... Frustratingly, I can't find the docs on where they output the logs to. Let me try a few things to see if I can empirically validate.

yorinasub17:
Ok yea both `docker-compose` and `docker logs` streams all the log output to `STDOUT`, including `STDERR`! This is especially frustrating considering that the logs are actually tracking the source (https://docs.docker.com/config/containers/logging/json-file/).

So it looks like this approach and function is not going to work 😞

yorinasub17:
We probably need to add a command that extracts the stdout and stderr logs using the raw log files. I found the command that gives you the path to the raw log file which has the metadata on where it came from

```
docker inspect --format='{{.LogPath}}' DOCKER_IMAGE_ID
```

But... this is only accessible from the VM on mac osx so you can't actually see it directly and we'd need to yak shave a way to get into the VM...

rhoboat:
@yorinasub17 However, when I source this change here locally in the TLS script, it's able to pull just the right info that's echoed to stdout, meaning the ARN. I was trying to figure out what the difference would be. Any ideas why it would work when used in service catalog but not here?

rhoboat:
Ah, I solved the problem. If I call the bash_script with `docker-compose run bash_script` rather than `up`, it works.

rhoboat:
🎉  Thanks @brikis98 and @yorinasub17! Ready for (final?) review.

modules/docker/docker_compose.go
```diff
@@ -16,15 +17,26 @@ type Options struct {
...
 	}
 	return out
 }
 
+// RunDockerComposeAndGetStdout runs docker-compose with the given arguments and options and returns only stdout.
+func RunDockerComposeAndGetStdOut(t testing.TestingT, options *Options, args ...string) string {
+	out, err := runDockerComposeE(t, true, options, args...)
+	require.NoError(t, err)
+	return out
+}
```
yorinasub17, line 32: Just realized that we should probably expose a `RunDockerComposeAndGetStdOutE` function as well.

yorinasub17 (approved):
Updates LGTM! Found one more nit where we should expose another function, but we can punt that to another PR to avoid another test cycle.

rhoboat:
Thanks for the review! Failures are not related to this change, so merging.

yorinasub17:
Don't forget to cut a new release!

rhoboat:
Okay! Thanks for that. I was waiting for feedback on this! Will go ahead now.

rhoboat:
Ah, before I do, did you see my comment in Slack? I wanted to make sure we were safe to do it, i.e. passing tests on the main branch. There were a few small failures in this branch but I merged because they were unrelated to this change. That's a bit of risky logic, though, and not how I'd normally like to do things. So my idea was to wait for green on `master` before cutting a release.

PR #669, opened 2020-10-06
Document Terragrunt as test dependency

### LINKED ISSUE #668: Documentation: terragrunt not listed as dependency for tests
The `terragrunt` binary is needed to run the tests for `TgApplyAll` and similar functions, but isn't listed as a dependency in the Contributor's guide. For faster setup, it would be good to have it listed.

Incredibly minor, but I wanted to raise a PR as it cost me some time to find this when I was getting set up. Raising an issue so I can link the PR to it: sorry for the extra few clicks.

OwenTuz:
Super minor! Fixes #668


brikis98 (approved):
Ah, good point. Thanks for the fix!

PR #671, opened 2020-10-06
Use DOCKER_HOST to find the correct address for exposed container ports

### LINKED ISSUE #670: Docker tests fail when Docker is not running on localhost
`TestStop` in the Docker module fails on my Mac as Docker runs on a VM (via docker-machine) and cannot be reached on localhost:
https://github.com/gruntwork-io/terratest/blob/0c883d74782fecaf8e5987a760f5aa9e06a04a7f/modules/docker/stop_test.go#L23

This is minor and fairly easy to work around by using `docker-machine ssh -L 13030:localhost:13030` before running tests, but in the interests of portability it would be nice to detect when the docker host is not localhost.

Probably the easiest approach is to get the IP address from the `DOCKER_HOST` environment variable (documented [here](https://docs.docker.com/engine/reference/commandline/cli/#environment-variables)) if it exists, and assume that we're talking to `localhost` if it doesn't.

I'll raise a PR to try and fix this.

OwenTuz:
Fixes #670 

This has only been tested on my Mac but I believe the test should now pass when using `docker-machine` (or other remote Docker hosts) as well as native Docker on both Linux and Windows.

### DIFF
--- a/modules/docker/host.go
+++ b/modules/docker/host.go
@@ -19,18 +19,14 @@ func getDockerHostFromEnv(env []string) string {
 
 	for _, item := range env {
 		envVar := strings.Split(item, "=")
-		if envVar[0] == "DOCKER_HOST" && len(envVar) == 2 {
+		if len(envVar) == 2 && envVar[0] == "DOCKER_HOST" {
 			dockerUrl = strings.Split(envVar[1], ":")
-			if len(dockerUrl) < 2 {
-				// DOCKER_HOST did not split, is empty or ended with ":"
-				return "localhost"
-			}
 			break
 		}
 	}
 
-	if len(dockerUrl) == 0 {
-		// no DOCKER_HOST variable, return default value
+	if len(dockerUrl) < 2 {
+		// DOCKER_HOST was empty, not present or not a valid URL
 		return "localhost"
 	}
 
--- a/modules/docker/host.go
+++ b/modules/docker/host.go
@@ -19,8 +19,12 @@ func getDockerHostFromEnv(env []string) string {
 
 	for _, item := range env {
 		envVar := strings.Split(item, "=")
-		if envVar[0] == "DOCKER_HOST" {
+		if envVar[0] == "DOCKER_HOST" && len(envVar) == 2 {
 			dockerUrl = strings.Split(envVar[1], ":")
+			if len(dockerUrl) < 2 {
+				// DOCKER_HOST did not split, is empty or ended with ":"
+				return "localhost"
+			}
 			break
 		}
 	}
@@ -34,7 +38,7 @@ func getDockerHostFromEnv(env []string) string {
 	case "tcp", "ssh", "fd":
 		return strings.TrimPrefix(dockerUrl[1], "//")
 	default:
-		// if DOCKER_HOST is empty, or not in one of the formats listed above
+		// if DOCKER_HOST is not in one of the formats listed above, return default
 		return "localhost"
 	}
 }
--- a/modules/docker/host.go
+++ b/modules/docker/host.go
@@ -1,0 +1,22 @@
+package docker
+
+import (
+	"os"
+	"strings"
+)
+
+// GetDockerHost returns the name or address of the host on which the Docker engine is running.
+func GetDockerHost() string {
+	// Parses the DOCKER_HOST environment variable to find the address
+	//
+	// For valid formats see:
+	// https://github.com/docker/cli/blob/6916b427a0b07e8581d121967633235ced6db9a1/opts/hosts.go#L69
+	dockerUrl := strings.Split(os.Getenv("DOCKER_HOST"), ":")
+	switch dockerUrl[0] {
+	case "tcp", "ssh", "fd":
+		return strings.TrimPrefix(dockerUrl[1], "//")
+	default:
+		// if DOCKER_HOST is empty, or not in one of the formats listed above
+		return "localhost"
+	}
+}
--- a/modules/docker/host.go
+++ b/modules/docker/host.go
@@ -7,11 +7,24 @@ import (
 
 // GetDockerHost returns the name or address of the host on which the Docker engine is running.
 func GetDockerHost() string {
+	return getDockerHostFromEnv(os.Environ())
+}
+
+func getDockerHostFromEnv(env []string) string {
 	// Parses the DOCKER_HOST environment variable to find the address
 	//
 	// For valid formats see:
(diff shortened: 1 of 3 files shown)


brikis98 (commented):
Ah, this is a good idea, thanks!

I think it makes sense to take the code you wrote and expose it in a method called something like `docker.GetHost`, as I'm sure we'll want to use this functionality in the future! You can then use that new method in this test, and all other Docker-related tests in the future.

OwenTuz:
Makes sense to me, I'll refactor and push that change.

modules/docker/host_test.go
```diff
@@ -0,0 +1,66 @@
...
+		{
+			"invalid::value::with::semicolons",
+			"localhost",
+		},
+	}
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("GetDockerHost: %s", test.Input), func(t *testing.T) {
+			// GetHost() uses the DOCKER_HOST environment variable, so we need to set
+			// it to our test input and then reset it afterwards for other tests
+			defer os.Setenv("DOCKER_HOST", os.Getenv("DOCKER_HOST"))
```
OwenTuz, line 56: perhaps it's worth clarifying that setting and resetting the environment variable means we can't safely use `t.Parallel()` here?

I believe it's safe to do this provided we don't call `Parallel()` but somebody with more experience here may well be able to correct me on that.

brikis98: Alternative idea I often use in this case:

1. Create a `getDockerHostFromEnv(env []string) string` method that takes in a list of environment variables (each one in the `key=value` format).
1. Update `GetDockerHost` to call `getDockerHost(os.Environ())` so it's using the real env vars.
1. Update the automated tests to pass in mock values.

modules/docker/host_test.go
```diff
@@ -0,0 +1,66 @@
...
+			os.Setenv("DOCKER_HOST", test.Input)
+
+			host := GetDockerHost()
+
+			if host != test.Expected {
+				t.Fatalf("Error: expected %s, got %s", test.Expected, host)
+			}
```
brikis98, line 63: NIT: you can use `assert.Equal(t, test.Expected, host)`

modules/docker/stop_test.go
```diff
@@ -20,7 +20,7 @@ func TestStop(t *testing.T) {
 
 	// choosing a unique port since 80 may not fly well on test machines
 	port := "13030"
-	testURL := "http://localhost:" + port
+	testURL := "http://" + GetDockerHost() + ":" + port
```
brikis98, line 23: NIT: I know the code originally was written with concat, but it's more idiomatic in Go to use:

```suggestion
	testURL := fmt.Sprintf("http://%s:%d", GetDockerHost(), port)
```

OwenTuz:
Thanks for the comments - I've incorporated these and also added some handling for the corner case where `DOCKER_HOST` is not just empty but unset. If we're in that state then something else is probably broken, but we can at least handle our end safely enough by returning a default value of "localhost".

I've also made a complete mess with my merge commits, sorry: left the commits for now to show my working but happy to squash this to a single commit and re-push if you would like.

modules/docker/host.go
```diff
@@ -0,0 +1,40 @@
...
+func getDockerHostFromEnv(env []string) string {
+	// Parses the DOCKER_HOST environment variable to find the address
+	//
+	// For valid formats see:
+	// https://github.com/docker/cli/blob/6916b427a0b07e8581d121967633235ced6db9a1/opts/hosts.go#L69
+	var dockerUrl []string
+
+	for _, item := range env {
+		envVar := strings.Split(item, "=")
+		if envVar[0] == "DOCKER_HOST" {
```
brikis98, line 22: Need to check that `len(envVar) == 2` as well.

modules/docker/host.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	if len(dockerUrl) == 0 {
+		// no DOCKER_HOST variable, return default value
+		return "localhost"
+	}
+
+	switch dockerUrl[0] {
+	case "tcp", "ssh", "fd":
+		return strings.TrimPrefix(dockerUrl[1], "//")
```
brikis98, line 35: Need to check `len(dockerUrl) > 1`

modules/docker/host_test.go
```diff
@@ -0,0 +1,76 @@
+package docker
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetDockerHostFromEnv(t *testing.T) {
```
brikis98, line 10: Can add `t.Parallel()` here and in all the `t.Run` below.

OwenTuz:
D'oh. Thanks for your patience with this one! I've made those changes, and got rid of the merge commits - this could still use being squashed on merge, but to hopefully make review a little easier.

modules/docker/host.go
```diff
@@ -0,0 +1,44 @@
...
+func getDockerHostFromEnv(env []string) string {
+	// Parses the DOCKER_HOST environment variable to find the address
+	//
+	// For valid formats see:
+	// https://github.com/docker/cli/blob/6916b427a0b07e8581d121967633235ced6db9a1/opts/hosts.go#L69
+	var dockerUrl []string
+
+	for _, item := range env {
+		envVar := strings.Split(item, "=")
+		if envVar[0] == "DOCKER_HOST" && len(envVar) == 2 {
```
brikis98, line 22: NIT: I happen to know `strings.Split` returns a non-empty array, but since we're checking length anyway, it's safer to do that length check first.

```suggestion
		if len(envVar) == 2 && envVar[0] == "DOCKER_HOST" {
```

modules/docker/host.go
brikis98: I know you're checking that `len(dockerUrl)` above is not 0, but that still doesn't mean there are at least two items here, so we need that check somewhere...

modules/docker/host_test.go
brikis98: Still need a `t.Parallel()` here, at the top of the main test method.

modules/docker/host.go
OwenTuz: This is checked on line 24 if/when we apply `strings.Split` - but could be clearer, I'll refactor it to a single check. Thanks!

brikis98:
OK, great, thank you! I'll kick off tests now.

brikis98 (approved):
There were some intermittent test failures unrelated to this PR, but on re-run, they passed, so merging now!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.30.11

PR #672, opened 2020-10-06
Add GetEcsClusterWithInclude

### LINKED ISSUE #643: Include Tags in GetEcsCluster
Is there a specific reason why tags are not returned in the `GetEcsCluster` method?

So far as I can see it would only require a little change to the SDK input:
https://github.com/gruntwork-io/terratest/blob/c5ca9ef8e4e66f723e40c8b4cab06f82310618cd/modules/aws/ecs.go#L25-L29

Adding the `TAGS` value to the `Include` field should do the trick.

ref: https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#DescribeClustersInput

particleflux:
Add a new method for getting ECS cluster info, including additional
optional information. This can be used, for example, to fetch tags along
with the cluster info. Therefore the tags property will no longer be
empty, and can be tested against as well.

Fixes #643

I tried to avoid code duplication by moving the main logic for invoking the API to the new method. As far as I see, this should be fully compatible.

Added a new test case for the new method as well.

### DIFF
--- a/modules/aws/ecs.go
+++ b/modules/aws/ecs.go
@@ -18,6 +18,20 @@ func GetEcsCluster(t testing.TestingT, region string, name string) *ecs.Cluster 
 
 // GetEcsClusterE fetches information about specified ECS cluster.
 func GetEcsClusterE(t testing.TestingT, region string, name string) (*ecs.Cluster, error) {
+	return GetEcsClusterWithIncludeE(t, region, name, []string{})
+}
+
+// GetEcsClusterWithInclude fetches extended information about specified ECS cluster.
+// The `include` parameter specifies a list of `ecs.ClusterField*` constants, such as `ecs.ClusterFieldTags`.
+func GetEcsClusterWithInclude(t testing.TestingT, region string, name string, include []string) *ecs.Cluster {
+	clusterInfo, err := GetEcsClusterWithIncludeE(t, region, name, include)
+	require.NoError(t, err)
+	return clusterInfo
+}
+
+// GetEcsClusterWithIncludeE fetches extended information about specified ECS cluster.
+// The `include` parameter specifies a list of `ecs.ClusterField*` constants, such as `ecs.ClusterFieldTags`.
+func GetEcsClusterWithIncludeE(t testing.TestingT, region string, name string, include []string) (*ecs.Cluster, error) {
 	client, err := NewEcsClientE(t, region)
 	if err != nil {
 		return nil, err
@@ -26,6 +40,7 @@ func GetEcsClusterE(t testing.TestingT, region string, name string) (*ecs.Cluste
 		Clusters: []*string{
 			aws.String(name),
 		},
+		Include: aws.StringSlice(include),
 	}
 	output, err := client.DescribeClusters(input)
 	if err != nil {
--- a/modules/aws/ecs_test.go
+++ b/modules/aws/ecs_test.go
@@ -3,6 +3,9 @@ package aws
 import (
 	"testing"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -20,4 +23,40 @@ func TestEcsCluster(t *testing.T) {
 
 	assert.Nil(t, err)
 	assert.Equal(t, "terratest", *c2.ClusterName)
+}
+
+func TestEcsClusterWithInclude(t *testing.T) {
+	t.Parallel()
+
+	region := GetRandomStableRegion(t, nil, nil)
+	clusterName := "terratest-" + random.UniqueId()
+	tags := []*ecs.Tag{&ecs.Tag{
+		Key:   aws.String("test-tag"),
+		Value: aws.String("hello-world"),
+	}}
+
+	client := NewEcsClient(t, region)
+	c1, err := client.CreateCluster(&ecs.CreateClusterInput{
+		ClusterName: aws.String(clusterName),
+		Tags:        tags,
+	})
+	assert.NoError(t, err)
+
+	defer DeleteEcsCluster(t, region, c1.Cluster)
+
+	assert.Equal(t, clusterName, aws.StringValue(c1.Cluster.ClusterName))
+
+	c2, err := GetEcsClusterWithIncludeE(t, region, clusterName, []string{ecs.ClusterFieldTags})
+	assert.NoError(t, err)
+
+	assert.Equal(t, clusterName, aws.StringValue(c2.ClusterName))
+	assert.Equal(t, tags, c2.Tags)
+	assert.Empty(t, c2.Statistics)
+
+	c3, err := GetEcsClusterWithIncludeE(t, region, clusterName, []string{ecs.ClusterFieldStatistics})
+	assert.NoError(t, err)
+
+	assert.Equal(t, clusterName, aws.StringValue(c3.ClusterName))
+	assert.NotEmpty(t, c3.Statistics)
+	assert.Empty(t, c3.Tags)
 }


modules/aws/ecs.go
```diff
@@ -18,6 +18,18 @@ func GetEcsCluster(t testing.TestingT, region string, name string) *ecs.Cluster
...
+}
+
+// GetEcsClusterWithInclude fetches extended information about specified ECS cluster.
+func GetEcsClusterWithInclude(t testing.TestingT, region string, name string, include []string) *ecs.Cluster {
+	clusterInfo, err := GetEcsClusterWithIncludeE(t, region, name, include)
+	require.NoError(t, err)
+	return clusterInfo
+}
+
+// GetEcsClusterWithIncludeE fetches extended information about specified ECS cluster.
```
brikis98, line 31: NIT: Could you provide a bit more detail in this comment about valid values for `include`? Or better yet, is there a set of `const` values we can use?

modules/aws/ecs_test.go
```diff
@@ -21,3 +23,33 @@ func TestEcsCluster(t *testing.T) {
...
 }
+
+func TestEcsClusterWithInclude(t *testing.T) {
+	t.Parallel()
+
+	region := GetRandomStableRegion(t, nil, nil)
+
+	client := NewEcsClient(t, region)
+	c1, err := client.CreateCluster(&ecs.CreateClusterInput{
+		ClusterName: aws.String("terratest-extended"),
```
brikis98, line 34: You need to use `random.UniqueId` in the name. See [namespacing](https://terratest.gruntwork.io/docs/testing-best-practices/namespacing/) for more info.

modules/aws/ecs_test.go
```diff
@@ -21,3 +23,33 @@ func TestEcsCluster(t *testing.T) {
...
+		Tags: []*ecs.Tag{&ecs.Tag{
+			Key:   aws.String("test-tag"),
+			Value: aws.String("hello-world"),
+		}},
+	})
+	assert.Nil(t, err)
+
+	defer DeleteEcsCluster(t, region, c1.Cluster)
+
+	assert.Nil(t, err)
```
brikis98, line 44: You already have this check on line 40.

modules/aws/ecs_test.go
```diff
@@ -21,3 +23,33 @@ func TestEcsCluster(t *testing.T) {
...
+	assert.Equal(t, "terratest-extended", *c1.Cluster.ClusterName)
```
brikis98, line 45: Any time you dereference something that comes back from the AWS SDK, please use methods like `aws.StringValue` to avoid possible `nil` pointer dereferences.

modules/aws/ecs_test.go
```diff
@@ -21,3 +23,33 @@ func TestEcsCluster(t *testing.T) {
...
+
+	c2, err := GetEcsClusterWithIncludeE(t, region, *c1.Cluster.ClusterName, []string{"TAGS"})
```
brikis98, line 47: As mentioned in a comment above, is `"TAGS"` defined in a `const` somewhere in the AWS SDK?

modules/aws/ecs_test.go
```diff
@@ -21,3 +23,33 @@ func TestEcsCluster(t *testing.T) {
...
+
+	assert.Nil(t, err)
+	assert.Equal(t, "terratest-extended", *c2.ClusterName)
```
brikis98, line 50: NIT: store cluster name in a var and use that var when creating the cluster and when comparing here. You'll need to do that anyway when you use `UniqueId`.

modules/aws/ecs_test.go
```diff
@@ -21,3 +23,33 @@ func TestEcsCluster(t *testing.T) {
...
+
+	tag := *c2.Tags[0]
+	assert.Equal(t, "test-tag", *tag.Key)
+	assert.Equal(t, "hello-world", *tag.Value)
```
brikis98, line 54: NIT: store the tags you're creating in a var and check for direct equality here to ensure. That'll ensure there aren't _other_ unexpected tags there. Also, it'll avoid an array out of bounds error in case `c2.Tags` is an empty array.

brikis98 (commented):
Thanks for the PR!

modules/aws/ecs_test.go
particleflux: Ha, I was wondering about that, but as the existing test didn't use it, I opted to not do it as well.

modules/aws/ecs_test.go
particleflux: Good catch :+1:

particleflux:
Pushed an update for the mentioned issues, thanks for the feedback!

Additionally added another test case with a different set of include options. Unfortunately, can't really easily test against the remaining two:

* `ClusterFieldAttachments` seems to be only filled when running stuff inside the cluster, which would be a bit of overkill for this simple test case imho.
* `ClusterFieldSettings` _should_ change the settings field, but that seems to be somehow always there, regardless of the include param.

modules/aws/ecs.go
particleflux: Wasn't sure about the desired format of these doc blocks, and since the others were one-line only I tried to keep it somehow short.

Added a second line describing the new `include` field

modules/aws/ecs_test.go
```diff
@@ -21,3 +24,39 @@ func TestEcsCluster(t *testing.T) {
...
+		Key:   aws.String("test-tag"),
+		Value: aws.String("hello-world"),
+	}}
+
+	client := NewEcsClient(t, region)
+	c1, err := client.CreateCluster(&ecs.CreateClusterInput{
+		ClusterName: aws.String(clusterName),
+		Tags:        tags,
+	})
+	assert.Nil(t, err)
```
brikis98, line 43: NIT: 

```suggestion
	assert.NoError(t, err)
```

This will show the error clearly if it's not `nil`, making debugging easier.

modules/aws/ecs_test.go
```diff
@@ -21,3 +24,39 @@ func TestEcsCluster(t *testing.T) {
...
+
+	defer DeleteEcsCluster(t, region, c1.Cluster)
+
+	assert.Equal(t, clusterName, aws.StringValue(c1.Cluster.ClusterName))
+
+	c2, err := GetEcsClusterWithIncludeE(t, region, clusterName, []string{ecs.ClusterFieldTags})
+	assert.Nil(t, err)
```
brikis98, line 50: Same NIT:

```suggestion
	assert.NoError(t, err)
```

modules/aws/ecs_test.go
```diff
@@ -21,3 +24,39 @@ func TestEcsCluster(t *testing.T) {
...
+
+	assert.Equal(t, clusterName, aws.StringValue(c2.ClusterName))
+	assert.Equal(t, tags, c2.Tags)
+	assert.Empty(t, c2.Statistics)
+
+	c3, err := GetEcsClusterWithIncludeE(t, region, clusterName, []string{ecs.ClusterFieldStatistics})
+	assert.Nil(t, err)
```
brikis98, line 57:
```suggestion
	assert.NoError(t, err)
```

brikis98 (commented):
A few more NITs and I can kick off tests!

particleflux:
Replaced those _asserts_ as well. Note that the other test in that file has the same issues, though was not added with this PR.

brikis98:
Great, thanks! Kicking off tests now.

brikis98:
Hit a couple test failures. I _think_ these actually might be broken on `master`... Looking into it.

brikis98:
Fix in progress in https://github.com/gruntwork-io/terratest/pull/676.

brikis98:
https://github.com/gruntwork-io/terratest/pull/676 is now merged. Could you rebase?

particleflux:
Sure. Done :+1:

brikis98:
Thanks! Kicking off tests again...

brikis98 (approved):
One test failed, but it was due to an unrelated intermittent issue with a Packer build, so I'm going to merge this. Thanks!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.30.8

PR #676, opened 2020-10-12
Remove bash_script reference from packer-docker-example

brikis98:
It looks like #665 added the ability to run `docker-compose` and get `stdout`. As part of this work, a new example in `examples/docker-compose-stdout-example` was added that had a `bash_script.sh` that was used for testing. All this works fine, but it looks like #665 _also_ added a reference to `bash_script.sh` to the `docker-compose.yml` in `examples/packer-docker-example`... But not `bash_script.sh` itself, so it's making tests fail.

This seems like a copy/paste error, or perhaps a leftover of some previous test, so I'm removing this mention.

### DIFF
--- a/examples/packer-docker-example/docker-compose.yml
+++ b/examples/packer-docker-example/docker-compose.yml
@@ -17,11 +17,3 @@ services:
     # Expose the sample app's port on the host OS
     ports:
       - "${SERVER_PORT}:${SERVER_PORT}"
-  bash_script:
-    # The name we use for the Docker image in build.json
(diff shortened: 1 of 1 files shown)


brikis98:
@rhoboat Could you confirm that this was just a copy/paste thing with that example?

brikis98:
Tests passed! I'd be grateful for a review.

yorinasub17 (approved):
LGTM!

brikis98:

Issue #675, opened 2020-10-08
Terratest AWS module for RDS has issue with GetParameterValueForParameterOfRdsInstance function 


smeyer53027:
Bug report:  github.com/gruntwork-io/terratest/modules/aws/rds.go has issue with GetParameterValueForParameterOfRdsInstance function and it's dependent function GetParameterValueForParameterOfRdsInstanceE. The process loads all the parameters from the parameter group for the specified RDS instance and tries to locate the specified parameter in the set, but in fact, it is not loading all parameters.  The dependent AWS API call DescribeDBParameters, by default, only returns the first 100 parameters (in alphabetical order) and if your parameter group has more than 100 parameters and the parameter you are looking for is not in the first 100, the function throw errors,  incorrectly reporting the parameter does not exist.

Recommend the functions support pagination using the Marker parameter to scan all the parameters, 100 at a time.

brikis98:
Thanks for reporting! Would you be up for a PR to fix this?

smeyer53027:
I have modified the related modules to make this work but it's a bit messy.  Being very new to Go I had difficulties with combining the results of the DescribeDBParameters call in the GetAllParametersOfRdsInstanceE function.   I wanted to have this function loop through the pages of the available parameters.  So to get it working I setup the GetParameterValueForParameterOfRdsInstanceE function to also return the Marker token to the calling process and then setup a loop in the GetParameterValueForParameterOfRdsInstanceE function to use the token to reference the next set. I can submit  PR with this fix.....

brikis98:
PR welcome, thank you! Note that we probably have other code in this repo, and code in many other Go repos, that use the Marker token pattern to fetch all results if you need a reference.

PR #683, opened 2020-10-20
Resolve issue #675 in aws/rds.go for GetParameterValueForParameterOfRdsInstance and related functions to scan all parameters

smeyer53027:
Change Detail:
- GetParameterValueForParameterOfRdsInstanceE uses Marker token from DescribeDBParameters API call to paginate through all possible parameters. Uses a flag (morepages) to monitor when there are no more pages to receive from the parameter group.Loops through all available pages of parameters received from Parameter Group. If the parameter is found in the current page, the function returns the parameter details then and ends 

- GetAllParametersOfRdsInstance renamed to GetParametersOfRdsInstance:  Modified to receive additional parameter for the Marker and to return the new Marker and an integer of the record count of the parameters returned from GetParametersOfRdsInstanceE .

- GetAllParametersOfRdsInstanceE renamed GetParametersOfRdsInstanceE: Has a new parameter to receive Marker to specify the starting point for the DescribeDBParameters call.  Extracts the value of the Marker returned from DescribeDBParameters and the number of parameters retrieved.  Returns values to calling process.

### DIFF
--- a/modules/aws/rds.go
+++ b/modules/aws/rds.go
@@ -121,6 +121,58 @@ func GetParameterValueForParameterOfRdsInstanceE(t testing.TestingT, parameterNa
 	return "", aws.ParameterForDbInstanceNotFound{ParameterName: parameterName, DbInstanceID: dbInstanceID, AwsRegion: awsRegion}
 }
 
+//------------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------------
+// GetAllParametersOfRdsInstance gets all the parameters defined in the parameter group for the RDS instance in the given region.
+// Modified to accept a Marker Token to pass on to GetAllParametersOfRdsInstanceE and return the ending marker and the item count in parameters
+func GetAllParametersOfRdsInstance(t testing.TestingT, dbInstanceID string, awsRegion string, startMarker string) ([]*rds.Parameter, string, int) {
+	parameters, endmarker, cnt, err := GetAllParametersOfRdsInstanceE(t, dbInstanceID, awsRegion, startMarker)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return parameters, endmarker, cnt
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------------
+// GetAllParametersOfRdsInstanceE gets all the parameters defined in the parameter group for the RDS instance in the given region.
+// Modified to accept a Marker Token to pass on to DescribeDBParameters API call. Returns rds parameters, the ending marker from search, and count of records pulled
+func GetAllParametersOfRdsInstanceE(t testing.TestingT, dbInstanceID string, awsRegion string, strtmarker string) ([]*rds.Parameter, string, int, error) {
+	//  Get RDS instance details which include Parameter Group Name
+	var lastmarker string = ""
+
+	//  ***Get Instance Details
+	dbInstance, dbInstanceErr := GetRdsInstanceDetailsE(t, dbInstanceID, awsRegion)
+	if dbInstanceErr != nil {
+		return []*rds.Parameter{}, "", 0, dbInstanceErr
+	}
+	// Extract parameter group name for defined instance
+	parameterGroupName := aws.StringValue(dbInstance.DBParameterGroups[0].DBParameterGroupName)
+	
+	// create RDSclient interface
+	rdsClient := aws.NewRdsClient(t, awsRegion)
+	
+	// set the input parameters for DescribeDBParameters api call.  Marker will be nil first pass
+	input := rds.DescribeDBParametersInput{
+		DBParameterGroupName: aws.String(parameterGroupName),
+		Marker: aws.String(strtmarker),
+		MaxRecords: aws.Int64(100),
+	}
+
+	//   *** call DescribeDBParameters api to retreive parmaters
+	p_output, err := rdsClient.DescribeDBParameters(&input)
+	if err != nil {
+		return []*rds.Parameter{}, "", 0, err
+	}
+	numofparms := len(p_output.Parameters)
+
+	//  check if marker was returned,  if not, leave lastmarker as nil as declared above
+	if p_output.Marker != nil {
+		lastmarker = *p_output.Marker
+	}
+	return p_output.Parameters, lastmarker, numofparms, nil
+}
+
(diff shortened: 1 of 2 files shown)


modules/aws/rds.go
```diff
@@ -86,15 +86,91 @@ func GetParameterValueForParameterOfRdsInstance(t testing.TestingT, parameterNam
 	return parameterValue
 }
 
+//------------------------------------------------------------------------------------------------------------------------------------------------
 // GetParameterValueForParameterOfRdsInstanceE gets the value of the parameter name specified for the RDS instance in the given region.
+// modified to use previos token and page through all parameters
```
yorinasub17, line 91: NIT:

```suggestion
// GetParameterValueForParameterOfRdsInstanceE gets the value of the parameter name specified for the RDS instance in the given region,
// accounting for pagination.
```

modules/aws/rds.go
```diff
@@ -86,15 +86,91 @@ func GetParameterValueForParameterOfRdsInstance(t testing.TestingT, parameterNam
...
 func GetParameterValueForParameterOfRdsInstanceE(t testing.TestingT, parameterName string, dbInstanceID string, awsRegion string) (string, error) {
-	output := GetAllParametersOfRdsInstance(t, dbInstanceID, awsRegion)
-	for _, parameter := range output {
-		if aws.StringValue(parameter.ParameterName) == parameterName {
-			return aws.StringValue(parameter.ParameterValue), nil
+	lastmarker := ""			// var to store the returned Marker value
+	morepages := true 			// initialize morepages to indicate more pages need to be pulled
+	pagecnt := 1				// initialize pagecnt to be used to limit the number if loops in the event the parameter cannot be found
```
yorinasub17, line 95:
```suggestion
	pagecnt := 1				// initialize pagecnt to be used to limit the number of loops in the event the parameter cannot be found
```

modules/aws/rds.go
```diff
@@ -86,15 +86,91 @@ func GetParameterValueForParameterOfRdsInstance(t testing.TestingT, parameterNam
...
+		t.Fatal(err)
+	}
+	return parameters, endmarker, cnt
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------------
+// GetParametersOfRdsInstanceE gets all the parameters defined in the parameter group for the RDS instance in the given region.
+// Modified to accept a Marker Token to pass on to DescribeDBParameters API call. Returns rds parameters, the ending marker from search, and count of records pulled
+func GetParametersOfRdsInstanceE(t testing.TestingT, dbInstanceID string, awsRegion string, strtmarker string) ([]*rds.Parameter, string, int, error) {
```
yorinasub17, line 140: I don't think it is very ergonomical to replace and push the pagination details up from this function, especially when the expected test use case would be for RDS instances with very few parameters.

It probably makes sense to create two functions, one that returns all objects (the original function but modified to collect all the pages), and another function that returns partial results, which is this one.

That said, I would probably tweak the function API to use the paginator pattern, where a function is called on each page. The function signature would be something like:

```
func GetParametersOfRdsInstancePaginatorE(t testing.TestingT, dbInstanceID string, awsRegion string, func([]* rds.Parameter) error) error
```

modules/aws/rds.go
```diff
@@ -86,15 +86,91 @@ func GetParameterValueForParameterOfRdsInstance(t testing.TestingT, parameterNam
...
+	pagecnt := 1				// initialize pagecnt to be used to limit the number if loops in the event the parameter cannot be found
+	for morepages == true {
+		//  *** on the first pass, the lastmarker is nil, if additional records exist, the api call will return a marker. cnt holds the number of items in output
+		output, marker, cnt := GetParametersOfRdsInstance(t, dbInstanceID, awsRegion, lastmarker)
+		// store the new marker to a local var
+		lastmarker = marker
+		//  check if the number of records returned are less than 100 or if a marker was not returned to have function stop on this current page
+		if (cnt < 100) || (marker == "") {
+			morepages = false
 		}
```
yorinasub17, line 104: When `cnt` is less than 100, is `marker` set? If not, you can probably simplify this to just `if marker == ""`.

modules/aws/rds.go
```diff
@@ -86,15 +86,91 @@ func GetParameterValueForParameterOfRdsInstance(t testing.TestingT, parameterNam
...
+		
+		// examine the parameters returned for any matches to the specified
+		for _, parameter := range output {
+			//  ### debug print each of the parameters to match to the parameterName 
+			// fmt.Println(aws.StringValue(parameter.ParameterName))
```
yorinasub17, line 109: NIT: dead code should be removed instead of commented out.

modules/aws/rds.go
```diff
@@ -86,15 +86,91 @@ func GetParameterValueForParameterOfRdsInstance(t testing.TestingT, parameterNam
...
+		//  *** on the first pass, the lastmarker is nil, if additional records exist, the api call will return a marker. cnt holds the number of items in output
+		output, marker, cnt := GetParametersOfRdsInstance(t, dbInstanceID, awsRegion, lastmarker)
+		// store the new marker to a local var
+		lastmarker = marker
+		//  check if the number of records returned are less than 100 or if a marker was not returned to have function stop on this current page
+		if (cnt < 100) || (marker == "") {
+			morepages = false
 		}
+		
+		// examine the parameters returned for any matches to the specified
```
yorinasub17, line 106: NIT: looks like this comment is missing a word at the end.

modules/aws/rds.go
```diff
@@ -86,15 +86,91 @@ func GetParameterValueForParameterOfRdsInstance(t testing.TestingT, parameterNam
...
+			//  ### debug print each of the parameters to match to the parameterName 
+			// fmt.Println(aws.StringValue(parameter.ParameterName))
+			if aws.StringValue(parameter.ParameterName) == parameterName {
+				return aws.StringValue(parameter.ParameterValue), nil
+			}
+		}
+		
+		//  the above code should prevent paging beyond the actual records but just in case, included a stop
+		pagecnt++
+		if pagecnt > 6 {
```
yorinasub17, line 117: Why 6 pages maximum? This seems a bit arbitrary.

yorinasub17 (commented):
Thanks so much for your contribution, and apologies for the long delay in reviewing this! I finally found the time to start going through the terratest backlog again and realized we had missed this PR...

The overall problem of pagination definitely needs to be solved and the code looks fine here for the most part, but I am concerned about the change in ergonomics of the existing API. See comments for more details and suggested updates.

james00012:

PR #687, opened 2020-10-23
Add support for SSM Parameter deletion

jdstuart:
Added support to delete and SSM parameter.


jdstuart:
@brikis98 I added support to delete a SSM parameter. Can you please review?

jdstuart:
@yorinasub17 Can this be merged?

modules/aws/ssm_test.go
```diff
@@ -21,5 +21,21 @@ func TestParameterIsFound(t *testing.T) {
 	keyValue := GetParameter(t, awsRegion, expectedName)
 	logger.Logf(t, "Found key with name %s", expectedName)
 	assert.Equal(t, expectedValue, keyValue)
+}
+
+func TestParameterIsDeleted(t *testing.T) {
+	expectedName := fmt.Sprintf("test-name-%s", random.UniqueId())
+	// awsRegion := GetRandomRegion(t, nil, nil)
+	awsRegion := "ca-central-1"
```
yorinasub17, line 29: Replace with commented code

yorinasub17 (commented):
Apologies for the long delay in responding! This PR got lost in the sea of other PRs and we lost track of it.

The implementation looks good to me, but we need to make a case on whether we should include this or not. On the one hand, this contribution fails our "complexity" and "checking infrastructure" [litmus tests](https://terratest.gruntwork.io/docs/community/contributing/#types-of-contributions), which makes it difficult to make the case to include. With that said, there is an argument to be made about "completeness" as you would typically want to temporarily be setting and destroying SSM parameters for the test, so I think we are ok to merge this in.

Could definitely use a second opinion though (@brikis98 what do you think?). In the meantime, I kicked off a build!

As a side note that typically, we recommend doing this kind of dependency injection using Terraform, which is a much more robust platform (handling retry logic and waiting) for creating and managing infrastructure.

jdstuart:
> The implementation looks good to me, but we need to make a case on whether we should include this or not... With that said, there is an argument to be made about "completeness" as you would typically want to temporarily be setting and destroying SSM parameters for the test, so I think we are ok to merge this in.
That’s exactly out use case. We have modules that require SSM to be present. We use this library to setup those parameters and feel that we’d prefer to use the same library to also tear down those same parameters.

> As a side note that typically, we recommend doing this kind of dependency injection using Terraform. 
Which kind of DI are you referring to? Can you please elaborate or point me in the right direction?

Thanks,

modules/aws/ssm_test.go
jdstuart: Removed the region specific code and replaced it with `GetRandomRegion`

yorinasub17:
> Which kind of DI are you referring to? Can you please elaborate or point me in the right direction?

The idea is to create wrapper modules that call the modules you are testing, and then use terraform to setup the dependencies of the underlying module. Then, you use terratest against the wrapper modules, which implicitly test the underlying modules.

This way, you can use terraform to setup all the resource dependencies that you need, giving you a much more robust testing platform.

An alternative approach that we also sometimes take is to create a separate module that sets up the dependencies, and use `terratest` to call them in sequence, relying on `terraform.Output` to pass through the data. This lets you more directly test the target module when the input variables are complex and hard to duplicate in a wrapper.

jdstuart:
@yorinasub17 when you say wrapper module, I think you follow a similar pattern with your `examples` and `modules` in that your examples spin up an implementation of the `module` and your infrastructure test then actually init and apply the example (wrapper module) instead of the module itself, correct? Is that what you're referring to?

jdstuart:
Nevertheless. Is this PR something you think that will be merged, or should I change the approach on my tests? (I prefer referencing terratest directly and not a forked version of it)

yorinasub17:
> when you say wrapper module, I think you follow a similar pattern with your examples and modules in that your examples spin up an implementation of the module and your infrastructure test then actually init and apply the example (wrapper module) instead of the module itself, correct? Is that what you're referring to?

Yup that is correct.

> Is this PR something you think that will be merged, or should I change the approach on my tests?

I think we can merge this in, based on the completeness litmus test. I just reviewed the test failure and it is a known issue unrelated to this change so will be merging and releasing shortly.

yorinasub17 (approved):

infraredgirl (approved):

PR #689, opened 2020-10-28
Adding getLogging functionality to s3 buckets

ina-stoyanova:
Adding `GetBucketLogging` function for the use of such functionality in this PR: https://github.com/gruntwork-io/terragrunt/pull/1400


ina-stoyanova:
adding a test for this now...

ina-stoyanova:
This is ready for review now. 

When this is merged & released, it will unblock us to unskip the `TestTerragruntSetsAccessLoggingForTfSTateS3BuckeToADifferentBucket` test in `terragrunt`

You can find the relevant commit with the skipped test [here](https://github.com/gruntwork-io/terragrunt/commit/426cb9462f0a1177f07cbccb9f57e41bbb91bbf4#diff-8118e44ac15fb5ab2e2ece2ab52fd23062a10b5cc9c55ee5b3620a1242cff27cR601)

modules/aws/s3.go
```diff
@@ -260,6 +263,81 @@ func EmptyS3BucketE(t testing.TestingT, region string, name string) error {
...
+			LoggingEnabled: &s3.LoggingEnabled{
+				TargetBucket: aws.String(logsTargetBucketName),
+				TargetPrefix: aws.String("/"),
+			},
+		},
+	}
+
+	_, err = s3Client.PutBucketLogging(params)
+	return err
+}
```
yorinasub17, line 310: Can you share what use case prompted you to add the put version? The reason I ask is that this technically doesn't pass [the litmus test](https://terratest.gruntwork.io/docs/community/contributing/#infrastructure-and-validation-helpers) for contributions because it is an API for modifications, something that isn't really necessary in a testing library.

modules/aws/s3.go
```diff
@@ -368,6 +446,29 @@ func AssertS3BucketPolicyExistsE(t testing.TestingT, region string, bucketName s
...
+
+	if targetBucket == "" {
+		return err
+	}
+
+	if targetBucket == logsTargetBucketName {
+		return nil
+	}
+	return err
+}
```
yorinasub17, line 470: This will return `nil` because `err` is `nil` at this point (it failes the first if condition, which checks for non-nil). 

That said, this is a relatively simple function on top of `GetS3BucketLoggingTarget`, so I think it fails the complexity litmus test for inclusion in terratest.

Instead, you can implement this inline in your test as:

```
targetBucket := aws.GetS3BucketLoggingTarget(...)
assert.Equal(t, targetBucket, logsTargetBucketName)
```

modules/aws/s3.go
```diff
@@ -260,6 +263,81 @@ func EmptyS3BucketE(t testing.TestingT, region string, name string) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	if res.LoggingEnabled != nil {
+		return aws.StringValue(res.LoggingEnabled.TargetBucket), nil
+	}
+
+	return "", err
+}
```
yorinasub17, line 339: I'm a bit on the fence about this one, given that it's a thin wrapper around the API, but since the s3 based functionality isn't really complete ([completeness litmus test](https://github.com/gruntwork-io/terratest/issues/421#issuecomment-567350190)) without looking for access logging, I think it is ok to include this getter function.

modules/aws/s3_test.go
```diff
@@ -29,6 +29,24 @@ func TestCreateAndDestroyS3Bucket(t *testing.T) {
...
+	region := GetRandomStableRegion(t, nil, nil)
+	id := random.UniqueId()
+	logger.Logf(t, "Random values selected. Region = %s, Id = %s\n", region, id)
+
+	s3BucketName := "gruntwork-terratest-" + strings.ToLower(id)
+	s3LogsBucketName := s3BucketName + "-logs"
+
+	CreateS3Bucket(t, region, s3BucketName)
+	CreateS3Bucket(t, region, s3LogsBucketName)
+	PutS3BucketLogging(t, region, s3BucketName, s3LogsBucketName)
```
yorinasub17, line 44: If we end up removing this function, you should implement this inline. Alternatively, you should enhance https://github.com/gruntwork-io/terratest/tree/master/examples/terraform-aws-s3-example with access logging and update https://github.com/gruntwork-io/terratest/blob/master/test/terraform_aws_s3_example_test.go

yorinasub17 (commented):
Implementation mostly LGTM! However, I think a few of these functions fail our [litmus test for contributions](https://terratest.gruntwork.io/docs/community/contributing/#infrastructure-and-validation-helpers).

modules/aws/s3.go
ina-stoyanova: Yeah, that's a very good question @yorinasub17 

Initially, my thinking was "don't overclutter terratest, and just add the code I needed" _- as in the `GetLoggingTargetBucketName function`_

As I got to this yesterday though, I felt uncomfortable adding in code that was not tested, which then prompted me to add the PutBucketLogging method.

It wasn't a lot of extra effort, so I did add it in the end. Do you think it's smarter at this point to maybe skip that testing bit entirely? Or am I missing an obvious way how I can test without the prior setup? 

As it was a difficult trade-off, I thought it will be safer to push code that is tested. But as this was my main wondering _(and the reason why I didn't add the tests initially when the PR was opened)_, I'm very happy to take guidance and learn from this!

modules/aws/s3.go
ina-stoyanova: sweet! that simplifies it a lot.

modules/aws/s3.go
ina-stoyanova: Just checking if I've understood you correctly - is it that the function is a bit too specific?

I guess a bit of context would help - [this is actually going to help me test using `terratest` the new `terragrunt` `accesslogging_bucket_name`. ](https://github.com/gruntwork-io/terragrunt/pull/1400#pullrequestreview-518125751)

However, after you've pointed me to the `litmus test`, I can clearly see this is only needed as functionality in that one specific scenario.

modules/aws/s3_test.go
ina-stoyanova: Yeah - if it's fine for the putting bucket logging to live inside the test only, as we don't need it for more than one scenario, I think it's better to go with it being inline! This way, we don't add extra logic into terratest, but I'm still able to test the `getLogging` function, which _was_ my original and main intention.

modules/aws/s3.go
yorinasub17: The specific litmus test I am on the fence about is the "complexity" test, which states:

> Complexity: we ask that you only contribute infrastructure and validation helpers for code that is relatively complex to do from scratch. For example, a helper that merely wraps an existing function in the AWS or GCP SDK is not a great choice, as the wrapper isn’t contributing much value, but is bloating the Terratest API. On the other hand, helpers that expose simple APIs for complex logic are great contributions: ssh.CheckSshCommand is a great example of this, as it provides a simple one-line interface for dozens of lines of complicated SSH logic.

This function is a thin wrapper around the AWS SDK, which is why it can be considered as bloating the Terratest API. For example, in your test code, you can do something like the following:

```go
s3Client := NewS3Client(t, awsRegion)
res, err := GetBucketLogging(&s3.GetBucketLoggingInput{Bucket: &bucket})
require.NoError(t, err)
// Assert that logging is enabled
require.NotNil(t, res.LoggingEnabled)
// Assert that logging is done to the expected target bucket
assert.Equal(t, expectedTargetBucket, aws.StringValue(res.LoggingEnabled.TargetBucket))
```

This code is readable, and more direct in the assertions. Of course, arguably one line is better than five lines, and it will be subjective which is better (having a terratest function that simplifies these 5 lines of code, or writing them inline). Moreover, access log checks is very common, so one could argue that the terratest API is incomplete without it. This is why I am on the fence, but leaning towards including it.

modules/aws/s3_test.go
yorinasub17: Coming back to this after a day, I actually think it might be better to do this in terraform (the alternative I suggested). That would probably be a more robust test and cleaner.

modules/aws/s3.go
yorinasub17: Ah gotcha! That makes sense.

So in general, we prefer to have terraform based testing for terratest functions (integration tests) as opposed to having unit tests. There are 2 main advantages for this:

- We get to use terraform, a robust platform for managing infra, to do the test setup and teardown of resources that terratest depends on. This allows us to keep the API focused on validating infra, as opposed to managing infra.

- Users tend to refer to the examples for inspiration on testing. As such, having a terraform driven example + test, it leads to better discoverability of the functions (as opposed to a unit test which might not be as visible).

modules/aws/s3.go
ina-stoyanova: OK, this is really insightful, Yori! Thanks!!

I've spent some time adding this as a terraform test instead. I've got a bit more cleanup to do, so will move to draft for now, and then move it back into "Ready for Review" as I work through stuff (most likely on Monday).

modules/aws/s3.go
ina-stoyanova: this is super useful to understand - thanks for explaining! 👍

ina-stoyanova:
I believe this is ready for another review 🙂 Moving from Draft to ready for Review.

examples/azure/terraform-azure-mysqldb-example/main.tf
```diff
@@ -30,8 +30,8 @@ resource "azurerm_resource_group" "mysql_rg" {
 # Random password is used as an example to simplify the deployment and improve the security of the database.
 # This is not as a production recommendation as the password is stored in the Terraform state file.
 resource "random_password" "password" {
-  length = 16
-  special = true
+  length           = 16
+  special          = true
```
yorinasub17, line 34: Ah sorry I caused hassle here. I just merged the fix for this in master so you might have conflicts with it.

examples/terraform-aws-s3-logging-enabled-example/README.md
```diff
@@ -0,0 +1,50 @@
+# Terraform AWS S3 Example
```
yorinasub17, line 1: Oh I feel bad saying this after all the effort into creating a new example, but I actually meant that you should update the existing example as opposed to creating a new one (as suggested here: https://github.com/gruntwork-io/terratest/pull/689#discussion_r518288068)

This is more an art instead of a policy, but the main reason for it is that most people will look at the S3 example for inspirations on testing s3 buckets, and it would be a bit of a sprawl to have a separate example/test just for the logging test (since it is a relatively small extension of the existing s3 example).

We want to limit the number of examples/tests we add because the overhead of another example is not just UX, but it also affects our testing infra as well, as that is more code to maintain when we do, for example, a terraform version update.

Again this is a bit difficult to explain given that it is more an art of balancing tradeoffs, but in this case I think it would be favorable to add access logging to the existing s3 example and add the additional check to the existing test.

modules/aws/s3.go
```diff
@@ -260,6 +260,35 @@ func EmptyS3BucketE(t testing.TestingT, region string, name string) error {
...
+	})
+	if err != nil {
+		return "", err
+	}
+
+	if res.LoggingEnabled != nil {
+		return aws.StringValue(res.LoggingEnabled.TargetBucket), nil
+	}
+
+	return "", err
```
yorinasub17, line 289: I missed this in the first pass, but I just realized that this won't return an error, but `nil` (because you had the `err` check before). It might be more favorable to return an error in this case, since logging is not enabled.

Can you make this return a custom error (e.g., `S3AccessLoggingNotEnabledErr`)?

yorinasub17 (commented):
Almost there! I had 2 more nits, but once that is addressed I think we can merge this in!

examples/azure/terraform-azure-mysqldb-example/main.tf
ina-stoyanova: no probs - I'll merge!

test/terraform_aws_s3_logging_enabled_example_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+// An example of how to test the Terraform module in examples/terraform-aws-s3-example using Terratest.
+func TestTerraformAwsS3LoggingEnabledExample(t *testing.T) {
+	t.Parallel()
+
+	// Give this S3 Bucket a unique ID for a name tag so we can distinguish it from any other Buckets provisioned
+	// in your AWS account
+	expectedName := fmt.Sprintf("terratest-aws-s3-logging-example-%s", strings.ToLower(random.UniqueId()))
+
+	// Give this S3 Bucket an environment to operate as a part of for the purposes of resource tagging
```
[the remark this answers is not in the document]
ina-stoyanova: hah good spot! I re-read this thing so many times and still missed it! Thanks :)

modules/aws/s3.go
ina-stoyanova: Good shout!

examples/terraform-aws-s3-logging-enabled-example/README.md
ina-stoyanova: gotcha! no worries - it's all good experience! 

Having less code to maintain is a really good reason that resonates, and also giving people less variations will also be more straightforward! Agree :)

test/terraform_aws_s3_logging_enabled_example_test.go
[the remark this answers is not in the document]
zackproser: Thanks, was also just confirming my own understanding :+1:

ina-stoyanova:
The tests for this keep breaking for no apparent reason. When run locally, obviously all passes 🙄  I'm looking into this, but will try re-running the tests a few times and see if anything concrete comes up as a reason.

ina-stoyanova:
Ok, now that the tests are passing, good for one more try @yorinasub17 @zackproser - would appreciate another few minutes of your time to get this merged! 🙂

examples/terraform-aws-s3-logging-enabled-example/README.md
yorinasub17: It looks like you still have the new example here. Was there some discussion where we decided to go with the new example instead of consolidating?

ina-stoyanova: oh no - that's a mistake! correcting it now..

ina-stoyanova: D'oh! I thought I pushed - but apparently hadn't done so. All pushed up now :) My bad!

examples/terraform-aws-s3-example/README.md
```diff
@@ -1,16 +1,17 @@
 # Terraform AWS S3 Example
 
 This folder contains a simple Terraform module that deploys resources in [AWS](https://aws.amazon.com/) to demonstrate
-how you can use Terratest to write automated tests for your AWS Terraform code. This module deploys an [S3
-Bucket](https://aws.amazon.com/s3/) and gives that Bucket a `Name` & `Environment` tag with the value specified in the
-`tag_bucket_name` and `tag_bucket_environment` variables, respectively.  This module also contains a terraform variable 
-that will create a basic bucket policy that will restrict the bucket to only accept SSL connections.
+how you can use Terratest to write automated tests for your AWS Terraform code. This module deploys 2 [S3
+Buckets](https://aws.amazon.com/s3/) - one S3 Bucket with enabled logging and versioning, and another "targetBucket" one to serve as a
```
yorinasub17, line 5: NIT:

```suggestion
Buckets](https://aws.amazon.com/s3/) - one S3 Bucket with logging and versioning enabled, and another "targetBucket" one to serve as a
```

examples/terraform-aws-s3-example/main.tf
```diff
@@ -53,6 +53,16 @@ data "aws_iam_policy_document" "s3_bucket_policy" {
...
 
+resource "aws_s3_bucket" "test-bucket-logs" {
+  bucket = "${local.aws_account_id}-${var.tag_bucket_name}-logs"
+  acl    = "log-delivery-write"
+
+  tags = {
+    Name        = "${local.aws_account_id}-${var.tag_bucket_name}-logs"
+    Environment = var.tag_bucket_environment
+  }
+}
```
yorinasub17, line 64: Might want to set `force_destroy = true` to ensure it can be destroyed.

examples/terraform-aws-s3-example/main.tf
```diff
@@ -53,6 +53,16 @@ data "aws_iam_policy_document" "s3_bucket_policy" {
   }
 }
 
+resource "aws_s3_bucket" "test-bucket-logs" {
```
yorinasub17, line 56: Our terraform style guide is to use snake case instead of kebob case for block labels.

```suggestion
resource "aws_s3_bucket" "test_bucket_logs" {
```

examples/terraform-aws-s3-example/outputs.tf
```diff
@@ -6,3 +6,6 @@ output "bucket_arn" {
   value = aws_s3_bucket.test_bucket.arn
 }
 
+output "logging_target_bucket" {
+  value = aws_s3_bucket.test_bucket.logging[*].target_bucket
```
yorinasub17, line 10: NIT: Use `[0]` so that we return a string instead of a list:
```suggestion
  value = aws_s3_bucket.test_bucket.logging[0].target_bucket
```

yorinasub17 (commented):
LGTM! Only had a few more style nits in the example, but otherwise looks good!

examples/terraform-aws-s3-example/main.tf
ina-stoyanova: good point!

examples/terraform-aws-s3-example/outputs.tf
ina-stoyanova: This is what i tried initially, but then came across [this issue:](https://github.com/hashicorp/terraform/issues/22504)
```
Block type "logging" is represented by a set of objects, and set elements do
        	            	not have addressable keys. To find elements matching specific criteria, use a
        	            	"for" expression with an "if" clause.
```

I thought the simplest way is leave this as is. We have a few other ways we could go about this to make it clearer though:

- add a comment above this line
- use `tolist(aws_s3_bucket.test_bucket.logging)[0].target_bucket` (I've tested this and it seemed to work well!)
- or try changing the lifecycle policy for this: _(as suggested in some of the issue threads)_
```
lifecycle {
     create_before_destroy = true
     ignore_changes = [
       # Ignore changes to first tag in list: "foo"
       tags[0]
     ]
   }
```

I'd probably go with the 2nd option - use the `tolist` tf function, so it's clear to users of this example this returns a list

yorinasub17: Ah gotcha. Yes the `tolist` approach works for me!

yorinasub17 (approved):
Updates LGTM!

ina-stoyanova:
Sweet! Thanks for the help and the patience @yorinasub17! 🎉

PR #694, opened 2020-10-30
Extend #484: Remove plan struct and implement plan file support with Show

yorinasub17:
This extends #484 with the following changes:

- Renames the `Out` attribute in the `Options` structs to the more descriptive `PlanFilePath`.
- Only processes the `PlanFilePath` setting if the command is compatible with plan files (tracked in the new list `TerraformCommandsWithPlanFileSupport`).
- Removes the `jsonplan` package as that violates MPL, and also because it is implemented in the [terraform-json](https://github.com/hashicorp/terraform-json) package.
- Introduce another convenience function, `InitAndPlanAndShow`.
- Adds a top level example showing how you can use it (NOTE: I opted not to link in `terraform-json` for now, given the concerns about the licensing, but once that is resolved, I can update the example to use that).


modules/terraform/format.go
```diff
@@ -38,9 +49,27 @@ func FormatArgs(options *Options, args ...string) []string {
...
+		// The plan file arg should be last in the terraformArgs slice. Some commands use it as an input (e.g. show, apply)
+		terraformArgs = append(terraformArgs, FormatTerraformPlanFileAsArg(commandType, options.PlanFilePath)...)
+	}
+
 	return terraformArgs
 }
 
+// FormatTerraformPlanFileAsArg formats the out variable as a command-line arg for Terraform (e.g. of the format
+// -out=/some/path/to/plan.out or /some/path/to/plan.out). Only plan supports passing in the plan file as -out; the
+// other commands expect it as the first positional argument.
```
brikis98, line 62: NIT: mention this returns an empty list if `outPath` is an empty string.

modules/terraform/options.go
```diff
@@ -59,8 +59,10 @@ type Options struct {
 	NoColor                  bool                   // Whether the -no-color flag will be set for any Terraform command or not
 	SshAgent                 *ssh.SshAgent          // Overrides local SSH agent with the given in-process agent
 	NoStderr                 bool                   // Disable stderr redirection
+	OutputMaxLineSize        int                    // The max size of one line in stdout and stderr (in bytes)
 	Logger                   *logger.Logger         // Set a non-default logger that should be used. See the logger package for more info.
 	Parallelism              int                    // Set the parallelism setting for Terraform
+	PlanFilePath             string                 // The path to output a plan file to
```
brikis98, line 65:
```suggestion
	PlanFilePath             string                 // The path to output a plan file to (for the plan command) or read one from (for the apply command)
```

modules/terraform/plan.go
```diff
@@ -41,6 +41,28 @@ func PlanE(t testing.TestingT, options *Options) (string, error) {
 	return RunTerraformCommandE(t, options, FormatArgs(options, "plan", "-input=false", "-lock=false")...)
 }
 
+// InitAndPlanAndShow runs terraform init, then terraform plan, and then terraform show with the given options, and
+// returns the json output of the plan file. This will fail the test if there is an error in the command.
+func InitAndPlanAndShow(t testing.TestingT, options *Options) string {
```
brikis98, line 46: Thinking out loud, what's the difference between:

1. `terraform show -json`
1. `terraform plan -out=x && terraform show -json x`

I guess (1) shows the contents of the state file in JSON format, while (2) is a way to see the contents of a `plan` in JSON format? And I guess the only reason we need (2) is because `plan` doesn't natively support a `-json` flag?

modules/terraform/show.go
```diff
@@ -0,0 +1,26 @@
+package terraform
+
+import (
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// Show runs terraform show with the given options and returns stdout/stderr.
```
brikis98, line 8: NIT: Mention it runs `show` with `-json`, so the output will be in JSON format. BTW, should this _only_ return `stdout` so you get solely the JSON output, and none of the logging around it?

brikis98: Oh, I just realized this description doesn't match the one for `Show` below...

modules/terraform/show.go
```diff
@@ -0,0 +1,26 @@
...
+// This will fail the test if there is an error in the command.
+func Show(t testing.TestingT, options *Options) string {
+	out, err := ShowE(t, options)
+	require.NoError(t, err)
+	return out
+}
+
+// ShowE calls terraform show for the given plan output file and returns it.
```
brikis98, line 16: Hm, shouldn't we respect booth versions of `show`? The one that just outputs the state file and the one that reads a `plan` file and outputs data about that?

modules/terraform/plan.go
yorinasub17: Yes that's right. E.g., if you do `terraform show -json` on an unapplied module, you won't see anything.

modules/terraform/format.go
yorinasub17: Done: 7a3339e6ae1b0a031913a09312c420f35bce923e

modules/terraform/show.go
yorinasub17: Updated to sync the two descriptions and also made it so that `Show` only returns stdout: 7a3339e6ae1b0a031913a09312c420f35bce923e

modules/terraform/show.go
yorinasub17: Fixed to allow running show without plan files: 7a3339e6ae1b0a031913a09312c420f35bce923e

yorinasub17:
Thanks for the review! I believe I addressed all the comments. I also made a major update in 8dc6d81, where:

- I implemented a new function that allows you to index into a json blob with json path and unmarshal the result into an object. This is using the jsonpath util library in kubernetes `client-go`, so I implemented it in the `k8s` module. Note that I needed to bump the dependency to the v19 series so that it outputs the path in valid json format (otherwise, it uses a weird human friendly version like `[1 2 3]`).

- I updated the plan test to use json path so I can avoid all the type checking. Note that if the plan was malformed, you get a friendly error message along the lines of `Error extracting json path: values is not found`.

modules/terraform/plan.go
brikis98: Roger

modules/k8s/jsonpath.go
```diff
@@ -0,0 +1,72 @@
...
+// jsonPath := "{.key.data[*]}"
+// var output []int
+// UnmarshalJSONPathE(t, jsonBlob, jsonPath, &output)
+// => output = []int{1,2,3}
+func UnmarshalJSONPathE(t testing.TestingT, jsonData []byte, jsonpathStr string, output interface{}) error {
+	// First, unmarshal the full json object. We use interface{} to avoid the type conversions, as jsonpath will handle
+	// it for us.
+	var blob interface{}
+	if err := json.Unmarshal(jsonData, &blob); err != nil {
+		logger.Logf(t, "Error unmarshaling original json blob")
```
brikis98, line 48: Why do we need to log the error we're returning? Seems redundant? Do we need more descriptive error types?

modules/k8s/jsonpath_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUnmarshalJSONPath(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name        string
+		jsonBlob    []byte
```
brikis98, line 16: NIT: Could make this a `string` and have the test code convert to a byte array.

modules/terraform/show.go
```diff
@@ -0,0 +1,30 @@
...
+// terraform module at options.TerraformDir.
+func ShowE(t testing.TestingT, options *Options) (string, error) {
+	// We manually construct the args here instead of using `FormatArgs`, because show only accepts a limited set of
+	// args.
+	args := []string{"show", "-no-color", "-json"}
+
+	// Attach plan file path if specified.
+	if options.PlanFilePath != "" {
+		args = append(args, options.PlanFilePath)
+	}
```
brikis98, line 28: NIT: should this use `FormatTerraformPlanFileAsArg`?

brikis98 (approved):
JSONPath is a nice addition! Code LGTM. Just some minor NITs to consider or ignore as you see fit.

modules/k8s/jsonpath.go
yorinasub17: I added these in because the error returned by these are not very descriptive, and also because we don't have a stack trace, so it isn't clear where in this routine it failed.

E.g., a json error would show something like "unexpected char {", but coincidentally, the jsonpath parser could emit a very similar error. So you wouldn't exactly know if the error is from json parsing, or jsonpath parsing.

But this isn't ideal either because the error log doesn't show up in the assertion failure so you have scroll through logs. I'll switch to using error types so that the error message shows up in the assert library.

yorinasub17: Updated ee1205c

modules/k8s/jsonpath_test.go
yorinasub17: Updated 0dab3d7

modules/terraform/show.go
yorinasub17: I decided not to switch it, because it doesn't really save much.

brikis98 (approved):
LGTM!

yorinasub17:
Tests failed due to https://github.com/gruntwork-io/terratest/pull/693

Should I wait for us to resolve that branch? Or can I merge? Note that this sort of blocks the architecture catalog unit testing.

brikis98:
The gotcha with merging is that the release fails, so the Terratest log parser binaries don't get published 😕 

Perhaps we should update the release steps to run even if tests fail?

brikis98:
(note the last several releases are missing binaries)

yorinasub17:
> The gotcha with merging is that the release fails, so the Terratest log parser binaries don't get published

If that is the only concern, then let me go ahead and merge this and manually generate the binaries for this release.

PR #702, opened 2020-11-09
Add two new functions for testing if existing file is a dir or file

yorinasub17:
This adds helper functions for ensuring that a given path exists and is a dir or file.

NOTE: This doesn't pass the complexity litmus test, but I think there is an argument to be made for the completeness litmus test where `FileExists` is sometimes not a sufficient test as you want to make sure the path is a directory.


infraredgirl (approved):
Code LGTM and I agree with the complexity vs completeness argument.

yorinasub17:

PR #713, opened 2020-11-25
fix pod availability checking logic

Hsn723:
The changes to `IsPodAvailable` introduced in https://github.com/gruntwork-io/terratest/pull/480 incorrectly reported pods as available when the readinessProbe is failing but the containers were in a started state. Namely, when there is no startupProbe, `containerStatus.Started` will always be true ([ref](https://godoc.org/k8s.io/api/core/v1#ContainerStatus)) and the pod will be incorrectly reported as available.

This PR fixes the availability checking logic and adds a few test cases to validate the fix.

PR #714, opened 2020-11-30
Fix IsPodAvailable readiness check

Jerska:
Hey there!

It appears to me as if the condition in `k8s.IsPodAvailable` should return false either if:
- the pod's container is not started
- or if the pod's container is not ready

However, the current implem will return false only if both conditions are true.
This PR fixes it.

### DIFF
--- a/modules/k8s/pod.go
+++ b/modules/k8s/pod.go
@@ -142,7 +142,7 @@ func IsPodAvailable(pod *corev1.Pod) bool {
 		isContainerStarted := containerStatus.Started
 		isContainerReady := containerStatus.Ready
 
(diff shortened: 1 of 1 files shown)


infraredgirl:
Hi @Jerska, thanks for the contribution! Code LGTM, could you also add a unit test to verify this fix? Thanks!

brikis98:
@infraredgirl I just noticed that this PR seems to be identical to https://github.com/gruntwork-io/terratest/pull/713, but the latter has tests. Closing in favor of that one!

infraredgirl:
Ah, good catch!

Issue #717, opened 2020-12-03
TF 0.14.0 String Output leaves double quotes


dee-kryvenko:
Using `v0.30.27`. Having a simple tf `v0.14.0` code:

```
output "hello_world" {
  value = "Hello, World!"
}
```

Trying to:

```
output := terraform.Output(t, terraformOptions, "hello_world")
assert.Equal(t, "Hello, World!", output)
```

Fails with:

```
                Error:          Not equal: 
                                expected: "Hello, World!"
                                actual  : "\"Hello, World!\""
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -Hello, World!
                                +"Hello, World!"
```

brikis98:
Ah, bummer, Terraform 0.14 changed its output format.

For:

```hcl
output "foo" {
  value = "Hello, World"
}
```

With Terraform 0.13:

```
$ terraform output foo
Hello, World
```

And with Terraform 0.14:

```
$ terraform output foo
"Hello, World"
```

I don't see that in their release notes, but that seems like a significant API change...

brikis98:
I filed https://github.com/hashicorp/terraform/issues/27100 to see if this change was intentional or not...

dee-kryvenko:
@brikis98 seems like they confirmed the change was intentional. Here's a proposed fix: https://github.com/gruntwork-io/terratest/pull/718

wylie-ucb:
I know I'm a little late to the party here, is there a recommended resolution for this?  I'm seeing this after upgrading to TF14, using 0.14.10.  Thanks so much.

PR #718, opened 2020-12-03
Fix terraform string output for 0.14 (use -json)

dee-kryvenko:
Fixing #717

### DIFF
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
-	"strings"
 
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


brikis98 (approved):
Thanks for the PR! This seems like a reasonable approach. I'm going to kick off tests.

brikis98:
Oh, whoops, just realized our CI config is still on Terraform 0.13! Could you please bump to `0.14.0`? https://github.com/gruntwork-io/terratest/blob/master/.circleci/config.yml#L5

modules/terraform/output.go
```diff
@@ -21,13 +20,10 @@ func Output(t testing.TestingT, options *Options, key string) string {
 
 // OutputE calls terraform output for the given variable and return its value.
 func OutputE(t testing.TestingT, options *Options, key string) (string, error) {
-	output, err := RunTerraformCommandAndGetStdoutE(t, options, "output", "-no-color", key)
+	var str string
+	err := terraform.OutputStructE(t, options, key, &str)
```
brikis98, line 24: Oh, this is a compile error. You're already in the `terraform` package, so no need to call out to it.

brikis98:
To get this wrapped up quickly, I'm going to open a new PR built on this one, but with some fixes.

Issue #720, opened 2020-12-04
GCP / SSH test failures


brikis98:
We're seeing a couple test failures related to GCP and SSH.

@robmorgan Not sure if anything below rings a bell?

```
=== RUN   TestPackerGCPBasicExample
=== PAUSE TestPackerGCPBasicExample
=== CONT  TestPackerGCPBasicExample
TestPackerGCPBasicExample 2020-12-04T11:56:41Z retry.go:91: Attempting to request a Google OAuth2 token
TestPackerGCPBasicExample 2020-12-04T11:56:41Z compute.go:606: Successfully retrieved default GCP client
TestPackerGCPBasicExample 2020-12-04T11:56:42Z logger.go:66: Running Packer to generate a custom artifact for template ../examples/packer-basic-example/build.json
TestPackerGCPBasicExample 2020-12-04T11:56:42Z retry.go:91: packer [build -machine-readable -var gcp_zone=us-west1-b -var gcp_project_id=terratest-214610 -only=googlecompute ../examples/packer-basic-example/build.json]
TestPackerGCPBasicExample 2020-12-04T11:56:42Z logger.go:66: Running command packer with args [build -machine-readable -var gcp_zone=us-west1-b -var gcp_project_id=terratest-214610 -only=googlecompute ../examples/packer-basic-example/build.json]
TestPackerGCPBasicExample 2020-12-04T11:56:46Z logger.go:66: 1607083006,,ui,say,==> googlecompute: Checking image does not exist...
TestPackerGCPBasicExample 2020-12-04T11:56:47Z logger.go:66: 1607083007,,ui,say,==> googlecompute: Creating temporary rsa SSH key for instance...
TestPackerGCPBasicExample 2020-12-04T11:56:56Z logger.go:66: 1607083016,,ui,say,==> googlecompute: Using image: ubuntu-1804-bionic-v20201201
TestPackerGCPBasicExample 2020-12-04T11:56:56Z logger.go:66: 1607083016,,ui,say,==> googlecompute: Creating instance...
TestPackerGCPBasicExample 2020-12-04T11:56:56Z logger.go:66: 1607083016,,ui,message,    googlecompute: Loading zone: us-west1-b
TestPackerGCPBasicExample 2020-12-04T11:56:56Z logger.go:66: 1607083016,,ui,message,    googlecompute: Loading machine type: n1-standard-1
[...]

PR #723, opened 2020-12-04
Fix terraform string output types conversion; Add all json output.

dee-kryvenko:
- Fix regression in `OutputE`: now it can also output bool and number as string. I am not sure about other terraform data types but their string representation would be useless and inconsistent anyway, I guess it's one of the reasons TF did that breaking change in the first place. Added suggestion to use `OutputStruct` for non-primitive types.
- `OutputE` and `Output` comment now explicitly says it returns string representation of a value.
- Added `OutputJsonE` and `OutputJson` that will return `-json` tf output as-is - might be useful if user wants to do anything fancy with it
- Some DRY cleanup - other output functions now use the new `OutputJsonE` function instead of `RunTerraformCommandAndGetStdoutE`
- Added missing tests for `Output` as well as for new `OutputJson`

### DIFF
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -12,7 +12,7 @@ import (
 )
 
 // Output calls terraform output for the given variable and return its string value representation.
-// It is only designed to work for primitive terraform types: string, number and bool.
+// It only designed to work with primitive terraform types: string, number and bool.
 // Please use OutputStruct for anything else.
 func Output(t testing.TestingT, options *Options, key string) string {
 	out, err := OutputE(t, options, key)
@@ -21,7 +21,7 @@ func Output(t testing.TestingT, options *Options, key string) string {
 }
 
 // OutputE calls terraform output for the given variable and return its string value representation.
-// It is only designed to work for primitive terraform types: string, number and bool.
+// It only designed to work with primitive terraform types: string, number and bool.
 // Please use OutputStructE for anything else.
 func OutputE(t testing.TestingT, options *Options, key string) (string, error) {
 	var val interface{}
@@ -263,7 +263,7 @@ func OutputForKeys(t testing.TestingT, options *Options, keys []string) map[stri
 
 // OutputJson calls terraform output for the given variable and returns the
 // result as the json string.
-// If v is an empty string, it will return entire output.
+// If v is an empty string, it will return the entire output.
 func OutputJson(t testing.TestingT, options *Options, key string) string {
 	str, err := OutputJsonE(t, options, key)
 	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


modules/terraform/output.go
```diff
@@ -257,6 +261,32 @@ func OutputForKeys(t testing.TestingT, options *Options, keys []string) map[stri
 	return out
 }
 
+// OutputJson calls terraform output for the given variable and returns the
+// result as the json string.
+// If v is an empty string, it will return the entire output.
```
brikis98, line 266:
```suggestion
// If key is an empty string, it will return all the output variables.
```

modules/terraform/output.go
```diff
@@ -257,6 +261,32 @@ func OutputForKeys(t testing.TestingT, options *Options, keys []string) map[stri
...
+func OutputJson(t testing.TestingT, options *Options, key string) string {
+	str, err := OutputJsonE(t, options, key)
+	require.NoError(t, err)
+	return str
+}
+
+// OutputJsonE calls terraform output for the given variable and returns the
+// result as the json string.
+// If v is an empty string, it will return the entire output.
```
brikis98, line 275:
```suggestion
// If key is an empty string, it will return all the output variables.
```

modules/terraform/output.go
```diff
@@ -257,6 +261,32 @@ func OutputForKeys(t testing.TestingT, options *Options, keys []string) map[stri
...
+	if key != "" {
+		args = append(args, key)
+	}
+
+	out, err := RunTerraformCommandAndGetStdoutE(t, options, args...)
+	if err != nil {
+		return "", err
+	}
+
+	return out, nil
```
brikis98, line 287:
```suggestion
	return RunTerraformCommandAndGetStdoutE(t, options, args...)
```

brikis98 (commented):
This is great, thank you for the fixes and clean up! I made some minor code clean up tweaks. Once those are merged, I can kick off tests.

dee-kryvenko:
@brikis98 thanks for the review - added all your fixes.

brikis98:
Thanks! I'll kick off tests shortly.

dee-kryvenko:
Hmmm these two failed tests `TestPackerGCPBasicExample` and `TestSshAccessToComputeInstance ` doesn't seem to be related...

brikis98 (approved):
Yea, those two tests are flaky now for some reason: https://github.com/gruntwork-io/terratest/issues/720. I'm OK ignoring them and merging. Thank you!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.31.1

PR #727, opened 2020-12-08
Allow extra args for helm Install/Upgrade

### LINKED ISSUE #726: Unable to pass extra arguments to `helm install`
I'd like to be able to pass arguments to the helm command when using the [`Install`](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/install.go#L15) function in the `helm` module. 

The use case is that we have a test that uses the `Install` function in the `helm` module, but that test is flaky because it needs more time to wait for the helm install. We'd like to be able to pass `--wait --timeout 15m` to the helm install command but we're currently not able to do that with the existing `Install` function.

We could add an option to [options.go](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/options.go#L17) called `ExtraArgs` and modify [`InstallE`](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/install.go#L34) to check if `options.ExtraArgs` is set, and pass those along to [`RunHelmCommandAndGetOutputE`](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/install.go#L44). For consistency this change could also be made to [`UpgradeE`](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/upgrade.go#L31). This would be backwards compatible since it doesn't change the signature of the `Install` or `Upgrade` functions in the helm module.

I'm happy to contribute this change! EDIT: https://github.com/gruntwork-io/terratest/pull/727

ndhanushkodi:
### GH Issue
[GH Issue 726](https://github.com/gruntwork-io/terratest/issues/726)

### Changes Proposed
This PR allows passing arguments to the helm command when using the `Install` and `Upgrade` functions in the `helm` module.

### Test output
The test output can be found in this [gist](https://gist.github.com/ndhanushkodi/006b904a188604099654f0f3280762c4)

### Description of code changes
Adds an option to [options.go](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/options.go#L17) called `ExtraArgs` and modifies [`InstallE`](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/install.go#L34)  and [`UpgradeE`](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/upgrade.go#L31) to check if `options.ExtraArgs` is set, and passes those along to [`RunHelmCommandAndGetOutputE`](https://github.com/gruntwork-io/terratest/blob/2afcf242dd02ba62bc1ca5fe555b4e6550dbffcf/modules/helm/install.go#L44). This would be backwards compatible since it doesn't change the signature of the `Install` or `Upgrade` functions in the helm module.

Closes #726

### DIFF
--- a/modules/helm/install.go
+++ b/modules/helm/install.go
@@ -32,6 +32,9 @@ func InstallE(t testing.TestingT, options *Options, chart string, releaseName st
 	// Declare err here so that we can update args later
 	var err error
 	args := []string{}
+	if options.ExtraArgs != nil {
+		args = append(args, options.ExtraArgs...)
+	}
 	args = append(args, getNamespaceArgs(options)...)
 	if options.Version != "" {
 		args = append(args, "--version", options.Version)
--- a/modules/helm/install_test.go
+++ b/modules/helm/install_test.go
@@ -65,6 +65,8 @@ func TestRemoteChartInstall(t *testing.T) {
 
 	// Fix chart version and retry install
 	options.Version = "2.3.0"
+	// Test that passing extra arguments doesn't error, by changing default timeout
+	options.ExtraArgs = []string{"--timeout", "5m1s"}
 	require.NoError(t, InstallE(t, options, helmChart, releaseName))
 	waitForRemoteChartPods(t, kubectlOptions, releaseName, 1)
 
--- a/modules/helm/options.go
+++ b/modules/helm/options.go
@@ -15,4 +15,5 @@ type Options struct {
 	EnvVars        map[string]string   // Environment variables to set when running helm
 	Version        string              // Version of chart
 	Logger         *logger.Logger      // Set a non-default logger that should be used. See the logger package for more info.
+	ExtraArgs      []string            // Extra arguments to pass to the helm install/upgrade command
 }
--- a/modules/helm/upgrade.go
+++ b/modules/helm/upgrade.go
@@ -29,6 +29,9 @@ func UpgradeE(t testing.TestingT, options *Options, chart string, releaseName st
 
 	var err error
 	args := []string{}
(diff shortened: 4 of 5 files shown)


infraredgirl (commented):
Code changes LGTM, I'll kick off the tests!

infraredgirl (approved):
The two failed tests are a known issue (#720). Approving and merging this now. Many thanks for the contribution @ndhanushkodi!

yorinasub17 (commented):
I know I am late to the review, but it probably makes more sense for `ExtraArgs` to be a map from subcommand to args list, and handle this in all the commands, not just install and update.

infraredgirl:
That's a good point Yori, my apologies for missing that in the review. I opened the issue #731 for this. 

@ndhanushkodi feel free to address this if you have any cycles. if not no worries, someone else will eventually get to it!

ndhanushkodi:
Thank you @infraredgirl and @yorinasub17 for your time in looking at this PR! @yorinasub17 your suggestion to make ExtraArgs a map from subcommand to args list makes sense and I've implemented that in a new PR here: https://github.com/gruntwork-io/terratest/pull/732

PR #729, opened 2020-12-08
AWS module - get Lambda function configuration

vasiliy-sharko:
This code allows getting Lambda function details from AWS account and comparing it to those created by IaC tools (Terraform).

### DIFF
--- a/modules/aws/lambda.go
+++ b/modules/aws/lambda.go
@@ -59,6 +59,30 @@ func (err *FunctionError) Error() string {
 	return fmt.Sprintf("%s error invoking lambda function: %v", err.Message, err.Payload)
 }
 
+//GetLambdaFunctionConfiguration gets ARN for a given function.
+func GetLambdaFunctionConfiguration(t testing.TestingT, region, functionName string) lambda.FunctionConfiguration {
+	out, err := GetLambdaFunctionArnE(t, region, functionName)
+	require.NoError(t, err)
+	return out
+}
+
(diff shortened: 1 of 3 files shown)


modules/aws/lambda.go
```diff
@@ -59,6 +59,30 @@ func (err *FunctionError) Error() string {
...
+//GetLambdaFunctionConfigurationE gets ARN for a given function.
+func GetLambdaFunctionConfigurationE(t testing.TestingT, region, functionName string) (lambda.FunctionConfiguration, error) {
+	lambdaClient, err := NewLambdaClient(t, region)
+
+	input := &lambda.GetFunctionInput{
+		FunctionName: &functionName
+	}
+
+	out, err = lambdaClient.GetFunction(input)
+	require.NoError(t, err)
```
yorinasub17, line 78: Remove this line, as the `E` flavor should be bubbling errors instead of failing tests.

yorinasub17 (commented):
Thanks for the contribution! This looks good to me other than the suggestion.

Can you also add a call to this function in https://github.com/gruntwork-io/terratest/blob/master/test/terraform_aws_lambda_example_test.go so it is tested?

MAINTAINER'S NOTE: Although this contribution does not pass our litmus test for contributions (specifically, it fails the `Complexity` test), I think it is ok to accept this on the grounds of `Completeness`.

vasiliy-sharko:
@yorinasub17 Uploaded a new commit according to your comments :)

james00012:

PR #730, opened 2020-12-08
WIP: "tspec" - Add a BDD-style interface to Terratest

robmorgan:
This PR adds a [BDD-style](https://cucumber.io/docs/bdd/) interface to Terratest which means you can write infrastructure tests in plain-English.

## RFC

Terratest is great and really powerful, but since it's written in Go, it can sometimes be a significant learning curve for a lot of people to properly test their infrastructure. Imagine if you could write infrastructure tests in plain-English, but take advantage of all of the Terratest modules under the hood? This PR adds a [BDD-style](https://cucumber.io/docs/bdd/) interface to Terratest which means you can write infrastructure tests without writing a single line of code.

Consider the [Hello, World test](https://github.com/gruntwork-io/terratest/blob/master/test/terraform_hello_world_example_test.go) for Terratest. It's 25+ lines for the simplest use case. Using `tspec` you could write a feature test as follows:

```Gherkin
Feature: Terraform Hello World

  Scenario: Run a simple test
    Given the Terraform module at "./examples/terraform-hello-world-example"
    When I run "terraform apply"
    Then the "hello_world" output is "Hello, World!"
```

Or to test something slightly more complicated you could write a feature test along the lines of:

```Gherkin
Feature: Terraform AWS Example

  Scenario: Run a simple test on AWS
    Given the Terraform module at "./examples/terraform-aws-example"
    And an input variable named "instance_name" with the value "test-instance-1"
    And an environment variable named "AWS_DEFAULT_REGION" with the value "us-east-1"
    When I run "terraform apply"
    Then the "instance_id" output should match "i-[0-9a-z]+(\,i-[0-9a-z]+)*"
```

When you execute `tspec` it will automatically copy the specified Terraform module to a temporary path, pass in the target parameters and ensure the tests pass as expected using Terratest methods under the hood.

## PoC

Under the hood, `tspec` borrows heavily from the [Godog](https://github.com/cucumber/godog) Cucumber library but includes a library of common step definitions so users don't need to write any Go code.

**Note:** Unfortunately this is a rather _large_ PR, so this RFC should serve as a potential discussion as to how we can "chop" this tool down into the smallest possible product increment.

### TODO
- [ ] Support for [Background](https://cucumber.io/docs/gherkin/reference/?sbsearch=background) grammar.
- [ ] Add more grammar for tools like Packer & Docker.

### Open Questions

1. Should we remove all of the various formatters like `junit`, `event` etc from the first version?
1. What should we name this? `tspec`, `infraspec`?


cmd/tspec/main.go
```diff
@@ -0,0 +1,98 @@
...
+   --log-level LEVEL  Set the log level to LEVEL. Must be one of: [panic fatal error warning info debug]
+                      (default: "info")
+   --outputdir value  Path to directory to output test output to. If unset will use the current directory.
+   --help, -h         show help
+`
+
+var opts = tspec.Options{
+	Output:      colors.Colored(os.Stdout),
+	Format:      "pretty", // can define default values
+	Concurrency: 1,
```
brikis98, line 33: What's this do?

cmd/tspec/main.go
```diff
@@ -0,0 +1,98 @@
...
+	opts.Paths = []string{cliContext.Args().Get(0)}
+
+	status := tspec.TestSuite{
+		Name:                 "tspec",
+		TestSuiteInitializer: tspec.InitializeTestSuite,
+		ScenarioInitializer:  tspec.InitializeScenario,
+		Options:              &opts,
+	}.Run()
+
+	logger.Infof("Exit code is: %d", status)
```
brikis98, line 61: Are you  supposed to exit with this exit code? So if tests fail, you exit with a non-zero code?

cmd/tspec/main.go
```diff
@@ -0,0 +1,98 @@
...
+	entrypoint.HelpTextLineWidth = 120
+
+	app.Name = "tspec"
+	app.Author = "Gruntwork <www.gruntwork.io>"
+	app.Description = `A BDD testing tool that uses Terratest under the hood.`
+	app.Action = run
+
+	currentDir, err := os.Getwd()
+	if err != nil {
+		logger.Fatalf("Error finding current directory: %s", err)
```
brikis98, line 78: I believe this will `panic` under the hood, which I've always found surprising with a `logger` method... Might be clearer to make this more explicit.

features/terraform_aws_example.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Terraform AWS Example
```
brikis98, line 1: Is putting these in a `features` folder a BDD idiom? Or a requirement? 

If it's not a requirement, I'd be tempted to put them under `test/features` instead...

modules/tspec/assert.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// assertExpectedAndActual is a helper function to allow the step function to call
+// assertion functions where you want to compare an expected and an actual value.
+func assertExpectedAndActual(a expectedAndActualAssertion, expected, actual interface{}, msgAndArgs ...interface{}) error {
+	var t asserter
+	a(&t, expected, actual, msgAndArgs...)
```
brikis98, line 13: NIT: try to avoid one-letter variable names like `a`. Even though it's defined just 2 lines above, but the time you read this line, it's completely opaque what `a` is supposed to be.

modules/tspec/colors/ansi_others.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2014 shiena Authors. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+// +build !windows
+
+package colors
```
brikis98, line 7: Is all this `colors` package stuff copied from somewhere?

modules/tspec/fmt.go
```diff
@@ -0,0 +1,76 @@
...
+
+type storageFormatter interface {
+	SetStorage(*storage.Storage)
+}
+
+// FormatterFunc builds a formatter with given
+// suite name and io.Writer to record output
+type FormatterFunc = formatters.FormatterFunc
+
+func printStepDefinitions(steps []*models.StepDefinition, w io.Writer) {
```
brikis98, line 55: What is this formatter stuff? And step definitions? I think I'm missing some context on how this stuff all comes together.

modules/tspec/internal/formatters/fmt.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// DefinitionID ...
+func DefinitionID(sd *models.StepDefinition) string {
+	ptr := sd.HandlerValue.Pointer()
+	f := runtime.FuncForPC(ptr)
+	file, line := f.FileLine(ptr)
+	dir := filepath.Dir(file)
+
+	fn := strings.Replace(f.Name(), dir, "", -1)
```
brikis98, line 82: Struggling to understand this code too. In part, due to lots of one and two letter variable names, and in part, as I'm missing some context of how this BDD API is defined...

modules/tspec/internal/formatters/fmt_base.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if text := f.Snippets(); text != "" {
+		fmt.Fprintln(f.out, "")
+		fmt.Fprintln(f.out, yellow("You can implement step definitions for undefined steps with these snippets:"))
+		fmt.Fprintln(f.out, yellow(text))
+	}
+}
+
+// Snippets ...
+func (f *Basefmt) Snippets() string {
```
brikis98, line 198: Is this all code you wrote? Or is it copied from somewhere?

brikis98 (commented):
First, the BDD functionality is FANTASTIC. So cool to see this. I think it'll extend the ability to use Terratest and test infrastructure code to so many more people. Thanks for putting this together Rob 🍺 

Second, I started to review this, but quickly ran into a lot of code that seemed to be copied from somewhere, or that I didn't fully understand, so I stopped. Could you provide a bit more context on this stuff? What to review? What the code is based on? Etc?

robmorgan:
@brikis98 hey Jim, thanks for a first pass! I feel you may have missed my remark in the RFC: the code **borrows heavily** from the [Godog Cucumber](https://github.com/cucumber/godog) library. I think a good approach now is possibly how I've extended this library CLI/tool and maybe we should focus on reviewing my custom parts. An additional step would be figuring out how to reduce the amount of code here. This would be super easy for me to explain on a call, so I'm wonder if we should briefly chat on Friday during Ticket Jam.

brikis98:
> @brikis98 hey Jim, thanks for a first pass! I feel you may have missed my remark in the RFC: the code **borrows heavily** from the [Godog Cucumber](https://github.com/cucumber/godog) library. 

I saw that, but coudn't figure out which parts were borrowed, and which are unique? Also, is it possible to use those parts directly from `godog`, pulling them in as dependencies?

> This would be super easy for me to explain on a call, so I'm wondering if we should briefly chat on Friday during Ticket Jam.

Sure, chatting live is fine too.

> p.s I'm not going to respond to your comments in the meantime.

What a jerk 😄

robmorgan:
@brikis98 actually on 2nd thoughts I'll leave a few comments here ahead of the call:

> I saw that, but couldn't figure out which parts were borrowed, and which are unique? Also, is it possible to use those parts directly from Godog, pulling them in as dependencies?

Yes definitely! I was unsure how much of the library I'd actually need to modify, but now I have a better sense I'm wondering if we should simply import the parts we need and only sprinkle in our custom logic? The downsides would be:

1. We need to add additional imports to Terratest.
2. You've mentioned some parts of the code aren't very clear to you, e.g: their interfaces and two-letter variable names. If we import the code, we'll miss the opportunity to throw away the parts we don't need and also explain things more clearly.

On a more positive note, Godog is MIT-licensed, so we can either import the parts we need or vendor them directly into Terratest.

brikis98:
> > I saw that, but couldn't figure out which parts were borrowed, and which are unique? Also, is it possible to use those parts directly from Godog, pulling them in as dependencies?
> 
> Yes definitely! I was unsure how much of the library I'd actually need to modify, but now I have a better sense I'm wondering if we should simply import the parts we need and only sprinkle in our custom logic? The downsides would be:
> 
> 1. We need to add additional imports to Terratest.
> 2. You've mentioned some parts of the code aren't very clear to you, e.g: their interfaces and two-letter variable names. If we import the code, we'll miss the opportunity to throw away the parts we don't need and also explain things more clearly.
> 
> On a more positive note, Godog is MIT-licensed, so we can either import the parts we need or vendor them directly into Terratest.

Adding imports is definitely preferred. In fact, this is a standalone CLI tool, so those imports shouldn't affect anyone not directly using the CLI tool.

francis-io:
Hello :wave:  

Did the idea of using Terratest in a BDD framework get off the ground beyond what's in this issue?

robmorgan:
@francis-io Hi Mark, sorry for the delayed reply. I've been away on vacation and am slowly getting caught up on everything. Short answer - No. Long answer: We've decided to prioritise other products and projects so this idea was temporarily put on hold. Also I don't have an ETA for when we'd pick it up again. You can somewhat try and use or take advantage of what I've pushed in this branch, but we wouldn't be able to offer any official support for it at this stage.

robmorgan:
Closing this PR as I've started to explore this idea independently over at https://github.com/robmorgan/infraspec

PR #732, opened 2020-12-09
Allow extra args for all helm commands: Install, Upgrade, Delete, and Rollback

### LINKED ISSUE #731: helm: Allow passing extra args to all the commands
Recent PR #727 implemented this for the Install and Upgrade commands, but not for others. We should add this support for other commands as well.

As part of this `ExtraArgs` should be refactored into a map.

For more context, see [this comment](https://github.com/gruntwork-io/terratest/pull/727#pullrequestreview-547335597).

ndhanushkodi:
This addresses @yorinasub17 's comment on my previous PR that ExtraArgs should be for every command, and implements their suggestion to have a map from subcommand to a list of args.

### GH Issue
[GH Issue 731](https://github.com/gruntwork-io/terratest/issues/731)
[GH Issue 726](https://github.com/gruntwork-io/terratest/issues/726)

### Changes Proposed
This PR allows passing arguments to the helm command when using the `Install`, `Upgrade`, `Delete`, and `Rollback` functions in the `helm` module.

### Test output
The test output can be found in this [gist](https://gist.github.com/ndhanushkodi/12e4c428b2f2d48d78decdb1eda00e67)

Closes #731

### DIFF
--- a/modules/helm/options.go
+++ b/modules/helm/options.go
@@ -15,5 +15,5 @@ type Options struct {
 	EnvVars        map[string]string   // Environment variables to set when running helm
 	Version        string              // Version of chart
 	Logger         *logger.Logger      // Set a non-default logger that should be used. See the logger package for more info.
-	ExtraArgs      []string            // Extra arguments to pass to the helm install/upgrade command
+	ExtraArgs      map[string][]string // Extra arguments to pass to the helm install/upgrade/rollback/delete command. The key signals the command (e.g., install) while the values are the extra arguments to pass through.
 }
--- a/modules/helm/delete.go
+++ b/modules/helm/delete.go
@@ -18,6 +18,11 @@ func DeleteE(t testing.TestingT, options *Options, releaseName string, purge boo
 	if !purge {
 		args = append(args, "--keep-history")
 	}
+	if options.ExtraArgs != nil {
+		if deleteArgs, ok := options.ExtraArgs["delete"]; ok {
+			args = append(args, deleteArgs...)
+		}
+	}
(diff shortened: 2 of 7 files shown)


modules/helm/options.go
```diff
@@ -15,5 +15,5 @@ type Options struct {
 	EnvVars        map[string]string   // Environment variables to set when running helm
 	Version        string              // Version of chart
 	Logger         *logger.Logger      // Set a non-default logger that should be used. See the logger package for more info.
-	ExtraArgs      []string            // Extra arguments to pass to the helm install/upgrade command
+	ExtraArgs      map[string][]string // Extra arguments to pass to the helm install/upgrade/rollback/delete command
```
yorinasub17, line 18:
```suggestion
	ExtraArgs      map[string][]string // Extra arguments to pass to the helm install/upgrade/rollback/delete command. The key signals the command (e.g., install) while the values are the extra arguments to pass through.
```

yorinasub17 (commented):
Thanks for the contribution! This LGTM! Had one minor comment suggestion. Will kick off a build shortly.

modules/helm/options.go
ndhanushkodi: Thanks for the suggestion, I made the change! Resolving this.

ndhanushkodi:
Thanks for the review @yorinasub17, I just addressed the comments, and let me know if there's anything else you'd like me to do before this is merged. Thanks!

yorinasub17 (approved):
Update LGTM and helm tests passed, so this is good to go! Will merge now. Thanks again for your contribution!

PR #733, opened 2020-12-09
adding function to extract terraform vars from var file

alpacamybags118:
This PR adds the following functionality to the terraform Options module:

* Exposes a function that allows you to get a map of all the variables contained in each terraform VarFile specified in a terraform options object. This would allow you to reference these easily later on in a test for comparison, etc.

This is my first time contributing to this repo, so I tried to follow the style guides and everything to the best of my ability. Any feedback would be greatly appreciated!


brikis98 (commented):
Thanks for the PR!

A few thoughts:

1. What is the use case you have for this? I'd like to hear concrete details to make sure we're adding support for the right thing.
1. Following from the previous question, my _guess_ is that it makes more sense to have a method to read a single value from a var file, rather than the whole var file. That way, you can parse it as a specific type (instead of `interface{}`), and use that value explicitly (whereas using the entire file seems unlikely). But again, this depends on the particular use case.
1. Please see our [error handling guidelines](https://terratest.gruntwork.io/docs/testing-best-practices/error-handling/). 
1. Please make sure to run `goimports`.

alpacamybags118:
Thanks for the feedback @brikis98 ! 

As for the use case, I have found frequently that I've wanted to access a setting I've set in a tfvars file later in a test for use in comparison for assertion/validation. For example, I was changing a module that exposes an option to set a CNAME record. I was planning to use the value we set the CNAME record to for some validation. Since I couldn't access the value from the tfvars file, I had to end up taking it out and hard-code the value in the test and pass in both the var file and that individual var into the terraform options.

So, while this functionally isn't *needed* since you can get around cases where you do need to access certain variables, you would always need to hard-code them into your test and add additional code to import both a var and a var file in your terraform options. (unless I'm totally missing another way to do this).

Overall my argument is this functionality would allow you to not have to write additional code for use cases like this and keeps configuration more separated from the code. Hopefully what I'm saying makes sense 🤣 

If you are OK with adding this, I'll go ahead and address the additional feedback you provided.

brikis98:
> Thanks for the feedback @brikis98 !
> 
> As for the use case, I have found frequently that I've wanted to access a setting I've set in a tfvars file later in a test for use in comparison for assertion/validation. For example, I was changing a module that exposes an option to set a CNAME record. I was planning to use the value we set the CNAME record to for some validation. Since I couldn't access the value from the tfvars file, I had to end up taking it out and hard-code the value in the test and pass in both the var file and that individual var into the terraform options.
> 
> So, while this functionally isn't _needed_ since you can get around cases where you do need to access certain variables, you would always need to hard-code them into your test and add additional code to import both a var and a var file in your terraform options. (unless I'm totally missing another way to do this).
> 
> Overall my argument is this functionality would allow you to not have to write additional code for use cases like this and keeps configuration more separated from the code. Hopefully what I'm saying makes sense 🤣
> 
> If you are OK with adding this, I'll go ahead and address the additional feedback you provided.

Makes sense to me!

alpacamybags118:
Crap, pushed too soon lol. Don't review yet - still need to update/add tests.

alpacamybags118:
@brikis98 happy (late) new year! I finally got a chance to come back to refactoring this PR a bit. I decided to structure it more like the output module and provide different functions for getting input file values as their different types (string, map, list), along with adding some new error types.

modules/terraform/options.go
```diff
@@ -102,3 +106,144 @@ func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options
...
+	variable, exists := variables[key]
+
+	if !exists {
+		return "", InputFileKeyNotFound(key)
+	}
+
+	return fmt.Sprintf("%v", variable), nil
+}
+
+// GetVariableAsMapFromVarFile Gets the map represention of a variable from a provided input file found in VarFile
```
brikis98, line 138: NIT: in both descriptions, it's probably worth calling out that this is specifically a map of strings. For maps with more complicated nested types, or other types in general, use `GetAllVariablesFromVarFile`.

modules/terraform/options.go
```diff
@@ -102,3 +106,144 @@ func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options
...
+	mapKeys := variables[key].([]map[string]interface{})[0]
+
+	for mapKey, mapVal := range mapKeys {
+		resultMap[mapKey] = fmt.Sprintf("%v", mapVal)
+	}
+
+	return resultMap, nil
+}
+
+// GetVariableAsListFromVarFile Gets the string list represention of a variable from a provided input file found in VarFile
```
brikis98, line 175: Same note here that this only works for lists of strings and to use `GetAllVariablesFromVarFile` for more complicated types.

modules/terraform/options.go
```diff
@@ -102,3 +106,144 @@ func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options
...
+// Will return error if GetAllVariablesFromVarFileE returns an error, the key provided does not exist, or the value associated with the key is not a list
+func GetVariableAsListFromVarFileE(t *testing.T, option *Options, fileName string, key string) ([]string, error) {
+	resultArray := []string{}
+	variables, err := GetAllVariablesFromVarFileE(t, option, fileName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if _, exists := variables[key]; !exists {
```
brikis98, line 193: Same NIT here about storing the return value instead of looking it up again several times below.

modules/terraform/options.go
```diff
@@ -102,3 +106,144 @@ func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options
...
+
+	for _, item := range variables[key].([]interface{}) {
+		resultArray = append(resultArray, fmt.Sprintf("%v", item))
+	}
+
+	return resultArray, nil
+}
+
+// GetAllVariablesFromVarFile Parses all data from a provided input file found in VarFile and returns them as a key-value map
+func GetAllVariablesFromVarFile(t *testing.T, option *Options, fileName string) map[string]interface{} {
```
brikis98, line 209: A `map[string]interface{}` is a bit unwiedly and error prone to use, as you have to cast all the values within the map. What are your thoughts instead on an interface like:

```
func GetAllVariablesFromVarFile(t *testing.T, option *Options, fileName string, returnValue interface{})
```

Where `returnValue` is a pointer the user can pass in with any custom struct, and we will parse the tfvar file into that struct, including all the types within it. This is similar to `OutputStruct`.

modules/terraform/options.go
```diff
@@ -102,3 +106,144 @@ func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options
...
+	variableMaps, err := GetAllVariablesFromVarFileE(t, option, fileName)
+	require.NoError(t, err)
+
+	return variableMaps
+}
+
+// GetAllVariablesFromVarFileE Parses all data from a provided input file found ind in VarFile and returns them as a key-value map
+// Retursn an error if the specified file does not exist, the specified file is not readable, or the specified file cannot be decoded from HCL
+func GetAllVariablesFromVarFileE(t *testing.T, option *Options, fileName string) (map[string]interface{}, error) {
```
brikis98, line 218: Why take in a `fileName` and `option` struct, and have to search for the file within it, rather than take in solely the path directly to a var file? This applies to all the methods in this PR.

brikis98 (commented):
Thanks! Left some comments/thoughts on the API.

modules/terraform/options.go
alpacamybags118: That makes sense. I think I wanted to use the options struct since I was originally thinking in just the scope of what is provided to options, but its just adding more unneeded logic. I'll adjust this.

modules/terraform/options.go
alpacamybags118: That's valid. It would certainly make it easier for the user. I did see to minimize using pointers in the best practices for this repo and this is a bit unique compared to the other functions, but it probably makes sense given the complexity of the return type.

modules/terraform/options.go
```diff
@@ -102,3 +106,144 @@ func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options
...
+	}
+
+	_, exists := variables[key]
+
+	if !exists {
+		return nil, InputFileKeyNotFound(key)
+	}
+
+	if reflect.TypeOf(variables[key]).String() != "[]map[string]interface {}" {
+		return nil, UnexpectedOutputType{Key: key, ExpectedType: "map", ActualType: reflect.TypeOf(variables[key]).String()}
```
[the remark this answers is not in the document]
alpacamybags118: Ah, good catch! Probably a bad copy-paste :P

modules/terraform/options.go
```diff
@@ -102,3 +106,144 @@ func WithDefaultRetryableErrors(t *testing.T, originalOptions *Options) *Options
...
+// Returns an error if GetAllVariablesFromVarFileE returns an error, the key provided does not exist, or the value associated with the key is not a map
+func GetVariableAsMapFromVarFileE(t *testing.T, option *Options, fileName string, key string) (map[string]string, error) {
+	resultMap := make(map[string]string)
+	variables, err := GetAllVariablesFromVarFileE(t, option, fileName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	_, exists := variables[key]
```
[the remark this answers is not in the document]
alpacamybags118: Good catch, will update.

alpacamybags118:
Thank you for the feedback! I'll get to making the updates later this week.

alpacamybags118:
@brikis98 I addressed all your feedback! I do have one thought, though, that I wanted to run by you. With these changes, it seems like maybe this code doesn't quite fit into the options file since it no longer interfaces with the options object. Maybe should be separated into its own module, like `input`? What are your thoughts?

brikis98:
> @brikis98 I addressed all your feedback! I do have one thought, though, that I wanted to run by you. With these changes, it seems like maybe this code doesn't quite fit into the options file since it no longer interfaces with the options object. Maybe should be separated into its own module, like `input`? What are your thoughts?

Sorry for the delay. Agreed it doesn't quite fit into `options.go`. Perhaps `var-files.go` or similar instead?

alpacamybags118:
Updated to move the code to a different file!

brikis98:
Great, thanks! Would you mind rebasing on master? I can then kick off tests.

alpacamybags118:
Done! @brikis98

go.mod
```diff
@@ -21,7 +21,11 @@ require (
 	github.com/google/go-containerregistry v0.0.0-20200110202235-f4fb41bf00a3
 	github.com/google/uuid v1.1.1
 	github.com/gruntwork-io/gruntwork-cli v0.7.0
+<<<<<<< HEAD
 	github.com/hashicorp/go-multierror v1.1.0
+=======
+	github.com/hashicorp/hcl v1.0.0
+>>>>>>> adding function to extract terraform vars from var file
```
brikis98, line 28: Doh, looks like a merge conflict got committed!

alpacamybags118: whoops, i blame vscode 🤣  . fixed now!

modules/terraform/var-file_test.go
```diff
@@ -0,0 +1,250 @@
...
+		t.FailNow()
+	}
+
+	defer os.Remove(randomFileName)
+
+	err = GetAllVariablesFromVarFileE(t, randomFileName, variables)
+
+	if err == nil {
+		t.FailNow()
+	}
```
brikis98, line 204: NIT: use `require.Error` instead.

modules/terraform/var-file_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+// Helper function to write a file to the filesystem
+// Will immediately fail the test if it could not write the file
+func WriteFile(t *testing.T, fileName string, bytes []byte) {
+	err := ioutil.WriteFile(fileName, bytes, 0644)
+
+	if err != nil {
+		fmt.Println(err.Error())
+		t.FailNow()
+	}
```
brikis98, line 249: This will produce a more readable error message, and is more concise:

```suggestion
	require.NoError(t, err)
```

brikis98 (commented):
Apologies, found two more NITs, and then we're good!

alpacamybags118:
Updated!

brikis98 (approved):
LGTM! I'll kick off tests now.

alpacamybags118:
Hm, looks like a test not related to my code changes failed `TestTerraformAwsRdsExample - test` . It does look like my tests passed though:

```
=== RUN   TestGetVariablesFromVarFilesAsString
--- PASS: TestGetVariablesFromVarFilesAsString (0.00s)
=== RUN   TestGetVariablesFromVarFilesAsStringKeyDoesNotExist
--- PASS: TestGetVariablesFromVarFilesAsStringKeyDoesNotExist (0.00s)
=== RUN   TestGetVariableAsMapFromVarFile
--- PASS: TestGetVariableAsMapFromVarFile (0.00s)
=== RUN   TestGetVariableAsMapFromVarFileNotMap
--- PASS: TestGetVariableAsMapFromVarFileNotMap (0.00s)
=== RUN   TestGetVariableAsMapFromVarFileKeyDoesNotExist
--- PASS: TestGetVariableAsMapFromVarFileKeyDoesNotExist (0.00s)
=== RUN   TestGetVariableAsListFromVarFile
--- PASS: TestGetVariableAsListFromVarFile (0.00s)
=== RUN   TestGetVariableAsListNotList
--- PASS: TestGetVariableAsListNotList (0.00s)
=== RUN   TestGetVariableAsListKeyDoesNotExist
--- PASS: TestGetVariableAsListKeyDoesNotExist (0.00s)
=== RUN   TestGetAllVariablesFromVarFileEFileDoesNotExist
--- PASS: TestGetAllVariablesFromVarFileEFileDoesNotExist (0.00s)
=== RUN   TestGetAllVariablesFromVarFileBadFile
--- PASS: TestGetAllVariablesFromVarFileBadFile (0.00s)
=== RUN   TestGetAllVariablesFromVarFile
--- PASS: TestGetAllVariablesFromVarFile (0.00s)
```

alpacamybags118:
Ah, looks like it might be using an instance that isn't valid? `2021-02-15T15:38:55Z logger.go:66: Error: Error creating DB Instance: InvalidParameterValue: Invalid DB Instance class: db.t2.micro` . I thought that was a valid one? https://aws.amazon.com/rds/instance-types/

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.32.6

WafflesMcDuff:
https://github.com/gruntwork-io/terratest/issues/784

PR #743, opened 2020-12-19
Check if folder exists before attempting to copy folders

### LINKED ISSUE #738: test_structure.CopyTerraformFolderToTemp - verify terraformModuleFolder exists
test_structure.CopyTerraformFolderToTemp doesn't currently validate that directories passed into the function exist.

If the terraformModuleFolder directory that is passed into `test_structure.CopyTerraformFolderToTemp` doesn't exist, the failure is handled during the `terraform init` (i.e `Error: No configuration files`).

It might be better to fail earlier if the directory doesn't exist and report a detailed error message to make debugging easier.

Let me know if this seems like a good idea and I can attempt to whip up PR one of these days!

cbuto:
This PR attempts to close #738.

I ran through the unit tests that were affected by these changes. I'm not sure if the output is super useful for those but I also ran some of the automated tests that use `CopyTerraformFolderToTemp` which look good. I was unable to run the full automated test suite right now, but I can look at doing at later if needed.

Here is the output from the automated test I did run:

https://gist.github.com/cbuto/13dd83b40d29503288b11192ee2c4149

### DIFF
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -2,7 +2,6 @@
 package files
 
 import (
-	"errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -80,7 +79,7 @@ func CopyFolderToTemp(folderPath string, tempFolderPrefix string, filter func(pa
 		return "", err
 	}
 	if !exists {
-		return "", errors.New("Provided folder does not exist: " + folderPath)
+		return "", DirNotFoundError{Directory: folderPath}
 	}
 
 	tmpDir, err := ioutil.TempDir("", tempFolderPrefix)
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -24,16 +24,6 @@ func FileExistsE(path string) (bool, error) {
 	return err == nil, nil
 }
 
-// DirExistsE returns true if the given file exists
-// It will return an error if os.Stat error is a ErrNotExist
-func DirExistsE(path string) (bool, error) {
-	_, err := os.Stat(path)
-	if err != nil && os.IsNotExist(err) {
-		return false, err
-	}
-	return err == nil, nil
-}
-
 // IsExistingFile returns true if the path exists and is a file.
 func IsExistingFile(path string) bool {
 	fileInfo, err := os.Stat(path)
@@ -84,9 +74,12 @@ func CopyTerragruntFolderToTemp(folderPath string, tempFolderPrefix string) (str
 // CopyFolderToTemp creates a copy of the given folder and all its filtered contents in a temp folder
 // with a unique name and the given prefix.
 func CopyFolderToTemp(folderPath string, tempFolderPrefix string, filter func(path string) bool) (string, error) {
-	_, err := DirExistsE(folderPath)
+	exists, err := FileExistsE(folderPath)
 	if err != nil {
 		return "", err
+	}
+	if !exists {
+		return "", os.ErrNotExist
 	}
 
 	tmpDir, err := ioutil.TempDir("", tempFolderPrefix)
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -2,6 +2,7 @@
 package files
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -79,7 +80,7 @@ func CopyFolderToTemp(folderPath string, tempFolderPrefix string, filter func(pa
 		return "", err
 	}
 	if !exists {
-		return "", os.ErrNotExist
+		return "", errors.New("Provided folder does not exist: " + folderPath)
 	}
 
 	tmpDir, err := ioutil.TempDir("", tempFolderPrefix)
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -24,6 +24,16 @@ func FileExistsE(path string) (bool, error) {
 	return err == nil, nil
 }
 
+// DirExistsE returns true if the given file exists
+// It will return an error if os.Stat error is a ErrNotExist
+func DirExistsE(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil && os.IsNotExist(err) {
+		return false, err
+	}
+	return err == nil, nil
+}
+
 // IsExistingFile returns true if the path exists and is a file.
 func IsExistingFile(path string) bool {
 	fileInfo, err := os.Stat(path)
@@ -74,6 +84,11 @@ func CopyTerragruntFolderToTemp(folderPath string, tempFolderPrefix string) (str
 // CopyFolderToTemp creates a copy of the given folder and all its filtered contents in a temp folder
 // with a unique name and the given prefix.
 func CopyFolderToTemp(folderPath string, tempFolderPrefix string, filter func(path string) bool) (string, error) {
+	_, err := DirExistsE(folderPath)
+	if err != nil {
+		return "", err
+	}
+
 	tmpDir, err := ioutil.TempDir("", tempFolderPrefix)
 	if err != nil {
 		return "", err
--- a/modules/files/files_test.go
+++ b/modules/files/files_test.go
@@ -24,21 +24,6 @@ func TestFileExists(t *testing.T) {
 	assert.False(t, FileExists("/not/a/real/path"))
 }
 
-func TestDirExistsE(t *testing.T) {
-	t.Parallel()
-
-	expectedError := new(os.PathError)
-	currentFile, err := filepath.Abs(os.Args[0])
-	require.NoError(t, err)
-
-	exists1, err := DirExistsE(currentFile)
-	assert.True(t, exists1)
-
-	exists2, err := DirExistsE("/not/a/real/path")
-	assert.False(t, exists2)
-	assert.IsType(t, expectedError, err)
-}
-
 func TestIsExistingFile(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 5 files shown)


modules/files/files.go
```diff
@@ -24,6 +24,16 @@ func FileExistsE(path string) (bool, error) {
 	return err == nil, nil
 }
 
+// DirExistsE returns true if the given file exists
```
brikis98, line 27:
```suggestion
// DirExistsE returns true if the given directory exists
```

brikis98: Actually, this method doesn't seem to check if the given `path` is a directory or not... So it seems to be identical to `FileExistsE` above?

modules/files/files.go
```diff
@@ -74,6 +84,11 @@ func CopyTerragruntFolderToTemp(folderPath string, tempFolderPrefix string) (str
 // CopyFolderToTemp creates a copy of the given folder and all its filtered contents in a temp folder
 // with a unique name and the given prefix.
 func CopyFolderToTemp(folderPath string, tempFolderPrefix string, filter func(path string) bool) (string, error) {
+	_, err := DirExistsE(folderPath)
```
brikis98, line 87: Shouldn't you be storing the boolean returned here and checking it?

modules/files/files_test.go
```diff
@@ -24,6 +24,21 @@ func TestFileExists(t *testing.T) {
...
 }
 
+func TestDirExistsE(t *testing.T) {
+	t.Parallel()
+
+	expectedError := new(os.PathError)
+	currentFile, err := filepath.Abs(os.Args[0])
+	require.NoError(t, err)
+
+	exists1, err := DirExistsE(currentFile)
```
brikis98, line 34: Will `os.Args[0]` be the path to the executable (file) or a directory? If it's a file, this check doesn't make much sense...

modules/files/files_test.go
```diff
@@ -48,6 +63,27 @@ func TestIsExistingDir(t *testing.T) {
...
+
+	currentFileDir := filepath.Dir(currentFile)
+	filter := func(path string) bool {
+		return !PathContainsHiddenFileOrFolder(path) && !PathContainsTerraformState(path)
+	}
+
+	_, err = CopyFolderToTemp("/not/a/real/path", tempFolderPrefix, filter)
+	assert.IsType(t, expectedErrorType, err)
+
+	folder, err := CopyFolderToTemp(currentFileDir, tempFolderPrefix, filter)
```
brikis98, line 82: We shouldn't copy anything to the current working dir, as that will leave the file system in a messy state. You can use `ioutil.TempDir` to create a dir in the system temp folder. 

Also, the checks in this test don't seem to work. One check is that `CopyFolderToTemp` handles in valid paths correctly, which is fine. But shouldn't the other check, the one related to this PR, be that it _doesn't_ overwrite an existing folder?

modules/test-structure/test_structure.go
```diff
@@ -72,6 +72,13 @@ func CopyTerraformFolderToTemp(t testing.TestingT, rootFolder string, terraformM
 		return filepath.Join(rootFolder, terraformModuleFolder)
 	}
 
+	fullTerraformModuleFolder := filepath.Join(rootFolder, terraformModuleFolder)
+
+	_, err := files.DirExistsE(fullTerraformModuleFolder)
+	if err != nil {
+		t.Fatal(err)
+	}
```
brikis98, line 80: Same issue here... Shouldn't you be using the bool returned?

brikis98 (commented):
Thanks for the PR!

modules/files/files.go
cbuto: yeah...😅 I removed this function and used `FileExistsE` instead. At first i created it so i could return an error for a directory not existing instead of just the bool.

modules/files/files.go
cbuto: fixed to use the bool from FileExistsE

modules/files/files_test.go
cbuto: I removed this test and method

modules/files/files_test.go
cbuto: Update to use a temp dir and also fixed the checks to cover what you described, let me know if those make more sense now!

modules/test-structure/test_structure.go
cbuto: fixed as well

modules/files/files.go
```diff
@@ -74,6 +74,14 @@ func CopyTerragruntFolderToTemp(folderPath string, tempFolderPrefix string) (str
 // CopyFolderToTemp creates a copy of the given folder and all its filtered contents in a temp folder
 // with a unique name and the given prefix.
 func CopyFolderToTemp(folderPath string, tempFolderPrefix string, filter func(path string) bool) (string, error) {
+	exists, err := FileExistsE(folderPath)
+	if err != nil {
+		return "", err
+	}
+	if !exists {
+		return "", os.ErrNotExist
```
brikis98, line 82: NIT: Could you create a new error that includes the path of the non-existent folder in the error message? That'll help quite a bit with debugging.

modules/files/files_test.go
```diff
@@ -48,6 +48,27 @@ func TestIsExistingDir(t *testing.T) {
...
+	require.NoError(t, err)
+
+	filter := func(path string) bool {
+		return !PathContainsHiddenFileOrFolder(path) && !PathContainsTerraformState(path)
+	}
+
+	folder, err := CopyFolderToTemp("/not/a/real/path", tempFolderPrefix, filter)
+	require.Error(t, err)
+	exists := FileExists(folder)
+	assert.False(t, exists)
```
brikis98, line 65: NIT:

```suggestion
	assert.False(t, FileExists(folder))
```

modules/files/files_test.go
```diff
@@ -48,6 +48,27 @@ func TestIsExistingDir(t *testing.T) {
...
+
+	folder, err = CopyFolderToTemp(tmpDir, tempFolderPrefix, filter)
+	assert.DirExists(t, folder)
+	assert.Nil(t, err)
```
brikis98, line 69:
```suggestion
	assert.NoError(t, err)
```

modules/test-structure/test_structure.go
```diff
@@ -72,6 +72,16 @@ func CopyTerraformFolderToTemp(t testing.TestingT, rootFolder string, terraformM
 		return filepath.Join(rootFolder, terraformModuleFolder)
 	}
 
+	fullTerraformModuleFolder := filepath.Join(rootFolder, terraformModuleFolder)
+
+	exists, err := files.FileExistsE(fullTerraformModuleFolder)
+	if err != nil {
+		t.Fatal(err)
+	}
```
brikis98, line 80:
```suggestion
	require.NoError(t, err)
```

modules/test-structure/test_structure.go
```diff
@@ -72,6 +72,16 @@ func CopyTerraformFolderToTemp(t testing.TestingT, rootFolder string, terraformM
...
+	if !exists {
+		t.Fatal(os.ErrNotExist, fullTerraformModuleFolder)
```
brikis98, line 82: Will this show a clear error message?

cbuto: This will result in: 

`TestTerraformAwsExamplePlan: test_structure.go:82: file does not exist ../no/examples/terraform-aws-example`

modules/files/files_test.go
cbuto: 👍  fixed

modules/test-structure/test_structure.go
cbuto: fixed!

modules/files/files_test.go
cbuto: done!

modules/files/files.go
cbuto: fixed this to include the folderPath. I wonder if I should also use that same error message [here](https://github.com/gruntwork-io/terratest/pull/743/files/5ade4a1b1766a64344ef183e05f64fd59147f71b#diff-822ca59b3639b5fce8df6421bcef822e23c19fcab3e497cfc423a9dfbb1e896a)

cbuto:
@brikis98 Thanks for taking the time to review this, I appreciate it!

modules/files/files.go
brikis98: Yes please! In fact, we typically define a new type that implements the `error` interface (has an `Error()` method), so that (a) you can check if the error you got back was of a specific type, rather than doing string parsing and (b) you can re-use that error type in more than one place.

modules/test-structure/test_structure.go
brikis98: That's decent, but would be slightly better if it said "directory" instead of file. If you extract + reuse the error type as mentioned above, that'll do the trick.

modules/files/files.go
cbuto: got it, I created `DirNotFoundError`. Let me know if that's what you were thinking!

brikis98 (dismissed):

brikis98:
Ah, doh, I forgot we have a couple random test failures, and we're working on a fix now: https://github.com/gruntwork-io/terratest/pull/756. Let's get that merged first, and then I can re-run the tests here.

cbuto:
I rebased this PR with master to pull in the test updates

brikis98:
Thanks! I'll re-run tests now.

brikis98 (approved):
Tests passed! Merging now, thank you!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.32.1

PR #754, opened 2021-01-07
Use terraform-json to have better terraform plan testing

yorinasub17:
Now that our opensource policy allows MPL, here is a PR that updates our examples to use `terraform-json`. Note that this also implements a struct that wraps [terraform-json](https://github.com/hashicorp/terraform-json) to provide a more ergonomic interface to navigate the plan struct by providing a map that maps the resource addresses to the corresponding struct values.

### DIFF
--- a/modules/terraform/plan_struct.go
+++ b/modules/terraform/plan_struct.go
@@ -9,14 +9,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// PlanStruct is a Go Struct representation of the plan object returned from Terraform. Unlike the raw plan representation
-// returned by terraform-json, this struct provides a map that maps the resource addresses to the changes and planned
-// values to make it easier to navigate the raw plan struct.
+// PlanStruct is a Go Struct representation of the plan object returned from Terraform (after running `terraform show`).
+// Unlike the raw plan representation returned by terraform-json, this struct provides a map that maps the resource
+// addresses to the changes and planned values to make it easier to navigate the raw plan struct.
 type PlanStruct struct {
+	// The raw representation of the plan. See
+	// https://www.terraform.io/docs/internals/json-format.html#plan-representation for details on the structure of the
+	// plan output.
 	RawPlan tfjson.Plan
 
+	// A map that maps full resource addresses (e.g., module.foo.null_resource.test) to the planned values of that
+	// resource.
 	ResourcePlannedValuesMap map[string]*tfjson.StateResource
-	ResourceChangesMap       map[string]*tfjson.ResourceChange
+
+	// A map that maps full resource addresses (e.g., module.foo.null_resource.test) to the planned actions terraform
+	// will take on that resource.
+	ResourceChangesMap map[string]*tfjson.ResourceChange
 }
 
 // parsePlanJson takes in the json string representation of the terraform plan and returns a go struct representation
--- a/modules/terraform/plan_struct.go
+++ b/modules/terraform/plan_struct.go
@@ -1,0 +1,108 @@
+package terraform
+
+import (
+	"encoding/json"
+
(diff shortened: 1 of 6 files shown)


modules/terraform/plan_struct.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"encoding/json"
+
+	"github.com/gruntwork-io/terratest/modules/testing"
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// PlanStruct is a Go Struct representation of the plan object returned from Terraform. Unlike the raw plan representation
```
brikis98, line 12: Is it from `plan` or from `show`?

modules/terraform/plan_struct.go
```diff
@@ -0,0 +1,108 @@
...
+// returned by terraform-json, this struct provides a map that maps the resource addresses to the changes and planned
+// values to make it easier to navigate the raw plan struct.
+type PlanStruct struct {
+	RawPlan tfjson.Plan
+
+	ResourcePlannedValuesMap map[string]*tfjson.StateResource
+	ResourceChangesMap       map[string]*tfjson.ResourceChange
```
brikis98, line 19: NIT: consider adding comments to what each of these fields represents / contains.

modules/terraform/plan_struct_test.go
```diff
@@ -0,0 +1,78 @@
...
+	basicJsonUrl      = "https://raw.githubusercontent.com/hashicorp/terraform-json/v0.8.0/testdata/basic/plan.json"
+	deepModuleJsonUrl = "https://raw.githubusercontent.com/hashicorp/terraform-json/v0.8.0/testdata/deep_module/plan.json"
+
+	changesJsonUrl = "https://raw.githubusercontent.com/hashicorp/terraform-json/v0.8.0/testdata/has_changes/plan.json"
+)
+
+func TestPlannedValuesMapWithBasicJson(t *testing.T) {
+	t.Parallel()
+
+	// Retrieve test data from the terraform-json project.
```
brikis98, line 21: Any reason not to copy the JSON files locally to avoid an external dependency in this test?

brikis98 (commented):
Nice improvement 👍

modules/terraform/plan_struct_test.go
yorinasub17: Just trying to be very conservative in how we use MPL code. I don't want to be in a situation where we have mixed licensing here, because technically, those json files will have to be MPL if they are checked in here.

modules/terraform/plan_struct.go
yorinasub17: Fixed in 77d1f0b8

modules/terraform/plan_struct.go
yorinasub17: Added comments in 77d1f0b8

modules/terraform/plan_struct_test.go
brikis98: Ah, makes sense. Perhaps mention that in a comment?

yorinasub17: Done in 57eb06dd

brikis98 (approved):
LGTM!

yorinasub17:
Going to hold off merging this until we resolve the build issues with https://github.com/gruntwork-io/terratest/pull/756

rhoboat:
Is the test failure because the instances in the ASG didn't come up, or something else?

rhoboat (approved):
LGTM!

yorinasub17:
Thanks for review. I ran the build a few times, and different tests fail each time, and they are all unrelated to this PR, so going to go ahead and merge this in.

PR #756, opened 2021-01-08
Fix tests

yorinasub17:
This fixes a consistent build failure in the `TestPackerBasicExample` and the helm tests.

Note that we still have consistent build failures with GCP, which appears to be caused by https://github.com/hashicorp/packer/issues/8939#issuecomment-604708639. ~@robmorgan can you investigate this when you get the chance?~ I am investigating this now since I just realized I have access to the underlying GCP project.


brikis98 (approved):
LGTM!

yorinasub17:
Ok I can't figure out the issue with the GCP SSH issue, so will need help from @robmorgan

robmorgan:
Looks like there might be a [relevant bug fix](https://github.com/hashicorp/packer/blob/master/CHANGELOG.md#bug-fixes) in the latest version of Packer, so I'm bumping it to 1.6.6. I'm also going to fix the way we authenticate using `gcloud`.

robmorgan:
Okay, no luck so far. For some reason, it seems something is updating the `resolv.conf` on the CircleCI VM. I have a feeling it's actually the base CircleCI Ubuntu image we are using:

```
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
nameserver 169.254.169.254
search c.circleci-picard-mahmood.internal google.internal
```

This actually allows HTTP requests like `curl http://metadata` to work. When I run the following, I get back the name of the VM CircleCI is using, e.g: `default-8d8da713-e49b-4dd9-937a-d2c3dd70970f` which makes me wonder if CircleCI is spinning up VMs on a Google Cloud host and somehow their private instance metadata is confusing our test suite:

```
$ curl -H "Metadata-Flavor: Google" http://metadata/computeMetadata/v1/instance/name
default-8d8da713-e49b-4dd9-937a-d2c3dd70970f
```

Note, if I query the `image` param, I get one of their private images:

```
$ curl -H "Metadata-Flavor: Google" http://metadata/computeMetadata/v1/instance/image
projects/circleci-picard-mahmood/global/images/ubuntu-1604-xenial-fixup-build-2072-us-east1
```

The problem with this is tools like Packer get confused and think they are actually running on a GCE instance (_which ironically they are_), then try to look up instance metadata for our project that isn't fully available. The VMs appear to be running in CircleCI's project - `circleci-picard-mahmood`.

The only workaround I can think of so far is trying to pass in an [account file](https://www.packer.io/docs/builders/googlecompute#precedence-of-authentication-methods) directly to Packer in the tests which should force Packer to use our credentials instead looking within CircleCI's GCP project. The downside to this is it likely makes our examples a little less portable as I'd need to pass in a credentials filepath somewhere.

brikis98:
Ohhhh... Nice work figuring that out Rob 👍 

> 2. Try and run GCP tests using the Docker executor, would should give better isolation.

This seems like the most promising option... We could add one more build job to run in parallel, and have it run in a Docker container. Perhaps not a ton of work, and will make the test suite a tiny bit faster and more stable?

> 3. Investigate CircleCI admin permissions to see if there are any settings around where machine executors run. I.e. can we run machine builds on AWS only?

I poked around and I'm not seeing anything that exposes where their underlying test runners run.

yorinasub17:
@robmorgan and @brikis98 can we merge what we have in here now, which fixes a few of the k8s tests? I need those fixes to sanity check https://github.com/gruntwork-io/terratest/pull/760, and I'd hate for this to block non-GCP PRs on GCP test instability.

brikis98 (dismissed):
Yup, @robmorgan could you do a separate PR for the GCP fixes?

robmorgan:
@yorinasub17 @brikis98 sure, I've removed the GCP specific fixes from this PR. I'll open another one.

brikis98 (approved):
Thx Rob!

robmorgan:
Merging, given the [build failures](https://app.circleci.com/pipelines/github/gruntwork-io/terratest/359/workflows/e2988790-f7ab-4918-ba06-445163e5ce34/jobs/4540/steps) were all GCP related.

PR #763, opened 2021-01-15
Houssem dellai patch 1

HoussemDellai:
I've added a link and image to youtube video demonstration (https://youtu.be/jeZCOqAvd0A).


yorinasub17:
Good to see some content around terratest, but unfortunately, we currently don't promote third party content on the terratest docs.

PR #767, opened 2021-01-22
Adding support to get details for TargetPrefix

ina-stoyanova:
This PR adds the functionality needed in this code changes: https://github.com/gruntwork-io/terragrunt/pull/1507

With this PR we can:
- get the `TargetPrefix` specified for a state bucket with enabled access logging

Issue linked: https://github.com/gruntwork-io/terragrunt/issues/1437

### DIFF
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -262,19 +262,27 @@ func EmptyS3BucketE(t testing.TestingT, region string, name string) error {
 }
 
 // GetS3BucketLoggingTarget fetches the given bucket's logging configuration status and returns it as a string
-func GetS3BucketLoggingTarget(t testing.TestingT, awsRegion string, bucket string) (string, string) {
-	loggingTarget, loggingObjectTargetPrefix, err := GetS3BucketLoggingTargetE(t, awsRegion, bucket)
+func GetS3BucketLoggingTarget(t testing.TestingT, awsRegion string, bucket string) string {
+	loggingTarget, err := GetS3BucketLoggingTargetE(t, awsRegion, bucket)
 	require.NoError(t, err)
 
-	return loggingTarget, loggingObjectTargetPrefix
+	return loggingTarget
+}
+
+// GetS3BucketLoggingTarget fetches the given bucket's logging configuration status and returns it as a string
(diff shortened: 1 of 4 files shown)


modules/aws/s3.go
```diff
@@ -262,33 +262,34 @@ func EmptyS3BucketE(t testing.TestingT, region string, name string) error {
...
+		return "", "", err
 	}
 
 	if res.LoggingEnabled == nil {
-		return "", S3AccessLoggingNotEnabledErr{bucket, awsRegion}
+		return "", "", S3AccessLoggingNotEnabledErr{bucket, awsRegion}
 	}
 
-	return aws.StringValue(res.LoggingEnabled.TargetBucket), nil
+	return aws.StringValue(res.LoggingEnabled.TargetBucket), aws.StringValue(res.LoggingEnabled.TargetPrefix), nil
```
brikis98, line 292: To maintain backwards compatibility, should there be a separate method to return just the target prefix? Or possibly the bucket + prefix?

infraredgirl: +1 to introduce a separate method that just retrieves the target prefix.

ina-stoyanova: That's a really good point! Thanks a lot both 🙂

ina-stoyanova:
Took me a while in the midst of support to come back here _(I thought it would require a lot more changes, but turns out it was this quick)._

Another look would be very appreciated!

modules/aws/s3.go
ina-stoyanova: So in terms of backwards compatibility - did you also mean to keep the naming of the method as it was originally? It was quite a generic name, so there's now 2 functions: `GetS3BucketLoggingTarget` & `GetS3BucketLoggingTargetPrefix`

brikis98: Yea, I think preserving the old name makes sense too. It's not an ideal name, but probably not bad enough it merits a backwards incompatible change.

ina-stoyanova: Ok, will change that too - thanks, Jim!

brikis98 (approved):
LGTM!

ina-stoyanova:
Thanks Jim!

PR #769, opened 2021-01-25
Add eu-north-1, which should be considered stable at this point

yorinasub17:


infraredgirl (approved):

infraredgirl:
The only failed test is a GCP one, so this is safe to merge!

yorinasub17:

PR #786, opened 2021-02-16
Use git branch --show-current because it works when there are no commits

rhoboat:
Needed for https://github.com/gruntwork-io/terraform-aws-architecture-catalog/pull/138

### DIFF
--- a/modules/git/git.go
+++ b/modules/git/git.go
@@ -26,17 +26,7 @@ func GetCurrentBranchNameE(t testing.TestingT) (string, error) {
 	cmd := exec.Command("git", "branch", "--show-current")
 	bytes, err := cmd.Output()
 	if err != nil {
-		cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD")
-		bytes, err := cmd.Output()
-		if err != nil {
-			return "", err
-		}
-		name := strings.TrimSpace(string(bytes))
-		if name == "HEAD" {
-			return "", nil
-		}
-
-		return name, nil
+		return GetCurrentBranchNameOldE(t)
 	}
 
 	name := strings.TrimSpace(string(bytes))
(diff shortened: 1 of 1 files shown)


modules/git/git.go
```diff
@@ -21,7 +21,7 @@ func GetCurrentBranchName(t testing.TestingT) string {
 // GetCurrentBranchNameE retrieves the current branch name or
 // empty string in case of detached state.
 func GetCurrentBranchNameE(t testing.TestingT) (string, error) {
-	cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD")
+	cmd := exec.Command("git", "branch", "--show-current")
```
yorinasub17, line 24: This option was introduced in git 2.22, which unfortunately is newer than the default that is installable with Ubuntu 18.04. So I don't think we should switch to this command completely.

Instead, we should do a fallback approach. Either:

- Check the version of git installed and use this if newer than 2.22, or fallback to the other one
- Run `git branch --show-current` and if it fails, fall back to `git rev-parse --abbrev-ref HEAD`.

rhoboat: Ah, good idea! Will change. Thanks for the review.

modules/git/git.go
```diff
@@ -20,7 +20,36 @@ func GetCurrentBranchName(t testing.TestingT) string {
...
+			return "", nil
+		}
+
+		return name, nil
+	}
+
+	name := strings.TrimSpace(string(bytes))
+	if name == "HEAD" {
+		return "", nil
+	}
```
yorinasub17, line 45: NIT: DRY with the logic above by extracting the git subprocess call into its own function.

yorinasub17 (commented):
Had one minor nit for code style, but other than that LGTM!

rhoboat:
I just force-pushed a rebase. Will take care of the nit now.

modules/git/git.go
rhoboat: Of course!

zackproser (approved):
LGTM

Issue #785, opened 2021-02-15
When using terragrunt binary state locking is explicitly disabled


dz-sourced:
We are using `terragrunt` together with Terratest for integration testing and ran into an issue where this function https://github.com/gruntwork-io/terratest/blob/88ba3f070a939aca8351eae616a3c899bfe28b91/modules/terraform/apply.go#L58 explicitly sets `-lock=false` 

It's not quite clear why this is specific to Terragrunt, but it is triggering this bug in Terraform https://github.com/hashicorp/terraform/issues/27771 when used in combination with Azure remote state backend.

Unless there is a specific reason to disable state locking when using `terragrunt` binary, I suggest leaving this options in control of the end users via `terraform.Options` 

If the above makes sense, I can create a small PR to fix this issue.

PR #788, opened 2021-02-17
Make sure Terragrunt commands can be configured with -lock=true

dz-sourced:
This fixes https://github.com/gruntwork-io/terratest/issues/785

In the current implementation, when using Terragrunt binary instead of Terraform, Terratest explicitly passes `-lock=false` to the CLI. 
This triggers a bug in Terraform (see issue for details on that), but also doesn't allow us to run integration tests with remote state
backend configured in the same way we would run actual deployment since Terraform uses `-lock=true` by default.

This change allows `Options.Lock` to be passed down to the Terragrunt `apply-all` and `destroy-all` commands. 

Default behavior in Terratest does not change, since `Options.Lock` is still initialized to `false`, but at least we can control it now.

### DIFF
--- a/modules/terraform/apply.go
+++ b/modules/terraform/apply.go
@@ -60,7 +60,7 @@ func TgApplyAllE(t testing.TestingT, options *Options) (string, error) {
 		return "", TgInvalidBinary(options.TerraformBinary)
 	}
 
-	return RunTerraformCommandE(t, options, FormatArgs(options, "apply-all", "-input=false", "-lock=false", "-auto-approve")...)
+	return RunTerraformCommandE(t, options, FormatArgs(options, "apply-all", "-input=false", "-auto-approve")...)
 }
 
 // ApplyAndIdempotent runs terraform apply with the given options and return stdout/stderr from the apply command. It then runs
--- a/modules/terraform/apply_test.go
+++ b/modules/terraform/apply_test.go
@@ -182,3 +182,21 @@ func TestParallelism(t *testing.T) {
 	duration := end.Sub(start)
 	require.Greater(t, int64(duration.Seconds()), int64(25))
 }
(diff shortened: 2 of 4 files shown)


brikis98 (commented):
Sanity check: it doesn't look like you changed anything in `RunTerraformCommandE`... So is the fix that `-lock=false` is no longer hard-coded and is now read from `options` instead? Or is the fix that `apply-all` is part of `TerraformCommandsWithLockSupport` now?

dz-sourced:
@brikis98 Yes, the fix is that `-lock=false` is no longer hardcoded and can be supplied via `options`, but for this to work with Terragrunt `apply-all` and `destroy-all` commands need to be added to the `TerraformCommandsWithLockSupport` because of this check here https://github.com/gruntwork-io/terratest/blob/2114c229b43e043b692d60893545de878342f836/modules/terraform/format.go#L39

brikis98 (dismissed):
Ah, got it, thanks for explaining! Changes LGTM. I'll kick off tests now.

dz-sourced:
@brikis98  I see that a couple of tests failed, but they don't seem to be related to my PR. Is there anything I need to do on my end to sort this out?

brikis98:
We just merged some test stability fixes. Could you pull in the latest from `master` and I'll re-run tests?

dz-sourced:
@brikis98 I have merged the latest from the upstream `master`. Should be good to go for another tests run.

brikis98:
Thanks! I just kicked off the tests again.

brikis98:
There was one test failure, but it looks unrelated/transient, so I'm going to merge.

brikis98 (approved):

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.32.16

Issue #784, opened 2021-02-15
modules/terraform/var-file.go:142:8: undefined: hcl.Decode


WafflesMcDuff:
Latest release is causing an error:
```
vendor/github.com/gruntwork-io/terratest/modules/terraform/var-file.go:142:8: undefined: hcl.Decode
FAIL	git.*******.com/mongodbatlas-tf/terratest/mongodbatlas-cluster [build failed]
FAIL

Exited with code exit status 2
```

https://github.com/gruntwork-io/terratest/blob/88ba3f070a939aca8351eae616a3c899bfe28b91/modules/terraform/var-file.go#L142

WafflesMcDuff:
@brikis98
@alpacamybags118

WafflesMcDuff:
Problem started with #733

WafflesMcDuff:
Apparently this version breaks compatibility with go dep.
Had to switch from
`dep init && dep ensure -vendor-only `
to
`go mod init git.rockfin.com/mongodbtatlas-tf && go mod vendor`

alpacamybags118:
I think the root of the issue is dep is trying to pull v2 of hcl while the go mod of the terratest library explicitly referenes v1, so dep isnt respecting the libraries required version

bgarcial:
@alpacamybags118, @WafflesMcDuff I got the same error when running `go get -u github.com/gruntwork-io/terratest/modules/terraform` in our pipeline. I was wondering how can I use the [previous release](https://github.com/gruntwork-io/terratest/releases/tag/v0.32.5)in the meantime. I will appreciate your piece of advise.

avishnyakov:
Confirming, having same problem with the recent changes. Testing, trying to find a workaround.

getSurreal:
As WafflesMcDuff mentioned, use

`go mod init`

instead of

`dep init`

yorinasub17:
This is indeed a bug as we should be using hclv2. Will work on updating the library function to use hcl v2 asap.

jnovick:
This issue is affecting me as well. Part of the pipeline we wrote ensures we always use the latest terratest version so it has been broken and preventing any tests from actually running.

yorinasub17:
Potential fix for this was released https://github.com/gruntwork-io/terratest/releases/tag/v0.32.8

Please try it out and see if it resolves this.

jnovick:
[...]

PR #790, opened 2021-02-18
Update to hcl2

yorinasub17:
Fixes https://github.com/gruntwork-io/terratest/issues/784

This updates the `var-file.go` routines to use hclv2 instead of hclv1. Note that this is a __backwards incompatible change__ due to the change to make `GetAllVariablesFromVarFileE` more ergonomic for both the user and code maintenance.


modules/terraform/var-file.go
```diff
@@ -117,33 +102,135 @@ func GetVariableAsListFromVarFileE(t *testing.T, fileName string, key string) ([
...
 
 	return resultArray, nil
 }
 
-// GetAllVariablesFromVarFile Parses all data from a provided input file found in VarFile and stores the result in the value pointed to by out
-func GetAllVariablesFromVarFile(t *testing.T, fileName string, out interface{}) {
-	err := GetAllVariablesFromVarFileE(t, fileName, out)
+// GetAllVariablesFromVarFileE Parses all data from a provided input file found in VarFile and returns a map of the
+// variable values where the keys are the var names and the value are the parsed var values.
+func GetAllVariablesFromVarFile(t *testing.T, fileName string) map[string]interface{} {
```
brikis98, line 115: Hm, this seems like a less ergonomic interface. In the old version, you could pass in a custom fully-typed struct, and have the HCL parsed into that. Now, your only option is an untyped map...

modules/terraform/var-file.go
```diff
@@ -117,33 +102,135 @@ func GetVariableAsListFromVarFileE(t *testing.T, fileName string, key string) ([
...
+		return nil, err
+	}
+	return out, nil
+}
 
+// parseAndDecodeHcl uses the HCL2 parser to parse the given varfile string into an HCL file body, and then decode it
+// into a map that maps var names to values.
+func parseAndDecodeVarFile(hclContents string, filename string) (out map[string]interface{}, err error) {
+	// The HCL2 parser and especially cty conversions will panic in many types of errors, so we have to recover from
+	// those panics here and convert them to normal errors
```
brikis98, line 141: Bummer, that's unfortunate

modules/terraform/var-file.go
```diff
@@ -117,33 +102,135 @@ func GetVariableAsListFromVarFileE(t *testing.T, fileName string, key string) ([
...
+// The following functions were lifted from terragrunt -
+// https://github.com/gruntwork-io/terragrunt/blob/874d2db4618afc7dc8e153ab0219282147865d4a/config/cty_helpers.go
 
-	err = hcl.Decode(out, string(fileContents))
+// This is a hacky workaround to convert a cty Value map to a Go map[string]interface{}. cty does not support this directly
+// (https://github.com/hashicorp/hcl2/issues/108) and doing it with gocty.FromCtyValue is nearly impossible, as cty
+// requires you to specify all the output types and will error out when it hits interface{}. So, as an ugly workaround,
+// we convert the given value to JSON using cty's JSON library and then convert the JSON back to a
+// map[string]interface{} using the Go json library.
+func parseCtyValueMapToGoMap(valMap map[string]cty.Value) (map[string]interface{}, error) {
```
brikis98, line 184: Ugh, I'm not a fan of perpetuating this hack. So the `hcl2` lib doesn't have the ability to parse data into arbitrary structs passed in by the user? 

Also, did this function in Terragrunt have unit tests? If so, we prob need to port those over too.

modules/terraform/var-file.go
yorinasub17: I'll see what I can do, but as you know, the HCL2 parser is significantly less ergonomical when it comes to transforming to structs so not sure if we can make it do that.

yorinasub17: In the worst case, we might have to ship this version first, and then improve later. I don't think it's ideal that these functions are using the HCL1 parser since that means they do not support HCL2 syntax and that can be a problem for any var-file implemented for tf12+.

modules/terraform/var-file.go
yorinasub17: The `hcl2` library supports parsing the data to arbitrary structs _except map[string]interface{}_. See the new implementation in f8477eb, where I special case the arbitrary map.

> Also, did this function in Terragrunt have unit tests? If so, we prob need to port those over too.

It looks like this function was implicitly tested in terragrunt. I think the implicit tests we have in here should be sufficient.

brikis98 (approved):
LGTM, thanks Yori!

modules/terraform/var-file.go
brikis98: Roger!

yorinasub17:
Thanks for review! Test passed (except for GCP, which is not relevant to this PR), so going to merge this in now.

PR #794, opened 2021-02-24
Implement Terraform Workspace deletion.

### LINKED ISSUE #583: Function for deleting workspace
https://github.com/gruntwork-io/terratest/pull/164

Function for removing workspace and returning to previous or default.

dduportal:
Hello there 👋 Thanks for this awesoe tool.

This PR fixes #583 : it implements 2 methods to handle Terraform Workspace's deletion.

Please note that deleting the workspace `default` is not supported as it triggers edgy use cases.

### DIFF
--- a/modules/terraform/workspace.go
+++ b/modules/terraform/workspace.go
@@ -60,3 +60,49 @@ func nameMatchesWorkspace(name string, workspace string) bool {
 	match, _ := regexp.MatchString(fmt.Sprintf("^\\*?\\s*%s$", name), workspace)
 	return match
 }
+
+// WorkspaceDelete removes the specified terraform workspace with the given options.
+// It returns the name of the current workspace AFTER deletion, and the returned error (that can be nil).
+// If the workspace to delete is the current one, then it tries to switch to the "default" workspace.
+// Deleting the workspace "default" is not supported.
+func WorkspaceDeleteE(t testing.TestingT, options *Options, name string) (string, error) {
+	currentWorkspace := RunTerraformCommand(t, options, "workspace", "show")
+
+	if name == "default" {
+		return currentWorkspace, fmt.Errorf("Deleting the workspace 'default' is not supported")
+	}
+
+	out, err := RunTerraformCommandE(t, options, "workspace", "list")
+	if err != nil {
+		return currentWorkspace, err
+	}
+	if !isExistingWorkspace(out, name) {
+		return currentWorkspace, fmt.Errorf("The workspace %q does not exist.", name)
+	}
+
+	// Switch workspace before deleting if it is the current
+	if currentWorkspace == name {
+		currentWorkspace = WorkspaceSelectOrNew(t, options, "default")
+	}
+
+	// delete workspace
+	_, err = RunTerraformCommandE(t, options, "workspace", "delete", name)
+
+	return currentWorkspace, err
+}
+
+// WorkspaceDelete removes the specified terraform workspace with the given options.
+// It returns the name of the current workspace AFTER deletion.
+// If the workspace to delete is the current one, then it tries to switch to the "default" workspace.
+// Deleting the workspace "default" is not supported and only return an empty string (to avoid a fatal error).
+func WorkspaceDelete(t testing.TestingT, options *Options, name string) string {
+	if name == "default" {
+		return name
+	}
+
+	out, err := WorkspaceDeleteE(t, options, name)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return out
+}
--- a/modules/terraform/workspace.go
+++ b/modules/terraform/workspace.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
 )
 
 // WorkspaceSelectOrNew runs terraform workspace with the given options and the workspace name
@@ -66,10 +67,13 @@ func nameMatchesWorkspace(name string, workspace string) bool {
 // If the workspace to delete is the current one, then it tries to switch to the "default" workspace.
 // Deleting the workspace "default" is not supported.
 func WorkspaceDeleteE(t testing.TestingT, options *Options, name string) (string, error) {
-	currentWorkspace := RunTerraformCommand(t, options, "workspace", "show")
+	currentWorkspace, err := RunTerraformCommandE(t, options, "workspace", "show")
+	if err != nil {
+		return currentWorkspace, err
+	}
 
 	if name == "default" {
-		return currentWorkspace, fmt.Errorf("Deleting the workspace 'default' is not supported")
+		return currentWorkspace, &UnsupportedDefaultWorkspaceDeletion{}
 	}
 
 	out, err := RunTerraformCommandE(t, options, "workspace", "list")
@@ -77,12 +81,15 @@ func WorkspaceDeleteE(t testing.TestingT, options *Options, name string) (string
 		return currentWorkspace, err
 	}
 	if !isExistingWorkspace(out, name) {
-		return currentWorkspace, fmt.Errorf("The workspace %q does not exist.", name)
+		return currentWorkspace, WorkspaceDoesNotExist(name)
 	}
 
 	// Switch workspace before deleting if it is the current
 	if currentWorkspace == name {
-		currentWorkspace = WorkspaceSelectOrNew(t, options, "default")
+		currentWorkspace, err = WorkspaceSelectOrNewE(t, options, "default")
+		if err != nil {
+			return currentWorkspace, err
+		}
 	}
 
 	// delete workspace
@@ -101,8 +108,6 @@ func WorkspaceDelete(t testing.TestingT, options *Options, name string) string {
 	}
 
 	out, err := WorkspaceDeleteE(t, options, name)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	return out
 }
--- a/modules/terraform/workspace.go
+++ b/modules/terraform/workspace.go
@@ -103,10 +103,6 @@ func WorkspaceDeleteE(t testing.TestingT, options *Options, name string) (string
 // If the workspace to delete is the current one, then it tries to switch to the "default" workspace.
 // Deleting the workspace "default" is not supported and only return an empty string (to avoid a fatal error).
 func WorkspaceDelete(t testing.TestingT, options *Options, name string) string {
-	if name == "default" {
-		return name
-	}
-
 	out, err := WorkspaceDeleteE(t, options, name)
 	require.NoError(t, err)
 	return out
--- a/modules/terraform/workspace_test.go
+++ b/modules/terraform/workspace_test.go
@@ -1,6 +1,7 @@
 package terraform
 
 import (
+	"errors"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


modules/terraform/workspace.go
```diff
@@ -60,3 +60,49 @@ func nameMatchesWorkspace(name string, workspace string) bool {
 	match, _ := regexp.MatchString(fmt.Sprintf("^\\*?\\s*%s$", name), workspace)
 	return match
 }
+
+// WorkspaceDelete removes the specified terraform workspace with the given options.
+// It returns the name of the current workspace AFTER deletion, and the returned error (that can be nil).
+// If the workspace to delete is the current one, then it tries to switch to the "default" workspace.
+// Deleting the workspace "default" is not supported.
+func WorkspaceDeleteE(t testing.TestingT, options *Options, name string) (string, error) {
+	currentWorkspace := RunTerraformCommand(t, options, "workspace", "show")
```
brikis98, line 69: Any `E` function must use other `E` functions underneath. So here, you should call `RunTerraformCommandE` and return any error that function may return.

modules/terraform/workspace.go
```diff
@@ -60,3 +60,49 @@ func nameMatchesWorkspace(name string, workspace string) bool {
...
+
+	if name == "default" {
+		return currentWorkspace, fmt.Errorf("Deleting the workspace 'default' is not supported")
```
brikis98, line 72: Please define a custom error type for this instead of using `fmt.Errorf`. That makes it possible to check returned errors by type, rather than string messages that may change.

brikis98: This applies here and to all errors below.

modules/terraform/workspace.go
```diff
@@ -60,3 +60,49 @@ func nameMatchesWorkspace(name string, workspace string) bool {
...
+	if err != nil {
+		return currentWorkspace, err
+	}
+	if !isExistingWorkspace(out, name) {
+		return currentWorkspace, fmt.Errorf("The workspace %q does not exist.", name)
+	}
+
+	// Switch workspace before deleting if it is the current
+	if currentWorkspace == name {
+		currentWorkspace = WorkspaceSelectOrNew(t, options, "default")
```
brikis98, line 85: Should use `E` version

modules/terraform/workspace.go
```diff
@@ -60,3 +60,49 @@ func nameMatchesWorkspace(name string, workspace string) bool {
...
+}
+
+// WorkspaceDelete removes the specified terraform workspace with the given options.
+// It returns the name of the current workspace AFTER deletion.
+// If the workspace to delete is the current one, then it tries to switch to the "default" workspace.
+// Deleting the workspace "default" is not supported and only return an empty string (to avoid a fatal error).
+func WorkspaceDelete(t testing.TestingT, options *Options, name string) string {
+	if name == "default" {
+		return name
+	}
```
brikis98, line 101: Why does this method differ from the `E` version with regards to the default workspace?

modules/terraform/workspace.go
```diff
@@ -60,3 +60,49 @@ func nameMatchesWorkspace(name string, workspace string) bool {
...
+
+	out, err := WorkspaceDeleteE(t, options, name)
+	if err != nil {
+		t.Fatal(err)
+	}
```
brikis98, line 106:
```suggestion
	require.NoError(t, err)
```

modules/terraform/workspace_test.go
```diff
@@ -130,3 +130,177 @@ func TestNameMatchesWorkspace(t *testing.T) {
 		assert.Equal(t, testCase.expected, actual, "Name: %q, Workspace: %q", testCase.name, testCase.workspace)
 	}
 }
+
+// Please note that this test depends on other functions that should be mocked to be a unit test.
```
brikis98, line 134: Not sure I follow. What other functions?

modules/terraform/workspace_test.go
```diff
@@ -130,3 +130,177 @@ func TestNameMatchesWorkspace(t *testing.T) {
...
+	// testCase describes a named test case with a state, args and expcted results
+	type testCase struct {
+		name                 string
+		initialState         state
+		toDeleteWorkspace    string
+		expectedCurrent      string
+		expectedErrorMessage string
+	}
+
+	testCases := []testCase{
```
brikis98, line 153: Nice tests 👍

modules/terraform/workspace_test.go
```diff
@@ -130,3 +130,177 @@ func TestNameMatchesWorkspace(t *testing.T) {
...
+				current:    "staging",
+			},
+			toDeleteWorkspace:    "default",
+			expectedCurrent:      "staging",
+			expectedErrorMessage: "Deleting the workspace 'default' is not supported",
+		},
+	}
+
+	for _, tt := range testCases {
+		testFolder, err := files.CopyTerraformFolderToTemp("../../test/fixtures/terraform-workspace", tt.name)
```
brikis98, line 197: These should all be done within a `t.Run` so each test case can be executed separately.

modules/terraform/workspace_test.go
```diff
@@ -130,3 +130,177 @@ func TestNameMatchesWorkspace(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
```
brikis98, line 200: Use `require.NoError(t, err)` here and in all similar blocks below.

modules/terraform/workspace_test.go
```diff
@@ -130,3 +130,177 @@ func TestNameMatchesWorkspace(t *testing.T) {
...
+			// Check for results
+			assert.Equal(t, tt.expectedCurrent, gotResult)
+			assert.False(t, isExistingWorkspace(RunTerraformCommand(t, options, "workspace", "list"), tt.toDeleteWorkspace))
+		}
+
+	}
+}
+
+// Please note that this test depends on other functions that should be mocked to be a unit test.
+func TestWorkspaceDelete(t *testing.T) {
```
brikis98, line 237: Testing just one of the `E` or non `E` is usually enough...

brikis98 (commented):
Thanks for the PR!

modules/terraform/workspace_test.go
dduportal: The implementation utilizes the method `RunTerraformCommandE`.

If the test is expected to be "unit test", then such a command should be mocked to avoid "chained failure".

But I feel like the test harness is not really unit test, but more integration/functional and does not require such level of detail (one might say that going to the "mock" might be too much fragile code).

If it is ok, I'm removing the comment here has it brings more confusion than it helps.

modules/terraform/workspace.go
dduportal: I'm not sure about the expected behavior of this method as it does not output. Could I have your advise?

If consumer calls `terraform.WorkspaceDelete(t, opts, "default")`:

* Should it return the string `default`, and the consumer should take care of comparing it to the input parameter? Or should it return an empty string to mark the "edge case"?

dduportal:
Hello @brikis98 , thanks a lot for the review.

I've updated the PR with a bunch of your feedbacks, but I need inputs on some.

Please note that I haven't squashed the commits yet, in order to allow further update or rollback as necessary.

modules/terraform/workspace.go
brikis98: Shouldn't it just call `WorkspaceDeleteE` directly and return/use whatever that function returns?

modules/terraform/workspace_test.go
```diff
@@ -130,3 +131,99 @@ func TestNameMatchesWorkspace(t *testing.T) {
...
+				current:    "staging",
+			},
+			toDeleteWorkspace: "default",
+			expectedCurrent:   "staging",
+			expectedError:     &UnsupportedDefaultWorkspaceDeletion{},
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
```
brikis98, line 197: Add `t.Parallel()` here.

modules/terraform/workspace_test.go
```diff
@@ -130,3 +131,99 @@ func TestNameMatchesWorkspace(t *testing.T) {
...
+				workspaces: []string{"staging", "production"},
+				current:    "staging",
+			},
+			toDeleteWorkspace: "default",
+			expectedCurrent:   "staging",
+			expectedError:     &UnsupportedDefaultWorkspaceDeletion{},
+		},
+	}
+
+	for _, tt := range testCases {
```
brikis98, line 196: Two notes:

1. I mention below calling `t.Parallel()`. That means you'll need to store a copy of `tt` here via a silly looking `tt := tt`. See https://github.com/gruntwork-io/terratest/blob/0c883d74782fecaf8e5987a760f5aa9e06a04a7f/modules/aws/ec2_test.go#L52-L54 for an example.
1. NIT: We typically avoid one or two letter variable names, except in very idiomatic cases (e.g., `i` in a for loop or `t` in a Go test case). So `testCase` may be a better name for this variable.

modules/terraform/workspace_test.go
```diff
@@ -130,3 +131,99 @@ func TestNameMatchesWorkspace(t *testing.T) {
...
+			require.NoError(t, err)
+
+			// Testing time, wooohoooo
+			gotResult, gotErr := WorkspaceDeleteE(t, options, tt.toDeleteWorkspace)
+
+			// Check for errors
+			if tt.expectedError != nil {
+				assert.EqualError(t, gotErr, tt.expectedError.Error())
+			} else {
+				assert.Nil(t, gotErr)
```
brikis98, line 221:
```suggestion
				assert.NoError(t, gotErr)
```

brikis98 (commented):
Almost there!

modules/terraform/workspace.go
brikis98: There are multiple ways to do it. The answers in [this SO thread](https://stackoverflow.com/questions/23796543/go-checking-for-the-type-of-a-custom-error) provide a good list of the options.

modules/terraform/workspace_test.go
dduportal: Wow, I'm impressed by the performance induced by adding this option (went to ~8s to ~2s on my machine).

modules/terraform/workspace.go
brikis98: Looks like this never got resolved?

dduportal: I'm sorry, I missed the comment with the folded reviews. Let me check and fix (and I'll check all the others feedbacks to ensure I did not miss another). Thanks for your eagle eyes!

dduportal: You are correct: this code should returns whatever `WorkspaceDeleteE` returns. I don't know why I over-complicated it 😅

modules/terraform/workspace.go
dduportal: Missed this one. Updating to test on the error type with `assert.True(t, errors.As(gotErr, &testCase.expectedError))`

dduportal:
Ok, I've checked all your feedbacks, it should be ok (also rebased from upstream). Let me know if I missed something else.
Thanks for all the feedbacks and the review!

brikis98 (approved):
Great, thank you! Code LGTM now. I'll kick off tests.

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.32.17

Issue #792, opened 2021-02-21
Build targets in docker module as a separate BuildOption


ThelonKarrde:
I want to suggest putting the `target` parameter of the `docker build` command as a separate parameter in `BuildOptions`.
I understand and see the note that `terratest` does not support most of the build parameters, but since `target` allows you to do truly unit testing for the Dockerfile. I know that you can pass it as a parameter to `OtherOptions`, but I think put it as a separate argument will be very handy.

I can implement it on my own, but it shows in your contribution guide - let me know what you think about it :)

brikis98:
Seems reasonable. PR welcome!

PR #798, opened 2021-02-28
Build target flag for docker module

ThelonKarrde:
This PR is an implementation for the issue: https://github.com/gruntwork-io/terratest/issues/792
Adding an optional `--target` flag for build options. A bit modified `Dockerfile` used for tests (does not break reverse-compatibility for other tests) and added an additional test to make sure that flag works.

Gist with tests output within `modules/docker` path: https://gist.github.com/ThelonKarrde/9eab7f3172edd8c71032e690618764e1

### DIFF
--- a/modules/docker/build.go
+++ b/modules/docker/build.go
@@ -15,6 +15,9 @@ type BuildOptions struct {
 	// Build args to pass the 'docker build' command
 	BuildArgs []string
 
+	// Target build arg to pass to the 'docker build' command
+	Target string
+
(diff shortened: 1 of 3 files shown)


brikis98 (approved):
LGTM, thanks! I'll kick off the tests now.

brikis98:
There were some unrelated, transient test failures; we'll have to deal with those separately. The relevant tests passed, so merging!

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.32.9

Issue #800, opened 2021-03-04
The GCP example in the terratest page points to the wrong file.


kanekotic:
The GCP example on the [live page](https://terratest.gruntwork.io/) points to the wrong url.
- Current Link: github.com/gruntwork-io/terratest/test/terraform_gcp_hello_world_example_test.go
- Correct Link:  github.com/gruntwork-io/terratest/test/gcp/terraform_gcp_hello_world_example_test.go

![Screenshot 2021-03-04 at 17 47 37](https://user-images.githubusercontent.com/3071208/109998647-5a69e880-7d09-11eb-9357-dda48177bda5.png)

kanekotic:
added a PR to solve this :)

kanekotic:
thanks @brikis98 I will close at its now live 😄

PR #801, opened 2021-03-04
#800 | Fix link to gcp example

kanekotic:
Fix link that shows the example for GCP in the terratest page.


brikis98 (approved):
LGTM, thanks for the fix!

PR #804, opened 2021-03-10
Add replicaset helper functions

### LINKED ISSUE #793: Feature Request: Helper functions for Kubernetes ReplicaSets
I'm not sure if this would pass the litmus tests defined in the contributor guide but I think it might be something worthwhile supporting in Terratest in terms of popularity/completeness.

We have tests implemented currently that rely on the outcome of creating deployments and assertions based on the associated ReplicaSet events. 

This would look much the daemonset implementation.

Let me know if this is something that sounds reasonable!

P.S Terratest has significantly improved our existing testing strategy, so I really appreciate this project and support from the maintainers!

cbuto:
Closes #793.

Thanks ahead of time for reviewing!

### DIFF
--- a/modules/k8s/replicaset_test.go
+++ b/modules/k8s/replicaset_test.go
@@ -1,0 +1,87 @@
+// +build kubeall kubernetes
+
+// NOTE: we have build tags to differentiate kubernetes tests from non-kubernetes tests. This is done because minikube
+// is heavy and can interfere with docker related tests in terratest. Specifically, many of the tests start to fail with
+// `connection refused` errors from `minikube`. To avoid overloading the system, we run the kubernetes tests and helm
+// tests separately from the others. This may not be necessary if you have a sufficiently powerful machine.  We
+// recommend at least 4 cores and 16GB of RAM if you want to run all the tests together.
+
+package k8s
+
+import (
+	"fmt"
+
+	"strings"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


modules/k8s/replicaset_test.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func TestGetReplicaSets(t *testing.T) {
+	t.Parallel()
+
+	uniqueID := strings.ToLower(random.UniqueId())
+	options := NewKubectlOptions("", "", uniqueID)
+	configData := fmt.Sprintf(EXAMPLE_REPLICASET_YAML_TEMPLATE, uniqueID, uniqueID)
+	KubectlApplyFromString(t, options, configData)
+	defer KubectlDeleteFromString(t, options, configData)
```
yorinasub17, line 37: NIT: the delete defer should be before the apply, so that it always runs even if the apply fails with error (which could partially deploy things).

yorinasub17 (commented):
Thanks for your contribution! LGTM other than the one nit about when to `defer`. I think we can merge this in once that is addressed!

Separately, I kicked off a build so we get one test cycle in the meantime.

yorinasub17 (approved):
FYI The kubernetes tests passed. Since the tests passed, and the issue mentioned above is very minor, I'll go ahead and merge this in (but would appreciate if you can address the above comment in a follow up PR!).

cbuto:
> FYI The kubernetes tests passed. Since the tests passed, and the issue mentioned above is very minor, I'll go ahead and merge this in (but would appreciate if you can address the above comment in a follow up PR!).

Awesome thanks! I'll put up PR for that fix here in a second

PR #815, opened 2021-03-17
Move defer before the apply

cbuto:
Follow up work for https://github.com/gruntwork-io/terratest/pull/804#pullrequestreview-613965137

thanks!


yorinasub17 (approved):
Thanks! The relevant tests passed, so will merge this in!

Issue #802, opened 2021-03-04
FeatureRequest:  Allow InvocationType argument to InvokeFunction()


kbalk:
## What I would like to do:
Add an argument to `InvokeFunction()` that can be used as the InvocationType argument to AWS SDK's `Invoke()`.

From the AWS CLI command reference (https://docs.aws.amazon.com/cli/latest/reference/lambda/invoke.html):

> invoke
> --function-name <value>
> [--invocation-type <value>]    <-----
> [--log-type <value>]
> [--client-context <value>]
> [--payload <value>]
> [--qualifier <value>]
> \<outfile\>

https://github.com/gruntwork-io/terratest/blob/8361e81e718c421388f76021603746485c972f30/modules/aws/lambda.go#L13    

## Why?

An InvocationType argument of "DryRun" provides a simple verification of a Lambda's installation.  It's also useful where it's difficult to generate the appropriate payload to test a lambda's invocation.

brikis98:
PR to add this, perhaps in a new function, is welcome!

brikis98:
Fixed in #817 and released in https://github.com/gruntwork-io/terratest/releases/tag/v0.32.21.

PR #817, opened 2021-03-23
Add function to invoke a lambda using an invocation type of "DryRun"

kbalk:
Provides the functionality to invoke a Lambda with an InvocationType of "DryRun".  This functionality provides a simple verification of a Lambda's installation.

This functionality was implemented through the new function `InvokeDryRunE()`, which returns a status code and an error.  I debated about the return values, but a status code and error seemed consistent with the return values from the AWS SDK.  When the the AWS SDK `Invoke()` is called with a invocation type of DryRun, `StatusCode` is the only field in the returned `InvokeOutput` struct that has a value, even when there is an error.  A `StatusCode` value of 204 indicates the DryRun was successful and further details are then found in the returned error:

>  // The HTTP status code is in the 200 range for a successful request. For the
    // RequestResponse invocation type, this status code is 200. For the Event invocation
    // type, this status code is 202. For the DryRun invocation type, the status
    // code is 204.
 StatusCode *int64 `location:"statusCode" type:"integer"`

I also debated whether there was a value to creating a `InvokeDryRun()` (no "E" at the end) version of the function to simply return the status code and no error.  I assumed if the dry run failed, the error was needed, so `InvokeDryRun()` wouldn't have much utility.

And I debated adding a test of an error condition.  The dry run can determine whether the user or role has permission to execute the Lambda, which would be more involved to set up.  I could specify a bad function name, but that's not really the point of a dry run invocation.

This pull request is a bit of a trial balloon to determine just what you'd like to see for this added functionality.

Issue:  [#802](https://github.com/gruntwork-io/terratest/issues/802) 
Test results:  https://gist.github.com/kbalk/fbb68f156d391eacf4d330b7870e1401

### DIFF
--- a/modules/aws/lambda.go
+++ b/modules/aws/lambda.go
@@ -53,10 +53,6 @@ type LambdaOptions struct {
 // fields may or may not have a value depending on the invocation type and
 // whether an error occurred or not.
 type LambdaOutput struct {
-	// If present, indicates that an error occurred during function execution.
-	// Error details are included in the response payload.
-	FunctionError *string
-
 	// The response from the function, or an error object.
 	Payload []byte
 
@@ -75,8 +71,26 @@ func InvokeFunction(t testing.TestingT, region, functionName string, payload int
 
 // InvokeFunctionE invokes a lambda function.
 func InvokeFunctionE(t testing.TestingT, region, functionName string, payload interface{}) ([]byte, error) {
-	input := &LambdaOptions{Payload: &payload}
-	out, err := InvokeFunctionWithParamsE(t, region, functionName, input)
+	lambdaClient, err := NewLambdaClientE(t, region)
+	if err != nil {
+		return nil, err
+	}
+
+	invokeInput := &lambda.InvokeInput{
+		FunctionName: &functionName,
+	}
+
+	if payload != nil {
+		payloadJson, err := json.Marshal(payload)
+
+		if err != nil {
+			return nil, err
+		}
+		invokeInput.Payload = payloadJson
+	}
+
+	out, err := lambdaClient.Invoke(invokeInput)
+	require.NoError(t, err)
 	if err != nil {
 		return nil, err
 	}
@@ -98,8 +112,10 @@ func InvokeFunctionWithParams(t testing.TestingT, region, functionName string, i
 }
 
 // InvokeFunctionWithParamsE invokes a lambda function using parameters
-// supplied in the LambdaOptions struct and returns values in a LambdaOutput
-// struct and the error.
+// supplied in the LambdaOptions struct.  Returns the status code and payload
+// in a LambdaOutput struct and the error.  A non-nil error will either reflect
+// a problem with the parameters supplied to this function or an error returned
+// by the Lambda.
 func InvokeFunctionWithParamsE(t testing.TestingT, region, functionName string, input *LambdaOptions) (*LambdaOutput, error) {
 	lambdaClient, err := NewLambdaClientE(t, region)
 	if err != nil {
@@ -111,8 +127,7 @@ func InvokeFunctionWithParamsE(t testing.TestingT, region, functionName string, 
 	// "RequestResponse".
 	invocationType, err := input.InvocationType.Value()
 	if err != nil {
-		msg := err.Error()
-		return &LambdaOutput{FunctionError: &msg}, err
+		return nil, err
 	}
 
 	invokeInput := &lambda.InvokeInput{
@@ -129,15 +144,23 @@ func InvokeFunctionWithParamsE(t testing.TestingT, region, functionName string, 
 	}
 
 	out, err := lambdaClient.Invoke(invokeInput)
-
-	// As this function supports different invocation types, so it must
-	// support different combinations of output.
-	lambdaOutput := LambdaOutput{
-		FunctionError: out.FunctionError,
-		Payload:       out.Payload,
-		StatusCode:    out.StatusCode,
+	if err != nil {
+		return nil, err
 	}
-	return &lambdaOutput, err
+
+	// As this function supports different invocation types, it must
+	// then support different combinations of output other than just
+	// payload.
+	lambdaOutput := LambdaOutput{
+		Payload:    out.Payload,
+		StatusCode: out.StatusCode,
+	}
+
+	if out.FunctionError != nil {
+		return &lambdaOutput, errors.New(*out.FunctionError)
+	}
+
+	return &lambdaOutput, nil
 }
 
 type FunctionError struct {
--- a/modules/aws/lambda.go
+++ b/modules/aws/lambda.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/service/lambda"
@@ -9,35 +10,60 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// LambdaOptions contains additional parameters for InvokeFunctionWithParams().
+// It contains a subset of the fields found in the lambda.InvokeInput struct.
+type LambdaOptions struct {
+	// FunctionName is a required field containing the lambda function name.
+	FunctionName *string
+
+	// InvocationType can be one of "RequestResponse" or "DryRun".
+	//    * RequestResponse (default) - Invoke the function synchronously.
+	//    Keep the connection open until the function returns a response
+	//    or times out.
+	//
+	//    * DryRun - Validate parameter values and verify that the user or
+	//    role has permission to invoke the function.
+	InvocationType *string
+
+	// Lambda function input; will be converted to JSON.
+	Payload interface{}
+}
+
+// LambdaOutput contains the output from InvokeFunctionWithParams().  The
+// fields may or may not have a value depending on the invocation type and
+// whether an error occurred or not.
+type LambdaOutput struct {
+	// If present, indicates that an error occurred during function execution.
+	// Error details are included in the response payload.
+	FunctionError *string
+
+	// The response from the function, or an error object.
+	Payload []byte
+
+	// The HTTP status code for a successful request is in the 200 range.
+	// For RequestResponse invocation type, the status code is 200.
+	// For the DryRun invocation type, the status code is 204.
+	StatusCode *int64
+}
+
 // InvokeFunction invokes a lambda function.
 func InvokeFunction(t testing.TestingT, region, functionName string, payload interface{}) []byte {
-	out, err := InvokeFunctionE(t, region, functionName, payload)
+	input := &LambdaOptions{
+		FunctionName: &functionName,
+		Payload:      &payload,
+	}
+	out, err := InvokeFunctionWithParams(t, region, input)
 	require.NoError(t, err)
-	return out
+	return out.Payload
 }
 
 // InvokeFunctionE invokes a lambda function.
 func InvokeFunctionE(t testing.TestingT, region, functionName string, payload interface{}) ([]byte, error) {
-	lambdaClient, err := NewLambdaClientE(t, region)
-	if err != nil {
-		return nil, err
-	}
-
-	invokeInput := &lambda.InvokeInput{
+	input := &LambdaOptions{
 		FunctionName: &functionName,
+		Payload:      &payload,
 	}
-
-	if payload != nil {
-		payloadJson, err := json.Marshal(payload)
-
-		if err != nil {
-			return nil, err
-		}
-		invokeInput.Payload = payloadJson
-	}
-
-	out, err := lambdaClient.Invoke(invokeInput)
-	require.NoError(t, err)
+	out, err := InvokeFunctionWithParams(t, region, input)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


brikis98 (commented):
Thanks for the PR! I'm a little worried about a sprawling collection of `InvokeXXX` functions for every permutation: e.g., one for dry run, one not dry run, one async, one not, etc. Perhaps the new thing to add is an `InvokeFunctionWithParams` function that lets you pass in the settings, such as `InvocationType`, in a `struct`?

kbalk:
> Thanks for the PR! I'm a little worried about a sprawling collection of `InvokeXXX` functions for every permutation: e.g., one for dry run, one not dry run, one async, one not, etc. Perhaps the new thing to add is an `InvokeFunctionWithParams` function that lets you pass in the settings, such as `InvocationType`, in a `struct`?

I would have taken that approach, but your comment to my issue said:  "PR to add this, perhaps in a new function, is welcome!"  The "new function" suggested you didn't want what you are now suggesting.  I can do it the other way ... I just want to make sure which way you truly want it done.

kbalk:
I should also point out that there are only three invocation types.  Adding `InvokeDryRun()` means there is no impact on the current `InvokeFunction()`, i.e., no optional `InvocationType` struct parameter.  It also means you have two of the three types defined.

brikis98:
> > Thanks for the PR! I'm a little worried about a sprawling collection of `InvokeXXX` functions for every permutation: e.g., one for dry run, one not dry run, one async, one not, etc. Perhaps the new thing to add is an `InvokeFunctionWithParams` function that lets you pass in the settings, such as `InvocationType`, in a `struct`?
> 
> I would have taken that approach, but your comment to my issue said: "PR to add this, perhaps in a new function, is welcome!" The "new function" suggested you didn't want what you are now suggesting. I can do it the other way ... I just want to make sure which way you truly want it done.

`InvokeFunctionWithParams` _is_ a new function.

kbalk:
My apologies for the misunderstanding.  I am confused so I want to make sure we're on the same page before I modify any more code.  And more apologies if I repeat what you already know ...

You're suggesting a new function `InvokeFunctionWithParams` that includes a parameter for the `InvocationType`.  The AWS SDK for Golang defines `InvocationType` enum values as shown below and the SDK's [`InvokeInput`](https://docs.aws.amazon.com/sdk-for-go/api/service/lambda/#InvokeInput) struct contains a field for the `InvocationType`.  In Terratest, `InvokeFunction` initializes an `InvokeInput` struct, but the `InvocationType` field is not initialized and the default value of "RequestResponse" is assumed.

```
const (
    // InvocationTypeEvent is a InvocationType enum value
    InvocationTypeEvent = "Event"

    // InvocationTypeRequestResponse is a InvocationType enum value
    InvocationTypeRequestResponse = "RequestResponse"

    // InvocationTypeDryRun is a InvocationType enum value
    InvocationTypeDryRun = "DryRun"
)
```

I initially assumed the enhancement request would involve a boolean argument to Terratest's `InvokeFunction` that would indicate whether or not a "DryRun" invocation was requested.  Silly me -- I've been working with Python where it'd be easy to add an optional argument without affecting existing callers of the API.

So to be clear, when you suggest the creation of an `InvokeFunctionWithParams` function, are you suggesting a duplicate of `InvokeFunction` with an argument that can be one of the `InvocationType` enum values?  `InvokeFunction` uses the default `InvocationType` of "RequestResponse" and I can add the support for "DryRun", but are you suggesting I also add support for "Event"?  If so, I would prefer that the addition of the "Event" logic be handled in a different PR as it's not something that I'm volunteering to do.

Regarding your concern about multiple flavors of `InvokeFunction`, I'd like to point out that the AWS SDK for Golang has 6 types of invocation functions:  `Invoke`, `InvokeAsync`, `InvokeAsyncRequest`, `InvokeAsyncWithContext`, `InvokeRequest` and `InvokeWithContext`.

Also, I see the "DryRun" invocation as a pass/fail sort of function.  It doesn't actually invoke the function; it just tests that it has permissions to do so.  So, it kind of does it's own thing.  When support for an an "Event" invocation is handled, that's a bit more complicated and you might end up splitting that out as well.  Just my two cents.

brikis98:
> I initially assumed the enhancement request would involve a boolean argument to Terratest's `InvokeFunction` that would indicate whether or not a "DryRun" invocation was requested. Silly me -- I've been working with Python where it'd be easy to add an optional argument without affecting existing callers of the API.

Yup, this is an annoying limitation of Go.

> So to be clear, when you suggest the creation of an `InvokeFunctionWithParams` function, are you suggesting a duplicate of `InvokeFunction` with an argument that can be one of the `InvocationType` enum values? 

Not quite. It's a new function that accepts a new `struct` called something like `LambdaOptions`. The `struct` allows us to support various other params we may want to pass in while invoking the Lambda function. One of those can be `InvocationType`, as you're asking for here, but having the `struct` allows us to add support for other options in the future. Under the hood, the original `InvokeFunction` would then be updated to call `InvokeFunctionWithParams`, passing the proper options `struct`.

kbalk:
I've added the function you requested and updated the gist with the test results.  When handling different invocation types, you have to be prepared for different types of output, so I needed a struct for the output (e.g., "DryRun" has no payload and the status is in the returned structure).  Here's the signature for the new function:

```
// InvokeFunctionWithParams invokes a lambda function using parameters
// supplied in the LambdaOptions struct and returns values in a LambdaOutput
// struct.
func InvokeFunctionWithParams(t testing.TestingT, region string, input *LambdaOptions) (*LambdaOutput, error) {
```

LambdaOptions and LambdaOutput are subsets of lambda.InvokeInput and lambda.InvokeOutput, respectively.

Note: this actually is an "E" function; I didn't create a non-E and E function, but I'm not sure it makes sense to have a non-E version.

modules/aws/lambda.go
```diff
@@ -2,51 +2,136 @@ package aws
...
 	"github.com/aws/aws-sdk-go/service/lambda"
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/stretchr/testify/require"
 )
 
+// LambdaOptions contains additional parameters for InvokeFunctionWithParams().
+// It contains a subset of the fields found in the lambda.InvokeInput struct.
+type LambdaOptions struct {
+	// FunctionName is a required field containing the lambda function name.
+	FunctionName *string
```
brikis98, line 17: NIT: Since this is actually required, I'd rather keep it as a separate parameter in the function signature, rather than an optional value in the "options" `struct`.

modules/aws/lambda.go
```diff
@@ -2,51 +2,136 @@ package aws
...
 func InvokeFunction(t testing.TestingT, region, functionName string, payload interface{}) []byte {
-	out, err := InvokeFunctionE(t, region, functionName, payload)
+	input := &LambdaOptions{
+		FunctionName: &functionName,
+		Payload:      &payload,
+	}
+	out, err := InvokeFunctionWithParams(t, region, input)
 	require.NoError(t, err)
-	return out
+	return out.Payload
```
brikis98, line 57: This should call `InvokeFunctionE`. No need to change it.

modules/aws/lambda.go
```diff
@@ -2,51 +2,136 @@ package aws
...
+	FunctionName *string
+
+	// InvocationType can be one of "RequestResponse" or "DryRun".
+	//    * RequestResponse (default) - Invoke the function synchronously.
+	//    Keep the connection open until the function returns a response
+	//    or times out.
+	//
+	//    * DryRun - Validate parameter values and verify that the user or
+	//    role has permission to invoke the function.
+	InvocationType *string
```
brikis98, line 26: NIT: Any reason not to use the AWS enum type here? Reduces error checking down below.

modules/aws/lambda.go
```diff
@@ -2,51 +2,136 @@ package aws
...
+	Payload interface{}
+}
+
+// LambdaOutput contains the output from InvokeFunctionWithParams().  The
+// fields may or may not have a value depending on the invocation type and
+// whether an error occurred or not.
+type LambdaOutput struct {
+	// If present, indicates that an error occurred during function execution.
+	// Error details are included in the response payload.
+	FunctionError *string
```
brikis98, line 38: Why include this in the output rather than returning it as an `error` as the function did before? It seems more Go idiomatic to wrap this in an `error`, rather than embedding it in the output struct.

modules/aws/lambda.go
```diff
@@ -2,51 +2,136 @@ package aws
...
+		return out.Payload, &FunctionError{Message: *out.FunctionError, StatusCode: *out.StatusCode, Payload: out.Payload}
+	}
+
+	return out.Payload, nil
+}
+
+// InvokeFunctionWithParams invokes a lambda function using parameters
+// supplied in the LambdaOptions struct and returns values in a LambdaOutput
+// struct.
+func InvokeFunctionWithParams(t testing.TestingT, region string, input *LambdaOptions) (*LambdaOutput, error) {
```
brikis98, line 81: This should use the `E` naming. But there should be a non-`E` version too.

modules/aws/lambda.go
kbalk: Done.

modules/aws/lambda.go
kbalk: Done.

modules/aws/lambda.go
kbalk: To be frank, I'm having a hard time figuring out what part of the AWS SDK you're trying to "shield" from the user and what part you're okay with exposing.  So I try something I think is inline with your approach and then adjust when you tell me I'm wrong.  It's not the best use of time for either of us.

Done

modules/aws/lambda.go
kbalk: Done

modules/aws/lambda.go
kbalk: [But that's how the AWS SDK represents a FunctionError in their InvokeOutput struct.](https://github.com/aws/aws-sdk-go/blob/v1.38.4/service/lambda/api.go#L11905-L11927).

`InvokeFunctionE()`'s `FunctionError` struct is equivalent to `InvokeFunctionWithParamsE()`'s `InvokeOutput`.  Which will be confusing, but my contention is that `InvokeFunctionE()`'s `FunctionError` is a misnomer.  For example, for DryRun, a non-nil `FunctionError` is not an error as it contains output, e.g., a status code.

The fact that any other kind of `FunctionError` from `InvokeFunctionE()` is treated as an error is why I needed to make these code changes in the first place.

modules/aws/lambda.go
brikis98: That still seems cleaner than having an untyped string though, right?

modules/aws/lambda.go
brikis98: Since the AWS API is very... odd here... Wouldn't it make more sense to have the new `InvokeFunctionWithParamsE` check for `FunctionError` and handle it accordingly? E.g., If a Dry Run `FunctionError` isn't _really_ an error, then handle it accordingly?

kbalk: I'm very confused and I'm not sure what you're asking or if I do understand it, if I agree with you.  

`InvokeFunctionWithParamsE()` returns `LambdaOutput` and `error`:

- `error` represents the error that could occur in the process of invoking the lambda, e.g., bad arguments passed to `InvokeFunctionWithParmsE()` itself. 

- `LambdaOutput` doesn't "judge" the output.  It's not trying to guess what is or is not a valid response from the lambda.  Where I work, we invoke lambdas that have do not return a payload and issue AWS SDK calls to verify the expected side effects take place, so we're only interested in the `error` return and maybe the status code in `LambdaOutput`.  

I don't want to call `LambdaOutput` `FunctionError` because that's assuming the contents represent an error, which is not always the case.  The output can have no payload and a status code of "200" and that's success, not an error.  For DryRun, there's no payload and a status code of "204".  That's success.

If the Lambda function returns a status code the caller is not expecting, then `LambdaOutput.FunctionError` provides details that can be logged for future debugging.

If what you're saying is that you want `LambdaOutput.FunctionError` to be of type `Error`, I can do that.

brikis98: > I'm very confused and I'm not sure what you're asking or if I do understand it, if I agree with you.
> 
> `InvokeFunctionWithParamsE()` returns `LambdaOutput` and `error`:
> 
> * `error` represents the error that could occur in the process of invoking the lambda, e.g., bad arguments passed to `InvokeFunctionWithParmsE()` itself.
> * `LambdaOutput` doesn't "judge" the output.  It's not trying to guess what is or is not a valid response from the lambda.  Where I work, we invoke lambdas that have do not return a payload and issue AWS SDK calls to verify the expected side effects take place, so we're only interested in the `error` return and maybe the status code in `LambdaOutput`.
> 
> I don't want to call `LambdaOutput` `FunctionError` because that's assuming the contents represent an error, which is not always the case. The output can have no payload and a status code of "200" and that's success, not an error. For DryRun, there's no payload and a status code of "204". That's success.
> 
> If the Lambda function returns a status code the caller is not expecting, then `LambdaOutput.FunctionError` provides details that can be logged for future debugging.
> 
> If what you're saying is that you want `LambdaOutput.FunctionError` to be of type `Error`, I can do that.

I think the mental model I have is that the signature is roughly:

```go
InvokeFunctionWithParams(...) (*LambdaOutput, error)
```

Where `LambdaOutput` is:

```go
type LambdaOutput struct {
  Payload []byte
  StatusCode *int64
}
```

Any error `InvokeFunctionWithParams` hits, whether while invoking the Lambda, or returned in `out.FunctionError`, is returned in the second `error` param. The rest goes into `LambdaOutput`. The idea of having to check for errors in the `error` param _and_ in `LambdaOutput` strikes me as very confusing.

kbalk: Thank you for the write up; that cleared up the confusion.  I think the confusion was due to `FunctionError` being the name of a struct and a field in another struct; I think we weren't talking about the same `FunctionError`.

I removed `FunctionError` from `LambdaOutput`.  I had to restore the original code for `InvokeFunctionE()` because the values returned from `InvokeFunctionWithParams()` no longer distinguish between parameter errors and invocation errors, and `InvokeFunctionE()` needs to know the difference to determine whether `FunctionError{}` should be populated.

test/terraform_aws_lambda_example_test.go
```diff
@@ -62,6 +62,83 @@ func TestTerraformAwsLambdaExample(t *testing.T) {
...
+	functionName := fmt.Sprintf("terratest-aws-lambda-withparams-example-%s", random.UniqueId())
+
+	// Pick a random AWS region to test in. This helps ensure your code works in all regions.
+	awsRegion := aws.GetRandomStableRegion(t, nil, nil)
+
+	// Construct the terraform options with default retryable errors to handle the most common retryable errors in
+	// terraform testing.
+	terraformOptions := terraform.WithDefaultRetryableErrors(t, &terraform.Options{
+		// The path to where our Terraform code is located
+		TerraformDir: "../examples/terraform-aws-lambda-example",
```
brikis98, line 82: Since these tests run in parallel, if you run more than one test in a given folder, it's possible for the state files and `.terraform` folders to clobber each other. Therefore, you'll need to use [this function](https://github.com/gruntwork-io/terratest/blob/master/modules/test-structure/test_structure.go#L70) in this test and the one above it to make a copy of this folder.

test/terraform_aws_lambda_example_test.go
```diff
@@ -62,6 +62,83 @@ func TestTerraformAwsLambdaExample(t *testing.T) {
...
+	invocationType = aws.InvocationTypeRequestResponse
+	input = &aws.LambdaOptions{
+		InvocationType: &invocationType,
+		Payload:        ExampleFunctionPayload{ShouldFail: true, Echo: "hi!"},
+	}
+	out, err := aws.InvokeFunctionWithParamsE(t, awsRegion, functionName, input)
+
+	// The Lambda executed, but should have failed.
+	require.NotNil(t, err)
+	assert.Contains(t, err.Error(), "Unhandled")
```
brikis98, line 124:
```suggestion
	require.Error(t, err, "Unhandled")
```

brikis98 (commented):
Thanks! Just a couple tweaks and then I can run tests.

test/terraform_aws_lambda_example_test.go
kbalk: Done

test/terraform_aws_lambda_example_test.go
kbalk: Good to know!  Done

brikis98 (approved):

brikis98:
Tests passed. Merging now.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.32.21

PR #863, opened 2021-04-08
updated keyvault ref from 2016 to 2019, verified sovereign support

shahedc:
* updated keyvault reference from 2016 version to newer 2019 version
* verified that client_factory.go has CreateKeyVaultManagementClientE() method that looks up env URI
* verified that keyvault.go has GetKeyVaultE() method that calls GetKeyVaultManagementClientE() which returns a vaultClient object (rather than a reference)


james00012:

PR #864, opened 2021-04-09
Module to support Yandex.Cloud

GennadySpb:


brikis98:
Thank you for the PR! However, as [per our contribution guidelines](https://terratest.gruntwork.io/docs/community/contributing/#types-of-contributions), right now, we're only focusing on AWS, GCP, and Azure. Unfortunately, at the moment, our team does not have the resources to support any other clouds or providers. Perhaps we'll be able to come back to this in the future, but I'm going to close this PR for now.

GennadySpb:
@brikis98  thank you for clarification! Terratest looks like a great product and we would like to use its to work with Yandex.Cloud. Please suggest the best way to accomplish this task.

brikis98:
@GennadySpb I think for now, the best bet would be to create a separate repo of Yandex.Cloud helpers. We could then link to it from the [Terratest tools and plugins page](https://terratest.gruntwork.io/docs/testing-best-practices/tools-and-plugins/). In the future, if our team grows, and we start focusing on Yandex.Cloud, we'd be more than happy to help out with this, but I think for now, a separate repo maintained by someone else is the best option.

Issue #855, opened 2021-04-08
[Terraform] Add Terraform Validate function


axilleastr:
From what I've seen `terraform validate` is missing from the terratest available terraform functions. 
Is this intentionally or is it something that could be added? 

Thanks! :)

brikis98:
Not intentional. PR to add it is welcome!

axilleastr:
Awesome, thanks :)

brikis98:
Fixed in #866 and released in https://github.com/gruntwork-io/terratest/releases/tag/v0.32.23.

PR #866, opened 2021-04-12
[Terraform] Add Terraform Validate function

axilleastr:
This PR addresses #855

Description:

**What has been added:**
- terraform validate functions, that were missing from the terraform module

### DIFF
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -1,0 +1,39 @@
+package terraform
+
+import (
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// Validate calls terraform validate and returns stdout/stderr.
+func Validate(t testing.TestingT, options *Options) string {
+	out, err := ValidateE(t, options)
+	require.NoError(t, err)
+	return out
+}
+
+// ValidateE calls terraform validate and returns stdout/stderr.
+func ValidateE(t testing.TestingT, options *Options) (string, error) {
+	return RunTerraformCommandE(t, options, FormatArgs(options, "validate")...)
+}
+
+// InitAndValidate runs terraform init and validate with the given options and returns stdout/stderr from the validate command.
+// This will fail the test if there is an error in the command.
+func InitAndValidate(t testing.TestingT, options *Options) string {
+	out, err := InitAndValidateE(t, options)
+	require.NoError(t, err)
+	return out
+}
+
+// InitAndValidateE runs terraform init and validate with the given options and returns stdout/stderr from the validate command.
+func InitAndValidateE(t testing.TestingT, options *Options) (string, error) {
+	if _, err := InitE(t, options); err != nil {
+		return "", err
+	}
+
+	if _, err := GetE(t, options); err != nil {
+		return "", err
+	}
+
+	return ValidateE(t, options)
+}
--- a/modules/terraform/validate_test.go
+++ b/modules/terraform/validate_test.go
@@ -1,0 +1,37 @@
+package terraform
+
+import (
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/files"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInitAndValidateWithNoError(t *testing.T) {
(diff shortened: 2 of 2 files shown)


modules/terraform/init.go
```diff
@@ -22,3 +22,17 @@ func InitE(t testing.TestingT, options *Options) (string, error) {
 	args = append(args, FormatTerraformPluginDirAsArgs(options.PluginDir)...)
 	return RunTerraformCommandE(t, options, args...)
 }
+
+// InitWithoutBackend calls terraform init and return stdout/stderr.
+func InitWithoutBackend(t testing.TestingT, options *Options) string {
```
brikis98, line 27: What is the use case for this?

modules/terraform/validate.go
```diff
@@ -0,0 +1,62 @@
+package terraform
+
+import (
+	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/stretchr/testify/require"
+)
+
+// Validate calls terraform validate and return stdout/stderr.
```
brikis98, line 8:
```suggestion
// Validate calls terraform validate and returns stdout/stderr.
```

modules/terraform/validate.go
```diff
@@ -0,0 +1,62 @@
...
+func Validate(t testing.TestingT, options *Options) string {
+	out, err := ValidateE(t, options)
+	if err != nil {
+		t.Fatal(err)
+	}
```
brikis98, line 13: NIT:

```suggestion
	require.NoError(err)
```

brikis98 (commented):
Thanks for the PR!

modules/terraform/init.go
axilleastr: Since validation requires an initialized working directory, I tried to add the functionality to have an initialized working directory without accessing any configured remote backend. ( ref: https://www.terraform.io/docs/cli/commands/validate.html )
However that may be absolutely redundant, since the main goal of this PR is to add `terraform validate` option and the existing Init functions are enough for that.
I could revert the commit that adds this functionality if you have concerns about it :)

axilleastr: commit with this change reverted :)

modules/terraform/validate_test.go
```diff
@@ -0,0 +1,39 @@
...
+	require.NoError(t, err)
+
+	options := &Options{
+		TerraformDir: testFolder,
+	}
+
+	InitAndValidate(t, options)
+	out, err := ValidateE(t, options)
+	require.NoError(t, err)
+	require.Contains(t, out, "Success! The configuration is valid")
```
brikis98, line 23:
```suggestion
	out := InitAndValidate(t, options)
	require.Contains(t, out, "Success! The configuration is valid")
```

modules/terraform/validate_test.go
```diff
@@ -0,0 +1,39 @@
...
+	testFolder, err := files.CopyTerraformFolderToTemp("../../test/fixtures/terraform-basic-configuration", t.Name())
+	require.NoError(t, err)
+
+	options := &Options{
+		TerraformDir: testFolder,
+	}
+
+	out, err := ValidateE(t, options)
+	require.Error(t, err)
+	require.Contains(t, out, "Plugin reinitialization required. Please run \"terraform init\"")
```
brikis98, line 38: This doesn't seem like the case we're most interested in... Instead, we should add a test fixture with _invalid_ Terraform code, call `InitAndValidate` on it, and ensure we get back an error and the expected text.

axilleastr: Agreed. 
How about using the `terraform-with-plan-error` test fixture for this case, instead of creating a new one? 
It's basically  a test case that should indicate the functionality we are looking for, since it is trying to output an undeclared input variable.

brikis98: Perfect, thanks!

brikis98 (approved):
Changes LGTM! I'll kick off tests now.

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.32.23

Issue #811, opened 2021-03-15
Support resolving base URIs for existing Azure modules


davidkpiano:
Placeholder issue for the ongoing work for supporting different clouds (e.g., sovereign cloud) resolved via `baseURI`:

- [x] Action Group (@davidkpiano)
- [x] AKS (@davidkpiano) 
- [x] Availability Set (@davidkpiano)
- [x] Compute (@davidkpiano)
- [x] Cosmos DB (@davidkpiano)
- [x] Disk (@davidkpiano)
- [x] Key Vault (@shahedc)
- [x] Load Balancer (@mjyeaney)
- [x] Log Analytics (@davidkpiano)
- [x] Monitor (@davidkpiano) 
- [x] MySQL (@davidkpiano)
- [x] Network Interface (@mjyeaney)
- [x] NSG (@mjyeaney)
- [x] Public IP Address (@mjyeaney)
- [x] Resource Group (@davidkpiano)
- [x] SQL (@davidkpiano)
- [x] Storage (@davidkpiano)
- [x] Virtual Network (@davesee)

yorinasub17:
Quick question: is it possible to keep the `Get...Client` name for now to avoid backwards incompatibility? And then at a later time once all the modules are updated, we have one PR + release that renames them to `Create...Client`?

Otherwise, we can't incrementally merge the changes without avoiding a backwards incompatible version bump each time.

davidkpiano:
> Quick question: is it possible to keep the `Get...Client` name for now to avoid backwards incompatibility? And then at a later time once all the modules are updated, we have one PR + release that renames them to `Create...Client`?
> 
> Otherwise, we can't incrementally merge the changes without avoiding a backwards incompatible version bump each time.

Good call; I'l make that change.

HadwaAbdelhalem:
@yorinasub17  would you please close it as all realted PR are either merged or closed.

PR #868, opened 2021-04-13
[Azure] Add sovereign cloud support for Storage

### LINKED ISSUE #861: [Azure] Add sovereign cloud support for storage
The Base URI is incorrectly resolved for the storage client, and should be fixed to read from the environment.

Addresses #811

davidkpiano:
This PR resolves the Base URI for the Resource Group module to support sovereign cloud.

Addresses #811, fixes #861 

cc. @rguthriemsft


hattan:

You can check the status of the CI Pipeline logs here ; https://github.com/aztfmod/terratest/actions/runs/745238923

modules/azure/storage.go
```diff
@@ -191,6 +192,7 @@ func GetStorageAccountClientE(subscriptionID string) (*storage.AccountsClient, e
 }
 
 // GetStorageBlobContainerClientE creates a storage container client.
```
HadwaAbdelhalem, line 194: remove the method GetStorageBlobContainerClientE  as it is no longer used

modules/azure/storage.go
```diff
@@ -174,6 +174,7 @@ func GetStorageAccountPropertyE(storageAccountName, resourceGroupName, subscript
 }
 
 // GetStorageAccountClientE creates a storage account client.
+// TODO: remove in next version
```
HadwaAbdelhalem, line 177: remove GetStorageAccountClientE method as it is no longer used

HadwaAbdelhalem (changes requested):
Thanks @davidkpiano , one last NIT change

modules/azure/storage.go
yorinasub17: I mentioned this somewhere, but I can't find where I said this, so reiterating it here:

Since this is a public method, removing it means we will have to cut a backwards incompatible release. It would be a bit of a pain for us to release a backwards incompatible release for each of these changes. I think it would be better to keep these in so that we can merge all of these changes as patch releases, and then at the very end once all the files are updated, we can have one big PR to remove all the `Get.*Client` functions and cut one backwards incompatible release.

HadwaAbdelhalem (approved):

modules/azure/storage.go
HadwaAbdelhalem: sounds good @yorinasub17 , I approved it and it is ready for  review  @gruntwork-terratest-terragrunt

yorinasub17 (approved):
LGTM as well! Going to merge and release this version.

PR #869, opened 2021-04-16
Temporary workaround for broken build

yorinasub17:
Golang is no longer available from where we were originally downloading from. We temporarily workaround the issue here until the permanent fix is available in gruntwork tooling.


rhoboat (approved):
LGTM! Also approved the permanent fix one.

yorinasub17:
Thanks for review! Merging this in so tests start passing.

Issue #758, opened 2021-01-11
k8s: Add function to return server version


tylerauerbeck:
Currently have a need to only run tests if running against a certain version of Kubernetes. This is particularly useful as we need to support multiple versions of Kubernetes, but only want to use / test features as they are available and make sure that we don't use things as they are deprecated.

brikis98:
This probably doesn't past the complexity requirements of our [litmus test](https://terratest.gruntwork.io/docs/community/contributing/#types-of-contributions), but I suspect checking the version could be a very common operation with K8S, so it might be worth having from a completeness standpoint anyway, so I think a PR is welcome!

tylerauerbeck:
@brikis98 Sounds good. I can take a look at this over the next few days when I get a chance.

james03160927:
It's implemented here: https://github.com/gruntwork-io/terratest/pull/876. Closing this.

PR #876, opened 2021-04-23
add k8s.GetKubernetesClusterVersion to get cluster version

falmotlag:
I found myself checking cluster version often so I wrote a helper function to do so. This also satisfy the requirements of this [issue](https://github.com/gruntwork-io/terratest/issues/758).

PR adds a function that returns Kubernetes cluster version.

I'm not sure how to write unit tests for code the depend on [Kubernetes client](https://github.com/gruntwork-io/terratest/blob/adca4a84bcddefbb7d6453fb8bd17da25f2ecd04/modules/k8s/client.go#L16), any pointers are greatly appreciated.

### DIFF
--- a/modules/k8s/version.go
+++ b/modules/k8s/version.go
@@ -3,7 +3,7 @@ package k8s
 import "github.com/gruntwork-io/terratest/modules/testing"
 
 // GetKubernetesClusterVersion returns the Kubernetes cluster version.
-func GetKubernetesClusterVersion(t testing.TestingT) (string, error) {
+func GetKubernetesClusterVersionE(t testing.TestingT) (string, error) {
 	clientset, err := GetKubernetesClientE(t)
 	if err != nil {
 		return "", err
--- a/modules/k8s/version.go
+++ b/modules/k8s/version.go
@@ -4,7 +4,19 @@ import "github.com/gruntwork-io/terratest/modules/testing"
 
 // GetKubernetesClusterVersion returns the Kubernetes cluster version.
 func GetKubernetesClusterVersionE(t testing.TestingT) (string, error) {
-	clientset, err := GetKubernetesClientE(t)
+	kubeConfigPath, err := GetKubeConfigPathE(t)
+	if err != nil {
+		return "", err
+	}
+
+	options := NewKubectlOptions("", kubeConfigPath, "default")
+
+	return GetKubernetesClusterVersionWithOptionsE(t, options)
+}
+
+// GetKubernetesClusterVersion returns the Kubernetes cluster version given a configured KubectlOptions object.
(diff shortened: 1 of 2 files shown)


modules/k8s/version.go
```diff
@@ -0,0 +1,18 @@
+package k8s
+
+import "github.com/gruntwork-io/terratest/modules/testing"
+
+// GetKubernetesClusterVersion returns the Kubernetes cluster version.
+func GetKubernetesClusterVersion(t testing.TestingT) (string, error) {
```
yorinasub17, line 6: NIT: We label functions that return errors with a `E`

```suggestion
func GetKubernetesClusterVersionE(t testing.TestingT) (string, error) {
```

yorinasub17 (commented):
Thanks for your contribution! Implementation LGTM, but we will want to have a unit test for the function. Can you add one that compares the result of this function with `kubectl version`?

modules/k8s/version.go
falmotlag: Totally agree. I'm having trouble figuring out how to unit test this function since it depends on [Kubernetes client](https://github.com/gruntwork-io/terratest/blob/adca4a84bcddefbb7d6453fb8bd17da25f2ecd04/modules/k8s/client.go#L16). 

I initially thought of accepting the client as input to the function to allow for passing a mock client for testing. However, this raises the complexity of the function that should be an easy to call helper function.

Got an idea on how to test this?

yorinasub17: Two thoughts:

- It might be a good idea to introduce an additional function, `GetKubernetesClusterVersionWithOptionsE(t testing.TestingT, kubectlOptions *KubectlOptions) (string, error)`, and use `GetKubernetesClientFromOptionsE` instead of `GetKubernetesClient`. This way, you can have full control over how to authenticate by constructing custom KubectlOptions. If you do do this, `GetKubernetesClusterVersionE` should call `GetKubernetesClientFromOptionsE`, constructing the options using the same way as https://github.com/gruntwork-io/terratest/blob/master/modules/k8s/client.go#L16

- You can actually write the tests ignoring injection, and rely on the default context against minikube. This is actually how most of the existing tests for the module are written.


I have a slight preference for the first option, but will accept this PR with either approach!

falmotlag: Implemented suggestion number 1 and added unit tests.

yorinasub17:
Updates LGTM! I'll kick off a regression test build and if it passes, we can merge this in! Thanks for your contribution!

yorinasub17 (approved):
build failed, but passed once I rebased on main so this should be good to go. Thanks again for your contribution!

Issue #840, opened 2021-03-30
modules/k8s, Ingress dependency on apiVersion


kuritka:
k8s module Ingress strongly depends on `k8s.io/api/extensions/v1beta1` although `extensions/v1beta1` is [deprecated](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/).

If I migrate an ingress to `networking.k8s.io/v1` or `networking.k8s.io/v1beta1` terratest retreives warnings like this:
```
W0326 15:52:08.878455   20111 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
TestK8gbRepeatedlyRecreatedFromIngress 2021-03-26T15:52:08Z retry.go:103: Wait for ingress test-gslb-failover-simple to be provisioned. returned an error: Ingress test-gslb-failover-simple is not available. 
```

The warning is caused by hardcoded  `clientset.ExtensionsV1beta1()` in `GetIngressE` function.
```go
func GetIngressE(t testing.TestingT, options *KubectlOptions, ingressName string) (*extensionsv1beta1.Ingress, error) {
   // ...
   return clientset.ExtensionsV1beta1().Ingresses(options.Namespace).Get(context.Background(), ingressName, metav1.GetOptions{})
}
```

Can you extend the k8s module to support multiple apiVersions ?

cbuto:
Ran into this today. @yorinasub17 I'm happy to take a stab at this but I'll likely need to do some brainstorming to make this backward compatible. Do you know of any existing patterns for supporting multiple API versions in Terratest? Otherwise, I'll come up with a proposal for a possible implementation!

yorinasub17:
Hi, sorry for the delay in responding to this. I think this is one of the few cases were having a backward incompatible release is fine, given that Kubernetes doesn't really support go modules that well and thus it is near impossible to support multiple API versions when we link the libraries like this.

That said, as far as supporting multiple versions go, I think the only way we can do that is to have multiple copies of the function for each library version.

I opened a PR to update the references: https://github.com/gruntwork-io/terratest/pull/879

yorinasub17:
Released as https://github.com/gruntwork-io/terratest/releases/tag/v0.34.0

PR #879, opened 2021-04-27
Use networkingv1 instead of extensionsv1beta1

yorinasub17:
Fixes https://github.com/gruntwork-io/terratest/issues/840

NOTE: due to the way the k8s APIs are designed, we have no choice but to duplicate the functions to support the separate API versions.


brikis98 (approved):
LGTM

yorinasub17:
Thanks for review! Build failed, but it looks like it is unrelated to this change, so going to merge + release this!

Issue #877, opened 2021-04-23
Running InitAndPlanE will always update modules


alan2wong:
The Terraform Options struct includes the option to disable upgrading during `terraform init`, but even if this is set to false, an upgrade still occurs due to `terraform get -upgrade` always being called during `InitAndPlanE`:

- `InitAndPlanE` calls both `InitE` and `GetE`: https://github.com/gruntwork-io/terratest/blob/adca4a84bcddefbb7d6453fb8bd17da25f2ecd04/modules/terraform/plan.go#L19-L29
- `GetE` is set to always call `terraform get -update` which forces an update: https://github.com/gruntwork-io/terratest/blob/adca4a84bcddefbb7d6453fb8bd17da25f2ecd04/modules/terraform/get.go#L17-L19

The expected behavior is if `Upgrade` is set to false or not set, no upgrade should be run by either `init` or `get`.

yorinasub17:
This is now fixed in https://github.com/gruntwork-io/terratest/releases/tag/v0.34.1, where the call to `get` is removed.

PR #880, opened 2021-04-27
Should not need to call terraform get

yorinasub17:
Fixes https://github.com/gruntwork-io/terratest/issues/877

We should not have to make a call to `GetE`, as `InitE` should be sufficient to handle all the necessary initialization procedures.


brikis98 (approved):
LGTM

yorinasub17:
Thanks for review! Merging this in now!

PR #898, opened 2021-05-13
Fix build

yorinasub17:
At some point, terratest build process stopped working (See https://app.circleci.com/pipelines/github/gruntwork-io/terratest/553/workflows/59862dba-539e-4d0a-8f9b-f0af33b7b4db for example). I still don't understand **why** this is happening given that I can't repro locally both in my Mac and Linux boxes. However, what appears to happen is that `go` completely ignores the `go.mod` and `go.sum` files located at the root in the circleci environment.

So I bumped golang to the latest version. This appears to have worked, but introduced a different problem where it thinks the `go.mod` file is out of date, even though it is up to date in my local env (when I run `go mod tidy` locally, nothing changes!).

In any case, the build appears to have been fixed when I added a call to `go mod tidy` prior to `build-go-binaries`.

As a bonus, this adds `darwin/arm64` binaries for `terratest_log_parser` and `pick-instance-type` bins.


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Merging this in now.

PR #899, opened 2021-05-13
Fix bug in terratest-log-parser that mixes result lines

yorinasub17:
This fixes two bugs in `terratest-log-parser`:

- `terratest-log-parser` was incorrectly handling result lines such that they leaked into the wrong test. You can see this behavior in the modifications made to the test fixtures.
- With newer go versions, the assertion failures were being reported before the fail result line. This change was previously not handled well, resulting in the error traces being ignored and not making it to any split out log file. This now properly feeds those lines to the right test file.

### DIFF
--- a/modules/logger/parser/parser.go
+++ b/modules/logger/parser/parser.go
@@ -41,7 +41,7 @@ func SpawnParsers(logger *logrus.Logger, reader io.Reader, outputDir string) {
 var (
 	regexResult  = regexp.MustCompile(`--- (PASS|FAIL|SKIP): (.+) \((\d+\.\d+)(?: ?seconds|s)\)`)
 	regexStatus  = regexp.MustCompile(`=== (RUN|PAUSE|CONT)\s+(.+)`)
-	regexSummary = regexp.MustCompile(`^(ok|FAIL)\s+([^ ]+)\s+(?:(\d+\.\d+)s|\(cached\)|(\[\w+ failed]))(?:\s+coverage:\s+(\d+\.\d+)%\sof\sstatements(?:\sin\s.+)?)?$`)
+	regexSummary = regexp.MustCompile(`(^FAIL$)|(^(ok|FAIL)\s+([^ ]+)\s+(?:(\d+\.\d+)s|\(cached\)|(\[\w+ failed]))(?:\s+coverage:\s+(\d+\.\d+)%\sof\sstatements(?:\sin\s.+)?)?$)`)
 	regexPanic   = regexp.MustCompile(`^panic:`)
 )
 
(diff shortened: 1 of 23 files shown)


yorinasub17:
NOTE: build is failing because of https://github.com/gruntwork-io/terratest/pull/898

modules/logger/parser/parser.go
```diff
@@ -140,6 +140,7 @@ func parseAndStoreTestOutput(
 
 			case isStatusLine(data):
 				testName := getTestNameFromStatusLine(data)
+				previousTestName = testName
```
yorinasub17, line 143: This is the fix for the lost error trace. The error trace line was happening right after a `=== CONT TESTNAME` line, which is a status line. Now we roll up any unmarked test lines that show up after a status line to the test that relates to the status line.

modules/logger/parser/parser.go
```diff
@@ -164,13 +164,15 @@ func parseAndStoreTestOutput(
 				previousTestName = "summary"
 				logWriter.writeLog(logger, "summary", data)
 
+			case isResultLine(data):
+				// We ignore result lines, because that is handled specially below.
```
yorinasub17, line 168: This fixes the bug where the results lines leak into unrelated test logs. The result lines were ending up in the base case right below this because there was no case guard that handled the result lines separately.

modules/logger/parser/parser.go
```diff
@@ -164,13 +164,15 @@ func parseAndStoreTestOutput(
...
+
 			case previousTestName != "":
 				// Base case: roll up to the previous test line, if it exists.
 				// Handles case where terratest log has entries with newlines in them.
 				logWriter.writeLog(logger, previousTestName, data)
 
-			case !isResultLine(data):
-				// Result Lines are handled below
+			default:
```
yorinasub17, line 175: The previous logic didn't make sense, so this was switched to `default` which makes more sense.

modules/logger/parser/parser.go
```diff
@@ -149,12 +150,11 @@ func parseAndStoreTestOutput(
...
 				vals := strings.Split(data, " ")
 				testName := vals[0]
-				logWriter.writeLog(logger, testName, data)
 				previousTestName = testName
+				logWriter.writeLog(logger, testName, data)
 
-			case isIndented && previousTestName != "summary":
-				// In a test result block, so collect the line into all the test results we have seen so far.
-				// Note that previousTestName would only be set to summary if we saw a panic line.
+			case isIndented && isResultLine(data):
```
yorinasub17, line 156: This is a more direct case guard for the specific case that this is trying to capture.

brikis98 (dismissed):
LGTM!

yorinasub17:
UPDATE: Rebased on `master` to trigger build with fix branch so we get some regression testing.

yorinasub17:
@brikis98 Can I get another approval now that the build is passing on the rebase?

brikis98 (approved):

yorinasub17:
Thanks for review! Merging this in now!

PR #760, opened 2021-01-15
Fallback to in-cluster authentication via service account token

msvechla:
This PR implements a fallback to in-cluster authentication via service account token when no valid kubeconfig can be found.

For more info see: https://github.com/gruntwork-io/terratest/issues/753

I manually tested this by running the following inside a Kubernetes pod:

```go
package test

import (
	"fmt"
	"testing"

	"github.com/gruntwork-io/terratest/modules/k8s"
	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

func TestInClusterAuth(t *testing.T) {
	options := k8s.NewKubectlOptions("", "", "default")

	pods := k8s.ListPods(t, options, v1.ListOptions{})
	for _, p := range pods {
		fmt.Println(p.Name)
	}
}
```

Which resulted in:

```
root@golang:/go/terratest# go test -v auth_test.go
=== RUN   TestInClusterAuth
TestInClusterAuth 2021-01-15T10:35:14Z client.go:35: Configuring kubectl using config file /root/.kube/config with context
TestInClusterAuth 2021-01-15T10:35:14Z client.go:39: Error loading api client config, falling back to in-cluster authentication via serviceaccount token: stat /root/.kube/config: no such file or directory
TestInClusterAuth 2021-01-15T10:35:14Z client.go:44: Configuring kubectl using in-cluster serviceaccount token
golang
nginx-deployment-c9dcf7b89-jhff4
nginx-deployment-c9dcf7b89-th97v
rabbitmq-0
redis-master-0
redis-slave-0
redis-slave-1
--- PASS: TestInClusterAuth (0.02s)
PASS
ok      command-line-arguments  0.038s
```

Issue #753, opened 2021-01-07
support in-cluster authentication


msvechla:
Hello, thanks for this awesome project!

Many of our end to end tests are running inside a Kubernetes cluster. 
As we want to interact with the Kubernetes API inside the same cluster where the tests are being executed, it would be helpful to have support for in-cluster authentication via service account tokens as well.

This is already widely supported in client-go, see: https://github.com/kubernetes/client-go/tree/master/examples/in-cluster-client-configuration#authenticating-inside-the-cluster

I would be open to help with a PR, if you have any guidance on how to best implement it within your framework.

My starting point would be inside [GetKubernetesClientFromOptionsE](https://github.com/gruntwork-io/terratest/blob/11c44c322ac13c0954118fb96bc730e27838b4cc/modules/k8s/client.go#L26) and maybe adding an additional config option to `KubectlOptions` to indicate in-cluster authentication.

brikis98:
I think we'd welcome a PR here. @yorinasub17 Thoughts on the approach to take?

yorinasub17:
This makes sense! In terms of approach, we can use a method similar to what we do in `kubergrunt`, which [already supports alternative authentication schemes](https://github.com/gruntwork-io/kubergrunt/blob/master/kubectl/options.go).

msvechla:
Sounds great, I will try to implement a first draft asap. Thanks for your reply!

msvechla:
@yorinasub17 I looked into it a bit more, unfortunately adding another option to `KubectlOptions` in `kubectl_options.go` would be a breaking change, as the signature of `NewKubectlOptions` would change.

We could avoid the breaking change, by simply falling back to in-cluster authentication, when `ContextName` and `ConfigPath` are unset. 

Do you have any preference or other ideas?

yorinasub17:
> I looked into it a bit more, unfortunately adding another option to KubectlOptions in kubectl_options.go would be a breaking change, as the signature of NewKubectlOptions would change.

I think we can introduce a new function `NewKubectlOptionsWithInClusterAuth` to maintain backwards compatibility, but the fallback approach also makes sense to me. Will try to review the PR today!

msvechla:
Awesome, let me know if you would prefer `NewKubectlOptionsWithInClusterAuth` over the current PR, then I will prepare one accordingly.

dee-kryvenko:
[...]

PR #901, opened 2021-05-16
Add new function to allow explicitly using in-cluster auth

cbuto:
Related to #753. 

This PR builds off of the work done in https://github.com/gruntwork-io/terratest/pull/760.

I took @yorinasub17's advice and implemented a new function `NewKubectlOptionsWithInClusterAuth` to make it a bit more straightforward for users to explicitly implement in-cluster authentication.

I also kept the initial fallback implementation to maintain backwards compatibility. 

Thanks for reviewing!

### DIFF
--- a/modules/k8s/client.go
+++ b/modules/k8s/client.go
@@ -28,20 +28,28 @@ func GetKubernetesClientFromOptionsE(t testing.TestingT, options *KubectlOptions
 	var err error
 	var config *rest.Config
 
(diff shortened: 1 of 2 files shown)


yorinasub17 (approved):
LGTM! I'll kick off a regression build, and if it passes, we can merge this in.

yorinasub17:
Regression build passed, so going to merge this in!

cbuto:
Thanks @yorinasub17!

PR #912, opened 2021-05-25
expand azure resource group module

### LINKED ISSUE #911: [AZURE] Expand Terratest resource group module
The current implementation of get resource group in terratest doesn't return the object itself. This change will change the module to 
- return a resource group object helper
- add helper to list resource groups by tags helper

HadwaAbdelhalem:
The current implementation of get resource group in terratest doesn't return the object itself. This change will change the module to

return a resource group object helper
add helper to list resource groups by tags helper

closes #911

### DIFF
--- a/modules/azure/resourcegroup.go
+++ b/modules/azure/resourcegroup.go
@@ -2,9 +2,10 @@ package azure
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
-	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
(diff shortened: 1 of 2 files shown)


yorinasub17 (approved):
Code changes LGTM! Can you comment with the CI build result?

HadwaAbdelhalem:
Thanks @yorinasub17  for reviewing. Here is the [CI](https://github.com/HadwaAbdelhalem/terratest/runs/2769168958?check_suite_focus=true) . it has a failure in the azure keuyvault tests  due to expired sample certificate used for the Key vault test and unrelated to this change.

PR #922, opened 2021-06-08
Implement ValidateAllTerraformModules function

zackproser:
- We need an easy way to generate a single test that runs validate
- It should always expand to catch all examples and modules

Relates to https://github.com/gruntwork-io/prototypes/pull/175

### DIFF
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -1,20 +1,25 @@
 package terraform
 
 import (
+	go_test "testing"
+
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/stretchr/testify/require"
 )
 
 // Automatically finds all folders specified in RootDir that contain .tf files and runs InitAndValidate in all of them
 // Excludes any folders specified in the ExcludeDirs slice
-func ValidateAllTerraformModules(t testing.TestingT, opts ValidationOptions) error {
+func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) error {
 	dirsToValidate, readErr := readModuleAndExampleSubDirs(opts)
 	if readErr != nil {
 		return readErr
 	}
 	for _, dir := range dirsToValidate {
-		tfOpts := &Options{TerraformDir: dir}
-		InitAndValidate(t, tfOpts)
+		t.Run(dir, func(t *go_test.T) {
+			t.Parallel()
+			tfOpts := &Options{TerraformDir: dir}
+			InitAndValidate(t, tfOpts)
+		})
 	}
 	return nil
 }
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -8,7 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// ValidateAllTerraformModules rutomatically finds all folders specified in RootDir that contain .tf files and runs
+// ValidateAllTerraformModules automatically finds all folders specified in RootDir that contain .tf files and runs
 // InitAndValidate in all of them.
 // Filters down to only those paths passed in ValidationOptions.IncludeDirs, if passed.
 // Excludes any folders specified in the ValidationOptions.ExcludeDirs. IncludeDirs will take precedence over ExcludeDirs
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -15,7 +15,7 @@ import (
 // Use the NewValidationOptions method to pass relative paths for either of these options to have the full paths built
 // Note that go_test is an alias to Golang's native testing package created to avoid naming conflicts with Terratest's
 // own testing package. We are using the native testing.T here because Terratest's testing.T struct does not implement Run
-func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) {
+func ValidateAllTerraformModules(t *go_test.T, opts *ValidationOptions) {
 	dirsToValidate, readErr := FindTerraformModulePathsInRootE(opts)
 	require.NoError(t, readErr)
 
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -2,6 +2,7 @@ package terraform
 
 import (
 	// We alias Golang's native testing package to go_test to avoid naming conflicts with terratest's own testing module
+
 	go_test "testing"
 
 	"github.com/gruntwork-io/terratest/modules/testing"
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -5,6 +5,20 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// Automatically finds all folders specified in RootDir that contain .tf files and runs InitAndValidate in all of them
+// Excludes any folders specified in the ExcludeDirs slice
+func ValidateAllTerraformModules(t testing.TestingT, opts ValidationOptions) error {
+	dirsToValidate, readErr := readModuleAndExampleSubDirs(opts)
+	if readErr != nil {
+		return readErr
+	}
+	for _, dir := range dirsToValidate {
+		tfOpts := &Options{TerraformDir: dir}
+		InitAndValidate(t, tfOpts)
+	}
+	return nil
+}
+
 // Validate calls terraform validate and returns stdout/stderr.
 func Validate(t testing.TestingT, options *Options) string {
 	out, err := ValidateE(t, options)
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -14,7 +14,7 @@ import (
 // Excludes any folders specified in the ValidationOptions.ExcludeDirs. IncludeDirs will take precedence over ExcludeDirs
 // Use the NewValidationOptions method to pass relative paths for either of these options to have the full paths built
 // Note that go_test is an alias to Golang's native testing package created to avoid naming conflicts with Terratest's
-// own testing package
+// own testing package. We are using the native testing.T here because Terratest's testing.T struct does not implement run
 func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) {
 	dirsToValidate, readErr := FindTerraformModulePathsInRootE(opts)
 	require.NoError(t, readErr)
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -1,19 +1,24 @@
 package terraform
 
 import (
+	// We alias Golang's native testing package to go_test to avoid naming conflicts with terratest's own testing module
 	go_test "testing"
 
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/stretchr/testify/require"
 )
 
-// Automatically finds all folders specified in RootDir that contain .tf files and runs InitAndValidate in all of them
-// Excludes any folders specified in the ExcludeDirs slice
-func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) error {
+// ValidateAllTerraformModules rutomatically finds all folders specified in RootDir that contain .tf files and runs
+// InitAndValidate in all of them.
+// Filters down to only those paths passed in ValidationOptions.IncludeDirs, if passed.
+// Excludes any folders specified in the ValidationOptions.ExcludeDirs. IncludeDirs will take precedence over ExcludeDirs
+// Use the NewValidationOptions method to pass relative paths for either of these options to have the full paths built
+// Note that go_test is an alias to Golang's native testing package created to avoid naming conflicts with Terratest's
+// own testing package
+func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) {
 	dirsToValidate, readErr := FindTerraformModulePathsInRootE(opts)
-	if readErr != nil {
-		return readErr
-	}
+	require.NoError(t, readErr)
+
 	for _, dir := range dirsToValidate {
 		dir := dir
 		t.Run(dir, func(t *go_test.T) {
@@ -22,7 +27,6 @@ func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) error {
 			InitAndValidate(t, tfOpts)
 		})
 	}
-	return nil
 }
 
 // Validate calls terraform validate and returns stdout/stderr.
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -10,7 +10,7 @@ import (
 // Automatically finds all folders specified in RootDir that contain .tf files and runs InitAndValidate in all of them
 // Excludes any folders specified in the ExcludeDirs slice
 func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) error {
-	dirsToValidate, readErr := readModuleAndExampleSubDirs(opts)
+	dirsToValidate, readErr := ReadModuleAndExampleSubDirs(opts)
 	if readErr != nil {
 		return readErr
 	}
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -14,7 +14,7 @@ import (
 // Excludes any folders specified in the ValidationOptions.ExcludeDirs. IncludeDirs will take precedence over ExcludeDirs
 // Use the NewValidationOptions method to pass relative paths for either of these options to have the full paths built
 // Note that go_test is an alias to Golang's native testing package created to avoid naming conflicts with Terratest's
-// own testing package. We are using the native testing.T here because Terratest's testing.T struct does not implement run
+// own testing package. We are using the native testing.T here because Terratest's testing.T struct does not implement Run
 func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) {
 	dirsToValidate, readErr := FindTerraformModulePathsInRootE(opts)
 	require.NoError(t, readErr)
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -10,11 +10,12 @@ import (
 // Automatically finds all folders specified in RootDir that contain .tf files and runs InitAndValidate in all of them
 // Excludes any folders specified in the ExcludeDirs slice
 func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) error {
-	dirsToValidate, readErr := ReadModuleAndExampleSubDirs(opts)
+	dirsToValidate, readErr := FindTerraformModulePathsInRootE(opts)
 	if readErr != nil {
 		return readErr
 	}
 	for _, dir := range dirsToValidate {
+		dir := dir
 		t.Run(dir, func(t *go_test.T) {
 			t.Parallel()
 			tfOpts := &Options{TerraformDir: dir}
--- a/modules/terraform/validate_struct.go
+++ b/modules/terraform/validate_struct.go
@@ -19,7 +19,7 @@ type ValidationOptions struct {
 	// Note that while the struct requires full paths, you can pass relative paths to the NewValidationOptions function
 	// which will build the full paths based on the supplied RootDir
 	IncludeDirs []string
-	// If you want to explicitly exclude certain sub directories, add them as paths relative to RootDir. For example, if the
+	// If you want to explicitly exclude certain sub directories of RootDir, add the absolute paths here. For example, if the
 	// RootDir is /home/project and you want to include everything EXCEPT /home/project/modules, add
 	// /home/project/modules to this slice. Note that ExcludeDirs is only considered when IncludeDirs is not passed
 	// Note that while the struct requires full paths, you can pass relative paths to the NewValidationOptions function
--- a/modules/terraform/validate_struct.go
+++ b/modules/terraform/validate_struct.go
@@ -29,7 +29,7 @@ type ValidationOptions struct {
 
 // NewValidationOptions returns a ValidationOptions struct, with override-able sane defaults. Note that the
 // ValidationOptions's fields IncludeDirs and ExcludeDirs must be absolute paths, but this method will accept relative paths
-// and build the full paths when instantiating the ValidationOptions struct,  making it the preferred means of configuring
+// and build the absolute paths when instantiating the ValidationOptions struct,  making it the preferred means of configuring
 // ValidationOptions.
 //
 // For example, if your RootDir is /home/project/ and you want to exclude "modules" and "test" you need
--- a/modules/terraform/validate_struct.go
+++ b/modules/terraform/validate_struct.go
@@ -68,7 +68,7 @@ func buildFullPathsFromRelative(rootDir string, relativePaths []string) []string
 }
 
 // FindTerraformModulePathsInRootE returns a slice strings representing the filepaths for all valid Terraform modules
-// in both the "modules" directory and "examples" directories in the project root, if they exist.
+// in the given RootDir, subject to the include / exclude filters.
 func FindTerraformModulePathsInRootE(opts ValidationOptions) ([]string, error) {
 	// Find all Terraform files from the configured RootDir
 	pattern := fmt.Sprintf("%s/**/*.tf", opts.RootDir)
--- a/modules/terraform/validate_struct.go
+++ b/modules/terraform/validate_struct.go
@@ -83,10 +83,9 @@ func ReadModuleAndExampleSubDirs(opts ValidationOptions) ([]string, error) {
 	return collections.ListSubtract(terraformModuleCandidates, opts.ExcludeDirs), nil
 }
 
-// filterTerraformModulesFromDirs accepts a slice of fs.DirEntry representing subDirectories
-// (under "modules" or "examples"), for instance, returning only those that contain a main.tf file in their root. This
-// is useful for filtering out any sub directories that might ship alongside Terraform modules, but actually be
-// Terraform modules themselves
+// IsTerraformModuleDirectory accepts a slice of string paths representing sub directories (under "modules" or "examples"),
+// for instance, returning true for any.tf files. This is useful for filtering out any sub directories that might ship
+// alongside Terraform modules, but not actually be Terraform modules themselves
 func IsTerraformModuleDirectory(path string) (bool, error) {
 	files, err := os.ReadDir(path)
 	if err != nil {
--- a/modules/terraform/validate_struct.go
+++ b/modules/terraform/validate_struct.go
@@ -1,0 +1,62 @@
+package terraform
+
+import (
+	"errors"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/gruntwork-io/terratest/modules/collections"
+)
+
+type ValidationOptions struct {
+	RootDir     string
+	ExcludeDirs []string
+}
+
+// readModuleAndExampleSubDirs returns a slice strings representing the filepaths for all valid Terraform modules
+// in both the "modules" directory and "examples" directories in the project root, if they exist.
+func readModuleAndExampleSubDirs(opts ValidationOptions) ([]string, error) {
+	if opts.RootDir == "" {
+		return nil, errors.New("RootDir must be defined in ValidationOptions passed to ValidateAllTerraformModules")
+	}
+	var validationCandidates []string
+	// We want to run InitAndValidate on all valid subdirectories of both the modules and examples dirs
+	modulesDir := filepath.Join(opts.RootDir, "modules")
+	examplesDir := filepath.Join(opts.RootDir, "examples")
+
+	moduleSubDirs, readModulesErr := os.ReadDir(modulesDir)
+	if readModulesErr != nil {
+		return nil, readModulesErr
+	}
+	exampleSubDirs, readExamplesErr := os.ReadDir(examplesDir)
+	if readExamplesErr != nil {
+		return nil, readExamplesErr
+	}
+
+	for _, m := range filterTerraformModulesFromDirs(modulesDir, moduleSubDirs) {
+		validationCandidates = append(validationCandidates, m)
+	}
+
+	for _, e := range filterTerraformModulesFromDirs(examplesDir, exampleSubDirs) {
+		validationCandidates = append(validationCandidates, e)
+	}
+
+	// Filter out any filepaths that were explicitly included in opts.ExcludeDirs
+	return collections.ListSubtract(validationCandidates, opts.ExcludeDirs), nil
+}
+
+// filterTerraformModulesFromDirs accepts a slice of fs.DirEntry representing subDirectories
+// (under "modules" or "examples"), for instance, returning only those that contain a main.tf file in their root. This
+// is useful for filtering out any sub directories that might ship alongside Terraform modules, but actually be
+// Terraform modules themselves
+func filterTerraformModulesFromDirs(rootDir string, subDirs []fs.DirEntry) []string {
+	var validTerraformModules []string
+	for _, m := range subDirs {
+		maybeMainTf := filepath.Join(rootDir, m.Name(), "main.tf")
+		if _, err := os.Stat(maybeMainTf); err == nil {
+			validTerraformModules = append(validTerraformModules, filepath.Join(rootDir, m.Name()))
+		}
+	}
+	return validTerraformModules
+}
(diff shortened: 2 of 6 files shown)


modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,62 @@
...
+type ValidationOptions struct {
+	RootDir     string
+	ExcludeDirs []string
+}
+
+// readModuleAndExampleSubDirs returns a slice strings representing the filepaths for all valid Terraform modules
+// in both the "modules" directory and "examples" directories in the project root, if they exist.
+func readModuleAndExampleSubDirs(opts ValidationOptions) ([]string, error) {
+	if opts.RootDir == "" {
+		return nil, errors.New("RootDir must be defined in ValidationOptions passed to ValidateAllTerraformModules")
```
yorinasub17, line 21: NIT: Use a concrete error type

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type ValidationOptions struct {
+	RootDir     string
+	ExcludeDirs []string
+}
+
+// readModuleAndExampleSubDirs returns a slice strings representing the filepaths for all valid Terraform modules
+// in both the "modules" directory and "examples" directories in the project root, if they exist.
+func readModuleAndExampleSubDirs(opts ValidationOptions) ([]string, error) {
```
yorinasub17, line 19: This can be a public function. I think this would be useful for other things.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,62 @@
...
+	examplesDir := filepath.Join(opts.RootDir, "examples")
+
+	moduleSubDirs, readModulesErr := os.ReadDir(modulesDir)
+	if readModulesErr != nil {
+		return nil, readModulesErr
+	}
+	exampleSubDirs, readExamplesErr := os.ReadDir(examplesDir)
+	if readExamplesErr != nil {
+		return nil, readExamplesErr
+	}
```
yorinasub17, line 35: Does this handle nested module/example folders? E.g., the service catalog and `terraform-aws-monitoring`.

As written, I think it will fail to work for those but I might be wrong.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,62 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/gruntwork-io/terratest/modules/collections"
+)
+
+type ValidationOptions struct {
+	RootDir     string
+	ExcludeDirs []string
+}
```
yorinasub17, line 15: Add comments. Specifically, does `ExcludeDirs` need to be absolute, or relative to `RootDir`, or relative to current working dir to work?

modules/terraform/validate_test.go
```diff
@@ -35,3 +38,57 @@ func TestInitAndValidateWithError(t *testing.T) {
...
+	require.NoError(t, cwdErr)
+
+	opts := ValidationOptions{
+		RootDir:     filepath.Join(cwd, "../../"),
+		ExcludeDirs: []string{},
+	}
+
+	subDirs, err := readModuleAndExampleSubDirs(opts)
+	require.NoError(t, err)
+	// There are many valid Terraform modules in the root/examples directory of the Terratest project, so we should get back 0 results
```
yorinasub17, line 63: NIT:
```suggestion
	// There are many valid Terraform modules in the root/examples directory of the Terratest project, so we should get back many results
```

modules/terraform/validate_test.go
```diff
@@ -35,3 +38,57 @@ func TestInitAndValidateWithError(t *testing.T) {
...
+	opts := ValidationOptions{
+		RootDir:     projectRootDir,
+		ExcludeDirs: exclusions,
+	}
+
+	subDirs, err := readModuleAndExampleSubDirs(opts)
+	require.NoError(t, err)
+	require.Greater(t, len(subDirs), 0)
+	for _, exclusion := range exclusions {
+		require.False(t, collections.ListContains(subDirs, exclusion))
```
yorinasub17, line 92: Use `assert.False` here so that you get all the included failures at once, instead of it halting on the first failure.

yorinasub17, line 92: One issue with this test is that it doesn't confirm that the exclusions are included in the first place. Consider adding a step that calls the function WITHOUT exclusions, and then verifying the exclusion list is included in the results.

Otherwise, if we ever rename or remove examples, this test will still pass as a tautology.

modules/terraform/validate_test.go
```diff
@@ -35,3 +38,57 @@ func TestInitAndValidateWithError(t *testing.T) {
...
+	}
+}
```
yorinasub17, line 94: Consider replacing these tests with fixtures instead of using the terratest repo to make it robust to reorganizations in the terratest repo.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,62 @@
...
+
+// filterTerraformModulesFromDirs accepts a slice of fs.DirEntry representing subDirectories
+// (under "modules" or "examples"), for instance, returning only those that contain a main.tf file in their root. This
+// is useful for filtering out any sub directories that might ship alongside Terraform modules, but actually be
+// Terraform modules themselves
+func filterTerraformModulesFromDirs(rootDir string, subDirs []fs.DirEntry) []string {
+	var validTerraformModules []string
+	for _, m := range subDirs {
+		maybeMainTf := filepath.Join(rootDir, m.Name(), "main.tf")
```
yorinasub17, line 56: NIT: Consider looking for any `tf` file (not just `main.tf`) to enhance flexibility.

zackproser:
Thanks for the detailed review @yorinasub17! Will implement your feedback.

modules/terraform/validate.go
```diff
@@ -5,6 +5,20 @@ import (
...
+// Automatically finds all folders specified in RootDir that contain .tf files and runs InitAndValidate in all of them
+// Excludes any folders specified in the ExcludeDirs slice
+func ValidateAllTerraformModules(t testing.TestingT, opts ValidationOptions) error {
+	dirsToValidate, readErr := readModuleAndExampleSubDirs(opts)
+	if readErr != nil {
+		return readErr
+	}
+	for _, dir := range dirsToValidate {
+		tfOpts := &Options{TerraformDir: dir}
+		InitAndValidate(t, tfOpts)
```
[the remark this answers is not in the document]
zackproser: Fixed in 746c760

modules/terraform/validate_struct.go
zackproser: Done in d1515bc

modules/terraform/validate_test.go
zackproser: Fixed in 9519b48

zackproser: Done in 9519b48

modules/terraform/validate_struct.go
zackproser: Clarified in 9519b48

modules/terraform/validate_struct.go
zackproser: Done in 9519b48

modules/terraform/validate_struct.go
zackproser: Made public in 9519b48

modules/terraform/validate_struct.go
zackproser: Recursive walking of directories added in 9519b48

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,62 @@
...
+// readModuleAndExampleSubDirs returns a slice strings representing the filepaths for all valid Terraform modules
+// in both the "modules" directory and "examples" directories in the project root, if they exist.
+func readModuleAndExampleSubDirs(opts ValidationOptions) ([]string, error) {
+	if opts.RootDir == "" {
+		return nil, errors.New("RootDir must be defined in ValidationOptions passed to ValidateAllTerraformModules")
+	}
+	var validationCandidates []string
+	// We want to run InitAndValidate on all valid subdirectories of both the modules and examples dirs
+	modulesDir := filepath.Join(opts.RootDir, "modules")
+	examplesDir := filepath.Join(opts.RootDir, "examples")
```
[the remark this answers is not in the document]
zackproser: Done in 9519b48

modules/terraform/validate_test.go
zackproser: Good call. Removed the usage of examples since the same functionality is just as thoroughly flexed by targeting existing test fixtures. ee73e61

modules/terraform/validate.go
```diff
@@ -1,10 +1,29 @@
...
 	"github.com/stretchr/testify/require"
 )
 
+// Automatically finds all folders specified in RootDir that contain .tf files and runs InitAndValidate in all of them
+// Excludes any folders specified in the ExcludeDirs slice
+func ValidateAllTerraformModules(t *go_test.T, opts ValidationOptions) error {
+	dirsToValidate, readErr := ReadModuleAndExampleSubDirs(opts)
+	if readErr != nil {
+		return readErr
+	}
```
yorinasub17, line 16: Don't return error and use `require.NoError`. Technically, you should introduce a `ValidateAllTerraformModulesE` (note the `E` suffix) function that returns errors, but that will be hard to do with the subtest design of this function, so we should just introduce the non-E flavor only here.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,99 @@
...
+		return vo, ValidationUndefinedRootDirErr{}
+	}
+	vo.RootDir = rootDir
+
+	// If no target sub directories are passed, default to recursively searching "modules" and "examples"
+	if len(targetSubDirs) == 0 {
+		vo.TargetSubDirs = []string{
+			"modules",
+			"examples",
+		}
```
yorinasub17, line 37: In general, I prefer explicit over implicit. In that world, would be better to throw an error here.

From there, either provide a global var that includes these two that users can use (e.g., `DefaultTargetSubDirs`), or a separate function (e.g., `NewValidationOptionsWithDefaults`).

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+type ValidationOptions struct {
+	RootDir string
+	//Sub-directories relative to the RootDir to walk recursively for all Terraform modules. For example, if your
+	// RootDir is /home/project and you want to validate all modules in home/project/terraform then pass "terraform"
+	// in TargetSubDirs
+	TargetSubDirs []string
+	//ExcludeDirs must be relative to the RootDir. For example if your RootDir is home/project and you want to exclude
+	// the path /home/project/test, pass "test" in ExcludeDirs
```
yorinasub17, line 17: It looks like this has to be the full path in the struct, but need to be relative in `NewValidationOptions`. In that case, this doc should mention that it needs to be full path, and the `NewValidationOptions` function will take relative paths and turn it into full paths.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return vo, nil
+}
+
+type ValidationUndefinedRootDirErr struct{}
+
+func (e ValidationUndefinedRootDirErr) Error() string {
+	return "RootDir must be defined in ValidationOptions passed to ValidateAllTerraformModules"
+}
```
yorinasub17, line 57: NIT: Push custom errors to the bottom of the file and denote section with `// Custom error types`.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,99 @@
...
+
+// readModuleAndExampleSubDirs returns a slice strings representing the filepaths for all valid Terraform modules
+// in both the "modules" directory and "examples" directories in the project root, if they exist.
+func ReadModuleAndExampleSubDirs(opts ValidationOptions) ([]string, error) {
```
yorinasub17, line 61: NIT: public functions in terratest that returns errors should have `E` suffix.

yorinasub17 (commented):
Very close to being done! Just had a few more nits.

modules/terraform/validate.go
```diff
@@ -1,10 +1,29 @@
...
+	for _, dir := range dirsToValidate {
```
brikis98, line 17: Don't you need to do `dir := dir`?

modules/terraform/validate.go
```diff
@@ -1,10 +1,29 @@
...
+		t.Run(dir, func(t *go_test.T) {
+			t.Parallel()
+			tfOpts := &Options{TerraformDir: dir}
```
brikis98, line 20: Do you have to copy over any settings from the original `Options` the user passed in?

modules/terraform/validate.go
brikis98: Heh, was going to say the same thing. Perhaps the `E` version of this method is a bit different and returns a list of test cases to run? And then the non `E` version loops over those test cases and calls `t.Run` on each one?

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/gruntwork-io/terratest/modules/collections"
+)
+
+type ValidationOptions struct {
+	RootDir string
+	//Sub-directories relative to the RootDir to walk recursively for all Terraform modules. For example, if your
+	// RootDir is /home/project and you want to validate all modules in home/project/terraform then pass "terraform"
+	// in TargetSubDirs
+	TargetSubDirs []string
```
brikis98, line 15: Are you going for include / exclude semantics with `TargetSubDirs` and `ExcludeDirs`? If so, then you may want to name them accordingly and update the docs to match.

modules/terraform/validate_struct.go
brikis98: I don't think we need to target `modules` and `examples` at all. I think the default should be to find all `.tf` files and run `validate` in the corresponding folders. If someone wants to target specific subfolders, then we can support include/exclude semantics in `Options`.

modules/terraform/validate_struct.go
brikis98: This function should be called something more generic: e.g., `FindTerraformModulePathsInRoot`

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,99 @@
...
+	return "RootDir must be defined in ValidationOptions passed to ValidateAllTerraformModules"
+}
+
+// readModuleAndExampleSubDirs returns a slice strings representing the filepaths for all valid Terraform modules
+// in both the "modules" directory and "examples" directories in the project root, if they exist.
+func ReadModuleAndExampleSubDirs(opts ValidationOptions) ([]string, error) {
+	var terraformModuleCandidates []string
+	// We want to run InitAndValidate on all valid subdirectories of both the modules and examples dirs
+
+	for _, dir := range opts.TargetSubDirs {
```
brikis98, line 65: Any reason not to use [glob](https://github.com/mattn/go-zglob) with `**/*.tf`?

modules/terraform/validate_struct.go
zackproser: Done in bc3031a

modules/terraform/validate_struct.go
zackproser: Done in c3e36ef

modules/terraform/validate_struct.go
zackproser: Seems to work well - done in c3e36ef

modules/terraform/validate_struct.go
zackproser: Done in 2ed0bb7

modules/terraform/validate.go
zackproser: Made the changes Yori propsoed in 2799cb1. Still thinking through Jim's suggestion.

modules/terraform/validate.go
zackproser: Added!

modules/terraform/validate.go
zackproser: I was originally thinking of support that - but I haven't given it deep thought yet. Right now we're only accepting the ValidationOptions. I figured most would follow our intended pattern of creating a single slim Go test that calls this method to run validate on every Terraform directory within their repo.

modules/terraform/validate.go
```diff
@@ -1,10 +1,34 @@
...
 	"github.com/stretchr/testify/require"
 )
 
+// ValidateAllTerraformModules rutomatically finds all folders specified in RootDir that contain .tf files and runs
+// InitAndValidate in all of them.
+// Filters down to only those paths passed in ValidationOptions.IncludeDirs, if passed.
+// Excludes any folders specified in the ValidationOptions.ExcludeDirs. IncludeDirs will take precedence over ExcludeDirs
+// Use the NewValidationOptions method to pass relative paths for either of these options to have the full paths built
+// Note that go_test is an alias to Golang's native testing package created to avoid naming conflicts with Terratest's
+// own testing package
```
yorinasub17, line 17: NIT: Mention why we are using the native testing.T and not the terratest testing.T

modules/terraform/validate.go
yorinasub17: I think that can be left as a future PR consideration based on usage in the wild.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,103 @@
...
+	// If you want to explicitly exclude certain sub directories, add them as paths relative to RootDir. For example, if the
+	// RootDir is /home/project and you want to include everything EXCEPT /home/project/modules, add
+	// /home/project/modules to this slice. Note that ExcludeDirs is only considered when IncludeDirs is not passed
+	// Note that while the struct requires full paths, you can pass relative paths to the NewValidationOptions function
+	// which will build the full paths based on the supplied RootDir
+	ExcludeDirs []string
+}
+
+// NewValidationOptions returns a ValidationOptions struct, with override-able sane defaults
+func NewValidationOptions(rootDir string, includeDirs, excludeDirs []string) (ValidationOptions, error) {
```
yorinasub17, line 31: NIT: Add comment indicating the `includeDirs` and `excludeDirs` are relative to the `rootDir`. I know you already mention this in the struct docs, but it is useful to mention it here was well, since some users might only look at the godocs for this function.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,103 @@
...
+
+func buildFullPathsFromRelative(rootDir string, relativePaths []string) []string {
+	var fullPaths []string
+	for _, relativePath := range relativePaths {
+		fullPaths = append(fullPaths, filepath.Join(rootDir, relativePath))
+	}
+	return fullPaths
+}
+
+// readModuleAndExampleSubDirs returns a slice strings representing the filepaths for all valid Terraform modules
```
yorinasub17, line 63:
```suggestion
// FindTerraformModulePathsInRootE returns a slice strings representing the filepaths for all valid Terraform modules
```

yorinasub17 (approved):
Looks good! Only had docs nits.

modules/terraform/validate.go
zackproser: Done in 95eb0a9

modules/terraform/validate_struct.go
zackproser: Makes sense - done in 95eb0a9

modules/terraform/validate.go
zackproser: Cool, makes sense to me.

modules/terraform/validate.go
```diff
@@ -1,10 +1,34 @@
...
 	"github.com/stretchr/testify/require"
 )
 
+// ValidateAllTerraformModules rutomatically finds all folders specified in RootDir that contain .tf files and runs
+// InitAndValidate in all of them.
+// Filters down to only those paths passed in ValidationOptions.IncludeDirs, if passed.
+// Excludes any folders specified in the ValidationOptions.ExcludeDirs. IncludeDirs will take precedence over ExcludeDirs
+// Use the NewValidationOptions method to pass relative paths for either of these options to have the full paths built
+// Note that go_test is an alias to Golang's native testing package created to avoid naming conflicts with Terratest's
+// own testing package. We are using the native testing.T here because Terratest's testing.T struct does not implement run
```
zackproser, line 17:
```suggestion
// own testing package. We are using the native testing.T here because Terratest's testing.T struct does not implement Run
```

modules/terraform/validate.go
```diff
@@ -1,10 +1,34 @@
...
 
 import (
+	// We alias Golang's native testing package to go_test to avoid naming conflicts with terratest's own testing module
+	go_test "testing"
+
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/stretchr/testify/require"
 )
 
+// ValidateAllTerraformModules rutomatically finds all folders specified in RootDir that contain .tf files and runs
```
brikis98, line 11:
```suggestion
// ValidateAllTerraformModules automatically finds all folders specified in RootDir that contain .tf files and runs
```

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+
+type ValidationOptions struct {
+	// The target directory to recursively search for all Terraform directories (those that contain .tf files)
+	// If you provide RootDir and do not pass entries in either IncludeDirs or ExcludeDirs, then all Terraform directories
+	// From the RootDir, recursively, will be validated
+	RootDir string
+	// If you only want to include certain sub directories, add them as paths relative to RootDir. For example, if the
+	// RootDir is /home/project and you want to only include /home/project/examples, add /home/project/examples here
+	// Note that while the struct requires full paths, you can pass relative paths to the NewValidationOptions function
+	// which will build the full paths based on the supplied RootDir
```
brikis98, line 20:
```suggestion
	// If you only want to include certain sub directories of RootDir, add the absolute paths here. For example, if the
	// RootDir is /home/project and you want to only include /home/project/examples, add /home/project/examples here
	// Note that while the struct requires full paths, you can pass relative paths to the NewValidationOptions function
	// which will build the full paths based on the supplied RootDir
```

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+	IncludeDirs []string
+	// If you want to explicitly exclude certain sub directories, add them as paths relative to RootDir. For example, if the
+	// RootDir is /home/project and you want to include everything EXCEPT /home/project/modules, add
+	// /home/project/modules to this slice. Note that ExcludeDirs is only considered when IncludeDirs is not passed
+	// Note that while the struct requires full paths, you can pass relative paths to the NewValidationOptions function
+	// which will build the full paths based on the supplied RootDir
```
brikis98, line 26:
```suggestion
	// If you want to explicitly exclude certain sub directories of RootDir, add the absolute paths here. For example, if the
	// RootDir is /home/project and you want to include everything EXCEPT /home/project/modules, add
	// /home/project/modules to this slice. Note that ExcludeDirs is only considered when IncludeDirs is not passed
	// Note that while the struct requires full paths, you can pass relative paths to the NewValidationOptions function
	// which will build the full paths based on the supplied RootDir
```

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+	ExcludeDirs []string
+}
+
+// NewValidationOptions returns a ValidationOptions struct, with override-able sane defaults. Note that the
+// ValidationOptions's fields IncludeDirs and ExcludeDirs must be full paths, but this method will accept relative paths
```
brikis98, line 31:
```suggestion
// ValidationOptions's fields IncludeDirs and ExcludeDirs must be absolute paths, but this method will accept relative paths
```

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+// and build the full paths when instantiating the ValidationOptions struct,  making it the preferred means of configuring
```
brikis98, line 32:
```suggestion
// and build the absolute paths when instantiating the ValidationOptions struct,  making it the preferred means of configuring
```

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+// ValidationOptions.
+//
+// For example, if your RootDir is /home/project/ and you want to exclude "modules" and "test" you need
+// only pass the relative paths in your excludeDirs slice like so:
+// opts, err := NewValidationOptions("/home/project", []string{}, []string{"modules", "test"})
+func NewValidationOptions(rootDir string, includeDirs, excludeDirs []string) (ValidationOptions, error) {
```
brikis98, line 38: NIT: We typically return a pointer to structs we are instantiating. That allows you to do `return nil, err` in case of errors.

modules/terraform/validate_struct.go
brikis98: > Note that ExcludeDirs is only considered when IncludeDirs is not passed

Why? The typical idiom with include / exclude is to:

1. Start with a list of included dirs. If the include list passed in by the user is empty, then your initial included dirs list is all subdirs that have `.tf` files.
1. Subtract the excluded dirs.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+		vo.ExcludeDirs = buildFullPathsFromRelative(vo.RootDir, excludeDirs)
+	}
+
+	return vo, nil
+}
+
+func buildFullPathsFromRelative(rootDir string, relativePaths []string) []string {
+	var fullPaths []string
+	for _, relativePath := range relativePaths {
+		fullPaths = append(fullPaths, filepath.Join(rootDir, relativePath))
```
brikis98, line 65: Check if the relative path is, in fact, relative, with `filepath.IsAbs`. If it's already an absolute path, don't change it.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return fullPaths
+}
+
+// FindTerraformModulePathsInRootE returns a slice strings representing the filepaths for all valid Terraform modules
+// in both the "modules" directory and "examples" directories in the project root, if they exist.
```
brikis98, line 71:
```suggestion
// FindTerraformModulePathsInRootE returns a slice strings representing the filepaths for all valid Terraform modules
// in the given RootDir, subject to the include / exclude filters.
```

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return matches, err
+	}
+	// Keep a unique set of the base dirs that contain Terraform files
+	terraformDirSet := make(map[string]bool)
+	for _, match := range matches {
+		// The glob match returns all full paths to every .tf file, whereas we're only interested in their root
+		// directories for the purposes of running Terraform validate
+		rootDir := path.Dir(match)
+		terraformDirSet[rootDir] = true
```
brikis98, line 85: For this to work, all dirs must be _canonical_ paths. That ensures the same path doesn't appear twice, once, say, with `../../foo` and once with `/foo`. So perhaps `NewValidationOptions` should call `filepath.Abs` (if not already absolute) and `filepath.Clean` on `RootDir` and all the include/exclude paths.

modules/terraform/validate_struct.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// Return the unique slice of Terraform directories found starting at opts.RootDir
+	terraformDirs := make([]string, 0, len(terraformDirSet))
+	for dir := range terraformDirSet {
+		terraformDirs = append(terraformDirs, dir)
+	}
```
brikis98, line 92: Is this not the same as:

```go
terraformDirs := keys(terraformDirSet)
```

modules/terraform/validate_struct.go
zackproser: > > Note that ExcludeDirs is only considered when IncludeDirs is not passed
> 
> Why? The typical idiom with include / exclude is to:
> 
>     1. Start with a list of included dirs. If the include list passed in by the user is empty, then your initial included dirs list is all subdirs that have `.tf` files.
> 
>     2. Subtract the excluded dirs.

This is actually how it works, so I think my description is confusing: 

1. Default behavior (if you pass neither IncludeDirs nor ExcludeDirs) is that all subdirs that have `.tf` files are found 
2. If you pass IncludeDirs, all subdirs that have `.tf` files are found, and only the intersection of IncludeDirs and all subdirs that have `.tf` files is returned. 
3. If you pass ExcludeDirs, all the subdirs that have `.tf` files are found, minus those included in ExcludeDirs. 

My main point is that if you for some reason pass IncludeDirs and ExcludeDirs, the code will return after processing your IncludeDirs. In other words, you should pass one or the other to achieve your desired targeting behavior.

zackproser: I'll update the comments to be more clear on this point.

brikis98: I think what you have is:

```
foldersToProcess = terraformModulesIn(rootDir)

if includeDirs
  return intersection(foldersToProcess, includeDirs)
else if excludeDirs
  return difference(foldersToProcess, excludeDirs)
else 
  return foldersToProcess
```

The more idiomatic include / exclude implementation is a bit different:

```
foldersToProcess = terraformModulesIn(rootDir)

if includeDirs
  foldersToProcess = intersection(foldersToProcess, includeDirs)

if excludeDirs
  foldersToProcess = difference(foldersToProcess, excludeDirs)

return foldersToProcess
```

modules/terraform/validate_struct.go
zackproser: I was looking for this `keys` method before I settled on the above implementation, but my current understanding is Go doesn't support this. What am I missing?

brikis98: Oh, sorry, brain fart. I thought Go had a native `keys` method.

`go-commons` does have this: https://github.com/gruntwork-io/go-commons/blob/master/collections/maps.go#L19. Not sure we use `go-commons` in Terratest yet?

modules/terraform/validate_struct.go
zackproser: Gotcha. Thanks for the explanation. Fixed in af208cc

modules/terraform/validate_struct.go
zackproser: Roger - fixed in 861ab23

modules/terraform/validate_struct.go
zackproser: Done in a1e0206

modules/terraform/validate_struct.go
zackproser: Roger. Implemented in d7e0bbe and tests are still passing afterwards.

modules/terraform/validate_struct.go
zackproser: Seems we do make liberal use of it - so cut over to it in 70e6c2d. Required slight non-material modification of the map type to satisfy the method signature of `Keys`.

brikis98 (commented):
Code LGTM... And I like the unit tests for `FindTerraformModulePathsInRoot`... But are there no tests for `ValidateAllTerraformModules` itself? We probably want some tests to check that it works on valid code and returns an error on invalid code. And, perhaps, we should call it on all the Terraform code in the Terratest repo itself?

zackproser:
> Code LGTM... And I like the unit tests for `FindTerraformModulePathsInRoot`... But are there no tests for `ValidateAllTerraformModules` itself? We probably want some tests to check that it works on valid code and returns an error on invalid code. And, perhaps, we should call it on all the Terraform code in the Terratest repo itself?

Thanks! Will add unit tests for `ValidateAllTerraformModules` and perhaps a separate one that calls `ValidateAllTerraformModules` against the Terratest repo root itself.

zackproser:
Added more unit test coverage in f3c1f44 and began running `ValidateAllTerraformModules` on Terratest itself. As noted in the commit message, we don't currently have a good way to prove that invalid Terraform code is caught by `ValidateAllTerraformModules` because of the method's use of `require.NoError` however we can be pretty confident given that we see validation failures when it's run against repos and it is a thin wrapper around Terraform `init` and `validate` and it fails as expected in local testing against invalid modules.

brikis98 (approved):
LGTM!

PR #925, opened 2021-06-10
Adapt count parser to tf 0.15.5

yorinasub17:
The plan output for no changes changed in tf 0.15.5, so we need to update the regex.


brikis98 (approved):
LGTM

yorinasub17:
Thanks for review! Build failed, but upon investigation they are transient errors unrelated to this change set, so will go ahead and merge this in!

PR #927, opened 2021-06-11
Fix regression bug using run-all in lock commands

yorinasub17:
Because we were including `run-all` in `TerraformCommandsWithLockSupport`, it was always appending `-lock=false` for any `run-call` command, even if the underlying tf command didn't support it (e.g., `run-all validate`).

This PR fixes that.


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Merging and releasing.

PR #930, opened 2021-06-15
Continue searching when we get a deleted bucket

### LINKED ISSUE #928: FindS3BucketWithTag can fail by listing tags on a deleted bucket
Occassionally, FindS3BucketWithTag will fail like this:
```
Error: Received unexpected error:
          NoSuchBucket: The specified bucket does not exist
          status code: 404, request id: Y4FJTRQYC6GEKP13, host id: U8Mu38...
```
It appears that this is because the ListBuckets API can return a bucket that has been deleted, but hasn't been removed from S3 yet: https://aws.amazon.com/premiumsupport/knowledge-center/s3-listing-deleted-bucket/

Terratest Version: `v0.35.6`

gigyas:
Check to see if the error we get is "NoSuchBucket"
Sometimes, ListBuckets will return buckets that have been deleted, but haven't actually been removed yet. This causes GetBucketTagging to return NoSuchBucket, and the function to return without checking any remaining buckets (which could be the one we actually want).
If we get "NoSuchBucket", just continue. A bucket that has been deleted can't be the one we are looking for (at least, it wouldn't be useful if it was). 

Fixes #928 

I'm not quite sure how to test this, as it it a transient error related to how S3 is working behind the scenes. It happened to me once in several days of use, but the issue persisted for an hour or more. The "phantom" bucket was gone the next morning.

### DIFF
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -37,6 +37,9 @@ func FindS3BucketWithTagE(t testing.TestingT, awsRegion string, key string, valu
 		tagResponse, err := s3Client.GetBucketTagging(&s3.GetBucketTaggingInput{Bucket: bucket.Name})
 
 		if err != nil {
+			if strings.Contains(err.Error(), "NoSuchBucket") {
+				continue
+			}
 			if !strings.Contains(err.Error(), "AuthorizationHeaderMalformed") &&
 				!strings.Contains(err.Error(), "BucketRegionError") &&
 				!strings.Contains(err.Error(), "NoSuchTagSet") {
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -38,6 +38,10 @@ func FindS3BucketWithTagE(t testing.TestingT, awsRegion string, key string, valu
 
 		if err != nil {
 			if strings.Contains(err.Error(), "NoSuchBucket") {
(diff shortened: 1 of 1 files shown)


modules/aws/s3.go
```diff
@@ -37,6 +37,9 @@ func FindS3BucketWithTagE(t testing.TestingT, awsRegion string, key string, valu
 		tagResponse, err := s3Client.GetBucketTagging(&s3.GetBucketTaggingInput{Bucket: bucket.Name})
 
 		if err != nil {
+			if strings.Contains(err.Error(), "NoSuchBucket") {
+				continue
+			}
```
brikis98, line 42: NIT: please add as a comment the description from the PR explaining why this error is safe to continue.

brikis98 (commented):
Thanks for the PR! Just one NIT to fix and then I can kick off tests.

modules/aws/s3.go
gigyas: Comment added

brikis98 (approved):
LGTM, thank you! I'll kick off tests now.

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terratest/releases/tag/v0.36.0

PR #931, opened 2021-06-16
Fix terraform validation issues

zackproser:
Fixes for the minor issues preventing a few modules from passing `terraform validate` as uncovered by the new `TestValidateAllTerraformModulesOnTerratest` test.

As can be seen in[ this test run](https://circleci.com/api/v1.1/project/github/gruntwork-io/terratest/5871/output/106/0?file=true&allocation-id=60ca4ed611a6f444d04271a5-0-build%2F2F80911B), all `TestValidateAllTerraformModulesOnTerratest` sub tests are now passing.


test/terraform_aws_s3_example_test.go
```diff
@@ -36,6 +36,8 @@ func TestTerraformAwsS3Example(t *testing.T) {
 			"tag_bucket_name":        expectedName,
 			"tag_bucket_environment": expectedEnvironment,
 			"with_policy":            "true",
+			// Passing region to Terraform here is required for the terraform validate command to pass
+			"region": awsRegion,
```
yorinasub17, line 40: If we're now setting the region this way, we should remove the `AWS_DEFAULT_REGION` env var configuration down below.

yorinasub17 (commented):
Had one more consistency fix to suggest. Once that is addressed, and the build is passing, this is good to go!

examples/terraform-aws-ecs-example/main.tf
```diff
@@ -2,6 +2,9 @@
 # PIN TERRAFORM VERSION TO >= 0.12
 # The examples have been upgraded to 0.12 syntax
 # ---------------------------------------------------------------------------------------------------------------------
+provider "aws" {
+  region = "us-east-1"
```
rhoboat, line 6: Did we want to make this a var instead of hardcoded string?

zackproser: Yes - thanks!

test/terraform_aws_s3_example_test.go
zackproser: Fixed in f5c6a3f

examples/terraform-aws-ecs-example/main.tf
zackproser: Fixed in f5c6a3f

yorinasub17 (approved):
LGTM, assuming the build passes!

zackproser:
> LGTM, assuming the build passes!

Thanks both for the reviews!

yorinasub17 (approved):
Approving, assuming build passes.

rhoboat (approved):

brikis98 (approved):

PR #935, opened 2021-06-18
Copy validated Terraform to /tmp to avoid conflict

zackproser:
- Some tests do not copy their Terraform modules to `/tmp` first, which can lead to conflicts when `ValidateAllTerraformModules` is called 
- Avoid conflicts by always copying Terraform modules to be validated to `/tmp`
- Fix an issue causing `FindTerraformModulePathsInRootE` to return `.terraform` hidden directories, leading to duplicate work and possible false positives.

### DIFF
--- a/modules/terraform/validate.go
+++ b/modules/terraform/validate.go
@@ -3,8 +3,10 @@ package terraform
 import (
 	// We alias Golang's native testing package to go_test to avoid naming conflicts with terratest's own testing module
 
+	"path/filepath"
 	go_test "testing"
 
+	"github.com/gruntwork-io/terratest/modules/files"
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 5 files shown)


modules/terraform/validate.go
```diff
@@ -24,7 +26,12 @@ func ValidateAllTerraformModules(t *go_test.T, opts *ValidationOptions) {
 		dir := dir
 		t.Run(dir, func(t *go_test.T) {
 			t.Parallel()
-			tfOpts := &Options{TerraformDir: dir}
+
+			//Copy the module to a tmp directory to avoid conflicts with tests that don't copy to /tmp
+			testFolder, err := files.CopyTerraformFolderToTemp(opts.RootDir, filepath.Base(dir))
```
zackproser, line 31: Note that, due to an import cycle error, I had to use the `files` package's `CopyTerraformFolderToTemp` method. `test_structure` imports the `terraform` package.

yorinasub17: Is this logic correct? I think this is wrong for cases where there is a subdir (e.g., `modules/agents/cloudwatch-agent`).

I think you need to use `filepath.Rel` here.

zackproser: Thanks! Revised as discussed.

yorinasub17 (approved):
LGTM assuming the `test` build passes (you can ignore the failing `gcp_test`). I had two nits, but they can be addressed in a subsequent PR.

Make sure to cut a backward incompatible release when you release this!

zackproser:
Thanks for the review! Going to merge this in now but will address your nits in a follow-up PR.

PR #938, opened 2021-06-23
Don't run validate tests in parallel

zackproser:
Terraform validate spins up lots of providers, each of which is CPU and memory hungry. Running these in parallel currently exhausts even beefy machines, consuming over 7GB of ram and pegging all CPUs. 

Removing the call to `parallel` resolves the problem locally, which I was able to repro prior to removal.


bwhaley (approved):
LGTM!

zackproser:
> LGTM!

Thanks for the review!

Issue #937, opened 2021-06-22
Namespace not passed to RunHelmCommandAndGetOutputE/RunHelmCommandAndGetStdOutE function


lperrin-obs:
I'm using terratest to make some test of kubernetes clusters.

I created this function to list the helm charts installed inside a specific namespace 

```
package testk8s

import (
	"fmt"
	"testing"

	"github.com/gruntwork-io/terratest/modules/helm"
	"github.com/gruntwork-io/terratest/modules/k8s"
	"github.com/stretchr/testify/require"
)

var monitoringChartName = "rancher-monitoring"

func TestMonitoringChart(t *testing.T) {
	t.Parallel()

	// Setup the kubectl config and context.
	kubectlOptions := k8s.NewKubectlOptions("dev-cluster", "/home/USER/.kube/dev-cluster", "dev-namespace")

	// Setup helm client
	helmOptions := &helm.Options{KubectlOptions: kubectlOptions}


	// Run helm to list chart inside the monitoringChartNamespace
	helmOutput, error := helm.RunHelmCommandAndGetOutputE(t, helmOptions, "list")
	require.NoError(t, error)

	fmt.Println(helmOutput)
}
```

But the namespace option set in `NewKubectlOptions` is not used by `RunHelmCommandAndGetOutputE` function, only kube-context  and kubeconfig :
```
go test -v monitoring_chart_test.go                                                                                    
=== RUN   TestMonitoringChart
=== PAUSE TestMonitoringChart
=== CONT  TestMonitoringChart
TestMonitoringChart 2021-06-22T15:39:26+02:00 logger.go:66: Running command helm with args [list --kube-context dev-cluster --kubeconfig /home/USER/.kube/dev-cluster]
TestMonitoringChart 2021-06-22T15:39:26+02:00 logger.go:66: WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/USER/.kube/dev-cluster
TestMonitoringChart 2021-06-22T15:39:26+02:00 logger.go:66: WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/USER/.kube/dev-cluster
TestMonitoringChart 2021-06-22T15:39:27+02:00 logger.go:66: NAME        NAMESPACE       REVISION        UPDATED STATUS  CHART   APP VERSION
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/USER/.kube/dev-cluster
[...]

PR #947, opened 2021-07-13
Always include namespace flag in helm

yorinasub17:
Fixes https://github.com/gruntwork-io/terratest/issues/937

In helm 3, namespace flag is a global option so we can always include it.


brikis98 (approved):
LGTM

yorinasub17:
Thanks for review! Will merge this in now.

NOTE: Test failure is unrelated to this change, as it is not in the kubernetes test container.

PR #951, opened 2021-07-16
Implement Terragrunt validate

zackproser:
TODO:
- [x] Implement [relaxed and strict mode for Terragrunt's `validate-inputs` command](https://github.com/gruntwork-io/terragrunt/issues/1603#issuecomment-805944075) via [PR](https://github.com/gruntwork-io/terragrunt/pull/1757)


modules/test-structure/validate_struct.go
```diff
@@ -69,6 +82,28 @@ func NewValidationOptions(rootDir string, includeDirs, excludeDirs []string) (*V
 	return vo, nil
 }
 
+// NewValidationOptions returns a ValidationOptions struct, with override-able sane defaults, configured to find
+// and process all directories containing .tf files
+func NewValidationOptions(rootDir string, includeDirs, excludeDirs []string) (*ValidationOptions, error) {
```
zackproser, line 87: By splitting `NewValidationOptions` into two separate functions, one that configures `ValidationOptions` with a `FileType` of `TF` for Terraform, and one that configures `ValidationOptions` with a `FileType` of `TG` for Terragrunt, we can a backward incompatible release for the tests already calling the `NewValidationOptions` function.

modules/terraform/validate.go
```diff
@@ -12,9 +12,15 @@ func Validate(t testing.TestingT, options *Options) string {
 	return out
 }
 
-// ValidateE calls terraform validate and returns stdout/stderr.
+// ValidateE calls terraform validate and returns stdout/stderr. Terragrunt's binary expects the `validate-inputs` command, whereas Terraform's correlative command is `validate`
```
yorinasub17, line 15: Ahhhh `validate-inputs` and `validate` are two separate commands, so we should probably have a `ValidateInputsE` function that calls `validate-inputs` and asserts the binary is terragrunt (similar to `TgApplyAllE`).

modules/test-structure/validate_struct.go
```diff
@@ -11,11 +11,24 @@ import (
...
 
+// ValidateFileType is the underlying module type to search for when performing validation. Either Terraform or Terragrunt
+// files are targeted during a given validation sweep
+type ValidateFileType string
+
+const (
+	// TF represents repositories that contain Terraform code
+	TF = "*.tf"
+	// TG represents repositories that contain Terragrunt code
+	TG = "*.hcl"
```
yorinasub17, line 22: This should search for `terragrunt.hcl`, not any `hcl` file. Otherwise, it will catch packer templates too (which end in `pkr.hcl`).

yorinasub17 (commented):
Looks like good progress. I have a couple high level thoughts:

- We want to run both `terragrunt validate-inputs` AND `terragrunt validate`. The latter is important to run to ensure `generate` blocks output valid terraform configuration.
- Can you do a speed test to compare if it is faster to run `validate` in this way, or if you run `run-all validate` like we are doing in the service catalog repo smoke test? I have a sneaking suspicion that `run-all validate` might be significantly faster, but I could be wrong.

modules/test-structure/validate_struct.go
zackproser: Thanks - I wasn't sure about this - and initially had it searching for `terragrunt.hcl`. I will switch it back.

zackproser: Done in 2b755e7

modules/terraform/validate.go
zackproser: Done in 2b755e7

yorinasub17 (approved):
Updates LGTM!

zackproser:
> Updates LGTM!

Thanks for the reviews! Going to merge this in now.

PR #953, opened 2021-07-20
Update f1-micro to e2-micro instance

yorinasub17:
I just got an email from GCP stating that the f1-micro is being retired from free tier access and needs to be updated to e2-micro. This PR makes that change.


yorinasub17:
Closing as outdated

PR #957, opened 2021-07-23
Add helper for extracting plan struct without logging and without specifying a plan file

yorinasub17:
I'm finding myself rewriting versions of this function more often than not, to the point that it probably makes sense to live in terratest.

The point of this function is:

- Avoid logging the planfile json to console. This is a really big log message that is not human readable, and is not very useful for debugging (there are other, easier ways to debug the plan file, such as by introspecting the struct using [delve](https://github.com/go-delve/delve)).

- Avoid having to specify where to store the plan file. After some usage, I realized that I never need the plan file on disk after I get the struct out.

### DIFF
--- a/modules/terraform/plan.go
+++ b/modules/terraform/plan.go
@@ -2,7 +2,10 @@ package terraform
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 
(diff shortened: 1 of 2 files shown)


zackproser (approved):
LGTM

yorinasub17:

PR #963, opened 2021-07-27
Updated CopyTerraformFolderToTemp to include .terraform-version file

### LINKED ISSUE #962: Include '.terraform-version' file when copying Terraform folder to temporary location
As a developer using Terratest and also using [tfenv](https://github.com/tfutils/tfenv) to ensure the correct version of Terraform is used when deploying, I would like to ensure that the `.terraform-version` file [used by `tfenv`](https://github.com/tfutils/tfenv#terraform-version-file) is copied when I use Terratest to make temporary copies of my Terraform code for testing. That way, when tests are run I can make sure they use the same version of Terraform that will be used in actual deployments.

yardbirdsax:
The CopyTerraformFolderToTemp method will now copy the '.terraform-version' file if it exists in the source folder. This adds support for users who use the 'tfenv' tool to enforce that a certain version of Terraform is used in deployments and testing. This fixes #962.

The output of the unit tests is available [here](https://gist.github.com/yardbirdsax/d100e41736f641788c78fb8af72988f5). Please note: I only ran the unit tests for the module containing the code I changed.

Thanks for your consideration!


modules/files/files.go
```diff
@@ -173,6 +173,17 @@ func PathContainsHiddenFileOrFolder(path string) bool {
...
 }
 
+// PathIsTerraformVersionFile returns true if the given path is the special '.terraform-version' file used by the [tfenv](https://github.com/tfutils/tfenv) tool.
+func PathIsTerraformVersionFile(path string) bool {
+	pathParts := strings.Split(path, string(filepath.Separator))
+	for _, pathPart := range pathParts {
+		if strings.Contains(pathPart, ".terraform-version") {
+			return true
+		}
+	}
```
brikis98, line 183: NIT: If `.terraform-version` is a file, wouldn't it be the final part of the path? And if so, could this be simplified to `filepath.Base(path) == ".terraform-version"`?

modules/files/files.go
```diff
@@ -39,12 +39,12 @@ func IsExistingDir(path string) bool {
...
-// the path to the temp folder with the copied contents. Hidden files and folders, Terraform state files, and
-// terraform.tfvars files are not copied to this temp folder, as you typically don't want them interfering with your
-// tests.
+// the path to the temp folder with the copied contents. Hidden files and folders (with the exception of the `.terraform-version` files used
+// by the [tfenv tool](https://github.com/tfutils/tfenv)), Terraform state files, and terraform.tfvars files are not copied to this temp folder,
+// as you typically don't want them interfering with your tests.
 func CopyTerraformFolderToTemp(folderPath string, tempFolderPrefix string) (string, error) {
 	filter := func(path string) bool {
-		return !PathContainsHiddenFileOrFolder(path) && !PathContainsTerraformStateOrVars(path)
+		return (!PathContainsHiddenFileOrFolder(path) || PathIsTerraformVersionFile(path)) && !PathContainsTerraformStateOrVars(path)
```
brikis98, line 47: NIT: this is getting hard to read / reason about with all the nots, ors, etc. Perhaps two lines will be easier to read?

```go
if PathIsTerraformVersionFile(path) {
  return true
} else if PathContainsHiddenFileOrFolder(path) || PathContainsTerraformStateOrVars(path) {
  return false
} else {
  return true
}

**Review by brikis98 — COMMENTED** [review]
Thanks for the PR!

File: modules/files/files.go (hunk 2 above)
Comment (yardbirdsax) [reply modules/files/files.go]: Fixed with a slight modification to match [the guidelines](https://gruntwork.io/guides/style%20guides/golang-style-guide/#control-structures).

File: modules/files/files.go (hunk 1 above)
Comment (yardbirdsax) [reply modules/files/files.go]: Fixed!

**Comment by yardbirdsax** [comment]
Thanks for the review @brikis98 ! I made some corrections based on your comments and pushed the changes. The unit tests still succeed fine.

**Review by brikis98 — APPROVED** [review]

**Comment by brikis98** [comment]
Looks like pre-commit checks failed:

```
- hook id: goimports

modules/files/files.go
```

Please run `goimports`!

yardbirdsax:
Whoops! Sorry about that, I re-ran it and pushed the branch again. Seems like my pre-commit install didn't work for some reason.

brikis98 (approved):
Sorry for the delay. Lost track of this one. I'll kick off tests again!

brikis98:
Hm, some tests are failing, and continue to fail on re-run, so it's not clear this is some transient issue. I'll have to come back to this to look into it to see if it's related to this PR or not.

yardbirdsax:
Yeah I saw that, which is very odd. If I have time I may run the tests locally both on the trunk branch and mine as well.

brikis98:
My apologies, I've been super buried, and completely lost track of this one. @infraredgirl Would you mind taking a look at the test failures here and seeing if they are relevant or if this is OK to merge?

infraredgirl:
I re-ran the tests, and the results are as follows:
- 3 flaky GPC tests are failing with the following errror: `Error resolving image name 'ubuntu-os-cloud/ubuntu-1604-lts': Could not find image or family ubuntu-os-cloud/ubuntu-1604-lts`. This is obviously unrelated to this change.
- `TestGitRefChecks/GetCurrentRefReturnsLightTagValue` has a failure that was fixed on `master` in the mean time.

Given the above, I'll go ahead and merge this PR. Thanks @yardbirdsax for your contribution and your patience!

infraredgirl (approved):

infraredgirl:
https://github.com/gruntwork-io/terratest/releases/tag/v0.38.8

PR #965, opened 2021-07-28
Add one more character to the commit SHA

bwhaley:
@yorinasub17 and I found that the `GetCurrentRefReturnsLightTagValue` test was failing because the number of abbreviated characters in a commit SHA returned by bit has changed due to the number of objects in the repo. This behavior is [documented in this comment](https://github.com/git/git/commit/e6c587c733b4634030b353f4024794b08bc86892). This PR fixes that by simply adding the expected character.


yorinasub17 (approved):
LGTM (assuming the test passes)!

bwhaley:
This fixed the relevant test so I'll go ahead and merge.

PR #970, opened 2021-08-05
Support multiarch docker image builds

yorinasub17:
I just got the M1 Macbook, and almost everything works except for Terratest tests that require pushing an image to AWS to run on amd64 machines (e.g., ECS Deploy Runner).

This PR updates the build functions in terratest to support multiarchitecture builds using the [buildx](https://docs.docker.com/buildx/working-with-buildx/) command that docker desktop supports.

### DIFF
--- a/modules/aws/ecr_test.go
+++ b/modules/aws/ecr_test.go
@@ -15,14 +15,14 @@ func TestEcrRepo(t *testing.T) {
 	t.Parallel()
 
 	region := GetRandomStableRegion(t, nil, nil)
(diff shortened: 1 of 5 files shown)


modules/aws/ecr_test.go
```diff
@@ -1,9 +1,12 @@
 package aws
```
yorinasub17, line 1: This is a bonus test stability fix I rolled in so I can get a green build.

zackproser (approved):
LGTM

yorinasub17:

PR #971, opened 2021-08-10
Call build --platform when only requesting a single cross platform image

yorinasub17:
Improve multiarchitecture support by avoiding the need for `buildx` if building a single architecture image, but cross platform. `docker build` has native support for creating a single architecture cross platform image in the `--platform` option.


yorinasub17:
Closing as cross platform building without buildx doesn't really work (at least not in circleci).

PR #975, opened 2021-08-16
Skip packer init for non-HCL templates

bwhaley:
Packer logs an error if `packer init` is attempted on a JSON template, so terratest should skip `init` if the template is JSON.


yorinasub17 (approved):
LGTM!

bwhaley:
Thx! Merging + releasing now.

PR #977, opened 2021-08-23
Update sqs_test.go

johyip:


infraredgirl:
Hey @johyip thanks for the PR! Can you explain the motivation behind this change? The `queueExists` is just a local helper function - any reason you want to expose it? Note also that after renaming the function, you'd need to make the corresponding change in the invocations of the function.

Issue #978, opened 2021-08-26
AWS/SSM module windows support


srg1177:
Hello

There is a problem with AWS/SSM module CheckSSMCommandWithClientE(), it does not support Windows platform becouse of the hardcoded [AWS-RunShellScript](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/ssm.go#L192) Commant Document.

`UnsupportedPlatformType: Cannot perform operation for instance id ########### of platform type Windows`

infraredgirl:
Thanks for the bug report, @srg1177! A PR contributing the fix for this would be most welcome.

srg1177:
hi @infraredgirl can you please grand me access ?)

infraredgirl:
Fixed in #981.

infraredgirl:
https://github.com/gruntwork-io/terratest/releases/tag/v0.37.9

PR #981, opened 2021-08-31
Add SSM CommandDoc support

srg1177:
Changes

- Adding `commandDocName` string variable for to support command execution on  multiple platforms.

[issue](https://github.com/gruntwork-io/terratest/issues/978)

### DIFF
--- a/modules/aws/ssm.go
+++ b/modules/aws/ssm.go
@@ -156,10 +156,9 @@ func WaitForSsmInstance(t testing.TestingT, awsRegion, instanceID string, timeou
 }
 
 // CheckSsmCommand checks that you can run the given command on the given instance through AWS SSM.
-func CheckSsmCommand(t testing.TestingT, awsRegion, instanceID, command string, commandDocName string, timeout time.Duration) *CommandOutput {
-	result, err := CheckSsmCommandE(t, awsRegion, instanceID, command, commandDocName, timeout)
-	require.NoErrorf(t, err, "failed to execute '%s' on %s (%v):]\n  stdout: %#v\n  stderr: %#v", command, instanceID, err, result.Stdout, result.Stderr)
-	return result
+// Deprecated
+func CheckSsmCommand(t testing.TestingT, awsRegion, instanceID, command string, timeout time.Duration) *CommandOutput {
+	return CheckSsmCommandWithDocument(t, awsRegion, instanceID, command, "AWS-RunShellScript", timeout)
 }
 
 // CommandOutput contains the result of the SSM command.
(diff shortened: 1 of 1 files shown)


infraredgirl:
Hey @srg1177 thanks so much for the contribution!

I have a concern about changing the signatures of these functions - it will break the existing function calls. I think we can avoid backwards incompatibility here by creating a new function, instead of changing the signatures of existing functions. E.g. you could create a new function `CheckSSMCommandWithClientWithDocumentNameE` that has the same parameters as `CheckSSMCommandWithClientE`, plus the new argument `commandDocName`. The old function can then call the new function, to avoid code duplication. 

Would you be willing to implement this change?

srg1177:
Hi @infraredgirl  that will be my pleasure

Just one more thing to clarify.
 
Did I get you correct and we need to keep the old `CheckSSMCommandWithClientE` function ?

And if yes thats mean that new `CheckSSMCommandWithClientWithDocumentNameE` will require new calling functions identical to `CheckSSMCommand` and `CheckSSMCommandE` but with  `commandDocName` argument, as this functions are calling  upon each other. 

Please let me know if I get you wrong.

Thanks you!

infraredgirl:
> Hi @infraredgirl that will be my pleasure
> 
> Just one more thing to clarify.
> 
> Did I get you correct and we need to keep the old `CheckSSMCommandWithClientE` function ?

Yes, we want to keep the old function signatures intact in order to avoid backwards incompatible changes. Unfortunately, in Golang, there's no way to change a function signature without it being a breaking change. So instead of altering existing functions signatures, we create new functions.

> 
> And if yes thats mean that new `CheckSSMCommandWithClientWithDocumentNameE` will require new calling functions identical to `CheckSSMCommand` and `CheckSSMCommandE` but with `commandDocName` argument, as this functions are calling upon each other.

Yes, I think it makes sense to create a `WithDocument` equivalent for each of these functions. As I said, you can avoid code duplication by having the old function call the new function.

srg1177:
Hi @infraredgirl 

My apologies for late response

Ive fixed function call issues, now old functions are one place but they are pointing to new ones with hardcoed Command Document type, like it was before.

Please let me know if any other changes are required.

infraredgirl (commented):
Hey @srg1177, sorry for the delay in reviewing this! Code changes LGTM and the tests are passing, I just had a few nits that I left inline regarding code comments. If you don't mind fixing those, we can merge this right after.

srg1177:
Hey @infraredgirl 

We are all set, I think this will do )

infraredgirl:
Excellent, LGTM! Kicking off the tests now.

infraredgirl:
Tests passed! I'm going to approve and merge.

infraredgirl (approved):

infraredgirl:
https://github.com/gruntwork-io/terratest/releases/tag/v0.37.9

PR #993, opened 2021-09-22
update job is succeeded to check completions

### LINKED ISSUE #992: k8s module: WaitUntilJobSucceed finishes before job has any completions
WaitUntilJobSucceed can finish before the job has had any successful completions in some cases.

The logic in [IsJobSucceeded](https://github.com/gruntwork-io/terratest/blob/v0.37.8/modules/k8s/job.go#L91) returns `true` when `job.Status.Active == 0 && job.Status.Failed == 0`. It can end up returning `true` very early on, before any pods have run.

Based on my reading of the [JobStatus](https://pkg.go.dev/k8s.io/api/batch/v1#JobStatus), `IsJobSucceeded` _could_ look at `Status.Succeeded` to make sure it is at least one, but I believe it would also need to look at [JobSpec](https://pkg.go.dev/k8s.io/api/batch/v1#JobSpec) `Completions` and `Parallelism` to determine if the right number of `Succeeded` Pods had run.

It looks like instead of using Active, Failed, and Succeeded, it could look at `job.Status.Conditons`:

```
// The latest available observations of an object's current state. When a Job
// fails, one of the conditions will have type "Failed" and status true. When
// a Job is suspended, one of the conditions will have type "Suspended" and
// status true; when the Job is resumed, the status of this condition will
// become false. When a Job is completed, one of the conditions will have
// type "Complete" and status true.
```

By checking `job.Status.Conditions` for a condition with type "Complete" and status true, `WaitUntilJobSucceed` could ensure that the job has run to completion before returning.

jeffmhastings:
When checking the job status in `IsJobSucceeded`, it will sometimes return true for jobs that are not complete.

This updates the check in IsJobSucceeded to check the job status conditions for a condition of type "Complete" with status "True".

resolves #992

### DIFF
--- a/modules/k8s/job.go
+++ b/modules/k8s/job.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/gruntwork-io/terratest/modules/logger"
@@ -87,7 +88,12 @@ func WaitUntilJobSucceedE(t testing.TestingT, options *KubectlOptions, jobName s
 	return nil
 }
 
-// IsJobSucceeded returns true if all containers in the job are completed & succeeded
+// IsJobSucceeded returns true when the job status condition "Complete" is true
 func IsJobSucceeded(job *batchv1.Job) bool {
-	return job.Status.Active == 0 && job.Status.Failed == 0
+	for _, condition := range job.Status.Conditions {
+		if condition.Type == batchv1.JobComplete && condition.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
 }
--- a/modules/k8s/job.go
+++ b/modules/k8s/job.go
@@ -88,7 +88,8 @@ func WaitUntilJobSucceedE(t testing.TestingT, options *KubectlOptions, jobName s
 	return nil
 }
 
-// IsJobSucceeded returns true when the job status condition "Complete" is true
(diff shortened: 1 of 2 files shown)


jeffmhastings:
I don't see any automated status checks, so I've run tests for the k8s module against minikube with

```
go test -tags kubeall,kubernetes ./modules/k8s
```

I am seeing a couple of failures, but I think they're related to the my with minikube, and not anything specific to these changes.

```
--- FAIL: TestGetServiceEndpointEReturnsAccessibleEndpointForNodePort (318.22s)
--- FAIL: TestWaitUntilIngressAvailableReturnsSuccessfully (313.03s)
FAIL
FAIL	github.com/gruntwork-io/terratest/modules/k8s	338.707s
FAIL
```

modules/k8s/job.go
```diff
@@ -87,7 +88,12 @@ func WaitUntilJobSucceedE(t testing.TestingT, options *KubectlOptions, jobName s
 	return nil
 }
 
-// IsJobSucceeded returns true if all containers in the job are completed & succeeded
+// IsJobSucceeded returns true when the job status condition "Complete" is true
```
yorinasub17, line 91:
```suggestion
// IsJobSucceeded returns true when the job status condition "Complete" is true. This behavior is documented in the kubernetes API reference:
// https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/job-v1/#JobStatus
```

yorinasub17 (commented):
Thanks for the contribution!

The update makes sense to me, but I think we can use an update to the function comment (see below).

Separately, I will kick off a sanity check build.

yorinasub17:
Build passed! So I think we can merge this in once the docs update suggestion is pushed.

modules/k8s/job.go
jeffmhastings: Updated!

yorinasub17 (approved):

PR #994, opened 2021-09-23
Add terragrunt example as a first class example on terratest.gruntwork.io

yorinasub17:
Addresses https://github.com/gruntwork-io/terratest/issues/122#issuecomment-924846867

Since we don't have preview for docs site, here are screenshots of how this looks like:

Screen Shot 2021-09-23 at 10 29 59 AM

Screen Shot 2021-09-23 at 10 30 18 AM

Screen Shot 2021-09-23 at 10 40 57 AM


brikis98 (commented):
Am I just blind, or is the actual `terragrunt.hcl` file missing from this PR?

yorinasub17:
> Am I just blind, or is the actual terragrunt.hcl file missing from this PR?

Nope not blind - that is correct and is intended. I am reusing the existing [terragrunt-example](https://github.com/gruntwork-io/terratest/tree/master/examples/terragrunt-example) that we already have on main.

brikis98 (approved):
Ohhh, I gotcha. OK, LGTM, thanks!

yorinasub17:
Thanks for review! Merging this in now.

Issue #995, opened 2021-09-23
No 'GetTagsForVpc()', helper function (and other 'GetTags...' functions)


SphenicPaul:
There doesn't appear to be any helper function for obtaining the 'Tags' for AWS VPCs - There seems to be something similar for AWS EC2 instances in [here](https://github.com/gruntwork-io/terratest/blob/737e05c04c19e2a4f01a460f91ea24161540c772/modules/aws/ec2.go#L224) but nothing for AWS VPCs.

Additionally, I suspect there are other resources that would benefit from being able to obtain the tags so these can be tested as part of builds/tests etc. - Subnet, users, roles etc. are others - Actually, there doesn't seem to be a file/script for 'subnet(s)' within this 'AWS' module - Is it worth creating one to support this initial functionality I have (or is there something located somewhere else that I'm missing)?

I might create a PR for some of the resources I initially require tags for but I just want to ensure that these would be suitable/wanted. Resources supporting tagging are on [this](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html) page.

Shout if you've any questions.

yorinasub17:
I think that is reasonable! For VPC, I might prefer updating [the Vpc struct](https://github.com/gruntwork-io/terratest/blob/master/modules/aws/vpc.go#L15-L20) that gets returned by the getter functions to include the tags.

Although these functions most likely fail the `Complexity` litmus test mentioned in [type of contributions we accept](https://terratest.gruntwork.io/docs/community/contributing/#types-of-contributions), we have a hidden criteria of `Completeness`, and retrieving tags so you can check them is a basic function of most testing, so it makes sense to add them in under that criteria.

PRs to add these are welcome!

PR #999, opened 2021-09-25
add functions to get s3 tags

### LINKED ISSUE #996: AWS S3 bucket tags test
Hi guys! Just started using terratest. I wanted to test that my AWS S3 bucket has certain keys in tags, but didn't find a function for this. There is a function [FindS3BucketWithTagE](https://github.com/gruntwork-io/terratest/blob/6d96ad5172920b1c81700987309b2100b8fa3821/modules/aws/s3.go#L25) that helps to find s3 bucket with the given key-value in tags, but I want something similar to [GetTagsForEc2InstanceE](https://github.com/gruntwork-io/terratest/blob/v0.37.10/modules/aws/ec2.go#L231), but for S3.

I can create the function, would you be interested in a PR?

asilvanovich:
Resolves #996 

[automated test run](https://gist.github.com/asilvanovich/3d024192505ff637e2abf372b276c940)

### DIFF
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -62,6 +62,36 @@ func FindS3BucketWithTagE(t testing.TestingT, awsRegion string, key string, valu
 	return "", nil
 }
 
+// GetS3BucketTagging fetches the given bucket's tags and returns them as a string map of strings.
+func GetS3BucketTagging(t testing.TestingT, awsRegion string, bucket string) map[string]string {
+	tags, err := GetS3BucketTaggingE(t, awsRegion, bucket)
+	require.NoError(t, err)
+
+	return tags
+}
+
+//  GetS3BucketTaggingE fetches the given bucket's tags and returns them as a string map of strings.
+func GetS3BucketTaggingE(t testing.TestingT, awsRegion string, bucket string) (map[string]string, error) {
+	s3Client, err := NewS3ClientE(t, awsRegion)
+	if err != nil {
+		return nil, err
+	}
+
+	out, err := s3Client.GetBucketTagging(&s3.GetBucketTaggingInput{
+		Bucket: &bucket,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	tags := map[string]string{}
+	for _, tag := range out.TagSet {
+		tags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
+	}
+
+	return tags, nil
+}
+
 // GetS3ObjectContents fetches the contents of the object in the given bucket with the given key and return it as a string.
 func GetS3ObjectContents(t testing.TestingT, awsRegion string, bucket string, key string) string {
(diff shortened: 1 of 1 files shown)


SphenicPaul:
I just noticed this PR as it's a similar requirement to Issue #995 and PR #1000. My thoughts...

I suspect there should be some tests added around the new functions added as part of this PR, and additionally I'm not sure what working should be used (e.g. `GetS3BucketTags()` or `GetS3BucketTagging()`? - Not sure there is consistency with some of these across the modules and dependencies.

Also, @asilvanovich and @yorinasub17 , is there a prefered/consistent way to implement use of Tags across resources? - I suspect there will be a lot of resources that could be making use of similar functionality and I'd guess some consistency and/or reuse of this functionality is preferred?

asilvanovich:
Regarding wording - I tried to be consistent with names of the other functions in the s3.go file, for example [GetS3BucketVersioningE](https://github.com/gruntwork-io/terratest/blob/24901c600137bf868d98544228a0a51dda7b27d9/modules/aws/s3.go#L372) that is based on AWS's `GetBucketVersioning()`. Personally, I'd prefer `GetS3BucketTags()`. @yorinasub17 What do you think?

I'm going to add tests for the new  s3 functions tomorrow :)

modules/aws/s3.go
```diff
@@ -62,6 +62,36 @@ func FindS3BucketWithTagE(t testing.TestingT, awsRegion string, key string, valu
 	return "", nil
 }
 
+// GetS3BucketTagging fetches the given bucket's tags and returns them as a string map of strings.
+func GetS3BucketTagging(t testing.TestingT, awsRegion string, bucket string) map[string]string {
```
yorinasub17, line 66: Let's go with `GetS3BucketTags`, so it's consistent with the other terratest resources.

yorinasub17 (commented):
Implementation LGTM! Thanks for your contribution! Just had one change request for the function name based on above discussion. Once that's addressed, I think this is good to go!

I'll also kick off a regression test build so we have one cycle.

yorinasub17:
> is there a prefered/consistent way to implement use of Tags across resources? - I suspect there will be a lot of resources that could be making use of similar functionality and I'd guess some consistency and/or reuse of this functionality is preferred?

Unfortunately, the original AWS APIs didn't really have a consistent interface for retrieving tags, so I'm not sure how much of the functionality can be reused across resources in different domains (e.g., EC2 vs S3). That said, the newly announced [AWS CloudControl API](https://aws.amazon.com/blogs/aws/announcing-aws-cloud-control-api/) might have a consistent interface for the resources? I haven't looked into it so not sure.

asilvanovich:
@yorinasub17 changed the names, please have a look.

yorinasub17 (approved):
Build passed, so will go ahead and merge this in. Thanks again for your contribution!

PR #1000, opened 2021-09-26
Added 'Tags' and supporting functions to VPC struct

### LINKED ISSUE #995: No 'GetTagsForVpc()', helper function (and other 'GetTags...' functions)

SphenicPaul:
## Summary

Updated `vpc.go.` and `vpc_test.go` to include support for `Tags` to resolve #995 (just for `VPC`, not `Subnet`):
* `Vpc` struct now has additional, `Tags` property
* Added additional `GetTagsForVpc()` and  `GetTagsForVpcE()` functions (to obtain tags for VPC)
* Added additional `GetTagValueForVpc()` and  `GetTagValueForVpcE()` functions (to obtain tag value for VPC tag)
* Tests added for `GetTagsForVpc()` and `GetTagValueForVpc()`.
* Test added for `GetTagValueForVpcE()` to assert error where no tag of specific key exists

## Test Evidence

image

### DIFF
--- a/modules/aws/vpc.go
+++ b/modules/aws/vpc.go
@@ -156,25 +156,20 @@ func GetSubnetsForVpcE(t testing.TestingT, vpcID string, region string) ([]Subne
 // GetTagsForVpc gets the tags for the specified VPC.
 func GetTagsForVpc(t testing.TestingT, vpcID string, region string) map[string]string {
 	tags, err := GetTagsForVpcE(t, vpcID, region)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
+
 	return tags
 }
 
 // GetTagsForVpcE gets the tags for the specified VPC.
 func GetTagsForVpcE(t testing.TestingT, vpcID string, region string) (map[string]string, error) {
 	client, err := NewEc2ClientE(t, region)
-	if err != nil {
-		return nil, err
-	}
+	require.NoError(t, err)
 
 	vpcResourceTypeFilter := ec2.Filter{Name: &vpcResourceTypeFilterName, Values: []*string{&vpcResourceTypeFilterValue}}
 	vpcResourceIdFilter := ec2.Filter{Name: &vpcResourceIdFilterName, Values: []*string{&vpcID}}
 	tagsOutput, err := client.DescribeTags(&ec2.DescribeTagsInput{Filters: []*ec2.Filter{&vpcResourceTypeFilter, &vpcResourceIdFilter}})
-	if err != nil {
-		return nil, err
-	}
+	require.NoError(t, err)
 
 	tags := map[string]string{}
 	for _, tag := range tagsOutput.Tags {
--- a/modules/aws/vpc.go
+++ b/modules/aws/vpc.go
@@ -27,6 +27,9 @@ type Subnet struct {
 }
 
 var vpcIDFilterName = "vpc-id"
+var vpcResourceIdFilterName = "resource-id"
+var vpcResourceTypeFilterName = "resource-type"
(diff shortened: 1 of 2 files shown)


SphenicPaul:
Please shout if you've any questions/suggestions etc. - I'm relatively new to Go but I've tried to follow conventions already present within the module.

modules/aws/vpc.go
```diff
@@ -143,6 +153,60 @@ func GetSubnetsForVpcE(t testing.TestingT, vpcID string, region string) ([]Subne
 	return subnets, nil
 }
 
+// GetTagsForVpc gets the tags for the specified VPC.
+func GetTagsForVpc(t testing.TestingT, vpcID string, region string) map[string]string {
+	tags, err := GetTagsForVpcE(t, vpcID, region)
+	if err != nil {
+		t.Fatal(err)
+	}
```
yorinasub17, line 161: I know this whole file doesn't follow this pattern, but we've been incrementally switching to usage of `require` here:
```suggestion
	require.NoError(t, err)
```

Can you update the other new functions to use this pattern too? Thanks!

Import path for `require` is: `"github.com/stretchr/testify/require"`.

modules/aws/vpc.go
```diff
@@ -143,6 +153,60 @@ func GetSubnetsForVpcE(t testing.TestingT, vpcID string, region string) ([]Subne
...
+	if err != nil {
+		return "", err
+	}
+
+	if tagValue, tagExists := tags[tagKey]; tagExists {
+		return tagValue, nil
+	} else {
+		return "", fmt.Errorf("VPC does not have any Tag with a key of %s", tagKey)
+	}
+}
```
yorinasub17, line 208: Hmm I'm not sure this function passes the **complexity** litmus test, and I also am not sure this fits the bill for *completeness* either. I'd be inclined to remove this so as not to bloat our API.

yorinasub17 (commented):
Thanks for your contribution! This mostly LGTM. Had a few nits, mostly around style. Once addressed, I think this is good to go!

Let me kick off a regression test build so you have one test cycle.

modules/aws/vpc.go
SphenicPaul: OK. Updated `GetTagsForVpcE` and `GetTagsForVpc` to make use of this.

image

modules/aws/vpc.go
SphenicPaul: OK. No worries. Removed both.

SphenicPaul:
Thanks @yorinasub17 . Updated those now.

Let me know if there are any further tweaks/changes required,

yorinasub17 (approved):
Updates LGTM! The build failed, but the test failures are known flaky tests that are unrelated to these changes, so can merge this in!

PR #1005, opened 2021-10-02
Added `GetTagsForSubnet()` (and `Tags` property to `Subnet` struct) into `aws` module (Resolves #1004)

### LINKED ISSUE #1004: No 'GetTagsForSubnet()', helper function
Similar to #995...There doesn't appear to be any helper function for obtaining the 'Tags' for AWS **subnets**.

I might create a PR for adding fuinction to obtain the the AWS subnet tags (taking a similar approach to PR #1000 that resolved #995 by amending the struct) for but I just want to ensure that these would be suitable/wanted (this is effectively a similar requirement to #995). 

Shout if you've any questions.

SphenicPaul:
Made some changes to implement `GetTagsForSubnet()` function and `Tags` to resolve #1004 (to provide support for obtaining `Tags` from `Subnet`.

- Added `GetTagsForSubnet()` function
- Added test for above `GetTagsForSubnet()` function
- Added `Tags` property into `Subnet` struct
- Updated `GetSubnetsForVpcE()` to populate `Tags` property in `Subnet` struct 

I've tried to make these changes in line with those made in a similar change for AWS, VPC Tags (#1000).

### DIFF
--- a/modules/aws/vpc.go
+++ b/modules/aws/vpc.go
@@ -31,6 +31,9 @@ var vpcIDFilterName = "vpc-id"
 var vpcResourceIdFilterName = "resource-id"
 var vpcResourceTypeFilterName = "resource-type"
 var vpcResourceTypeFilterValue = "vpc"
+var subnetResourceIdFilterName = "resource-id"
+var subnetResourceTypeFilterName = "resource-type"
+var subnetResourceTypeFilterValue = "subnet"
 var isDefaultFilterName = "isDefault"
 var isDefaultFilterValue = "true"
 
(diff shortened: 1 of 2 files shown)


modules/aws/vpc.go
```diff
@@ -22,14 +22,18 @@ type Vpc struct {
...
+	Tags             map[string]string // The tags associated with the subnet
 }
 
 var vpcIDFilterName = "vpc-id"
 var vpcResourceIdFilterName = "resource-id"
 var vpcResourceTypeFilterName = "resource-type"
 var vpcResourceTypeFilterValue = "vpc"
+var subnetResourceIdFilterName = "resource-id"
+var subnetResourceTypeFilterName = "resource-type"
+var subnetResourceTypeFilterValue = "subnet"
```
SphenicPaul, line 36: I wasn't sure if this near-duplication is wanted or not? - I've implemented `subnet` tags similar to the way the `vpc` tags were done but don't foresee additional/other duplication of these sorts of things as there are only `vpc` and `subnet` within this `vpc` module.

SphenicPaul:
Please shout if there's anything that wants amending/updating.

modules/aws/vpc.go
infraredgirl: Two thoughts:
- Any reason all these vars should not be `const`s?
- Since `vpcResourceIdFilterName` and `subnetResourceIdFilterName` have identical values, we could have just one `resourceIdFilterName` instead. Same for `vpcResourceTypeFilterName`/`subnetResourceTypeFilterName`.

infraredgirl (commented):
Looks great, thanks for the contribution! I only have one question - see inline.

modules/aws/vpc.go
SphenicPaul: I tend to agree - I’ll update the PR and consolidate these as suggested.

SphenicPaul:
@infraredgirl  - All of the `var`s defined for filter names/Ids within the `vpc` module seemed to fall into the same bracket/situation so are now updated to `const`s.

Hope that's OK. Shout if there is anything else - I'm a little new to Go and this module so thanks for the hand-holding 😄.

image

PR #1008, opened 2021-10-05
Pull request 1003

### LINKED ISSUE #1003: cve-2020-26160 vulnerability in dgrijalva/jwt-go
There is a CVE (cve-2020-26160) detected  in the library dgrijalva/jwt-go used by one of required library that you use (present in the go.sum file).

warning from github:
```
CVE-2020-26160
high severity
Vulnerable versions: <= 3.2.0
Patched version: No fix

jwt-go allows attackers to bypass intended access restrictions in situations with []string{} for m["aud"] (which is allowed by the specification). Because the type assertion fails, "" is the value of aud. This is a security problem if the JWT token is presented to a service that lacks its own audience check. There is no patch available and users of jwt-go are advised to migrate to golang-jwt at version 3.2.1
```

HadwaAbdelhalem:
Fixes #1003
Closes #1003

This PR updates the Auzre/autorest version to rid of the github.com/dgrijalva/jwt-go. 

Link to [CI ](https://github.com/HadwaAbdelhalem/terratest/runs/3798333690?check_suite_focus=true) note the failures in the current CI is the same on the Main branch [CI](https://github.com/HadwaAbdelhalem/terratest/runs/3806138795?check_suite_focus=true) . 

The failures are are AKS due to this open [issue ](https://github.com/Azure/AKS/issues/2584)on Azure/AKS. While the remaining are Azure VM related which I am currently investigating and will create issues to track them


go.sum
```diff
@@ -125,6 +127,8 @@ github.com/dgrijalva/jwt-go v3.2.0+incompatible h1:7qlOGliEKZXTDg6OTjfoBKDXWrumC
 github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
```
yorinasub17, line 127: It looks like this doesn't fully resolve the issue since the vulnerable package is still being pulled in, but `go mod why` indicates that we are not using this code path since the packages we import are not pulling it in. So I think that is ok.

yorinasub17 (commented):
Updates LGTM! Can you run `go mod tidy` at the root of this project? I ran it and it detected some additional packages to clear.

go.sum
HadwaAbdelhalem: @yorinasub17 , Yes I am not sure why it is still pulled.

yorinasub17: Digging into this further, this may actually be pulled in as a transient dependency of the azure-sdk-go version we are currently at: https://github.com/Azure/azure-sdk-for-go/blob/v46.0.0/Gopkg.lock#L31-L37

It looks like this specific dependency was updated in v50.0.0.

I'm not sure how disruptive it is to update the `azure-sdk-for-go` package, but we should consider updating to that version and above. Let me take a crack at that.

I'll go ahead and kick off a regression build with our test pipeline, and if that passes, I'll merge this in and start a new PR attempting to update the Azure SDK.

yorinasub17 (approved):
Precommit hook passed, so going to merge this in. Afterwards, I'll start hacking away at upgrading the azure-sdk-for-go dependency to see if it will bump the `jwt-go` dep.

yorinasub17:
Thanks for the effort on this by the way!

PR #1009, opened 2021-10-06
Bump azure-sdk-go and various related packages

yorinasub17:
This bumps `azure-sdk-go`, `go-containerregistry`, and `aws-sdk-go`, and the related packages. These were using fairly old versions and needed a refresh, so I went ahead and updated them.

Note that I tried to minimize bumping a lot of versions at once. In particular, I bumped `azure-sdk-go` to `v50.x` to minimize the amount of breakage.

cc @HadwaAbdelhalem


yorinasub17:
Awesome glad to hear that worked! Thanks for closing the loop here!

PR #1016, opened 2021-10-14
fix: adjust "Host" via http helper

estahn:
I would like to do this:

```golang
headers := map[string]string{"Host" : "gloo-running-7xhydj.foobar"}
code, result := http_helper.HTTPDo(t, "GET", "http://randomelb.elb.ap-southeast-2.amazonaws.com/", nil, headers, nil)

```


estahn:
@robmorgan Can you have a look at this one?

yorinasub17 (approved):
LGTM! Thanks for the contribution and apologies for the delay in reviewing this! I'll kick off a regression build and if it passes, we can merge this in.

PR #1026, opened 2021-11-09
Add support for go-getter syntax when calling opa.Eval

yorinasub17:
This updates `opa.Eval` to support downloading a remote location for an OPA policy by supporting go-getter syntax. This allows users to centralize their OPA policies across the org in a single repo that all their IaC modules source from.

### DIFF
--- a/modules/opa/download_policy.go
+++ b/modules/opa/download_policy.go
@@ -1,0 +1,73 @@
+package opa
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	getter "github.com/hashicorp/go-getter"
+
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/testing"
+)
+
+var (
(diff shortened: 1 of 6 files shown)


modules/opa/download_policy.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/testing"
+)
+
+var (
+	// A map that maps the go-getter base URL to the temporary directory where it is downloaded.
+	policyDirCache sync.Map
+)
+
+// downloadPolicy takes in a rule path written in go-getter syntax and downloads it to a temporary directory so that it
```
yorinasub17, line 21:
```suggestion
// downloadPolicyE takes in a rule path written in go-getter syntax and downloads it to a temporary directory so that it
```

brikis98 (approved):
Nice!

yorinasub17:
Thanks for review! Test failure is unrelated GCP issue, so merging now.

PR #1042, opened 2021-12-30
Fix broken Terraform example

josh-padnick:
The previous version of this code failed with Terraform 0.12+ on an M1 Mac with the following error:

> Error: Incompatible provider version
> 
> Provider registry.terraform.io/hashicorp/template v2.2.0 does not have a
> package available for your current platform, darwin_arm64.
> 
> Provider releases are separate from Terraform CLI releases, so not all
> providers are available for all platforms. Other versions of this provider
> may have different platforms supported.

The issue is that the `template_file` resource is now deprecated and no download exists for darwin_amd64 for this provider.

I considered using the now current `templatefile` function, but we don't have any files from which we're reading templates in this example, so I used the more idiomatic variable interpolation.

### DIFF
--- a/examples/terraform-basic-example/main.tf
+++ b/examples/terraform-basic-example/main.tf
@@ -15,21 +15,13 @@ terraform {
 # See test/terraform_aws_example.go for how to write automated tests for this code.
 # ---------------------------------------------------------------------------------------------------------------------
 
-data "template_file" "example" {
-  template = var.example
-}
-
-data "template_file" "example2" {
-  template = var.example2
-}
-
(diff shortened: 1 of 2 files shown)


brikis98 (commented):
Why were there `template_file` data sources in the first place? Were we rendering any templates in the examples or tests?

yorinasub17 (approved):
LGTM! I suspect we had those `template_file` data sources to make the test more "interesting."

The build failure is unrelated to this PR, so will go ahead and merge this in.

PR #1046, opened 2022-01-11
Typo fix: s/kubectly/kubectl/

dwickr:
As much as I would like for `kubectly` to be a real thing, I thought I'd fix this typo.


denis256 (approved):

infraredgirl:
Thanks for the fix!

Issue #902, opened 2021-05-17
Feature request: parametrise timeout of HttpHelper


darmach:
Default 10 seconds of timeout of http.Client here: https://github.com/gruntwork-io/terratest/blob/737e05c04c19e2a4f01a460f91ea24161540c772/modules/http-helper/http_helper.go#L164 is a bit too short for me -  it's used on part of the infrastructure that has some intermittent lags, and I can't do anything about it (vendor issue)

HttpDo is referenced by other HttpDo* request functions, so this would be useful in all of them - can the timeout be parametrized, and default to 10 if not defined otherwise?

brikis98:
It can, but it's not easy to do so in Go in a backwards compatible way. I'm also worried about adding yet another parameter to the already long function signatures in `http_helper`. 

Probably what we need to do is refactor the `http_helper` methods to take in an `Options` struct. That struct can contain `tlsConfig`, `retries`, `sleepBetweenRetries`, this new `timeout` setting, etc. What's nice is that the struct would be extensible, so we could add new params in a backwards compatible way, without too much verbosity, in the future.

A PR to do this is welcome.

darmach:
This makes a lot of sense, I will try to do that if I have some spare time.

denis256:
Options for HTTP requests were introduced in v0.40.11

https://github.com/gruntwork-io/terratest/releases/tag/v0.40.11

PR #1055, opened 2022-02-01
[Backwards Incompatible] Use options classes for HttpHelper methods

tonerdo:
This PR adds option classes for the HttpHelper methods so that it can be more easily extended in a backwards compatible way. 

Addresses #902

### DIFF
--- a/modules/http-helper/http_helper.go
+++ b/modules/http-helper/http_helper.go
@@ -15,17 +15,17 @@ import (
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
+type HttpGetOptions struct {
+	Url       string
+	TlsConfig *tls.Config
+	Timeout   int
+}
+
 type HttpDoOptions struct {
 	Method    string
 	Url       string
 	Body      io.Reader
 	Headers   map[string]string
-	TlsConfig *tls.Config
-	Timeout   int
-}
-
-type HttpGetOptions struct {
-	Url       string
 	TlsConfig *tls.Config
 	Timeout   int
 }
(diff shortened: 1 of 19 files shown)


modules/http-helper/http_helper.go
```diff
@@ -16,10 +16,25 @@ import (
...
+	Headers   map[string]string
+	TlsConfig *tls.Config
+	Timeout   int
+}
+
 // HttpGet performs an HTTP GET, with an optional pointer to a custom TLS configuration, on the given URL and
 // return the HTTP status code and body. If there's any error, fail the test.
-func HttpGet(t testing.TestingT, url string, tlsConfig *tls.Config) (int, string) {
-	statusCode, body, err := HttpGetE(t, url, tlsConfig)
+func HttpGet(t testing.TestingT, options *HttpGetOptions) (int, string) {
```
brikis98, line 36: I really like the `Options` struct approach, but these sorts of breaking changes are quite painful. Is it worth adding new methods that take in the `Options` struct and initially deprecate the old methods?

elgatito:
@brikis98 @tonerdo I would take into consideration the idea of using `context` instead of a custom timeout and move that into a separate method, then call from old methods (with current method signatures) new methods that use contexts.

modules/http-helper/http_helper.go
tonerdo: Spending some time thinking through this

tonerdo: @brikis98 #1078 uses your suggested approach and adds new methods that accept the options structs

tonerdo:
Closing this in favour of #1078

PR #1061, opened 2022-02-11
Add function to require an env var is defined in test

yorinasub17:
This adds a new function that asserts that a given environment variable is defined and available.


denis256 (approved):

yorinasub17:
Thanks for review! Merging in now.

PR #1062, opened 2022-02-11
Add functions useful for interacting with docker images and building remote images

yorinasub17:
This PR introduces a few functions that are useful for interacting with docker images, and building remote images:

- `ListImages` for listing images in the docker daemon
- `DeleteImage` for cleaning up images in the docker daemon
- `DoesImageExist` for checking whether a given image exists in the daemon
- `GitCloneAndBuild` for building a docker image from a remote git repo

### DIFF
--- a/modules/docker/build_test.go
+++ b/modules/docker/build_test.go
@@ -2,8 +2,12 @@ package docker
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
+	"github.com/gruntwork-io/terratest/modules/git"
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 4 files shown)


modules/docker/build_test.go
```diff
@@ -60,3 +64,23 @@ func TestBuildWithTarget(t *testing.T) {
 	out := Run(t, tag, &RunOptions{Remove: true})
 	require.Contains(t, out, text1)
 }
+
+func TestGitCloneAndBuild(t *testing.T) {
```
denis256, line 68: `t.Parallel()` was removed by intention?

denis256 (dismissed):
I was wondering if will make sense to add an additional function to explicitly pull image, it may be helpful in cases when we have `latest` tag and want to avoid caching locally.

modules/docker/build_test.go
yorinasub17: Ah good catch! Was not intentional. Added in! [443f3f8](https://github.com/gruntwork-io/terratest/pull/1062/commits/443f3f807fd24fcf28c2cdf3c1cd35831cafcf7d)

denis256 (approved):

yorinasub17:
Thanks for review! Build failures are unrelated to this change, so will go ahead and merge this in.

PR #1078, opened 2022-03-09
Add HttpHelper methods that accept options structs

tonerdo:
This PR adds option classes for the HttpHelper methods so that it can be more easily extended in a backwards compatible way.

Addresses https://github.com/gruntwork-io/terratest/issues/902

### DIFF
--- a/modules/http-helper/http_helper.go
+++ b/modules/http-helper/http_helper.go
@@ -371,7 +371,24 @@ func HTTPDoWithValidationRetry(
 	body []byte, headers map[string]string, expectedStatus int,
 	expectedBody string, retries int, sleepBetweenRetries time.Duration, tlsConfig *tls.Config,
 ) {
-	err := HTTPDoWithValidationRetryE(t, method, url, body, headers, expectedStatus, expectedBody, retries, sleepBetweenRetries, tlsConfig)
+	options := HttpDoOptions{
+		Method:    method,
+		Url:       url,
+		Body:      bytes.NewReader(body),
+		Headers:   headers,
+		TlsConfig: tlsConfig,
+		Timeout:   10}
+
+	HTTPDoWithValidationRetryWithOptions(t, options, expectedStatus, expectedBody, retries, sleepBetweenRetries)
+}
+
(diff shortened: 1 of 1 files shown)


modules/http-helper/http_helper.go
```diff
@@ -205,17 +329,40 @@ func HTTPDoWithRetryE(
...
+// HTTPDoWithRetryWithOptionsE repeatedly performs the given HTTP method on the given URL until the given status code and body are
+// returned or until max retries has been exceeded.
+// The function compares the expected status code against the received one and fails if they don't match.
+func HTTPDoWithRetryWithOptionsE(
+	t testing.TestingT, options HttpDoOptions, expectedStatus int,
+	retries int, sleepBetweenRetries time.Duration,
+) (string, error) {
+	// The request body is closed after a request is complete. options.
+	// Extract the underlying data and cache it so we can reuse for retried requests
+	data, err := io.ReadAll(options.Body)
```
denis256, line 352: The value of `err` is ignored

denis256 (approved):

dkuzmenok:
@denis256 
@tonerdo 
https://github.com/gruntwork-io/terratest/blob/master/modules/http-helper/http_helper.go#L211
There was a `[]byte` used to pass request body, then it was used to create a byte reader and pass to `http.Request` constructor.
When we require to use byte reader directly from HttpDoOptions, then on request retry we will try to use a reader that is already read in a previous request attempt.

I do see 2 options to stay safe:
1) Add additional field to HttpDoOptions to set `BodyReader` value, and `Body` to be able to use both a provided reader or the byte slice.
2) Change `Body` to become `[]byte` and create Reader from it, like it is already done.

Personally I would prefer 1) as it gives flexibility.

tonerdo:
@dkuzmenok the changes in this PR already account for the scenario you mentioned. If you take a look at https://github.com/gruntwork-io/terratest/blob/0acf7fc8e970dc7472eed6040307f54da69ff37b/modules/http-helper/http_helper.go#L352-L357 we read the data out of the io.Reader and cache it for use in subsequent requests.

dkuzmenok:
@tonerdo Strange I did not notice that, when I was tracing method calls. Then it looks fine. Thanks.

PR #1082, opened 2022-03-10
Updated GCP tests to skip zones that can't run N1 instances

denis256:
Updated GCP tests to skip zones that can't run N1 instances


yorinasub17 (approved):
LGTM!
EDIT: (assuming tests pass)

denis256:
image

PR #1083, opened 2022-03-11
Additional description for module helm.Options - .logger.Discard

sqerison:
Hello team.
This description will be helpful for engineers who are working with Helm.
I've spent a few hours looking for the solution to avoid logging while running tests inside the CI.

Example of use:

Import the logging module:
```go
import (
    "github.com/gruntwork-io/terratest/modules/logger"
)
```

Then inside your function append the options with log parameters:
```go
    options := &helm.Options{
        SetValues: map[string]string{
            "global.nameOverride": "test",
        },
        Logger: logger.Discard, // <- this line will disable console logging for the helm templates and will print only errors and PASS results.
    }
```


infraredgirl (approved):
LGTM! Test failure is unrelated so I'll go ahead and merge this.

Thanks for the contribution!

PR #1084, opened 2022-03-12
Update deprecated circleci images to latest

yorinasub17:
Update deprecated circleci images to latest

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ env: &env
 defaults: &defaults
   machine:
     enabled: true
-    image: "ubuntu-2004:202107-02"
+    image: ubuntu-2004:202111-02
   <<: *env
 
 setup_minikube: &setup_minikube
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -112,8 +112,8 @@ jobs:
             - $HOME/go/src/
 
       # The weird way you have to set PATH in Circle 2.0
-      - run: echo 'export PATH=$HOME/.local/bin:$HOME/go/bin:$HOME/terraform:$HOME/packer:$PATH'
-          >> $BASH_ENV
+      - run: |
```
marinalimeira, line 115: What difference does it make to use the `|`?

marinalimeira (approved):
LGTM

.circleci/config.yml
yorinasub17: `|` ensures the string is treated as "raw" multiline, and prevents scripts from reflowing the text to keep column width.

yorinasub17:
Thanks for review! Merging now.

PR #1099, opened 2022-04-11
Install goimports with go install

yorinasub17:
`go get` for `goimports` no longer works in our CI environment, so switch to using `go install`. Same with `gox`, which still works but emits a warning.


yorinasub17:
Thanks for review! There is still a build failure, but I'll investigate it in another branch.

PR #1114, opened 2022-05-04
Add utility function for validating slack integrations

yorinasub17:
This migrates one of our internal validation functions (from [terraform-aws-monitoring](https://github.com/gruntwork-io/terraform-aws-monitoring/blob/master/test/cloudwatch_to_slack_test.go#L208)) to `terratest` so that it can be reused across out modules.

The new function `slack.ValidateExpectedSlackMessageE` can be used to check if a particular message text has been posted by a bot in the given channel, which is useful for integration testing of slack based workflows.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -363,6 +363,7 @@ workflows:
       - test:
           context:
             - Gruntwork Admin
+            - Slack Token For Test
           requires:
             - setup
           filters:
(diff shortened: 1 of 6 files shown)


yorinasub17:
NOTE: I investigated the test failures and they don't seem to be related to this. In particular, the new test `TestValidateSlackMessage` passed, which is the only one that matters for this change set since the code is isolated in a new package.

infraredgirl (approved):
LGTM

Issue #1116, opened 2022-05-05
Terraform lock file should be copied by CopyTerraformFolderToTemp()


Erouan50:
Hi everyone! 

I'm a heavy user of the function `test_structure.CopyTerraformFolderToTemp()` to run my tests in parallel. However, I noticed that the `terraform.lock.hcl` is not copied and can lead to some surprises. Terraform gets the latest version of the provider during the' terraform init', but the version might not match the one locked in the `terraform.lock.hcl` of the test infrastructure. However, if I skip a step (with "SKIP_teardown" for example), the copy is skipped, and the terraform init would use the lock file to get the version. That can lead to a weird situation where I have a test that would fail in normal mode but succeed in skip mode.

So I think the function should copy the .terraform.lock.hcl to have consistent behavior, but I might be wrong. What do you think?

I'm ready to open a pull request :)

PR #1117, opened 2022-05-06
Include .terraform.lock.hcl in CopyTerraformFolderToDest

Erouan50:
This commit adds in the filter from the function
CopyTerragruntFolderToDest an exclusion exception for the lock file.

The `.terraform.lock.hcl` used by Terraform to lock the versions of
the providers is not included during the copy. It can bring weird
behaviors when the copy is skipped (check the function
test_structure.CopyTerraformFolderToDest) where version 1.1 of a
provider would be used in the example folder but might be updated during
the `terraform init` due to the copy without the lock file. The lock
file should be included within the copy to keep the same providers'
versions.

Fix: #1116

I ran the test from the `modules/files`. Here is the output:

```
Test completed without error, however due to a limitation in GNU diff < 3.3.0, directories have not been compared for equivalency.
PASS
ok      github.com/gruntwork-io/terratest/modules/files 0.005s
```
I try to run the tests from the `modules/terraform`, but a lot of them fail 
(they fail as well from master) and don't seem to be related to my changes.

### DIFF
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -43,12 +43,12 @@ func IsExistingDir(path string) bool {
 // This is useful when running multiple tests in parallel against the same set of Terraform files to ensure the
 // tests don't overwrite each other's .terraform working directory and terraform.tfstate files. This method returns
 // the path to the dest folder with the copied contents. Hidden files and folders (with the exception of the `.terraform-version` files used
-// by the [tfenv tool](https://github.com/tfutils/tfenv)), Terraform state files, and terraform.tfvars files are not copied to this temp folder,
-// as you typically don't want them interfering with your tests.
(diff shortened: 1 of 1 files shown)


yorinasub17 (approved):
LGTM! Thanks for your contribution! I'll kick off a regression test and if it passes, will merge this in.

yorinasub17:
Build failed, but for a known failure on `master`. I rebased your test branch and kicked off a new build here:

https://app.circleci.com/pipelines/github/gruntwork-io/terratest/988/workflows/1fef1dfe-f43d-4f3c-81c4-6cd68ee6d9d3

yorinasub17:
The updated build failed, but for a known failure from `master` unrelated to this change, so we can merge this in! Thanks again for the contribution!

Erouan50:
You're welcome, thank you!

PR #1122, opened 2022-05-18
Implement a helper funciton to check minimum binary versions.

### LINKED ISSUE #566: Feature Request: Packer/Terraform/Docker/Other? Minimum Version Checks
Add helper functions for Packer/Terraform/Docker (and maybe others) to ensure a minimum version is installed for use on a system.

hongil0316:
The overall approach is to run a shell command to extract version string: 
* run a shell command (e.g., `docker --version`) 
* use regex string matcher to extract the exact version string

Right now, it only supports `docker`, `packer`, and `terraform` and the caller need to ensure that those binary exists in the working directory. If we want to support other binaries in the future that potentially has different ways to extract the version string, we can branch it out in the main method e.g., :

```
version, err
switch binary: 
case Docker, Packer, Terrafor: 
  version, err = getVersionWithShellCommand(t, params)
case ...:
  version, err = getVersionX(t, params)

// other remaining code
```

TODO: explore ways to mock `shell.RunCommandAndGetOutputE` so we can test handling different potential output from the shell command end to end. 

Fixes #566

### DIFF
--- a/modules/version-checker/version_checker.go
+++ b/modules/version-checker/version_checker.go
@@ -2,11 +2,12 @@ package version_checker
 
 import (
 	"fmt"
+	"regexp"
+
 	"github.com/gruntwork-io/terratest/modules/shell"
 	"github.com/gruntwork-io/terratest/modules/testing"
 	"github.com/hashicorp/go-version"
 	"github.com/stretchr/testify/require"
-	"regexp"
 )
 
 // VersionCheckerBinary is an enum for supported version checking.
--- a/modules/version-checker/version_checker.go
+++ b/modules/version-checker/version_checker.go
@@ -4,12 +4,23 @@ import (
 	"fmt"
 	"github.com/gruntwork-io/terratest/modules/shell"
 	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/hashicorp/go-version"
 	"github.com/stretchr/testify/require"
 	"regexp"
-	"strconv"
-	"strings"
 )
 
+// VersionMismatchErr is an error to indicate version mismatch.
+type VersionMismatchErr struct {
+	expectedVersion string
+	actualVersion   string
+}
+
+func (r *VersionMismatchErr) Error() string {
+	return fmt.Sprintf("found version mismatch {%s} when "+
+		"expecting min version of {%s}", r.expectedVersion, r.actualVersion)
+}
+
+// VersionCheckerBinary is an enum for supported version checking.
 type VersionCheckerBinary int
 
 // List of binaries supported for version checking.
@@ -20,65 +31,63 @@ const (
 )
 
 const (
-	// VersionRegexMatcher is a regex used to extract version string from shell command output.
-	VersionRegexMatcher = "\\d+(\\.\\d+)+"
-	// DefaultVersionArg is a default arg to pass in to get version output from shell command.
-	DefaultVersionArg = "--version"
+	// versionRegexMatcher is a regex used to extract version string from shell command output.
+	versionRegexMatcher = `\d+(\.\d+)+`
+	// defaultVersionArg is a default arg to pass in to get version output from shell command.
+	defaultVersionArg = "--version"
 )
 
 type CheckVersionParams struct {
-	binary          VersionCheckerBinary
-	expectedVersion string
-	workingDir      string
+	Binary          VersionCheckerBinary
+	ExpectedVersion string
+	WorkingDir      string
 }
 
-// CheckVersionE checks whether the given binary version is greater or equal
+// CheckVersionE checks whether the given Binary version is greater than or equal
 // to the given expected version.
 func CheckVersionE(
 	t testing.TestingT,
 	params CheckVersionParams) error {
 
-	err := validateParams(params)
+	if err := validateParams(params); err != nil {
+		return err
+	}
+
+	binaryVersion, err := getVersionWithShellCommand(t, params)
 	if err != nil {
 		return err
 	}
 
-	version, err := getVersionWithShellCommand(t, params)
-	if err != nil {
+	if err := checkMinimumBinaryVersion(binaryVersion, params.ExpectedVersion); err != nil {
 		return err
 	}
 
-	return validateBinaryVersionGreaterOrEqual(version, params.expectedVersion)
+	return nil
 }
 
-// CheckVersion checks whether the given binary version is greater or equal to the
+// CheckVersion checks whether the given Binary version is greater than or equal to the
 // given expected version and fail if it's not.
 func CheckVersion(
 	t testing.TestingT,
 	params CheckVersionParams) {
-	err := CheckVersionE(t, params)
-	require.NoError(t, err)
+	require.NoError(t, CheckVersionE(t, params))
 }
 
 // Validate whether the given params contains valid data to check version.
 func validateParams(params CheckVersionParams) error {
 	// Check for empty parameters
-	if len(params.expectedVersion) == 0 {
-		return fmt.Errorf("empty expectedVersion found")
-	} else if len(params.workingDir) == 0 {
-		return fmt.Errorf("empty workingDir found")
-	} else if params.binary < 0 {
-		return fmt.Errorf("empty binary found")
+	if params.ExpectedVersion == "" {
+		return fmt.Errorf("set ExpectedVersion in params")
+	} else if params.WorkingDir == "" {
+		return fmt.Errorf("set WorkingDir in params")
+	} else if params.Binary < 0 {
+		return fmt.Errorf("set Binary in params")
 	}
 
 	// Check the format of the expected version.
-	match, err := regexp.MatchString(VersionRegexMatcher, params.expectedVersion)
-	if err != nil {
+	if _, err := version.NewVersion(params.ExpectedVersion); err != nil {
 		return fmt.Errorf(
-			"unexpected regex matcher compilation failure: %w", err)
-	} else if !match {
-		return fmt.Errorf(
-			"invalid expected version format found {%s}", params.expectedVersion)
+			"invalid version format found {%s}", params.ExpectedVersion)
 	}
 
 	return nil
@@ -86,10 +95,11 @@ func validateParams(params CheckVersionParams) error {
 
 // getVersionWithShellCommand get version by running a shell command.
 func getVersionWithShellCommand(t testing.TestingT, params CheckVersionParams) (string, error) {
-	// Set appropriate binary, versionArg and extract version function
+	// Set appropriate Binary, versionArg and extract version function
 	// based on the given parameters.
-	var binaryName, versionArg = "", DefaultVersionArg
-	switch params.binary {
+	var binaryName = ""
+	var versionArg = defaultVersionArg
+	switch params.Binary {
 	case Docker:
 		binaryName = "docker"
 	case Packer:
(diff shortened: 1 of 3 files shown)


modules/version-checker/version_checker_test.go
```diff
@@ -0,0 +1,112 @@
+package version_checker
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestParamValidation(t *testing.T) {
```
zackproser, line 8: I like how many cases you're testing for here. One NIT to consider: this test might be more easily maintained going forward as a ["table driven test"](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) - such that: 

* Quickly scanning the function visually for which cases are implemented and which are not is easier
* It's easier for other engineers in the future to append their own cases, since the overall test harness you've created is handling the test run plumbing itself. They'd only need to define new structs for inputs and the associated expected outputs.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+			"command output {%s}: %w", output, err)
+	}
+
+	return version, nil
+}
+
+// extractVersionFromShellCommandOutput extracts version with regex string matching
+// from the given shell command output string.
+func extractVersionFromShellCommandOutput(output string) (string, error) {
+	regexMatcher, err := regexp.Compile(VersionRegexMatcher)
```
zackproser, line 127: NIT: why use `regexp.Compile` here versus `regexp.MustCompile`?

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+	var binaryName, versionArg = "", DefaultVersionArg
+	switch params.binary {
+	case Docker:
+		binaryName = "docker"
+	case Packer:
+		binaryName = "packer"
+	case Terraform:
+		binaryName = "terraform"
+	default:
+		t.Fatalf("unsupported binary for checking versions.")
```
zackproser, line 100: NIT: it would be great if this error message could let you know which binaries are currently supported, so the user has even less work to do to look this value up / grep through our source for it.

yorinasub17 (commented):
Looks like a great start! I had a bunch of nits and comments here but they are mostly about style. The core logic seems reasonable. I think this is usable for most cases.

However, to be feature complete, we really want to support [flexible version syntax like that supported by Terraform](https://www.terraform.io/language/expressions/version-constraints) (e.g., being able to say `~> 1.0.0` to limit to `> 1.0, < 1.1`). Can you look into implementing that?

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+	params CheckVersionParams) {
+	err := CheckVersionE(t, params)
+	require.NoError(t, err)
+}
+
+// Validate whether the given params contains valid data to check version.
+func validateParams(params CheckVersionParams) error {
+	// Check for empty parameters
+	if len(params.expectedVersion) == 0 {
+		return fmt.Errorf("empty expectedVersion found")
```
zackproser, line 67: NIT: enhance this error message to explain to the user what they need to do to remediate their error.

zackproser: Same comment on the next two instances of `fmt.Errorf` in this block.

modules/version-checker/version_checker.go
hongil0316: SGTM. updated.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+const (
+	// VersionRegexMatcher is a regex used to extract version string from shell command output.
+	VersionRegexMatcher = "\\d+(\\.\\d+)+"
+	// DefaultVersionArg is a default arg to pass in to get version output from shell command.
+	DefaultVersionArg = "--version"
+)
+
+type CheckVersionParams struct {
```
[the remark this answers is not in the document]
hongil0316: Decided to use this library for version checking - https://github.com/hashicorp/go-version

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+	binary          VersionCheckerBinary
+	expectedVersion string
+	workingDir      string
```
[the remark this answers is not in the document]
hongil0316: SGTM. Updated.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+	err := CheckVersionE(t, params)
+	require.NoError(t, err)
+}
+
+// Validate whether the given params contains valid data to check version.
+func validateParams(params CheckVersionParams) error {
+	// Check for empty parameters
+	if len(params.expectedVersion) == 0 {
+		return fmt.Errorf("empty expectedVersion found")
+	} else if len(params.workingDir) == 0 {
```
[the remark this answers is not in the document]
hongil0316: SGTM. Updated.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+// CheckVersionE checks whether the given binary version is greater or equal
+// to the given expected version.
+func CheckVersionE(
+	t testing.TestingT,
+	params CheckVersionParams) error {
+
+	err := validateParams(params)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
hongil0316: SGTM. Updated.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return nil
+}
+
+// getVersionWithShellCommand get version by running a shell command.
+func getVersionWithShellCommand(t testing.TestingT, params CheckVersionParams) (string, error) {
+	// Set appropriate binary, versionArg and extract version function
+	// based on the given parameters.
+	var binaryName, versionArg = "", DefaultVersionArg
```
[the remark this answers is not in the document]
hongil0316: SGTM. Updated.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+// List of binaries supported for version checking.
+const (
+	Docker VersionCheckerBinary = iota
+	Terraform
+	Packer
+)
+
+const (
+	// VersionRegexMatcher is a regex used to extract version string from shell command output.
+	VersionRegexMatcher = "\\d+(\\.\\d+)+"
```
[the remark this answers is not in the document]
hongil0316: no longer needed.

modules/version-checker/version_checker.go
hongil0316: Good idea. Updated.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+	default:
+		t.Fatalf("unsupported binary for checking versions.")
+	}
+
+	// Run a shell command to get the version string.
+	output, err := shell.RunCommandAndGetOutputE(t, shell.Command{
+		Command:    binaryName,
+		Args:       []string{versionArg},
+		WorkingDir: params.workingDir,
+		Env:        map[string]string{},
```
rhoboat, line 108: NIT/Question: How do we normally write this? If we don't pass in additional Env Vars, do we usually leave off this parameter, or is it better to explicitly set it `Env: map[string]string{}`? cc @zackproser @yorinasub17

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+
+	for i := 0; i < length; i++ {
+		v1, err := strconv.Atoi(versionNums1[i])
+		if err != nil {
+			return fmt.Errorf("invalid format version found {%s}", versionStr1)
+		}
+
+		v2, err := strconv.Atoi(versionNums2[i])
+		if err != nil {
+			return fmt.Errorf("invalid format version found {%s}", versionStr2)
```
rhoboat, line 170: Same NIT:
```suggestion
			return fmt.Errorf("invalid version format found {%s}", versionStr2)
```

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,190 @@
...
+//
+//    validateBinaryVersionGreaterOrEqual(t, 1.0.31, 1.0.27) - no error
+//    validateBinaryVersionGreaterOrEqual(t, 1.0.10, 1.0.27) - error
+//    validateBinaryVersionGreaterOrEqual(t, 1.0, 1.0.10) - error
+func validateBinaryVersionGreaterOrEqual(versionStr1 string, versionStr2 string) error {
+	if len(versionStr1) == 0 {
+		return fmt.Errorf("empty versionStr1 found")
+	} else if len(versionStr2) == 0 {
+		return fmt.Errorf("empty versionStr2 found")
+	}
```
rhoboat, line 151: Super NIT: It's probably nothing more than preference, but suggesting these two ways:
```suggestion
	if len(versionStr1) == 0 || len(versionStr2) == 0 {
		return fmt.Errorf("empty versions found")
	}
```
Or
```suggestion
	if versionStr1 == "" || versionStr2 == "" {
		return fmt.Errorf("empty versions found")
	}
```

rhoboat (commented):
Left some nits! Looks like a great start.

modules/version-checker/version_checker.go
hongil0316: SGTM.

modules/version-checker/version_checker.go
hongil0316: Changed a bit so error messages are a bit different to indicate which argument is causing error.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,174 @@
...
+// VersionMismatchErr is an error to indicate version mismatch.
+type VersionMismatchErr struct {
+	expectedVersion string
+	actualVersion   string
+}
+
+func (r *VersionMismatchErr) Error() string {
+	return fmt.Sprintf("found version mismatch {%s} when "+
+		"expecting min version of {%s}", r.expectedVersion, r.actualVersion)
+}
```
yorinasub17, line 21: NIT: put this in a separate file called `errors.go` so it's easier to find all the custom error types.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,174 @@
...
+	switch params.Binary {
+	case Docker:
+		binaryName = "docker"
+	case Packer:
+		binaryName = "packer"
+	case Terraform:
+		binaryName = "terraform"
+	default:
+		t.Fatalf("unsupported Binary for checking versions {%s}.", params.Binary)
+	}
```
yorinasub17, line 111: Ah I just realized that using the `enum` here to figure out the binary to use may not be the best approach. This is specifically true for `terraform`, where the `Options` struct supports arbitrary terraform binary paths (https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/options.go#L41)

The main use case for this is multi version projects, where users have different ways of installing multiple terraform versions and explicitly will point the binary path to something else.

Perhaps the struct should have two fields, one that is `BinaryType` which is the enum, and a `BinaryPath` field that is optional? And use this logic only if `BinaryPath` is empty?

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if err := checkMinimumBinaryVersion(binaryVersion, params.ExpectedVersion); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CheckVersion checks whether the given Binary version is greater than or equal to the
+// given expected version and fail if it's not.
```
zackproser, line 69:
```suggestion
// given expected version and fails if it's not.
```

modules/version-checker/version_checker_test.go
```diff
@@ -0,0 +1,149 @@
...
+		if tc.containError {
+			require.EqualError(t, err, tc.expectedErrorMessage, testCaseContextStr)
+		} else {
+			require.NoError(t, err, testCaseContextStr)
+			require.Equal(t, tc.expectedVersionStr, versionStr, testCaseContextStr)
+		}
+	}
+}
+
+func TestValidateBinaryVersionGreaterOrEqual(t *testing.T) {
```
zackproser, line 104: NIT ( so, feel free to ignore or address later as a nice-to-have, or even defer by filing a GitHub issue for): for even greater legibility, you could optionally extend all your test cases with a `Name` field that could contain a human-legible string explaining what that test case is checking for / ensuring

Then, you still loop through your defined test cases, but also leverage `t.Run`, passing the test case name as the first param, as demonstrated in [this article](https://medium.com/code-chasm/unit-test-in-go-8d17d3910818): 

```golang
for _, tt := range tests {
  t.Run(tt.Name, func(t *testing.T) {
   ...
  }
}
```

This way, both in the source code, and in the logs when we're looking at CI/CD failures, we'll be able to quickly see which cases we just broke with a recent change.

modules/version-checker/version_checker.go
hongil0316: SGTM. Updated.

modules/version-checker/version_checker_test.go
hongil0316: Yeah SGTM. Updated.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,229 @@
...
+	binaryVersion, err := getVersionWithShellCommand(t, params)
+	if err != nil {
+		return err
+	}
+
+	switch params.VersionCheckerType {
+	case MinimumVersion:
+		err = checkMinimumVersion(binaryVersion, params.MinimumVersion)
+	case VersionConstraint:
+		err = checkVersionConstraint(binaryVersion, params.VersionConstraint)
```
yorinasub17, line 72: This split in type is not necessary, since the version constraint implementation is a superset of minimum version. That is, you can easily implement minimum version using `VersionConstraint: "> MINIMUM_VERSION"`.

yorinasub17 (commented):
Updates LGTM! I think we are getting closer! I had one more UX nit, but everything else looks good.

The one other meta comment I have is that we should probably have integration tests for each of the three binaries. You already have one for `terraform`, but would be nice to have one for `docker` and `packer` as well. The integration tests will make sure that: (a) the regex works for the three main binaries, (b) the `--version` flag works for each of the three main binaries

You can pretty much assume all three are setup correctly in the test environment, since `terratest` is mostly an integration testing library that runs the binaries (e.g., `docker` package calls out to `docker` using `shell.RunCommand`).

modules/version-checker/version_checker.go
hongil0316: Great suggestion. Made the change.

modules/version-checker/version_checker.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return err
+	}
+
+	var versionConstraint string
+	switch params.VersionCheckerType {
+	case MinimumVersion:
+		versionConstraint = ">" + params.MinimumVersion
+	case VersionConstraint:
+		versionConstraint = params.VersionConstraint
```
yorinasub17, line 73: Ah what I actually meant was to remove the type altogether from the struct and only support the version constraint interface. It makes for a better UX because it removes one more thing the user has to configure.

Does that make sense?

yorinasub17:
Thanks for adding the integration tests! I had a clarification for my other suggestion though.

Let me kick off a regression build on this branch since I believe it's ready to go.

modules/version-checker/version_checker.go
hongil0316: Ah got it. updated.

yorinasub17:
Build failed, but the test failure is unrelated to this change. I have to do one more final pass before approving, but I think this is good to go!

yorinasub17 (approved):
Implementation LGTM! Thanks for your contribution! Going to merge and release this now.

modules/version-checker/version_checker_test.go
```diff
@@ -1,112 +1,219 @@
...
+		{
+			name:                 "empty output string",
+			outputStr:            "",
+			expectedVersionStr:   "",
+			containError:         true,
+			expectedErrorMessage: "failed to find version using regex matcher",
+		},
+	}
+
+	for _, tc := range tests {
```
rhoboat, line 113: Just a note that these tests are not running in parallel, unless I'm mistaken. However, they're quick and aren't doing anything like spinning up infrastructure, so running them serially is fine! 

If the `extractVersionFromShellCommandOutput` could become a long-running function it might be worth parallelizing. (Same with the `TestCheckVersionConstraint` and `TestCheckVersionEndToEnd` tests.)

For reference, to run them in parallel, you'd need to use `t.Run`, as in:
```go
for _, tc := range tests {
    t.Run(tc.name, func(t *testing.T) {
        tc := tc
        // run the test here...
    })
}
```

Issue #1048, opened 2022-01-12
Upgrade to docker compose v2.x


yariksheptykin:
[General Availability](https://docs.docker.com/compose/cli-command/#transitioning-to-ga-for-compose-v2) (GA) for docker compose v2 is ["Almost There"](https://github.com/docker/roadmap/issues/256). With GA we could consider upgrading docker compose examples ([sample project](https://github.com/gruntwork-io/terratest/tree/master/examples/docker-compose-stdout-example) and [test](https://github.com/gruntwork-io/terratest/blob/master/test/docker_stdout_example_test.go)) as well as the [docker package](https://github.com/gruntwork-io/terratest/blob/master/modules/docker/docker_compose.go) to use v2.

The upgrade should be quite stright forward as it mostly requires replacing "docker-compose" (dash) command with "docker compose" (no dash) and linking to the [official migration instructions](https://docs.docker.com/compose/cli-command-compatibility/) for porting tests.

yorinasub17:
Seems reasonable! I subscribed to the GA ticket to track it, and we can prioritize this as soon as it is released as GA.

yariksheptykin:
Anquick heads up:
compose V2 has finally transitioned to GA. See https://github.com/docker/roadmap/issues/256#issuecomment-1110984993

nialdaly:
Hi @yariksheptykin and @yorinasub17, I've raised a [PR](https://github.com/gruntwork-io/terratest/pull/1125) to resolve this issue. I hope that's okay? I was also wondering if it's worth renaming instances to `docker-compose.yml` to `compose.yml` as discussed [here](https://github.com/compose-spec/compose-spec/blob/master/spec.md#compose-file)?

I wanted to get your thoughts on the `version` property present in the existing `docker-compose.yml` files and if they are necessary with the move to V2? Thanks!

yariksheptykin:
> I wanted to get your thoughts on the version property present in the existing docker-compose.yml files and if they are necessary with the move to Docker Compose V2? Thanks!

I would suggest dropping `version` altogether. This will set the version to the compose spec, which should be a sane default.

PR #1125, opened 2022-05-22
`docker-compose` command changed to `docker compose` for upgrade to Docker Compose V2

nialdaly:
### Summary
This PR includes changes made to replace the Docker Compose command, `docker-compose` with `docker compose` for the proposed upgrade to Docker Compose V2. This is a proposed solution to issue [1048](https://github.com/gruntwork-io/terratest/issues/1048).


yariksheptykin:
LGTM. Thanks for taking care of  #1048 !

nialdaly:
Thanks! @yariksheptykin I've updated the `runDockerComposeE` function in *modules/docker/docker_compose.go* in an effort to fix the failing [tests](https://app.circleci.com/pipelines/github/gruntwork-io/terratest/1002/workflows/038a71e4-51be-4bfc-b2fc-bfdeb7b9e362/jobs/7682/tests).

I've also removed the `version` property from the `docker-compose.yml` files, as suggested!

denis256:
I think this PR will require additional work to install "docker compose v2" inside of test images

Related failed tests in CI:

```
TestDockerComposeStdoutExample
Failed
    docker_compose.go:22: error while running command: exec: "docker compose": executable file not found in $PATH; 


TestPackerDockerExampleLocal
Failed
TestPackerDockerExampleLocal 2022-05-24T12:17:56Z logger.go:66: Running Packer to generate a custom artifact for template ../examples/packer-docker-example/build.pkr.hcl
TestPackerDockerExampleLocal 2022-05-24T12:17:56Z logger.go:66: Creating a temporary directory for Packer plugins
TestPackerDockerExampleLocal 2022-05-24T12:17:56Z logger.go:66: Running command packer with args [-version]
TestTerraformAwsHelloWorldExample 2022-05-24T12:17:56Z retry.go:91: terraform [init -upgrade=false]
TestTerraformAwsHelloWorldExample 2022-05-24T12:17:56Z logger.go:66: Running command terraform with args [init -upgrade=false]
    docker_compose.go:22: error while running command: exec: "docker compose": executable file not found in $PATH; 
TestPackerDockerExampleLocal 2022-05-24T12:18:46Z logger.go:66: Running command docker compose with args [--project-name TestPackerDockerExampleLocal down]
    docker_compose.go:22: error while running command: exec: "docker compose": executable file not found in $PATH; 
```

nialdaly:
@denis256 Is this something that I can do or is it CircleCI related? I'm just not sure what you mean by "test images". Thanks!

yariksheptykin:
@nialdaly I'm not a big CicleCI expert, but I thing we will need to [install compose plugin](https://docs.docker.com/compose/install/#install-compose-on-linux-systems) for docker in a similarly to how we [install buildx](https://github.com/gruntwork-io/terratest/blob/master/.circleci/config.yml#L66) in the CI. I expect that this can be accomplished by duplicating buildx block and adjusting it for installing compose.

Something this this I imagine:
```yaml

env: &env
  environment:
    DOCKER_CONFIG: ~/.docker/
    DOCKER_COMPOSE_VERSION: v2.5.0

...
# Installation script for docker compose plugin. See: https://docs.docker.com/compose/install/#install-compose-on-linux-systems
install_docker_compose: &install_docker_compose
  name: install docker compose
  command: |
    mkdir -p $DOCKER_CONFIG/cli-plugins
   curl -SL https://github.com/docker/compose/releases/download/"$DOCKER_COMPOSE_VERSION"/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
    chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
    chmod +x /usr/local/lib/docker/cli-plugins/docker-compose
    docker compose version

...
  test:
...
      - run:
          <<: *install_docker_compose
```
This is just a thought experiment thought, this CI config might not work.

nialdaly:
Thanks for the guidance! @yariksheptykin I've made a change to the `.circleci/config.yml` based on your suggestion, fingers crossed.

.circleci/config.yml
```diff
@@ -74,6 +76,16 @@ install_docker_buildx: &install_docker_buildx
...
     docker buildx create --use
 
+# Installation script for the docker compose plugin. See: https://docs.docker.com/compose/install/#install-compose-on-linux-systems
+install_docker_compose: &install_docker_compose
+  name: install docker compose
+  command: |
+    mkdir -p ${DOCKER_CONFIG}/cli-plugins
+    curl -SL https://github.com/docker/compose/releases/download/"${DOCKER_COMPOSE_VERSION}"/docker-compose-linux-x86_64 -o ${DOCKER_CONFIG}/cli-plugins/docker-compose
+    chmod +x ${DOCKER_CONFIG}/cli-plugins/docker-compose
+    chmod +x /usr/local/lib/docker/cli-plugins/docker-compose
```
denis256, line 86: File `/usr/local/lib/docker/cli-plugins/docker-compose` may not exists since `curl` download file to `${DOCKER_CONFIG}/cli-plugins/docker-compose`

examples/docker-compose-stdout-example/docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: '2.0'
```
denis256, line 1: Not sure why should be removed `version` field? It is not supported by new compose version?

I tested locally with restored field and tests seems to pass

modules/docker/docker_compose.go
```diff
@@ -24,24 +26,24 @@ func RunDockerCompose(t testing.TestingT, options *Options, args ...string) stri
...
 
-// RunDockerComposeE runs docker-compose with the given arguments and options and return stdout/stderr.
+// RunDockerComposeE runs docker compose with the given arguments and options and return stdout/stderr.
 func RunDockerComposeE(t testing.TestingT, options *Options, args ...string) (string, error) {
 	return runDockerComposeE(t, false, options, args...)
 }
 
 func runDockerComposeE(t testing.TestingT, stdout bool, options *Options, args ...string) (string, error) {
 	cmd := shell.Command{
-		Command: "docker-compose",
```
denis256, line 41: This will break backward compatibility and will make unusable "old" environments that have only "docker-compose" command.

Was considered to add to "Options" struct feature flags to allow fallback to usage of "docker-compose" ?

denis256:
CI job failed, on step `install docker compose`

```
chmod: cannot access '/usr/local/lib/docker/cli-plugins/docker-compose': No such file or directory
```

examples/docker-compose-stdout-example/docker-compose.yml
nialdaly: Sure I can add the version property back!

.circleci/config.yml
nialdaly: @denis256 I'm not sure what you mean here?

nialdaly: I'll try to test this locally/on my own CircleCI setup to make sure it works! Bit of a slow feedback loop on this since I can't trigger the pipeline to test.

modules/docker/docker_compose.go
nialdaly: Could you provide some guidance on how you think this should be implemented? @denis256

examples/docker-compose-stdout-example/docker-compose.yml
yariksheptykin: The support for `version` property remains in compose v2. The version here defines the version of the docker compose "Schema" (here 2.0) which is quite old now. Compose-file versions evolved from v1.x, to v2.x, to v3.x, and now stabilized in compose specification. Docker recommends using compose spec when crafting compose.yaml files. In order to instruct engine to apply spec compose.yaml file should not contain version field any more.
See https://docs.docker.com/compose/compose-file/compose-versioning/ for more details on compose versioning.

This change however has nothing to do with upgrade to docker compose v2. Both v1 and v2 support all composefile versions up to the spec. I would recommend dropping version field to make use of schema validation from spec. This should be done in a separate MR to keep the scope of this MR more focused.

modules/docker/docker_compose.go
yariksheptykin: This is of course your decision, but I would not recommend trying to provide backward compatibility for docker-compose users. Here is why:
1. although both v1 and v2 have nearly identical API, some of the command --flags were not ported from v1 to v2, or ported partly. See: https://docs.docker.com/compose/cli-command-compatibility/.
2. docker-compose v1 will only receive security updates but no new features. I expect docker to sunset v1 soon
3. maintaining compatibility between v1 and v2 will soon become quite time-consuming as v2 will start accumulating new features not present in v1. The maintenance burden will be on your side as docker is not planning to continue developing v1

I recommend to consider this MR a "breaking change" and require users of docker-compose functionality to upgrade their installations. We could add migrations notes, which should mostly point to https://docs.docker.com/compose/cli-command-compatibility/ and installation instructions for v2.

But this is something you as maintainers should decide on your own based on the number of your active users of compose-based tests.

denis256: Hi,
since `docker compose` was introduced recently, I would say to keep backward compatibility as much as possible maybe by checking if `docker compose version` is available or comparing docker version 

References: https://www.docker.com/blog/announcing-compose-v2-general-availability/

denis256: Hi, 
an error in case of `docker compose` may appear because of multiple reasons not because compose plugin is not installed...

Can be invoked for example `docker compose version` and checked status code to get right status about plugin installation

nialdaly: Thanks for the feedback! @denis256 By status code do you mean this after running `docker compose version`?
```
Docker Compose version v2.5.1
```

nialdaly: Would this be better? 
```golang
func runDockerComposeE(t testing.TestingT, stdout bool, options *Options, args ...string) (string, error) {
	cmd := shell.Command{
		Command: "docker",
		// We append --project-name to ensure containers from multiple different tests using Docker Compose don't end
		// up in the same project and end up conflicting with each other.
		Args:       append([]string{"compose", "--project-name", strings.ToLower(t.Name())}, args...),
		WorkingDir: options.WorkingDir,
		Env:        options.EnvVars,
		Logger:     options.Logger,
	}

	dockerComposeVersionCmd := icmd.Command("docker compose version")

	result := icmd.RunCmd(dockerComposeVersionCmd)

	if result.ExitCode != 0 {
		cmd = shell.Command{
			Command:    "docker-compose",
			Args:       append([]string{"--project-name", strings.ToLower(t.Name())}, args...),
			WorkingDir: options.WorkingDir,
			Env:        options.EnvVars,
			Logger:     options.Logger,
		}
	}

	if stdout {
		return shell.RunCommandAndGetStdOut(t, cmd), nil
	}
	return shell.RunCommandAndGetOutputE(t, cmd)
}
```

denis256: Hi,
I think this approach is better, not sure why should be created each time `cmd` struct with `docker-compose`, maybe first run `docker compose version` and then create respective structure

nialdaly: Thanks for the feedback, would we not need to have two commands since there are two conditions?
1. `docker compose version` command fails so use `docker-compose` command
2. `docker compose version` succeeds, so use `docker` command with `compose` as an argument

nial-harbr: And `cmd` needs to be initialised outside of the `if` block, otherwise the following error is returned:
```
cmd declared but not used compiler(UnusedVar)
```

nialdaly: Is this better?
```golang
func runDockerComposeE(t testing.TestingT, stdout bool, options *Options, args ...string) (string, error) {
	var cmd shell.Command

	dockerComposeVersionCmd := icmd.Command("docker", "compose", "version")

	if icmd.RunCmd(dockerComposeVersionCmd).ExitCode != 0 {
		cmd = shell.Command{
			Command: "docker",
			// We append --project-name to ensure containers from multiple different tests using Docker Compose don't end
			// up in the same project and end up conflicting with each other.
			Args:       append([]string{"compose", "--project-name", strings.ToLower(t.Name())}, args...),
			WorkingDir: options.WorkingDir,
			Env:        options.EnvVars,
			Logger:     options.Logger,
		}
	}
	cmd = shell.Command{
		Command:    "docker-compose",
		Args:       append([]string{"--project-name", strings.ToLower(t.Name())}, args...),
		WorkingDir: options.WorkingDir,
		Env:        options.EnvVars,
		Logger:     options.Logger,
	}

	if stdout {
		return shell.RunCommandAndGetStdOut(t, cmd), nil
	}
	return shell.RunCommandAndGetOutputE(t, cmd)
}

```

nialdaly: Is this okay to go ahead? @denis256

denis256 (approved):

nialdaly:
No problem, thanks @yariksheptykin and @denis256 for the guidance!

PR #1189, opened 2022-10-12
Navbar update - added "we're hiring"

klijakub:
This PR adds `We're Hiring!` nav item.

Screen:
image
image


josh-padnick (approved):
LGTM!

PR #1196, opened 2022-11-08
Bump Go to 1.18

infraredgirl:


modules/k8s/config_test.go
```diff
@@ -114,6 +114,7 @@ preferences: {}
 users:
 - name: minikube
   user:
+    as-user-extra: null
```
infraredgirl, line 117: For some reason, without this change, five k8s tests fail. I don't really understand why, but I've confirmed that they fail when run with go 1.18 but pass with go 1.17. Very strange.

infraredgirl:
The rest of the failing tests are failing on master as well.

denis256 (approved):
Will take a look on failing tests

infraredgirl:
Thanks - going to merge this in.

PR #1282, opened 2023-05-04
Add UniqueLowercaseId to random module

### LINKED ISSUE #1281: Lowercase UniqueId
When creating a random uniqueID from the random package (`random.UniqueId()`)as a postfix to an S3 bucket name during a test, the deploy fails since the bucket name must be all lowercase.

Suggest that a UniqueLowecaseID is created which uses a smaller character set, but potentially with more characters to ensure enough clashes.

Happy to make the change and provide the PR

devsebops:
## Description

Fixes #1281
Adds a new uniqueLowercaseId function to the random module

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Added a new uniqueLowercaseId function to the random module

### Migration Guide

N/A

### DIFF
--- a/modules/random/random.go
+++ b/modules/random/random.go
@@ -41,6 +41,23 @@ func UniqueId() string {
 	return out.String()
 }
 
+const base36chars = "0123456789abcdefghijklmnopqrstuvwxyz"
+const uniqueLowercaseIDLength = 6 // Should be good for 36^6 = 2+ billion combinations
+
+// UniqueLowercaseId returns a unique (ish) id we can attach to resources and tfstate files so they don't conflict with each other
+// Uses base 36 to generate a 6 character string that's unlikely to collide with the handful of tests we run in
+// parallel. Based on code here: http://stackoverflow.com/a/9543797/483528
+func UniqueLowercaseId() string {
+	var out bytes.Buffer
+
+	generator := newRand()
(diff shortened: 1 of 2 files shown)


zackproser:
Thanks for the PR! I'm wary of introducing an entire new method to Terratest that  duplicates the functionality of the original `UniqueId` method. 

Is there a particular use case or issue you're trying to address?

Can I ask if you could wrap your calls to `UniqueId` in the `strings.ToLower` function like so:

```go
lowerUniqueId := strings.ToLower(random.UniqueId())
```

Alternatively, if you feel strongly that this needs to be a separate method, you could simplify this PR significantly by creating a new method that would call the original `random.UniqueId` but wrap its return value in a call to `strings.ToLower` - something like this: 

```go 
func UniqueIdLower() string {
  return strings.ToLower(random.UniqueId())
}
```

devsebops:
That will solve my problem and be much simpler. Probably should have looked around using standard go before trying this. Definitely need to look into learning go a bit more before pushing terratest.

Thanks for the help

zackproser:
> That will solve my problem and be much simpler. Probably should have looked around using standard go before trying this. Definitely need to look into learning go a bit more before pushing terratest.
> 
> Thanks for the help

@devsebops No problem at all!  We really do appreciate you engaging with us and opening a PR. :raised_hands:

PR #1319, opened 2023-06-28
feat(test-structure): add save test data if dne

### LINKED ISSUE #1318: Support optionally not overwriting saved terraformOptions if present
In some cases uses test-structures its possible to accidentally create duplicate resources or corrupt/overwrite state files due to overwriting saved test data.

For example, if you have a test structure stage that uses a random aws region and you re-run your tests without destroying, you then overwrite the terraform options and apply with the new ones. This could totally orphan the existing resources, duplicating resources and making it a challenge to harness the power of terratest/terraform when cleaning up resources post test.

bt-macole:
## Description

SaveTestData is a powerful feature. But with great power comes great responsibility. If a user is not careful, this feature can cause unintended side effects including duplicating resources and overwriting state.

Adding support for only saving if a file does not exists prevent overwriting contents and provides terratest users protection from themselves.

resolves #1318

Adds SaveTerraformOptionsIfNotPresent and updates SaveTestData to include an additional `overwrite bool` attribute.

This change is backwards compatible.

- [x] Update the docs.

## Release Notes (draft)

Added `SaveTerraformOptionsIfNotPresent`

### DIFF
--- a/modules/test-structure/save_test_data.go
+++ b/modules/test-structure/save_test_data.go
@@ -21,7 +21,14 @@ import (
 // SaveTerraformOptions serializes and saves TerraformOptions into the given folder. This allows you to create TerraformOptions during setup
 // and to reuse that TerraformOptions later during validation and teardown.
 func SaveTerraformOptions(t testing.TestingT, testFolder string, terraformOptions *terraform.Options) {
-	SaveTestData(t, formatTerraformOptionsPath(testFolder), terraformOptions)
+	SaveTestData(t, formatTerraformOptionsPath(testFolder), true, terraformOptions)
+}
+
+// SaveTerraformOptionsIfNotPresent serializes and saves TerraformOptions into the given folder if the file does not exist or the json is
+// empty. This allows you to create TerraformOptions during setup and to reuse that TerraformOptions later during validation and teardown,
+// but will prevent overwritting the contents and potentially duplicating resources.
+func SaveTerraformOptionsIfNotPresent(t testing.TestingT, testFolder string, terraformOptions *terraform.Options) {
+	SaveTestData(t, formatTerraformOptionsPath(testFolder), false, terraformOptions)
 }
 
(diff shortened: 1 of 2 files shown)


denis256:
```
modules/test-structure/save_test_data.go:203:13: cannot use t (variable of type "github.com/gruntwork-io/terratest/modules/testing".TestingT) as type string in argument to t.Fatalf

```

bt-macole:
> ```
> modules/test-structure/save_test_data.go:203:13: cannot use t (variable of type "github.com/gruntwork-io/terratest/modules/testing".TestingT) as type string in argument to t.Fatalf
> ```

fixed:

```go
--- PASS: TestSaveAndLoadTestData (0.00s)
PASS
ok      github.com/gruntwork-io/terratest/modules/test-structure        0.587s
```

bt-macole:
@denis256 anything else needed here? not urgent, but we'd like to start using this feature in some of our test suite to avoid some accidental duplicated resources.

denis256:
Hello,
the biggest concern is that PR doesn't have any test which will validate 

```
$ grep -irn "SaveTerraformOptionsIfNotPresent" .
./modules/test-structure/save_test_data.go:27:// SaveTerraformOptionsIfNotPresent serializes and saves TerraformOptions into the given folder if the file does not exist or the json is
./modules/test-structure/save_test_data.go:30:func SaveTerraformOptionsIfNotPresent(t testing.TestingT, testFolder string, terraformOptions *terraform.Options) {
$
```

Implementation of `SaveTerraformOptionsIfNotPresent` can be changed in the future, so it will not be only a call of `SaveTestData` and there will be no tests to validate that `SaveTerraformOptionsIfNotPresent` will continue to work as before

bt-macole:
@denis256 great point. added a test to test that functionality directly as well as a test that ensures the original overwritting behavior of `SaveTerraformOptions` is unchanged.

denis256 (approved):

Issue #1021, opened 2021-10-26
Copy hidden directories using CopyTerraformFolderToTemp


ztki:
Currently - `CopyTerraformFolderToTemp` allows us to copy around terraform modules to temp locations - which is particularly nice for running a bunch of parallel tests, however - it doesn't copy hidden directories such as `.terraform` which prevents me from trying to cache both my plugins and modules prior to running my tests.

Something I've run into lately when running some tests in parallel, when running `InitAndApply` is hitting infrequent - but enough to be annoying transient network errors for fetching the various modules used that are basically `Failed to download module` type errors.

I ended up hitting a wall with causing the errors to go away since it's out of my immediate control, my first approach to solving this issue was to run a single `terraform init` prior to running my tests to populate `.terraform/plugins` and `.terraform/modules` prior to running the tests and just use those - reducing my number of Inits, and also improving the efficiency of the pipeline slightly.

I ended up finding out however that `CopyTerraformFolderToTemp` doesn't end up copying the `.terraform` directory (and seems to exclude all hidden directories [here](https://github.com/gruntwork-io/terratest/blob/cb603f158b50492e429b17ea17bbbb1399a68a73/modules/files/files.go#L48)). While ultimately I'm able to use `RetryableTerraformErrors` to just retry the step to get around the problem for my for specific use-case, I still wanted to make the suggestion to potentially allow `CopyTerraformFolderToTemp` to copy hidden directories (or just `.terraform`) to the temp directories.

While Terraform's native functionality does offer the ability to have a separate directory for the plugin-cache by setting a `plugin_cache_dir` env var - there (currently) exists no matching functionality to be able to manage a separate `modules-cache` type directory - so all downloaded modules are expected to be in `.terraform/modules` (see https://github.com/hashicorp/terraform/issues/16268). 

I think a solution built into `CopyTerraformFolderToTemp`, or a separate function to allow this would be more desirable than a hacky workaround where we copy modules around manually for those who want to make use of a module cache while still being able to use `CopyTerraformFolderToTemp`

I don't mind (attempting to) implement something, but wanted to put the idea out there to collect any sort of feedback prior to making any sort of attempts.

Thanks!

brikis98:
[...]

PR #1353, opened 2023-10-05
feat: add `*WithFilter` helper functions

AgustinRamiroDiaz:
## Description

Fixes https://github.com/gruntwork-io/terratest/issues/1021

I've tested this in my local test suite and it works :ok_hand: 

Question for the reviewers: would it make sense to approach this with the builder design pattern? It'd break the currently API but I think it'd be easier to maintain to add more of this configurations/features and more configurable

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Added `Copy*WithFilter` family funcitons like `CopyTerraformFolderToTempWithFilter`

### DIFF
--- a/modules/files/files.go
+++ b/modules/files/files.go
@@ -38,8 +38,13 @@ func IsExistingDir(path string) bool {
 	return err == nil && fileInfo.IsDir()
 }
 
-func CopyTerraformFolderToDestWithFilter(folderPath string, destRootFolder string, tempFolderPrefix string, fileFilter func(string) bool) (string, error) {
-	destFolder, err := CopyFolderToDest(folderPath, destRootFolder, tempFolderPrefix, fileFilter)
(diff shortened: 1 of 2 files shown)


denis256:
Hello,
to reduce the number of function arguments, can be introduced `options` struct which will have inside all required fields, to keep backward compatibility - existing functions can instantiate `options` struct and pass it to functions with options argument.

It would also be useful to have tests that can check if the new functions will keep working in the future.

Reference:
https://github.com/gruntwork-io/terratest/blob/master/modules/helm/options.go

AgustinRamiroDiaz:
> to reduce the number of function arguments, can be introduced options struct which will have inside all required fields, to keep backward compatibility - existing functions can instantiate options struct and pass it to functions with options argument.

Great idea! Do you think it'd be useful to do that?

> It would also be useful to have tests that can check if the new functions will keep working in the future.

I'll try to add some tests :+1:

james03160927:
Will close this PR since there were no activity for the past 2 years. I can pick it up and make the remaining change.

Issue #746, opened 2020-12-21
Extend auth options for k8s helpers / KubectlOptions


kirecek:
Hello,

AFAIK, currently, KubectlOptions configuration supports support only auth by using kubeconfig file, right?

I wonder if it would make sense to add also option to support auth without using kubeconfig file and use endpoint, token, cert data directly.

I am using terraform+kubernetes and I would like to get rid of kubeconfig file generators for providers defined like this:

```hcl
provider "kubernetes" {
  load_config_file          = false
  host                            = "https://${google_container_cluster.cluster.endpoint}"
  cluster_ca_certificate = base64decode(google_container_cluster.cluster.master_auth.0.cluster_ca_certificate)
  token                          = data.google_client_config.current.access_token
}
```

Or auth creds pulled from remote state etc..

No idea what is the recommended approach here tbh.

WDYT?

yorinasub17:
This seems reasonable. I think one approach we can take is to use [a similar function we defined in kubergrunt](https://github.com/gruntwork-io/kubergrunt/blob/b696756da9e41b96770ff2edbfbad68f0a745872/kubectl/config.go#L155) to manage the auth flow, although I am not 100% sure what the best way to wrap the `kubergrunt` API in terratest would be such that it is backwards compatible but also DRY (given that the structs are different).

charlierm:
This is definitely something that would be very handy, currently having to make a call to `aws eks update-kubeconfig` which relies on having the CLI installed on our CI which isn't ideal. 

I am happy to have a stab at this but not sure it would it be the best.

PR #1354, opened 2023-10-07
add an option for passing custom k8s rest config

kirecek:
## Description

I met again this issue https://github.com/gruntwork-io/terratest/issues/746 so trying draft some solution.

I think, the easiest for users would be just to pass their rest config. Or should I wrap it somehow?

```go
	options := k8s.NewKubectlOptionsWithRestConfig(
		&rest.Config{
			Host: server,
			TLSClientConfig: rest.TLSClientConfig{
				CAData: ca,
			},
			BearerToken: token,
		},
		"terratest",
	)
```

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Added
- `NewKubectlOptionsWithRestConfig` to extend auth options for kubernetes client

### DIFF
--- a/modules/k8s/client.go
+++ b/modules/k8s/client.go
@@ -34,6 +34,9 @@ func GetKubernetesClientFromOptionsE(t testing.TestingT, options *KubectlOptions
 			return nil, err
 		}
 		logger.Log(t, "Configuring Kubernetes client to use the in-cluster serviceaccount token")
+	} else if options.RestConfig != nil {
+		config = options.RestConfig
+		logger.Log(t, "Configuring Kubernetes client to use provided rest config object set with API server address: %s", config.Host)
 	} else {
 		kubeConfigPath, err := options.GetConfigPath(t)
 		if err != nil {
--- a/modules/k8s/kubectl_options.go
+++ b/modules/k8s/kubectl_options.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/testing"
+	"k8s.io/client-go/rest"
 )
 
 // KubectlOptions represents common options necessary to specify for all Kubectl calls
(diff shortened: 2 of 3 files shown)


modules/k8s/kubectl_options.go
```diff
@@ -13,6 +14,7 @@ type KubectlOptions struct {
 	Env           map[string]string
 	InClusterAuth bool
 	Logger        *logger.Logger
+	restConfig    *rest.Config
```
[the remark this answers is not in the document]
kirecek: > should be a private field?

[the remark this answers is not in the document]
kirecek: Thank you for review.

> Why restConfig should be a private field? other fields are public

Yeah, I was confused by that. I just could not think why it should be exported so went with the private approach.

Updated ✅ 

> Will be helpful to have tests that will track that field is used

Added simple test generating rest config from config file 🙈 Not sure about this approach

modules/k8s/client.go
```diff
@@ -34,6 +34,9 @@ func GetKubernetesClientFromOptionsE(t testing.TestingT, options *KubectlOptions
 			return nil, err
 		}
 		logger.Log(t, "Configuring Kubernetes client to use the in-cluster serviceaccount token")
+	} else if options.RestConfig != nil {
+		config = options.RestConfig
+		logger.Log(t, "Configuring Kubernetes client to use provided rest config object")
```
denis256, line 39: Will be helpful to log which config file was used, which will help in future debugging

kirecek: Thanks for reply. However, I'm sorry, I think I don't understand what exactly are you suggesting to log.

RestConfig don't use the kubeconfig file, but instead, it uses direct credentials (API address, cert file, etc..) from a provided object as you can see in the description.

__Note:__ _RestConfig can be generated also from the kubeconfig file the same as I did it in the test. But goal of the PR is that it does not matter where the config comes from._

Maybe I could improve the log by providing atleast a kube API address. Will check what can be useful here as there are more parameters.

kirecek: Added server address to log as it seems the only attr that should be always present.

denis256 (approved):

PR #1376, opened 2023-11-30
Fix root path lookup in validate-all

### LINKED ISSUE #1374: ValidateAllTerraformModules has a number of issues
1. The `ValidateAllTerraformModules` function doesn't work on all repo structures. In particular, it calls [`runValidateOnAllTerraformModules`](https://github.com/gruntwork-io/terratest/blob/master/modules/test-structure/test_structure.go#L190C6-L190C38) under the hood, which tries to find the "git root" through a 100% hard-coded path of "go two folders up": `filepath.Abs(filepath.Join(cwd, "../../"))`. This does not work for any Terraform code that isn't exactly two folders from the Git root! In fact, in a repo I tried this on, the code was just one folder down, and two folders up was a folder with ~100GB of code, so when this function then tried to copy all this to temp, it would hang. This code should use [`git.GetRepoRoot`](https://github.com/gruntwork-io/terratest/blob/master/modules/git/git.go#L101) instead.
1. (Minor) The `ValidateAllTerraformModules` method is in the `test_structure` package. Since this is a function for validating _Terraform_, it should be in the `terraform` package.

brikis98:
## Description

Fixes #1374. 

I've updated the `runValidateOnAllTerraformModules` function, which is used under the hood by `ValidateAllTerraformModules` and `OPAEvalAllTerraformModules`, to:

1. Find the Git root using `git.GetRepoRootForDirE`, instead of a hard-coded `../../` file path.
2. Copy the Git root to a temp folder once, rather than once per sub-test.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/modules/git/git.go
+++ b/modules/git/git.go
@@ -2,6 +2,7 @@
 package git
 
 import (
+	"os"
 	"os/exec"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


denis256 (approved):
Failed tests seems unrelated

Issue #1228, opened 2023-01-24
Order of combined output form shell.RunCommandAndGetOutputE


mimazaka:
Hi guys, 

I would like to asked if I should expect specific order of lines in output for func `shell.RunCommandAndGetOutputE` ?

I noticed `output` struct returns internal  `merged` struct on call of `output.Combined()`. The second one contains comment that parallel writes should not be allowed.

I noticed that during manual debug of `shell.RunCommandAndGetOutputE` , especially with some slowdowns inside  execution of code within wait group defined in `readStdoutAndStderr`, I obtained output with mixed lines form `stdout `and `stderr`.

Is there any solution to have an output in specific order e.g. standard error data stream first and standard output second?

denis256:
Hi,
AFAIK there is no function to return both `stdout` and `stderr` - this can be a good improvement for terratest

PR #1482, opened 2024-12-06
add RunCommandAndGetStdOutErr

james03160927:
## Description

Fixes https://github.com/gruntwork-io/terratest/issues/1228.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)


modules/shell/command.go
```diff
@@ -84,6 +84,27 @@ func RunCommandAndGetStdOutE(t testing.TestingT, command Command) (string, error
...
+	return stdout, stderr
+}
+
+// RunCommandAndGetStdOutErrE runs a shell command and returns solely its stdout and stderr as a string. The stdout
+// and stderr of that command will also be printed to the stdout and stderr of this Go program to make debugging easier.
+// Any returned error will be of type ErrWithCmdOutput, containing the output streams and the underlying error.
+func RunCommandAndGetStdOutErrE(t testing.TestingT, command Command) (stdout string, stderr string, err error) {
+	output, err := runCommand(t, command)
+	if err != nil {
+		return output.Stdout(), output.Stderr(), &ErrWithCmdOutput{err, output}
```
denis256, line 102: 'output' might have 'nil' value?


In some cases can be returned error and nil...

https://github.com/gruntwork-io/terratest/blob/main/modules/shell/command.go#L109


it may require to change in multiple places

denis256 (approved):

modules/shell/command.go
james03160927: Got it. Will address this in a follow-up PR.

james03160927: Looking at the code again, it is actually safe for output to be nil since it will return empty string
- https://github.com/gruntwork-io/terratest/blob/main/modules/shell/output.go#L30
- https://github.com/gruntwork-io/terratest/blob/main/modules/shell/output.go#L38
- https://github.com/gruntwork-io/terratest/blob/main/modules/shell/output.go#L46

PR #1505, opened 2025-01-20
Use HasSuffix to check for suffix

amangale:
For checking the existence of a workspace, the existing regular expression match is not required and is prone to failure. Given that the workspace name is essentially a suffix, the idiomatic way would be to use `HasSuffix`.


james03160927:
Hi @amangale, I believe this change might not be backward compatible since it alters the existing behavior. Could you elaborate on why this change is necessary and clarify what you mean by "prone to failure"?

amangale:
> Hi @amangale, I believe this change might not be backward compatible since it alters the existing behavior. Could you elaborate on why this change is necessary and clarify what you mean by "prone to failure"?

Hi @james03160927 - 
The earlier code was looking for the ws name at the end of the string which is a suffix.
We recently upgraded to TF v1.10 and Terragrunt v0.72 and modified the logging. This broke `WorkspaceSelectOrNewE` since the output of `terraform workspace list` was a different string with the ws name at the end.
I have this sample code:  https://go.dev/play/p/g8JiafkjXmU 
It demonstrates backward compatibility. If there is a specific use case you can share, I can modify the PR to accommodate it.

james03160927:
If we can modify the code to support the same functionality while also accommodating new use cases with the upgrade, it will ensure that existing users do not encounter sudden errors, especially in scenarios where their automatic setups update to the latest Terratest version.

amangale:
> If we can modify the code to support the same functionality while also accommodating new use cases with the upgrade, it will ensure that existing users do not encounter sudden errors, especially in scenarios where their automatic setups update to the latest Terratest version.

Hi @james03160927  - 
In the playground link, all the existing test cases are passing. This ensures backward compatibility and none of the existing functionality will break because of this change. Is there any other way we can ensure complete and total backward compatibility. Let me know, I would be happy to try that.

amangale:
@james03160927 - If it helps, here is the change in `workspace list` I am referring to:
(with terragrunt version v0.72.0 and terraform version v1.10.4)
`12:57:34.856 STDOUT terraform:   default
12:57:34.857 STDOUT terraform:   workspace01-dasdasd
12:57:34.857 STDOUT terraform:   workspace02-sdfdsfds
12:57:34.857 STDOUT terraform:   workspace03-cvxcvxc
12:57:34.857 STDOUT terraform:   workspace04-qweqwe`
(with terragrunt version v0.58.2 and terraform version v1.9.8)
`* default
  sample01workspace01-dasdasd  
  sample02workspace01-gdfgghgfh 
  sample03workspace01-ouiouioiu `
Given that the ws name is always the suffix, I have proposed this change.

james03160927:
Thanks for the clarification. Will trigger the test pipeline and see if it passes all the other tests. 
In the meantime, can you click the "update branch" button to merge with the latest change in main?

james03160927:
It's failing the precommit test

```
- hook id: goimports
```

amangale:
@james03160927 - 
The PR has been updated and the imports have been fixed.

james03160927:
Triggering test again.

james03160927:
Failing tests doesn't seem to be related to your change. LGTM.

james03160927 (approved):
LGTM

PR #1507, opened 2025-02-03
Stop DynamoDB methods with E Failing Immediately

robmorgan:
This PR changes the DynamoDB methods ending in `E` to return an error explicitly rather than calling other methods which fail the test immediately by calling `require.NoError`.


james03160927 (approved):
LGTM

robmorgan:
@james03160927 thanks, will do 👍

james03160927:
> @james03160927 thanks, will do 👍

Hmm still seems to be out of date.

robmorgan:
@james03160927 there's been a few commits since, might be better now?

james03160927 (approved):
LGTM

robmorgan:
@james03160927 thanks! 👍

Issue #1508, opened 2025-02-04
Terragrunt log formatting env vars are ignored when set at OS level


bt-macole:
**Describe the bug**
When trying to set log formatting variables for Terragrunt at the OS level they are ignored in terratest.

**To Reproduce**
`export TERRAGRUNT_LOG_FORMAT=bare`
run terratest
observe logs are still `key-value` format.

**Expected behavior**
Expect the OS env vars to be used for terragrunt log format config if not explicitly overwritten with options.envVars

**Versions**
- Terratest version: v0.48+

bt-macole:
@denis256 @yhakbar any insights into when this will be included in a release? 🙏

yhakbar:
We'll look to cut a release within the next week.

denis256:
Hello,
changes are available in https://github.com/gruntwork-io/terratest/releases/tag/v0.49.0

PR #1509, opened 2025-02-04
fix(cmd): support custom TG logger settings

bt-macole:
Terragrunt log formatting was only supported within the options.envVars. If settings were defined at the OS env level, they were ignored.

This will now check options.envVar and then os.env vars before using the default settings for TG log formatting in terratest.

Fixes: #1508

## Release Notes (draft)

Updated Terragrunt log formatting configuration

### DIFF
--- a/modules/terraform/cmd.go
+++ b/modules/terraform/cmd.go
@@ -2,6 +2,7 @@ package terraform
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


modules/terraform/cmd.go
```diff
@@ -188,3 +178,34 @@ func hasWarning(opts *Options, out string) error {
...
+		options.EnvVars = map[string]string{}
+	}
+
+	_, inOpts := options.EnvVars[tgLogFormatKey]
+	if !inOpts {
+		_, inEnv := os.LookupEnv(tgLogFormatKey)
+		if !inEnv {
+			// key-value format for terragrunt logs to avoid colors and have plain form
+			// https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-log-format
+			options.EnvVars[tgLogFormatKey] = "key-value"
```
cbowlby-bt, line 200: The documented default is 'pretty'

bt-macole: `key-value` format for terragrunt logs to avoid colors and have plain form by default inside terratest

Issue #1023, opened 2021-11-02
`OutputMapOfObjects` throws "Type switching to map[string]interface{} failed." when output objects contain values of type list


sbgillett:
I have a test which grabs a Terraform output which is a list of objects (Azure public IP addresses).

```terraform
output "public_ip_addresses" {
  value = azurerm_public_ip.all
}
```

The referenced resource `azurerm_public_ip.all` contains the `for_each` argument, so the output is a list of objects. So `terraform output -json public_ip_addresses` would give something like

```
{
  "pip1": {
    "allocation_method": "Static",
    "availability_zone": "2",
    "domain_name_label": null,
    "fqdn": null,
    "id": "/subscriptions/someguid/resourceGroups/example/providers/Microsoft.Network/publicIPAddresses/pip-example-1",
    "idle_timeout_in_minutes": 15,
    "ip_address": "1.2.3.4",
    "ip_tags": null,
    "ip_version": "IPv4",
    "location": "westeurope",
    "name": "pip-example-1",
    "public_ip_prefix_id": null,
    "resource_group_name": "example",
    "reverse_fqdn": null,
    "sku": "Standard",
    "sku_tier": "Regional",
    "tags": {
      "foo": "bar"
    },
    "timeouts": null,
    "zones": [
      "2"
    ]
  },
  "pip2": {
    "allocation_method": "Static",
    ...etc...
  }
}
```

I have a unit test that tries to retrieve this output in a useful form using

```go
tfOutput := terraform.OutputMapOfObjects(t, terraformOptions, "public_ip_addresses")
```

But it throws an error `"Type switching to map[string]interface{} failed."`

This seems to happen at the point it tries to parse the `zones` property of the first object in the map, as it is a list: https://github.com/gruntwork-io/terratest/blob/cb603f158b50492e429b17ea17bbbb1399a68a73/modules/terraform/output.go#L62

asnowfix:
Do you know of any work-around to validate a map content if it happens to contain a list?

sbgillett:
It's not a great workaround, but I just added new outputs to my Terraform confirm which effectively return only the map properties I'm interested in (which happen not to be those of list type in my case).

pinaki-accenture:
Facing the same issue in my case also. I have an output file with the list of data disk names of the linux_vm module. Any help, please.

bsch150:
Just stumbled into the same issue. We will also workaround with a separate output but it's not what I would have preferred

[...]

PR #1529, opened 2025-03-23
Update parseListOfMaps to handle non-map data structures

james03160927:
## Description

Fixes https://github.com/gruntwork-io/terratest/issues/1023.

## TODOs

- [ ] Update the docs.
- [ ] Make a plan for release of the functionality in this PR. If it delivers value to an end user, you are responsible for ensuring it is released promptly, and correctly. If you are not a maintainer, you are responsible for finding a maintainer to do this for you.

## Release Notes (draft)

### DIFF
--- a/modules/terraform/output.go
+++ b/modules/terraform/output.go
@@ -2,7 +2,6 @@ package terraform
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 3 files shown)


modules/terraform/output.go
```diff
@@ -131,6 +100,32 @@ func parseMap(m map[string]interface{}) (map[string]interface{}, error) {
 	return result, nil
 }
 
+func parseList(items []interface{}) (_ []interface{}, err error) {
+	for i, v := range items {
+		rv := reflect.ValueOf(v)
+		switch rv.Kind() {
```
denis256, line 106: Missing case for `default` :thinking: 

should be value copied as is?
```
default:
  result[i] = v
```

james03160927: no we don't need the default case, since we are not copying or moving the array item, but instead overriding the item(s) if it match certain conditions. so if doesn't match the conditions, we don't do anything and the value is intact. Similar to this https://go.dev/play/p/D6cnrJMiltP

denis256 (approved):

PR #1561, opened 2025-05-30
Updated the subnet struct and corresponding function to fetch the subnet CIDR range

Atchuthmullapudi:
Description:

1.Updated subnet struct to include CIDR block variable 
2.Updated the function GetSubnetsForVpcE to assign the CIDR block value to the struct variable before returning.

By this update users can fetch the subnet CIDR along with the other details for writing tests, particularly this is usefull when user trying to check if the CIDR allocated to the created subnet is expected or not with out writing any extra code.


james03160927 (approved):
LGTM the failing test doesn't seem to be related to this change.

PR #1621, opened 2025-11-18
Simplify terragrunt function names

### LINKED ISSUE #1618: Rename run-all functions to use Tg prefix for consistency
## Problem

The terragrunt module has inconsistent function naming:
- Stack commands: `TgStackGenerate`, `TgStackRun`, `TgStackClean`
- Init command: `TgInit`
- Run-all commands: `ApplyAll`, `DestroyAll`, `PlanAllExitCode` (missing `Tg` prefix)

This inconsistency is confusing and makes the API less discoverable.

## Proposal

Rename run-all functions to use the `Tg` prefix:
- `ApplyAll` → `TgApplyAll`
- `ApplyAllE` → `TgApplyAllE`
- `DestroyAll` → `TgDestroyAll`
- `DestroyAllE` → `TgDestroyAllE`
- `PlanAllExitCode` → `TgPlanAllExitCode`
- `PlanAllExitCodeE` → `TgPlanAllExitCodeE`

Keep old function names as deprecated aliases for backward compatibility:

```go
// Deprecated: Use TgApplyAll instead.
func ApplyAll(t testing.TestingT, options *Options) string {
    return TgApplyAll(t, options)
}
```

## Benefits

- Consistent naming across all terragrunt module functions
- Better API discoverability
- Clearer distinction from terraform module functions
- Easier to identify terragrunt-specific functions in code

## Breaking Change

This would be a breaking change requiring a major version bump. The deprecation period would give users time to migrate.

james00012:
## Description

Simplifies function names by removing the `Tg` prefix, following Terragrunt's own simplification of removing `terragrunt-` prefixes from CLI flags.

⚠️ **BREAKING CHANGE**: All function names have been updated. Users will need to update their test code when upgrading.

## Changes

**Function renames:**
- `ApplyAll` (was `TgApplyAll`)
- `DestroyAll` (was `TgDestroyAll`)
- `PlanAllExitCode` (was `TgPlanAllExitCode`)
- `Init` (was `TgInit`)
- `StackGenerate`, `StackRun`, `StackClean` (removed `Tg` prefix)
- `Output`, `OutputJson`, `OutputAll` (removed `Tg` prefix)

All tests and documentation updated.

## Migration Guide

```go
// Before
terragrunt.TgApplyAll(t, options)
defer terragrunt.TgDestroyAll(t, options)
terragrunt.TgInit(t, options)

// After
terragrunt.ApplyAll(t, options)
defer terragrunt.DestroyAll(t, options)
terragrunt.Init(t, options)
```

Simple find-replace: Remove `Tg` prefix from all terragrunt function calls.

## Benefits

- Simpler, cleaner API
- Package name already indicates these are terragrunt functions
- Consistent with Terragrunt CLI simplification
- Less typing, easier to read

Fixes #1618

### DIFF
--- a/modules/terragrunt/stack_clean.go
+++ b/modules/terragrunt/stack_clean.go
@@ -4,18 +4,18 @@ import (
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
-// TgStackClean calls tg stack clean to remove the .terragrunt-stack directory
+// StackClean calls terragrunt stack clean to remove the .terragrunt-stack directory
 // This command cleans up the generated stack files created by stack generate or stack run
-func TgStackClean(t testing.TestingT, options *Options) string {
-	out, err := TgStackCleanE(t, options)
+func StackClean(t testing.TestingT, options *Options) string {
+	out, err := StackCleanE(t, options)
 	if err != nil {
 		t.Fatal(err)
 	}
 	return out
 }
 
-// TgStackCleanE calls tg stack clean to remove the .terragrunt-stack directory
+// StackCleanE calls terragrunt stack clean to remove the .terragrunt-stack directory
 // This command cleans up the generated stack files created by stack generate or stack run
-func TgStackCleanE(t testing.TestingT, options *Options) (string, error) {
+func StackCleanE(t testing.TestingT, options *Options) (string, error) {
 	return runTerragruntStackCommandE(t, options, "clean")
 }
--- a/modules/terragrunt/stack_generate.go
+++ b/modules/terragrunt/stack_generate.go
@@ -4,16 +4,16 @@ import (
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
-// TgStackGenerate calls tg stack generate and returns stdout/stderr
-func TgStackGenerate(t testing.TestingT, options *Options) string {
-	out, err := TgStackGenerateE(t, options)
+// StackGenerate calls terragrunt stack generate and returns stdout/stderr
+func StackGenerate(t testing.TestingT, options *Options) string {
+	out, err := StackGenerateE(t, options)
 	if err != nil {
 		t.Fatal(err)
 	}
 	return out
 }
 
-// TgStackGenerateE calls tg stack generate and returns stdout/stderr
-func TgStackGenerateE(t testing.TestingT, options *Options) (string, error) {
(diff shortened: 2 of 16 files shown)


yhakbar:
@james00012 Why do this? We just removed the `terragrunt-` prefixes for flags in Terragrunt :)

It seems odd that we'd prefix them like this `terragrunt.TgApplyAll(t, options)`

james00012:
> @james00012 Why do this? We just removed the `terragrunt-` prefixes for flags in Terragrunt :)
> 
> It seems odd that we'd prefix them like this `terragrunt.TgApplyAll(t, options)`

Yeah that's a good idea.

modules/terragrunt/stack_clean.go
```diff
@@ -4,18 +4,18 @@ import (
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
-// TgStackClean calls tg stack clean to remove the .terragrunt-stack directory
+// StackClean calls terragrunt stack clean to remove the .terragrunt-stack directory
 // This command cleans up the generated stack files created by stack generate or stack run
-func TgStackClean(t testing.TestingT, options *Options) string {
-	out, err := TgStackCleanE(t, options)
+func StackClean(t testing.TestingT, options *Options) string {
```
denis256: StackClean() is not invoked in any test

modules/terragrunt/stack_generate.go
```diff
@@ -4,16 +4,16 @@ import (
 	"github.com/gruntwork-io/terratest/modules/testing"
 )
 
-// TgStackGenerate calls tg stack generate and returns stdout/stderr
-func TgStackGenerate(t testing.TestingT, options *Options) string {
-	out, err := TgStackGenerateE(t, options)
+// StackGenerate calls terragrunt stack generate and returns stdout/stderr
+func StackGenerate(t testing.TestingT, options *Options) string {
```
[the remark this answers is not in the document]
james00012: Hi @denis256, most of the current tests are focused on calling Stack*E. If needed, I can add tests for StackGenerate as well, even without checking for errors—but since it’s just a simple wrapper, I think covering one of the functions should be sufficient.

modules/terragrunt/stack_clean.go
james00012: ditto.

PR #1632, opened 2025-12-01
Modularize Terratest into 27 independent modules

james00012:
## Summary

This PR modularizes Terratest into 27 independent Go modules with `/v2` versioning, enabling users to import only the modules they need while maintaining full backward compatibility.

### Key Changes

- **Added `/v2` suffix** to all module paths (e.g., `github.com/gruntwork-io/terratest/modules/aws/v2`)
- **Restored root `go.mod`** for backward compatibility - `go get github.com/gruntwork-io/terratest` continues to work
- Split into 27 independent modules under `modules/`
- Each module has its own `go.mod` with explicit dependencies
- Added `go.work` for workspace-based development
- Updated documentation with `/v2` import paths

### Module Tiers

- **Tier 0 (no deps):** collections, testing
- **Tier 1 (minimal deps):** logger, random, retry, shell, files, environment, git
- **Tier 2:** ssh, http-helper, dns-helper, oci, opa, packer, docker, database, slack
- **Tier 3:** terraform, terragrunt, aws, azure, gcp, k8s, helm, test-structure, version-checker

### Backward Compatibility

| Command | Result |
|---------|--------|
| `go get github.com/gruntwork-io/terratest` | ✅ Works (gets root module v0.x.x) |
| `go get github.com/gruntwork-io/terratest/modules/aws/v2` | ✅ Works (gets modular v2.x.x) |

### After Merge - Tagging

```bash
git tag modules/aws/v2.0.0
git tag modules/terraform/v2.0.0
# ... for all modules
git push origin --tags
```

## Testing Performed

Validated `/v2` module versioning with a test repository:

| Test | Command | Result |
|------|---------|--------|
| ✅ `/v2` import | `go get .../random/v2@v2.0.0` | Downloads module correctly |
| ✅ Root module | `go get ...@v0.1.0` | Backward compatible |
| ✅ Without `/v2` | `go get .../random@v2.0.0` | Fails with clear error message |
| ✅ Latest | `go get .../random/v2@latest` | Finds latest v2.x.x tag |

Additional verification:
- `go build ./...` passes
- `go vet ./...` passes
- Unit tests pass (collections, random, files modules)

## Test Plan

- [x] CircleCI passes all tests
- [x] Verify modules can be imported independently with `/v2` paths


MIGRATION.md
```diff
@@ -0,0 +1,22 @@
+# Migration Guide: Terratest v1.0.0
+
+Starting with v1.0.0, Terratest is split into independent Go modules.
+
+## Migration
+
+**Before:**
+```bash
+go get github.com/gruntwork-io/terratest
```
ThisGuyCodes: I feel strongly that this command needs to continue to work for existing users.

[The accepted way to do breaking changes is to add a version suffix to the import path](https://go.dev/blog/v2-go-modules). You don't have to do this with a new directory, you can just change the module declaration in `go.mod`:

```
module github.com/gruntwork-io/terratest/v1
```

And make sure you tag appropriately.

Go tooling will allow the non-suffix'd `go get` command to continue to work, by pulling the latest `v0.x.x` tag. New users (or folks who want to upgrade) would run:
```bash
go get github.com/gruntwork-io/terratest/v1
```

And they'll get the latest `v1.x.x` version that's tagged.

This is where we need to decide if we're versioning each module individually (in which case we also need to be _really_ careful about never letting module A return non-interface types from module B), or version all of Terratest together (this is what I would advocate for).

thisguycodes (changes requested):
I consider this command:
```bash
go get github.com/gruntwork-io/terratest
```
as part of the API. Causing this command to break existing users will cause a ton of friction and frustration. We should lean into ecosystem conventions here and version the module import path.

It may seem like we need to make a new directory + copy everything, keeping both in the repo, but the only reason [this is recommended](https://go.dev/blog/v2-go-modules) is to support non-module-aware go tooling (the post linked is from 2019); afaik we have no intention of supporting non-module-aware tooling as all of that is way out of official support at this point.

First we need to decide if we're going to version each module independently (in which case we also need to be really careful about never letting module A return non-interface types from module B), or version all of Terratest together (this is what I would advocate for).

Then we place a major version number in the module directive in each `go.mod` file. Typically this is placed at the end:

```
module github.com/gruntwork-io/terratest/modules/aws/v1
```

the bare `go get` command (without the version suffix) will continue to pull the latest `v0.x.x` tag, allowing us to even branch + release bug fixes for old versions. And so long as we tag the `/v1` suffixed modules with `v1.x.x` then `go get github.com/gruntwork-io/terratest/modules/aws/v1` will continue to work when we release a `/v2` etc.

thisguycodes (approved):
Approved! Though I request that we remove the root `go.mod`, as it may confuse folks to have `go get github.com/gruntwork-io/terratest@v2` work but have `import "github.com/gruntwork-io/terratest"` not.

However it can be removed later without it having been there in previous v2 versions without ill effect. At minimum I think if we `go mod tidy` the root `go.mod` it'll end up empty.

Also, I'm approving the code changes, but since we intend to version the whole repo as one, I do not agree with the tagging plan. We can just make one `v2.0.0` tag, there is no need to tag all the submodules separately (this mechanism exists purely to allow the submodules to be independently versioned / released).

Just be sure *not* to tag any new commits with a `v0` tag!

PR #1642, opened 2025-12-25
Fix bugs in database, docker, and slack modules

james00012:
## Summary
- Fix database rows defer before nil check (would panic on error)
- Fix docker multierror.Append missing first argument (errors weren't accumulating)
- Fix slack returning wrong error type (was fmt.Errorf, should be MessageNotFoundErr)


james00012:
Superseded by #1646

PR #1756, opened 2026-04-11
feat(azure): add WithClient variants and azfake unit tests for all modules

james00012:
## Summary

Adds exported `WithClient` variants for Azure module functions that accept a pre-built SDK client, enabling unit testing with fake servers. Each existing `ContextE` function now delegates its SDK call to the `WithClient` variant.

Pattern established across all modules:
- `GetDiskContextE(ctx, name, rg, sub)` -- convenience, creates its own client
- `GetDiskWithClient(ctx, client, rg, name)` -- testable, accepts injected client

Functions that extract fields from response objects (not clients) drop the `WithClient` suffix:
- `GetAvailabilitySetFaultDomainCount(avs)`
- `GetLoadBalancerFrontendIPConfigNames(lb)`
- `GetNetworkInterfacePrivateIPs(nic)`
- `GetIPOfPublicIPAddressByName(pip)`
- `GetVirtualNetworkDNSServerIPs(vnet)`

All WithClient functions are tested using Azure SDK's azfake framework (httptest for servicebus beta SDK), running in CI without credentials.

**Modules updated (16 files):** disk, availabilityset, loadbalancer, networkinterface, publicaddress, virtualnetwork, cosmosdb, servicebus

### Breaking changes

None. All changes are additive — new exported functions only. No existing function signatures were modified.

## Validated

- [x] `go build ./...`
- [x] `go test ./modules/azure/...` -- all unit tests pass
- [x] `make lint` -- 0 issues

### DIFF
--- a/modules/azure/loadbalancer.go
+++ b/modules/azure/loadbalancer.go
@@ -2,6 +2,7 @@ package azure
 
 import (
 	"context"
+	"errors"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
 	"github.com/gruntwork-io/terratest/modules/testing"
@@ -89,22 +90,7 @@ func GetLoadBalancerFrontendIPConfigNamesContextE(ctx context.Context, loadBalan
 		return nil, err
 	}
 
-	// Get the Frontend IP Configurations
-	feConfigs := lb.Properties.FrontendIPConfigurations
-
-	if len(feConfigs) == 0 {
-		// No Frontend IP Configuration present
-		return nil, nil
-	}
-
-	// Get the names of the Frontend IP Configurations present
-	configNames := make([]string, len(feConfigs))
-
-	for i, config := range feConfigs {
-		configNames[i] = *config.Name
-	}
-
-	return configNames, nil
+	return ExtractLoadBalancerFrontendIPConfigNames(lb), nil
 }
 
 // GetIPOfLoadBalancerFrontendIPConfig gets the IP and LoadBalancerIPType for the specified Load Balancer Frontend IP Configuration.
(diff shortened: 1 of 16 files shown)


james00012 (commented):
-

modules/azure/loadbalancer.go
```diff
@@ -285,6 +268,61 @@ func GetLoadBalancerContextE(ctx context.Context, loadBalancerName string, resou
...
+	feConfigs := lb.Properties.FrontendIPConfigurations
+
+	if len(feConfigs) == 0 {
+		return nil
+	}
+
+	configNames := make([]string, len(feConfigs))
+
+	for i, config := range feConfigs {
+		configNames[i] = *config.Name
```
denis256: Makes sense to add a nil check on `config.Name` before dereferencing ? (skip or error)

modules/azure/loadbalancer.go
```diff
@@ -285,6 +268,61 @@ func GetLoadBalancerContextE(ctx context.Context, loadBalancerName string, resou
...
+// to resolve the public IP address.
+func GetIPOfLoadBalancerFrontendIPConfigWithClient(ctx context.Context, feConfig *armnetwork.FrontendIPConfiguration, pipClient *armnetwork.PublicIPAddressesClient, resourceGroupName string) (string, LoadBalancerIPType, error) {
+	if feConfig.Properties == nil {
+		return "", NoIP, errors.New("frontend IP configuration has nil properties")
+	}
+
+	feProps := feConfig.Properties
+
+	if feProps.PublicIPAddress != nil {
+		pipName := GetNameFromResourceID(*feProps.PublicIPAddress.ID)
```
denis256, line 306: Makes sense to check for feProps.PublicIPAddress.ID == nil before GetNameFromResourceID ?

denis256 (dismissed):

modules/azure/loadbalancer.go
james00012: Added — switched to `append` with a nil check so configs with nil names are skipped.

modules/azure/loadbalancer.go
james00012: Added — the condition now checks both `PublicIPAddress != nil && PublicIPAddress.ID != nil` before dereferencing.

denis256 (approved):

PR #1770, opened 2026-04-17
fix(aws): align auth.go with package convention for v1 readiness

james00012:
## Why

The auth.go surface shipped with three deviations from the convention every other module in `modules/aws` follows:

- No `testing.TestingT` parameter on any function
- No panic variants (only error-returning forms)
- `Context` suffix meant "takes ctx, returns error" — the opposite of the rest of the package, where `Context` panics and `ContextE` errors

v1.0.0 freezes the public surface, so this is the last window to land the idiomatic shape without a compat story.

Adding the idiomatic variants in place wasn't possible: the existing names were already taken by non-idiomatic signatures and Go can't overload. So the new canonical surface lands under renamed families that sidestep each collision — and every pre-existing public function is preserved byte-for-byte as a `// Deprecated:` shim pointing at its replacement, so no external caller breaks.

## Renames

| Old (kept as deprecated shim)              | New canonical                     | Reason                                                                        |
| ------------------------------------------ | --------------------------------- | ----------------------------------------------------------------------------- |
| `NewAuthenticatedSession*`                 | `NewAuthConfig*`                  | Name collision; also aligns with v2 SDK terminology (`*aws.Config`, not Session) |
| `CreateAwsSessionWith{Creds,Mfa}*`         | `NewAuthConfigWith{Creds,Mfa}*`   | Naming consistency within the family                                          |
| `GetTimeBasedOneTimePassword`              | `GenerateMfaToken[E]`             | Name collision                                                                |
| `ReadPasswordPolicyMinPasswordLength*`     | `GetPasswordPolicyMinLength*`     | Name collision                                                                |

The canonical surface now provides the standard 4-variant shape (panic / error / with-ctx-panic / with-ctx-error), with `testing.TestingT` as the first parameter. `NewAuthConfigWithCreds` intentionally has no `Context` variant — the function performs no I/O.

## Internal migration

22 call sites across 21 files moved from `NewAuthenticatedSessionContext(ctx, region)` → `NewAuthConfigContextE(t, ctx, region)`, and 2 `GetTimeBasedOneTimePassword(mfaDevice)` calls in `iam.go` → `GenerateMfaTokenE(t, mfaDevice)`. Every call site was inside a `New*ClientContextE(t, ctx, region)` helper where `t` was already in scope, so migrations are mechanical single-line swaps.

## Test plan

- [x] `go build ./modules/aws/...` — clean
- [x] `go vet ./modules/aws/...` — clean
- [x] `go test -count=1 -run='^$' ./modules/aws/...` — tests compile
- [x] Grepped for remaining references to old names outside `auth.go` — none
- [x] Confirmed all 13 pre-existing public functions still present with original signatures
- [x] Confirmed `CredentialsError` struct defined exactly once
- [ ] Integration smoke on any auth-dependent test in CI

### DIFF
--- a/modules/aws/account.go
+++ b/modules/aws/account.go
@@ -75,7 +75,7 @@ func ExtractAccountIDFromARN(arn string) (string, error) {
 // NewStsClientContextE creates a new STS client.
 // The ctx parameter supports cancellation and timeouts.
 func NewStsClientContextE(t testing.TestingT, ctx context.Context, region string) (*sts.Client, error) {
-	sess, err := NewAuthenticatedSessionContext(ctx, region)
+	sess, err := NewAuthConfigContextE(t, ctx, region)
 	if err != nil {
 		return nil, err
 	}
--- a/modules/aws/acm.go
+++ b/modules/aws/acm.go
@@ -64,7 +64,7 @@ func GetAcmCertificateArnE(t testing.TestingT, awsRegion string, certDomainName 
 // NewAcmClientContextE creates a new ACM client.
 // The ctx parameter supports cancellation and timeouts.
 func NewAcmClientContextE(t testing.TestingT, ctx context.Context, region string) (*acm.Client, error) {
-	sess, err := NewAuthenticatedSessionContext(ctx, region)
+	sess, err := NewAuthConfigContextE(t, ctx, region)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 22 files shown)


james00012:
Closing — auth.go is internal plumbing in practice (only callers are other files in modules/aws). The convention/rename polish doesn't justify the churn or the deprecation warnings it would emit to any external callers that do exist. Revisiting for v2 if a real need emerges.

PR #1772, opened 2026-04-19
fix(aws): honor context cancellation in SSM retries and IAM MFA sleeps

james00012:
## Why

The SSM and IAM helpers accept a `ctx context.Context` but silently ignore cancellation in places that loop or sleep, causing parallel tests to hang past their ctx deadline. This is a v1-readiness blocker.

## What

- **SSM:** `WaitForSsmInstanceWithClientContextE` and `CheckSSMCommandWithClientWithDocumentContextE` now short-circuit on `ctx.Err()` via `retry.FatalError` inside each retry closure, so cancellation is observed within one retry interval instead of the full timeout. Both functions also check ctx once at entry. The client parameter is narrowed to a new `SsmAPI` interface so these behaviors can be regression-tested without AWS credentials — `*ssm.Client` satisfies the interface structurally, so existing callers are unaffected.
- **IAM:** `EnableMfaDeviceContextE` replaces its two blocking `time.Sleep` calls (30s + 10s) with a new `sleepWithContext` helper that selects on `ctx.Done()`. The helper also pre-checks ctx and prefers cancellation when timer and ctx fire simultaneously, so cancellation always wins.
- **Tests:** New regression tests under `modules/aws/` assert that cancellation ends the retry or sleep within one retry interval, and that both `context.Canceled` and `context.DeadlineExceeded` propagate via `errors.Is`.

## Test plan

- [x] `go build ./modules/aws/...`
- [x] `go vet ./modules/aws/...`
- [x] `go test -count=1 -race -run 'TestSleep|TestWaitForSsm|TestCheckSSMCommand' ./modules/aws/...` — 9/9 pass, each completes in ~100ms with cancelled ctx (well under the 2s retry interval)
- [x] `golangci-lint run ./modules/aws/...` — 0 issues


james00012:
Closing — on further review, the ctx-cancellation gap in EnableMfaDeviceContextE is minor enough that it doesn't justify the extra surface. Callers rarely cancel mid-MFA-setup; when they do, ~40s of extra waiting is annoying but not a correctness bug. Can revisit if a real use case surfaces.

PR #1773, opened 2026-04-19
fix(aws): prevent RDS/IAM panics and paginate ACM certificate lookups

james00012:
## Why

Three correctness bugs in AWS helpers that panic or silently return wrong results.

## What

- **RDS** — guard four unchecked `[0]` slice accesses in `GetOptionGroupNameOfRdsInstanceContextE`, `GetOptionsOfOptionGroupContextE`, `GetAllParametersOfRdsInstanceContextE`, `GetRdsInstanceDetailsContextE`. Return a descriptive error instead of panicking.
- **ACM** — `GetAcmCertificateArnWithClientContextE` now paginates on `NextToken` so matches past page 1 are no longer missed. Extends the existing mock test with a two-page case.
- **IAM** — `GetIamPolicyDocumentContextE` guards `*version.VersionId` against nil, and fails fast with a clear error when no default version is found (previously fell through with an empty `VersionId` and surfaced a generic AWS validation error).
- Drops a stray `fmt.Println(err)` in `rds_test.go`.

## Test plan

- [x] `go build / vet / test -race ./modules/aws/...`
- [x] `golangci-lint run ./modules/aws/...` — 0 issues

Stacks on #1771.

### DIFF
--- a/modules/aws/acm_test.go
+++ b/modules/aws/acm_test.go
@@ -3,6 +3,7 @@ package aws_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 
 	awsSDK "github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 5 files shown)


modules/aws/acm_test.go
```diff
@@ -14,16 +14,27 @@ import (
...
+	callCount              int
 }
 
 func (m *mockAcmClient) ListCertificates(_ context.Context, _ *acm.ListCertificatesInput, _ ...func(*acm.Options)) (*acm.ListCertificatesOutput, error) {
 	if m.ListCertificatesErr != nil {
 		return nil, m.ListCertificatesErr
 	}
 
+	if len(m.ListCertificatesPages) > 0 {
+		page := m.ListCertificatesPages[m.callCount]
```
denis256, line 37: mock test will panic with index-out-of-range on over-call

denis256 (dismissed):

denis256 (approved):

PR #1774, opened 2026-04-19
fix(aws)!: align public API for v1 stable release

james00012:
## Why
Finalizes the `modules/aws/` public API for v1.0.0. After v1, these
changes become breaking.

## What
- auth.go: align every public function with the package's 4-tier
  convention (testing.TestingT first arg, Context/ContextE variants,
  deprecated non-context shims).
- vpc.go: add the missing `GetVpcsContext` panic variant.
- CredentialsError: stop embedding the underlying SDK error in the
  user-facing message; expose it via Unwrap().

## Test plan
- [x] `go build ./...` — clean.
- [x] `go vet ./...` — clean.
- [x] `go test -count=1 -race ./modules/aws/...` — one pre-existing,
      unrelated failure in `TestGetS3BucketOwnershipControls/NotExist`
      (AWS now always returns a default ownership control; the test's
      expectation that `GetS3BucketOwnershipControlsE` returns an error
      for a bucket without explicit controls is stale and reproduces
      on `origin/main`). No code paths touched in this PR are involved
      in that test.
- [ ] `golangci-lint run ./modules/aws/...` — not run locally: the
      installed `golangci-lint` (v2.6.2, built with Go 1.25) refuses
      the repo's `go 1.26` directive. CI should run it.

### DIFF
--- a/modules/aws/account.go
+++ b/modules/aws/account.go
@@ -75,7 +75,7 @@ func ExtractAccountIDFromARN(arn string) (string, error) {
 // NewStsClientContextE creates a new STS client.
 // The ctx parameter supports cancellation and timeouts.
 func NewStsClientContextE(t testing.TestingT, ctx context.Context, region string) (*sts.Client, error) {
-	sess, err := NewAuthenticatedSessionContext(ctx, region)
+	sess, err := NewAuthConfigContextE(t, ctx, region)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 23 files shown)


james00012:
Closing — auth.go isn't on the end-user call path, so aligning it with the package convention has no real impact. The breaking-change risk isn't worth the cleanup for code users don't invoke directly.

PR #1780, opened 2026-04-20
fix(oci): nil-safety and pagination in list calls

james00012:
## Why
- `imageSort.Less` panics when an OCI image has nil `TimeCreated` (SDK field is `*common.SDKTime`).
- Two list helpers call `ListVcns` / `ListImages` once and ignore `OpcNextPage`, silently missing results past the first page.

## What
- Guard `TimeCreated` nil-deref in `imageSort.Less`; treat nil as oldest.
- Loop on `OpcNextPage` in `GetAllVcnIDsContextE` (`ListVcns`) and the `ListImages` caller in `compute.go`.
- `ListAvailabilityDomains` was not changed: the SDK's `ListAvailabilityDomainsRequest` in this version has no `Page` field, so there is no way to request subsequent pages. Availability domains per compartment are small (3-6 in all real regions) and fit in one response.

## Test plan
- [x] `go build ./modules/oci/...`
- [x] `go vet ./modules/oci/...`
- [x] `go test -count=1 -race ./modules/oci/...` (no unit tests in package; build-tagged integration tests unchanged)
- [x] `golangci-lint run ./modules/oci/...`

### DIFF
--- a/modules/oci/network.go
+++ b/modules/oci/network.go
@@ -23,16 +23,35 @@ func GetAllVcnIDsContextE(t testing.TestingT, ctx context.Context, compartmentID
 
 	request := core.ListVcnsRequest{CompartmentId: &compartmentID}
 
-	response, err := client.ListVcns(ctx, request)
-	if err != nil {
-		return nil, err
+	var allItems []core.Vcn
+
+	for {
+		response, err := client.ListVcns(ctx, request)
+		if err != nil {
+			return nil, err
+		}
+
+		allItems = append(allItems, response.Items...)
+
+		// Stop when no next page, when the server returns an empty token, or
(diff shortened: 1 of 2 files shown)


modules/oci/network.go
```diff
@@ -23,16 +23,28 @@ func GetAllVcnIDsContextE(t testing.TestingT, ctx context.Context, compartmentID
 
 	request := core.ListVcnsRequest{CompartmentId: &compartmentID}
 
-	response, err := client.ListVcns(ctx, request)
-	if err != nil {
-		return nil, err
+	var allItems []core.Vcn
+
+	for {
```
denis256, line 28: Possible infinite loop

denis256 (dismissed):

denis256 (approved):

PR #1781, opened 2026-04-20
fix(ssh): harden FetchContentsOfFile against injection and fix error-path bugs

james00012:
## Why
Three defensive fixes in modules/ssh:
- Command injection via `FetchContentsOfFileContextE` interpolating user-supplied filePath directly into a shell command.
- Index-out-of-range panic in `listFileInRemoteDir` when the remote command returns empty output.
- `copyFileFromRemote` logs an SDK error but continues, writing nil/partial bytes to the local file.

## What
- Add an unexported `shellQuote` helper; use it around filePath in `FetchContentsOfFileContextE`.
- Guard the trailing-newline trim with `len > 0`.
- Return early on `ssh.Session.Output` error in `copyFileFromRemote`; close the session via defer placed before the Output call.
- Unit test for `shellQuote` covering special chars and injection attempts.

## Test plan
- [x] go build / vet / test -race / golangci-lint

### DIFF
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -646,11 +646,18 @@ func FetchContentsOfFileContext(t testing.TestingT, ctx context.Context, host *H
 	return out
 }
 
+// shellQuote returns s wrapped in single quotes, with any embedded single quote
+// replaced by the four-character sequence quote-backslash-quote-quote.
+// Safe for POSIX sh, bash, zsh.
+func shellQuote(s string) string {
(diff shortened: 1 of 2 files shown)


james00012:
Closing — terratest helpers take filePath from the test author, not external input, so the injection scenario is self-directed. The two correctness fixes (bounds check in listFileInRemoteDir, return-on-error in copyFileFromRemote) are worth revisiting separately if they surface in practice.

PR #1792, opened 2026-04-23
ci: automate weekly Terragrunt pin updates

james00012:
Adds `.github/workflows/update-terragrunt-pin.yml`, which runs every Monday (and on manual dispatch), reads the latest Terragrunt release tag from GitHub, and — if it differs from the pin in `mise.toml` — opens a PR bumping it. Mirrors the existing `update-lint-config.yml` pattern.

The CircleCI `TERRAGRUNT_VERSION` in `.circleci/config.yml` is intentionally untouched: CircleCI is being phased out and `mise.toml` is the authoritative pin for GHA jobs.

## Tested

Ran end-to-end on a personal fork by rolling `mise.toml` back to `0.93.10` and dispatching the workflow. It ran in 11s and opened a PR with the single-line diff `terragrunt = "0.93.10"` → `"1.0.2"`, correct title, and a working release-notes link. `actionlint` + `shellcheck` + YAML parse all clean.

Closes OSS-3157.


.github/workflows/update-terragrunt-pin.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    steps:
+    - uses: actions/checkout@v5
+
+    - name: Bump Terragrunt pin in mise.toml
+      id: bump
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        latest=$(gh api repos/gruntwork-io/terragrunt/releases/latest --jq '.tag_name' | sed 's/^v//')
```
yhakbar: NIT: Prefer storing this in a dedicated script in `.github/scripts` so that we can run shellcheck/shfmt on it.

.github/workflows/update-terragrunt-pin.yml
```diff
@@ -0,0 +1,40 @@
...
+        current=$(sed -nE 's/^terragrunt = "([^"]+)".*/\1/p' mise.toml)
+        echo "latest=$latest" >> "$GITHUB_OUTPUT"
+        echo "current=$current" >> "$GITHUB_OUTPUT"
+        sed -i -E "s/^(terragrunt = \")[^\"]+(\")/\\1${latest}\\2/" mise.toml
+
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v7
```
yhakbar: This isn't conditional, right? Shouldn't we be evaluating whether the two are different?

yhakbar (changes requested):
I think we need the pull request step to be conditional. Also, consider having the actions pinned to commits.

.github/workflows/update-terragrunt-pin.yml
james00012: Updated.

.github/workflows/update-terragrunt-pin.yml
james00012: Updated.

.github/scripts/update-terragrunt-pin.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Fetch the latest Terragrunt release and update the pin in mise.toml.
+
+set -e
```
denis256: Not sure about `set -e` without `pipefail` allows `gh api ... | sed ...` to continue if gh fails.

james00012: Updated.

.github/scripts/update-terragrunt-pin.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
yhakbar: NIT: prefer `/usr/bin/env bash` so that this script will accept any `bash` in `PATH`.

yhakbar (dismissed):
Looks great!

denis256 (dismissed):

.github/scripts/update-terragrunt-pin.sh
james00012: Updated.

yhakbar (approved):

PR #1793, opened 2026-04-24
ci: add Azure integration tests GitHub Actions workflow

james00012:
## Summary

Adds `.github/workflows/azure-integration-tests.yml` to run `go test -tags azure ./test/azure/...` against a real Azure subscription.

- Triggers: push to `main`, same-repo PRs, `workflow_dispatch` (paths-filtered to Azure-relevant changes).
- Auth: OIDC via `azure/login@v2` using the `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` / `AZURE_SUBSCRIPTION_ID` repo secrets — no client secret stored.
- Fork-PR exposure blocked at job level (`if: head.repo.full_name == github.repository`).
- `concurrency:` group cancels superseded runs on the same ref to avoid overlapping Azure resource provisioning.
- 27/28 tests in `test/azure/` will run; `terraform_azure_sqlmanagedinstance_example_test.go` is excluded by its `azure_ci_excluded` build tag (6–8h provision time).
- Complementary to existing `azure-tests.yml`, which only compiles `./test/azure/` and runs `./modules/azure/...` unit tests.

## Out of scope

- Verifying the Azure AD federated credential trust policy is scoped (single-subject vs. wildcard) — Azure-side configuration owned by Yousif.
- Deleting the stale `AZURE_CREDENTIALS` secret (2 years old, unused now that OIDC is wired up).

## Test plan

- [ ] Workflow runs on this PR and Azure login step authenticates.
- [ ] Integration test job completes (or fails informatively) within the 60m timeout.
- [ ] Fork PRs do not see secrets (verified by `if:` guard; cannot test without a fork PR).

PR #1794, opened 2026-04-24
ci: add Azure integration tests GitHub Actions workflow

james00012:
Adds `.github/workflows/azure-integration-tests.yml` — runs Azure integration tests against a real subscription via GitHub OIDC + `azure/login@v2` + `ARM_USE_OIDC=true`. No client secret stored. Fork PRs are blocked at the job level so secrets are never exposed.

Complementary to existing `azure-tests.yml` (compile-only on `test/azure/`, unit tests on `modules/azure/`). Replaces #1793 (closed; was opened from a fork, where the same fork-PR guard prevented the workflow from running).

## Scope

The AzureRM provider gained OIDC support in 3.7.0. 21 of 23 Azure example modules pin AzureRM 2.x and cannot auth via OIDC. This PR therefore scopes the run to the two examples already on 3.x:

- `TestTerraformAzureContainerAppExample`
- `TestTerraformAzureKeyVaultExample`

Bringing the remaining 21 under CI requires a separate provider-upgrade effort.

## Subscription prerequisites (to land green)

- Register `Microsoft.App` resource provider on the subscription (`az provider register --namespace Microsoft.App`) — currently blocks Container App with 409 `MissingSubscriptionRegistration`.
- `terraform-azure-keyvault-example` imports a hardcoded test cert that's expired — separate fix needed in the example.

## Out of scope

- Upgrading 21 example modules from AzureRM 2.x.
- Verifying Azure AD federated credential trust scope (Yousif).
- Deleting stale `AZURE_CREDENTIALS` secret (Yousif).


.github/workflows/azure-integration-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+          ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          ARM_USE_OIDC: "true"
+        run: |
+          set -o pipefail
+          mkdir -p /tmp/logs
+          # Scope: OIDC-compatible examples only (AzureRM provider >= 3.7).
+          # Examples on AzureRM 2.x cannot authenticate via OIDC and are
+          # gated behind a follow-up provider-upgrade effort.
+          go test -v -p 1 -tags azure -count=1 -timeout 45m \
+            -run '^(TestTerraformAzureContainerAppExample|TestTerraformAzureKeyVaultExample)$' \
```
denis256, line 85: Looks fragile - can we have some patterns for test names 
Like:
https://github.com/gruntwork-io/terragrunt/blob/main/.github/workflows/integration-test.yml#L109C19-L109C31

denis256 (approved):

.github/workflows/azure-integration-tests.yml
james00012: Got it. Will address in a follow-up

PR #1797, opened 2026-04-27
fix(aws): honor context cancellation in IAM MFA sleeps

james00012:
## Summary

`EnableMfaDeviceContextE` accepts a `context.Context` but its two `time.Sleep` calls did not honor it. Cancelling a context could not interrupt a 30-second MFA propagation wait.

## Changes

- New unexported `sleepContext(ctx, d)` helper in `modules/aws/iam.go` that selects on `ctx.Done()` and `time.After(d)`, returning `ctx.Err()` on cancellation.
- Both sleeps in `EnableMfaDeviceContextE` now use the helper.
- `iam_internal_test.go`: one test verifying that cancelling mid-sleep returns promptly with `context.Canceled`.

## Note on SSM

The SSM waits I originally suspected as also non-honoring already use `retry.DoWithRetryableErrorsContextE`, which honors context internally. No change needed there.

## Test plan

- [ ] CI passes
- [x] `go test -run TestSleepContext ./modules/aws/...` passes locally

Linear: LIB-4933

### DIFF
--- a/modules/aws/iam.go
+++ b/modules/aws/iam.go
@@ -244,7 +244,9 @@ func EnableMfaDeviceContextE(t testing.TestingT, ctx context.Context, iamClient 
 	const mfaEnableWait = 30 * time.Second
 
 	logger.Default.Logf(t, "Waiting 30 seconds for a new MFA Token to be generated...")
-	time.Sleep(mfaEnableWait)
+	if err := sleepContext(ctx, mfaEnableWait); err != nil {
+		return err
+	}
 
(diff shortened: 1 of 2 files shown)


james00012:
Closing without merge: this is over-engineered for a test library. Test-level `-timeout` already covers cancellation; `time.Sleep` in MFA propagation waits is bounded (a few seconds), and test authors rarely manually cancel the ctx they pass into terratest helpers. The 45-line addition (helper + test) doesn't earn its keep.

Linear LIB-4933 marked Cancelled and dropped from v1 milestone.

PR #1809, opened 2026-05-04
fix(azure): correct Manged typo in SQL client factory functions

james00012:
## Summary
- Add correctly-spelled `CreateSQLManagedInstanceClient{,Context}` and `CreateSQLManagedDatabasesClient{,Context}` as canonical names.
- Convert the misspelled `CreateSQLManged*` quartet to `// Deprecated:` aliases delegating to the new functions.
- Update internal call sites in `modules/azure/sql_managedinstance.go` to use the new spellings.

Tagging v1.0 with the typo would lock it into the stable API until a major version bump.

Closes OSS-3452.

## Test plan
- [x] `go build ./modules/azure/...` passes
- [ ] CI lint + Azure integration tests

PR #1813, opened 2026-05-04
fix(api): rename *Context functions returning errors to *ContextE

james00012:
## Summary
Thirteen exported functions returned `(T, error)` under names ending in `*Context`, violating the `*ContextE` convention used elsewhere. Tagging v1.0 with the convention violation would lock it into the stable API.

For each function:
- The body moves to a new canonical `*ContextE` function.
- The original `*Context` name stays as a `// Deprecated:` alias delegating to the new one, so existing callers keep compiling.
- Internal callers and the pre-existing non-Context deprecated wrappers are repointed at the new `*ContextE` directly.

Renamed (full list in commit message):
- `modules/aws/auth.go`: 6 functions in the `NewAuthenticatedSession*`, `CreateAwsSession*`, and `ReadPasswordPolicyMinPasswordLength` families.
- `modules/azure/client_factory.go`: 5 SQL/database/Action Group factories.
- `modules/azure/sql.go`: `GetSQLServerClientContext`, `GetDatabaseClientContext`.

Closes OSS-3451.

## Test plan
- [x] `go build ./modules/aws/... ./modules/azure/...` passes
- [ ] CI lint + integration tests

### DIFF
--- a/modules/aws/account.go
+++ b/modules/aws/account.go
@@ -93,7 +93,7 @@ func ExtractAccountIDFromARN(arn string) (string, error) {
 // NewStsClientContextE creates a new STS client.
 // The ctx parameter supports cancellation and timeouts.
 func NewStsClientContextE(t testing.TestingT, ctx context.Context, region string) (*sts.Client, error) {
-	sess, err := NewAuthenticatedSessionContext(ctx, region)
+	sess, err := NewAuthenticatedSessionContextE(ctx, region)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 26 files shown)


james00012:
Note for reviewers: this PR conflicts with #1809 (the Manged typo fix) since both touch the SQL managed-instance/database client functions in modules/azure/client_factory.go and the call sites in modules/azure/sql_managedinstance.go. Recommend merging #1809 first, then rebasing this PR on top.

james00012:
Closing without merge.

The 13 renamed functions (modules/aws/auth.go session factories, modules/azure/{client_factory,sql}.go client getters) are all internal plumbing wrapped by higher-level helpers. Zero direct usage in examples/ or test/ or any other documented call site. The convention violation is only visible in godoc to someone browsing the underlying client factories — which end users don't typically need to do.

The cost (13 deprecated aliases living in v1.x forever, 22 internal caller migrations) outweighs the benefit. Accepting as a v1 wart, in line with the other casing aliases (Vpc.Id, SshAgent, HttpGetOptions, SetJsonValues) we're already shipping with.

Issue #1762, opened 2026-04-15
Support for S3BucketServerSideEncryption


creiche:
**Describe the solution you'd like**
I would like to be able to assert Bucket SSE is enabled. It appears it was added in this PR https://github.com/gruntwork-io/terratest/pull/1494 but the removed by this commit https://github.com/gruntwork-io/terratest/commit/fb3ba0a56d6177f569aacac930eba70b12479e58.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1816, opened 2026-05-08
feat(aws): add AssertS3BucketServerSideEncryption helpers

james00012:
## Summary
Restores the SSE assertion helpers (originally added in #1494, dropped in fb3ba0a) and brings them in line with the modern Context/ContextE pattern. Adds a typed `BucketServerSideEncryptionNotEnabledError`.

Public surface:
- `AssertS3BucketServerSideEncryptionContext` / `AssertS3BucketServerSideEncryptionContextE`
- Deprecated `AssertS3BucketServerSideEncryption` / `AssertS3BucketServerSideEncryptionE` (delegate to the Context variants)
- `BucketServerSideEncryptionNotEnabledError` + `NewBucketServerSideEncryptionNotEnabledError`

Closes [LIB-4875](https://linear.app/gruntwork/issue/LIB-4875). Upstream issue: gruntwork-io/terratest#1762.

## Test plan
- [ ] `go vet ./modules/aws/...`
- [ ] `go build ./modules/aws/...`
- [ ] CI: `TestAssertS3BucketServerSideEncryption` runs against AES256 and aws:kms; verifies positive match and negative-match error path

### DIFF
--- a/modules/aws/s3.go
+++ b/modules/aws/s3.go
@@ -912,6 +912,64 @@ func AssertS3BucketPolicyExistsE(t testing.TestingT, region string, bucketName s
 	return AssertS3BucketPolicyExistsContextE(t, context.Background(), region, bucketName)
 }
 
+// AssertS3BucketServerSideEncryptionContextE checks if the given S3 bucket has server-side encryption configured with
+// the given algorithm, and returns an error if it does not. The ctx parameter supports cancellation and timeouts.
+func AssertS3BucketServerSideEncryptionContextE(t testing.TestingT, ctx context.Context, region string, bucketName string, algorithm types.ServerSideEncryption) error {
+	s3Client, err := NewS3ClientContextE(t, ctx, region)
+	if err != nil {
+		return err
+	}
+
+	out, err := s3Client.GetBucketEncryption(ctx, &s3.GetBucketEncryptionInput{
+		Bucket: aws.String(bucketName),
+	})
+	if err != nil {
+		return err
+	}
+
+	if out.ServerSideEncryptionConfiguration == nil {
+		return NewBucketServerSideEncryptionNotEnabledError(bucketName, region, algorithm, "")
+	}
+
+	for _, rule := range out.ServerSideEncryptionConfiguration.Rules {
+		if rule.ApplyServerSideEncryptionByDefault == nil {
+			continue
+		}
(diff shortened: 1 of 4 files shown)


modules/aws/s3_test.go
```diff
@@ -327,6 +328,55 @@ func TestGetS3BucketOwnershipControls(t *testing.T) {
...
+							ApplyServerSideEncryptionByDefault: &types.ServerSideEncryptionByDefault{
+								SSEAlgorithm: algorithm,
+							},
+						},
+					},
+				},
+			})
+			require.NoError(t, err)
+
+			terraaws.AssertS3BucketServerSideEncryption(t, region, s3BucketName, algorithm)
```
denis256, line 365: Looks like we don't have a test for case of no SSE config

modules/aws/s3.go
```diff
@@ -912,6 +914,75 @@ func AssertS3BucketPolicyExistsE(t testing.TestingT, region string, bucketName s
...
+//
+// The algorithm is matched exactly: an expectation of `aws:kms` will not match a bucket configured with
+// `aws:kms:dsse`, and vice versa.
+func AssertS3BucketServerSideEncryptionContextE(t testing.TestingT, ctx context.Context, region string, bucketName string, algorithm types.ServerSideEncryption) error {
+	s3Client, err := NewS3ClientContextE(t, ctx, region)
+	if err != nil {
+		return err
+	}
+
+	out, err := s3Client.GetBucketEncryption(ctx, &s3.GetBucketEncryptionInput{
```
denis256, line 928: Was wondering why cant be invoked GetS3BucketServerSideEncryptionContextE()

modules/aws/s3_test.go
```diff
@@ -327,6 +328,55 @@ func TestGetS3BucketOwnershipControls(t *testing.T) {
...
+	algorithms := []types.ServerSideEncryption{
+		types.ServerSideEncryptionAes256,
+		types.ServerSideEncryptionAwsKms,
+	}
+	for i, algorithm := range algorithms {
+		algorithm := algorithm
+		t.Run(string(algorithm), func(t *testing.T) {
+			t.Parallel()
+
+			s3BucketName := fmt.Sprintf("gruntwork-terratest-sse-%d-%s", i, strings.ToLower(random.UniqueID()))
```
denis256, line 346: Loop index `i` is interpolated into the bucket name, but adds no uniqueness

denis256 (approved):

PR #1837, opened 2026-06-15
deprecate collections and environment packages

james00012:
Part 2 of 3 of the Terratest V2 deprecation work.

## Summary
Mark the `collections` and `environment` packages for removal in v2 and migrate every in-repo caller off them to the standard library.

## Intent
The V2 proposal narrows Terratest to infrastructure testing. `collections` and `environment` are stdlib wrappers (slice/map helpers, env-var lookups) that no longer pay rent now that Go 1.21+ ships `slices`. Deprecating them signals the direction to consumers a full release ahead of removal.

## Decision
Deprecate now, remove in v2. Because surviving packages (aws, azure, gcp, k8s, test-structure) imported these, the deprecation would trip staticcheck SA1019 in CI unless the callers move at the same time, so the migration ships in this PR. The shared list logic that several packages needed lives in a single private `internal/collections` package rather than being duplicated per file or kept on the public API.

## What changed
- `// Deprecated:` notices on every exported symbol in `modules/collections` and `modules/environment`, each naming the stdlib replacement.
- New `internal/collections` package with generic `Intersection`/`Subtract` over `slices`, with unit tests.
- Callers migrated: `aws/region.go`, `azure/region.go`, `gcp/region.go`, `test-structure/validate_struct.go` use `internal/collections`; `azure/resourceid.go` uses `strings.Split`; `gcp/provider.go` and `k8s/config.go` use `os.Getenv`.
- Added a direct test for `azure.GetNameFromResourceIDE` (it now owns its error construction).

## Illustration
```go
// before
regionsToPickFrom = collections.ListSubtract(regionsToPickFrom, forbiddenRegions) // public, deprecated pkg
// after
import "github.com/gruntwork-io/terratest/internal/collections" // private, single source
regionsToPickFrom = collections.Subtract(regionsToPickFrom, forbiddenRegions)
```

## Validation
- `go build ./...`: clean.
- `go vet` on all changed packages: clean.
- `go test ./internal/collections/...`: pass (covers dedup, empty-not-nil, no input mutation).
- `go test ./modules/collections/...`: pass (original behavior unchanged).
- `go test ./modules/azure -run TestGetNameFromResourceID`: pass (success and error path).
- Behavior equivalence reviewed against each original helper (dedup, order, nil-vs-empty, no input mutation, error path): no differences.
- `golangci-lint` could not run locally (installed build targets go1.25, repo targets go1.26). SA1019 is provably unreachable: zero cross-package files import the deprecated packages after this change. Full lint runs in CI.
- E2E against phxdevops: not applicable. No real AWS/cloud behavior changes; the migrated code is pure region-list filtering and env-var lookups.

## Known limitations / follow-ups
Actual package removal happens in v2. When the per-domain module split lands, `internal/collections` moves into the shared core module (or each module inlines it), since root `internal/` is not importable across separate modules.

## Summary by CodeRabbit

* **Deprecations**
  * Expanded deprecation notices for legacy collections and environment helpers, with guidance to use Go standard-library alternatives (`slices`, `os.Getenv`, `strings.Split`).
* **New Features**
  * Added generic set-style helpers for intersection and subtraction, enabling consistent deduplicated filtering of regions/zones.
* **Improvements**
  * Updated region/zone filtering across cloud modules to use the new helpers.
  * Improved Azure resource ID name parsing with clearer “not found” error handling.
  * Simplified environment variable access in a few modules/tests.
* **Tests**
  * Added/expanded parallel table-driven tests and examples for set operations and Azure resource ID parsing.


internal/collections/collections.go
```diff
@@ -0,0 +1,34 @@
...
+		if slices.Contains(list2, item) && !slices.Contains(out, item) {
+			out = append(out, item)
+		}
+	}
+
+	return out
+}
+
+// Subtract returns the items in list1 that are not in list2.
+func Subtract[T comparable](list1 []T, list2 []T) []T {
```
yhakbar, line 28: We can improve this function by doing the following:

```go
    lookups := make(map[T]struct{}, len(list2))
	for _, item := range list2 {
		lookups[item] = struct{}{}
	}

    out := make([]T, 0, len(list1))

    for _, item := range list1 {
		if _, found := lookups[item]; !found {
			out = append(out, item)
		}
	}

	return out
```

We don't know how large list1 and list2 will be on average, so we are probably OK assuming that using a map lookup will work out better than slice iteration.

internal/collections/collections.go
```diff
@@ -0,0 +1,34 @@
...
+// public modules/collections package is deprecated and scheduled for removal in
+// v2.
+package collections
+
+import "slices"
+
+// Intersection returns the items present in both lists, de-duplicated, in the
+// order they appear in list1.
+func Intersection[T comparable](list1 []T, list2 []T) []T {
+	out := []T{}
```
yhakbar: We can improve this function by doing the following:

```go
    lookups := make(map[T]struct{}, len(list2))
	for _, item := range list2 {
		lookups[item] = struct{}{}
	}

	out := make([]T, 0, min(len(list1), len(list2)))

	for _, item := range list1 {
		if _, found := lookups[item]; found {
			out = append(out, item)
			delete(lookups, item) 
		}
	}

	return out
```

This dodges the cost of iteration over both lists for every element in list1.

internal/collections/collections_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"testing"
+
+	"github.com/gruntwork-io/terratest/internal/collections"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIntersection(t *testing.T) {
+	t.Parallel()
+
+	assert.Equal(t, []string{"b", "c"}, collections.Intersection([]string{"a", "b", "c"}, []string{"b", "c", "d"}))
```
yhakbar: This should use a test table so that new cases can be added easily.

internal/collections/collections_test.go
```diff
@@ -0,0 +1,27 @@
...
+	assert.Equal(t, []string{"a"}, collections.Intersection([]string{"a", "a"}, []string{"a"}), "dedups output")
+	assert.Equal(t, []string{}, collections.Intersection([]string{"a"}, []string{"b"}), "returns empty, not nil")
+}
+
+func TestSubtract(t *testing.T) {
+	t.Parallel()
+
+	assert.Equal(t, []string{"a"}, collections.Subtract([]string{"a", "b", "c"}, []string{"b", "c"}))
```
yhakbar: This should use a test table so that new cases can be added easily.

modules/azure/resourceid.go
```diff
@@ -15,10 +18,11 @@ func GetNameFromResourceID(resourceID string) string {
 // GetNameFromResourceIDE gets the Name from an Azure Resource ID.
 // This function would fail the test if there is an error.
 func GetNameFromResourceIDE(resourceID string) (string, error) {
-	id, err := collections.GetSliceLastValueE(resourceID, "/")
-	if err != nil {
-		return "", err
+	if !strings.Contains(resourceID, "/") {
+		return "", fmt.Errorf("could not resolve name from resource ID %q", resourceID)
```
yhakbar: Avoid inline errors like this. It should return a typed error with field `resourceID`.

modules/azure/resourceid.go
```diff
@@ -15,10 +18,11 @@ func GetNameFromResourceID(resourceID string) string {
 // GetNameFromResourceIDE gets the Name from an Azure Resource ID.
 // This function would fail the test if there is an error.
 func GetNameFromResourceIDE(resourceID string) (string, error) {
```
yhakbar, line 19: This function can be done without the extra heap allocation of `strings.Split` by using `i := strings.LastIndex(resourceID, "/")` and returning `return resourceID[i+1:], nil` if `i` isn't `-1`.

modules/azure/resourceid_test.go
```diff
@@ -16,3 +16,14 @@ func TestGetNameFromResourceID(t *testing.T) {
 	resultBadSeparator := azure.GetNameFromResourceID("noresourcepresent")
 	assert.Empty(t, resultBadSeparator)
 }
+
+func TestGetNameFromResourceIDE(t *testing.T) {
+	t.Parallel()
+
+	name, err := azure.GetNameFromResourceIDE("this/is/a/long/slash/separated/string/ResourceID")
+	assert.NoError(t, err)
```
yhakbar:
```suggestion
	require.NoError(t, err)
```

modules/azure/resourceid_test.go
```diff
@@ -16,3 +16,14 @@ func TestGetNameFromResourceID(t *testing.T) {
...
+	assert.Equal(t, "ResourceID", name)
+
+	_, err = azure.GetNameFromResourceIDE("noresourcepresent")
+	assert.Error(t, err)
```
yhakbar:
```suggestion
	require.Error(t, err)
```

modules/collections/collections.go
```diff
@@ -1,2 +1,7 @@
 // Package collections allows to interact with lists of things.
+//
+// Deprecated: The collections package is scheduled for removal in Terratest v2.
+// Go's standard library covers these helpers as of Go 1.21+ (slices.Contains,
+// slices.Index, and friends); use those, or strings.Split for the string-slice
+// helpers. See the Terratest v2 migration notes for details.
```
yhakbar: This should be a link to the notes.

modules/collections/lists.go
```diff
@@ -4,6 +4,9 @@ import "slices"
 
 // ListIntersection returns all the items in both list1 and list2. Note that this will dedup the items so that the
 // output is more predictable. Otherwise, the end list depends on which list was used as the base.
+//
+// Deprecated: scheduled for removal in Terratest v2. Use the slices package
+// (Go 1.21+) to build this in a few lines at the call site.
```
yhakbar: Didn't we recreate this as `collections.Intersection`?

modules/collections/lists.go
```diff
@@ -18,6 +21,9 @@ func ListIntersection[T comparable](list1 []T, list2 []T) []T {
 }
 
 // ListSubtract removes all the items in list2 from list1.
+//
+// Deprecated: scheduled for removal in Terratest v2. Use the slices package
+// (Go 1.21+) to build this in a few lines at the call site.
```
yhakbar: Didn't we recreate this as `collections.Subtract`?

modules/collections/lists.go
```diff
@@ -32,7 +38,7 @@ func ListSubtract[T comparable](list1 []T, list2 []T) []T {
 
 // ListContains returns true if the given list of strings (haystack) contains the given string (needle).
 //
-// Deprecated: Use slices.Contains instead.
+// Deprecated: Use slices.Contains instead. Scheduled for removal in Terratest v2.
```
yhakbar: It would be a good idea to just demo the replacement code here using the standard library so that people know exactly how to make the update by pattern matching on the text from the hover.

internal/collections/collections_test.go
```diff
@@ -0,0 +1,27 @@
+package collections_test
```
yhakbar, line 1: Consider adding `Example` prefixed tests here to make it easier to learn how to use these when reading the go documentation.

modules/collections/stringslicevalue.go
```diff
@@ -5,6 +5,9 @@ import (
 )
 
 // GetSliceLastValueE will take a source string and returns the last value when split by the separator char.
+//
+// Deprecated: scheduled for removal in Terratest v2. Use strings.Split and index
```
yhakbar: It would be a good idea to just demo the replacement code here using the standard library so that people know exactly how to make the update by pattern matching on the text from the hover.

modules/collections/stringslicevalue.go
```diff
@@ -17,6 +20,9 @@ func GetSliceLastValueE(source string, separator string) (string, error) {
 
 // GetSliceIndexValueE will take a source string and returns the value at the given index when split by
 // the separator char.
+//
+// Deprecated: scheduled for removal in Terratest v2. Use strings.Split and index
```
yhakbar: It would be a good idea to just demo the replacement code here using the standard library so that people know exactly how to make the update by pattern matching on the text from the hover.

modules/environment/envvar.go
```diff
@@ -8,6 +8,8 @@ import (
 )
 
 // GetFirstNonEmptyEnvVarOrFatal returns the first non-empty environment variable from envVarNames, or throws a fatal
+//
+// Deprecated: scheduled for removal in Terratest v2. Range over os.Getenv at the call site.
```
yhakbar: It would be a good idea to just demo the replacement code here using the standard library so that people know exactly how to make the update by pattern matching on the text from the hover.

modules/environment/envvar.go
```diff
@@ -19,6 +21,8 @@ func GetFirstNonEmptyEnvVarOrFatal(t testing.TestingT, envVarNames []string) str
 
 // GetFirstNonEmptyEnvVarOrEmptyString returns the first non-empty environment variable from envVarNames, or returns the
 // empty string
+//
+// Deprecated: scheduled for removal in Terratest v2. Range over os.Getenv at the call site.
```
yhakbar: It would be a good idea to just demo the replacement code here using the standard library so that people know exactly how to make the update by pattern matching on the text from the hover.

modules/environment/envvar.go
```diff
@@ -30,6 +34,8 @@ func GetFirstNonEmptyEnvVarOrEmptyString(t testing.TestingT, envVarNames []strin
 }
 
 // RequireEnvVar fails the test if the specified environment variable is not defined or is blank.
+//
+// Deprecated: scheduled for removal in Terratest v2. Use require.NotEmpty with os.Getenv at the call site.
```
yhakbar: It would be a good idea to just demo the replacement code here using the standard library so that people know exactly how to make the update by pattern matching on the text from the hover.

modules/gcp/provider.go
```diff
@@ -33,20 +34,42 @@ var googleIdentityEmailEnvVars = []string{
...
+
+	return ""
+}
+
+// firstNonEmptyEnvVarOrFatal returns the value of the first non-empty environment variable from names, failing the test
+// if all of them are empty.
+func firstNonEmptyEnvVarOrFatal(t testing.TestingT, names []string) string {
+	value := firstNonEmptyEnvVar(names)
+	if value == "" {
+		t.Fatalf("All of the following env vars %v are empty. At least one must be non-empty.", names)
```
yhakbar: Use a require here.

yhakbar (changes requested):
The one major issue is that we're not accurate in saying that users should use a short inline replacement for `Subtract` and `Intersect`. We _are_ providing replacements.

internal/collections/collections.go
james00012: Done.

internal/collections/collections.go
james00012: Done.

internal/collections/collections_test.go
james00012: Done.

internal/collections/collections_test.go
james00012: Done.

modules/azure/resourceid.go
james00012: Done.

modules/azure/resourceid.go
james00012: Done.

modules/azure/resourceid_test.go
james00012: Done.

modules/azure/resourceid_test.go
james00012: Done.

modules/collections/collections.go
james00012: There's no v2 migration notes page published yet, so I inlined the actual stdlib replacement in the comment instead. Can swap in a link once the guide exists.

modules/collections/lists.go
james00012: Yeah, but those live in `internal/collections`, so consumers can't import them. The deprecation points external callers at the stdlib `slices` pattern instead. The internal copy just keeps our own callers DRY.

modules/collections/lists.go
james00012: Yeah, but those live in `internal/collections`, so consumers can't import them. The deprecation points external callers at the stdlib `slices` pattern instead. The internal copy just keeps our own callers DRY.

modules/collections/lists.go
james00012: Done.

internal/collections/collections_test.go
james00012: Done.

modules/collections/stringslicevalue.go
james00012: Done.

modules/collections/stringslicevalue.go
james00012: Done.

modules/environment/envvar.go
james00012: Done.

modules/environment/envvar.go
james00012: Done.

modules/environment/envvar.go
james00012: Done.

modules/gcp/provider.go
james00012: Done.

modules/azure/resourceid.go
```diff
@@ -15,10 +17,10 @@ func GetNameFromResourceID(resourceID string) string {
...
-	id, err := collections.GetSliceLastValueE(resourceID, "/")
-	if err != nil {
-		return "", err
+	i := strings.LastIndex(resourceID, "/")
+	if i == -1 {
+		return "", NewResourceIDNameNotFoundError(resourceID)
 	}
 
-	return id, nil
+	return resourceID[i+1:], nil
```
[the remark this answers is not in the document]
james00012: Good catch, fixed. Trailing-slash IDs now return the not-found error.

james00012:
Good point. The catch is those recreated versions live in `internal/collections`, so consumers can't import them, they're just there to keep our own callers DRY. For external users there's no public replacement, so inlining with `slices` is the honest path (we're dropping the public collections package in v2). Happy to reword the comment to call that out explicitly if it reads as misleading.

james00012:
Reworded the ListIntersection/ListSubtract deprecations to say there's no public replacement (the package is being dropped), so the inline slices snippet reads as intentional rather than hiding a replacement.

yhakbar (approved):

PR #1840, opened 2026-06-15
ci: allow manual dispatch of the stale workflow

james00012:
## What

Adds a `workflow_dispatch` trigger to the stale workflow so it can be run on demand from the Actions tab (or via `gh workflow run`), not only on the daily schedule.

## Why

Useful for sweeping the existing backlog immediately and for testing the workflow without waiting for the 01:30 UTC cron.

## Summary by CodeRabbit

* **Chores**
  * Enhanced repository maintenance workflow to support manual triggering alongside scheduled automated runs.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,6 +2,7 @@ name: 'Close stale issues and PRs'
 on:
   schedule:
     - cron: '30 1 * * *'
+  workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


james00012:
Not needed. The workflow runs daily on its schedule, so we'll let the next scheduled run do the first sweep rather than dispatching manually.

PR #1843, opened 2026-06-21
feat(helm): support install --generate-name by omitting empty release name

### LINKED ISSUE #1412: Use helm install without chart
Hello everyone, again:)

Sometimes I use helm install like this
```
helm install --wait --generate-name \
    -n gpu-operator --create-namespace \
    nvidia/gpu-operator
```
get from [this](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/getting-started.html#procedure)

you can see that in this example doesn't use helm chart

If I use helm.Install(t, options, "", "") I got an error 
```
Error: INSTALLATION FAILED: expected at most two arguments, unexpected arguments: <> <> <- here 2 empty element
```
because [append](https://github.com/gruntwork-io/terratest/blob/2851217e39f9479414749017faa4eddac97586ae/modules/helm/install.go#L53)

in this case, is it preferable to use **RunHelmCommandAndGetOutputE** or may be add check in Install function that **chart** and **releaseName** are empty?

Thanks in advance!
hope you are doing well!

james00012:
Fixes #1412.

helm.Install required an explicit release name and always appended it as a positional argument, so calling it with an empty release name injected a stray empty argument and helm failed with "expected at most two arguments". This blocked the helm install --generate-name pattern (installing a chart and letting helm name the release).

The release name is now omitted when empty, so callers can pass --generate-name through ExtraArgs and let helm generate the name. Callers that pass a release name are unaffected.

Arg building moved into an installArgs helper, with unit tests covering both cases. The tests are untagged so they run without a cluster.

## Release Notes

* **New Features**
  * Helm installations now support omitting the release name; when omitted, you can use Helm’s `--generate-name` via extra install arguments.

* **Bug Fixes**
  * Improved Helm install argument construction to avoid passing an empty positional release name and to preserve correct argument ordering (including extra install arguments, version, and values).

* **Tests**
  * Added unit tests to verify argument ordering with a provided vs. empty release name, and to ensure extra install arguments are retained.

* **Documentation**
  * Updated the README’s Terratest wording and removed the “What Terratest is for” section.


denis256:
Lint failures

```
	                        ^
modules/helm/install.go:64:2: missing whitespace above this line (too many lines above return) (wsl_v5)
	return err
	^
modules/helm/install.go:91:2: missing whitespace above this line (invalid statement above assign) (wsl_v5)
	args = append(args, chart)
	^
3 issues:
* contextcheck: 1
* wsl_v5: 2
make: *** [Makefile:13: lint] Error 1
```

denis256 (approved):

PR #1851, opened 2026-06-25
feat(test-structure): add opt-in parallel mode to ValidateAllTerraformModules

### LINKED ISSUE #1236: ValidateAllTerraformModules in parallel
Thoughts to making the validate function run in parallel?

https://github.com/gruntwork-io/terratest/blob/2bac268390c1bb3c1bf507d7f92018ddb8928a9d/modules/test-structure/test_structure.go#L190

james00012:
Fixes #1236.

ValidateAllTerraformModules validates each discovered module sequentially. This adds a ValidationOptions.Parallel flag: when set, each module runs in its own parallel subtest via t.Parallel, so validation of many modules can run concurrently. Go caps the concurrency at the -parallel flag (GOMAXPROCS by default), so it is bounded.

The flag defaults to false, so existing callers are unaffected. The flag is carried through CloneWithNewRootDir (the internal clone the function performs), otherwise it would never take effect.

Tests:
- A unit test that the Parallel flag survives the internal clone.
- An integration test that runs the existing valid fixture with Parallel enabled, exercising the t.Parallel path against real terraform.

### DIFF
--- a/modules/test-structure/test_structure.go
+++ b/modules/test-structure/test_structure.go
@@ -278,6 +278,10 @@ func runValidateOnAllTerraformModulesContext(
 
 	for _, dir := range dirsToValidate {
 		t.Run(strings.TrimLeft(dir, "/"), func(t *go_test.T) {
+			if clonedOpts.Parallel {
+				t.Parallel()
+			}
+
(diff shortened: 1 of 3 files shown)


james00012:
Closing without merging. We decided not to add a built in parallel mode to ValidateAllTerraformModules. See #1236.

PR #1856, opened 2026-06-29
Add v2 modularization CI gates, release workflow, and runbook

james00012:
Adds the CI tooling that will enforce the v2 module layout, plus the release runbook. This lands ahead of the split so the gates are reviewed on their own and are live the moment the modules arrive.

## Gate scripts (`scripts/`)
- `check-acyclic-deps.sh` — imports must flow downward only (core, helpers, tooling, platforms, IaC).
- `check-single-source.sh` — each package path is defined in exactly one module.
- `check-siv-placement.sh` — every module path carries the `/v2` suffix.
- `check-no-replaces.sh` — release commits carry no local `replace` directives.
- `check-release-mode.sh` — every module builds in release mode (`GOWORK=off`) and an external consumer importing all of them builds, without needing published tags.

Each gate that inspects module structure short-circuits with a "skipped (no /v2 submodules present yet)" message until the split lands, so this PR is green and the gates start doing real work automatically once the modules exist.

## Workflows (`.github/workflows/`)
- `v2-checks.yml` — runs the five gates on push and PR.
- `v2-release.yml` — manual (`workflow_dispatch`) lockstep tag push for cutting a v2 release.

## Docs
- `docs/v2-release-runbook.md` — the release procedure, including version-seeding and the tag format.

## Not in this PR
- `go.work` and the modules themselves land with the first split PR (they reference module directories that do not exist yet).

## Summary by CodeRabbit

* **New Features**
  * Added CI workflows for v2 modularization checks and a manual v2 release process, including dry-run tagging, ordered tag publishing, and post-publish visibility verification.
  * Introduced release-time validation gates covering dependency layering (acyclic imports), single-source-of-truth module declarations, correct `/v2` import placement, and release hygiene that blocks forbidden local `replace` directives.
* **Documentation**
  * Added a “Terratest v2 Release Runbook” covering coordinated pre-flight validation, version selection/pinning, tagging order, external consumer verification, and recovery guidance.


.github/workflows/v2-checks.yml
```diff
@@ -0,0 +1,54 @@
+name: v2 Modularization Checks
+
+on:
+  push:
+  pull_request:
```
yhakbar, line 6: Why are we doing this both on pushes and pull request events?

.github/workflows/v2-checks.yml
```diff
@@ -0,0 +1,54 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  acyclic-deps:
+    name: Acyclic dependency graph
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
yhakbar: This is a pretty old version of the action. We should use a modern, pinned version.

.github/workflows/v2-checks.yml
```diff
@@ -0,0 +1,54 @@
...
+      - run: bash scripts/check-acyclic-deps.sh
```
yhakbar, line 24: Can we not commit the scripts as executable?

.github/workflows/v2-checks.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Set up mise
+        uses: jdx/mise-action@v3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Release-mode build (all 16 modules pin + external consumer, GOWORK=off)
+        run: bash scripts/check-release-mode.sh
+
+  release-prep-guard:
+    name: No local terratest replaces (release commits only)
```
yhakbar: We're not using replaces anyways, right?

.github/workflows/v2-checks.yml
```diff
@@ -0,0 +1,54 @@
...
+      - run: bash scripts/check-siv-placement.sh
+
+  consumer-simulation:
+    name: External consumer simulation (GOWORK=off)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up mise
+        uses: jdx/mise-action@v3
```
yhakbar: Let's use a modern, pinned version of this.

scripts/check-no-replaces.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+# check-no-replaces.sh — fails the build if a release commit still contains local
```
yhakbar: Why do we have this? We're going with go.work anyways, right?

scripts/check-no-replaces.sh
```diff
@@ -0,0 +1,25 @@
...
+# Match both the single-line form (`replace github.com/... => ...`) and the
+# block form (`replace (` ... `github.com/... => ...` ... `)`). Both put the
+# directive's `=>` on the same line as the terratest path, while `require`
+# lines never have `=>`, so this catches every internal replace regardless of
+# style.
+matches=$(grep -nHE 'github\.com/gruntwork-io/terratest.*=>' modules/*/go.mod cmd/*/go.mod 2>/dev/null || true)
+
+if [ -n "$matches" ]; then
+  echo "::error::Local terratest replace directives present in release commit:"
+  echo "$matches" | sed 's/^/    /'
```
yhakbar: NIT: Useless use of echo.

scripts/check-release-mode.sh
```diff
@@ -0,0 +1,62 @@
...
+# check-release-mode.sh — validates the v2 modules build in release mode
+# (GOWORK=off) WITHOUT needing published tags.
+#
+# For each module it: (1) adds throwaway internal `replace`s for every sibling,
+# (2) seeds the module with the ROOT go.mod's exact dependency versions so naive
+# tidy cannot float a dep to an incompatible newer release, then (3) tidies.
+# Finally it builds an external consumer that imports every module. All changes
+# are reverted on exit; nothing is committed.
+set -uo pipefail
+cd "$(git rev-parse --show-toplevel)"
```
yhakbar: NIT: Avoid messing with parent working directory state within scripts like this. Use pushd/popd or use cd in a subshell.

scripts/check-release-mode.sh
```diff
@@ -0,0 +1,62 @@
...
+export GOFLAGS="${GOFLAGS:--tags=aws,azure,azure_ci_excluded,azureslim,compute,gcp,helm,kubeall,kubernetes,network}"
+B="github.com/gruntwork-io/terratest/modules"
```
yhakbar: Why are we using B as the name of this variable?

scripts/check-release-mode.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Pre-split guard: until the /v2 submodules exist, there is nothing to validate.
+# This lets the gate land and run green before the modularization commit, and do
+# the real release-mode build afterward.
+if ! ls modules/*/go.mod >/dev/null 2>&1; then
+  echo "release-mode check: skipped (no /v2 submodules present yet)"
+  exit 0
+fi
+
+cleanup() {
```
yhakbar, line 22: NIT: Why are we mixing global procedures with function declarations like this? the function should precede the procedural stuff.

scripts/check-release-mode.sh
```diff
@@ -0,0 +1,62 @@
...
+  # Revert each path independently. `git checkout -- a b` aborts entirely if any
+  # pathspec matches no tracked file, so a missing go.work.sum would otherwise
+  # leave the rewritten modules/*/go.mod files dirty.
+  git checkout -- modules/ >/dev/null 2>&1 || true
+  git checkout -- go.work.sum >/dev/null 2>&1 || true
+  git status --porcelain 2>/dev/null | awk '/^\?\?.*modules\/.*\/go\.sum$/{print $2}' | xargs -r rm -f
+  [ -n "${TMP:-}" ] && rm -rf "$TMP"
```
yhakbar: What are we doing here? Deleting the entire tmp directory? Why?

scripts/check-release-mode.sh
```diff
@@ -0,0 +1,62 @@
...
+# Exact dependency versions the code is tested against, from the root module.
+REQARGS=$(go mod edit -json go.mod | jq -r '.Require[]? | "-require=\(.Path)@\(.Version)"' | tr '\n' ' ')
+
+fail=0
+for m in $ORDER; do
+  ( cd "modules/$m"
+    for s in $ORDER; do [ "$s" = "$m" ] || go mod edit -replace="$B/$s/v2=../$s"; done
+    # shellcheck disable=SC2086
+    go mod edit $REQARGS
+    GOWORK=off go mod tidy ) 2>/tmp/cr_$m.err || { echo "::error::release-mode tidy failed: modules/$m"; tail -4 /tmp/cr_$m.err; fail=1; }
```
yhakbar: Avoid polluting `/tmp` like this if you can. Just create a temporary directory that we clean up on success.

scripts/check-release-mode.sh
```diff
@@ -0,0 +1,62 @@
...
+
+TMP=$(mktemp -d)
+{ echo "module releasecheckconsumer"; echo "go 1.26"; } > "$TMP/go.mod"
+for s in $ORDER; do
+  go mod edit -modfile="$TMP/go.mod" -require="$B/$s/v2@v2.0.0" -replace="$B/$s/v2=$(pwd)/modules/$s"
+done
+{
+  echo "package main"; echo "import ("
+  echo "  _ \"$B/core/v2/random\""; echo "  _ \"$B/core/v2/files\""; echo "  _ \"$B/core/v2/collections\""; echo "  _ \"$B/core/v2/formatting\""
+  for s in ssh httphelper dnshelper docker packer database opa aws azure gcp k8s helm terraform terragrunt teststructure; do echo "  _ \"$B/$s/v2\""; done
```
yhakbar: Why are we duplicating this list in both places?

yhakbar (changes requested):
There are a few changes that are essential, and some NITs.

.github/workflows/v2-checks.yml
james00012: Pinned it to the mise action SHA we already use in the repo.

scripts/check-release-mode.sh
james00012: Switched to a single mktemp -d that we clean up on exit, so nothing is left in /tmp.

.github/workflows/v2-checks.yml
james00012: Good catch. It now runs on pull requests and on pushes to main only, so feature branches do not run it twice.

scripts/check-release-mode.sh
james00012: That is a mktemp -d scratch dir, not a shared path. Renamed it to WORKDIR with a comment so it is obvious.

.github/workflows/v2-checks.yml
james00012: Pinned it to the same checkout SHA we already use elsewhere in the repo.

scripts/check-no-replaces.sh
james00012: Dropped it. Removed the script, the release-prep-guard job, and the pre-flight call. We will rely on go.work keeping go.mods clean during dev and on stripping replaces at release-prep.

.github/workflows/v2-checks.yml
james00012: Dropped it, along with the release-prep-guard job.

scripts/check-release-mode.sh
james00012: Moved the cd into main() so it stays scoped to the script and does not leak out.

scripts/check-release-mode.sh
james00012: Renamed it to MODULE_BASE.

.github/workflows/v2-checks.yml
james00012: Done. Dropped the executable bit since they run via bash scripts anyway.

scripts/check-release-mode.sh
james00012: Removed the duplication. The consumer imports are now generated from the ORDER list.

scripts/check-no-replaces.sh
james00012: Kept the OK line as a log marker so the CI output shows the check actually ran and passed. Can drop it if you prefer.

scripts/check-release-mode.sh
james00012: Moved the function declarations above the procedural code.

scripts/check-no-replaces.sh
james00012: Dropped it.

scripts/check-release-mode.sh
```diff
@@ -0,0 +1,87 @@
...
+WORKDIR=""
+
+cleanup() {
+  # Revert each path independently. `git checkout -- a b` aborts entirely if any
+  # pathspec matches no tracked file, so a missing go.work.sum would otherwise
+  # leave the rewritten modules/*/go.mod files dirty.
+  git checkout -- modules/ >/dev/null 2>&1 || true
+  git checkout -- go.work.sum >/dev/null 2>&1 || true
+  git status --porcelain 2>/dev/null | awk '/^\?\?.*modules\/.*\/go\.sum$/{print $2}' | xargs -r rm -f
+  [ -n "$WORKDIR" ] && rm -rf "$WORKDIR"
```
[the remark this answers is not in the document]
james00012: Good catch. Added a clean-worktree guard that refuses to run if modules/ or go.work.sum is dirty, so a local run cannot wipe uncommitted work.

.github/workflows/v2-checks.yml
```diff
@@ -0,0 +1,55 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  acyclic-deps:
+    name: Acyclic dependency graph
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v4.2.2
```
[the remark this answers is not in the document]
james00012: Set persist-credentials: false on the read-only checkouts (all v2-checks jobs and the release verify job). The tag-push job keeps them since it needs to push.

.github/workflows/v2-checks.yml
```diff
@@ -0,0 +1,55 @@
+name: v2 Modularization Checks
+
+on:
+  pull_request:
+  push:
+    branches: [main]
```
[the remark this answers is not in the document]
james00012: Dropped the gate and this job entirely, so this is no longer relevant.

.github/workflows/v2-checks.yml
yhakbar: Just have it run on push. This will result in it running twice on merge

scripts/check-release-mode.sh
yhakbar: FYI, this still changes the working directory of the process calling the script. 

You have to either call cd in a subshell (e.g. `(cd something)` or use `pushd`/`popd` (e.g. `pushd something >/dev/null` and `popd something >/dev/null`) to avoid changing the working directory of the process calling the script, which is what we're trying to avoid here.

Not worth fixing at this stage.

scripts/check-no-replaces.sh
yhakbar: I'm not saying you need to remove the log lines. You don't need to use `echo`/`cat` to generate stdout to operate on in other programs.

See this:
https://www.shellcheck.net/wiki/SC2002

.github/workflows/v2-checks.yml
yhakbar: Generally, I would recommend making the script executable so that the shebang is respected.

yhakbar (approved):

PR #1860, opened 2026-07-05
fix(random): prevent UniqueID collisions from per-call time seeding

zuwasi:
## Description

`random.UniqueID()` (and `Random` / `RandomInt` / `RandomString`, via the internal `newRand()`) created a fresh `*rand.Rand` seeded with `time.Now().UnixNano()` on **every call**:

```go
func newRand() *rand.Rand {
	return rand.New(rand.NewSource(time.Now().UnixNano()))
}
```

When these functions are called in a tight loop or in parallel on a machine with coarse timer resolution (e.g. Windows), consecutive calls read the same timestamp, get the same seed, produce the same random sequence, and therefore return **identical "unique" IDs**. In practice this can cause duplicate resource names / tfstate keys and flaky parallel tests.

### Reproduction

`TestUniqueID` fails reliably on Windows (Go 1.26):

```
--- FAIL: TestUniqueID (0.00s)
    random_test.go:65: map[string]bool{"Gncypw":true} should not contain "Gncypw"
    ...
```

### Fix

Use the process-global, auto-seeded (Go 1.20+) and concurrency-safe top-level `math/rand` functions (`rand.Intn`). This removes the shared-seed collision source and also avoids allocating a new generator on every call.

- **No public API changes.**
- Output remains non-deterministic — the previous code already seeded from the wall clock, so nothing relied on a deterministic/reproducible seed.

### Testing

- `go test -count=5 -run TestUniqueID ./modules/random/...` → PASS (previously FAIL)
- `go test ./modules/random/...` → PASS
- `go build ./...` and `go vet ./...` → PASS

## Release Notes (draft)

Fixed `random.UniqueID()` occasionally returning duplicate IDs when called rapidly or in parallel (per-call time-based seeding replaced with the auto-seeded, concurrency-safe global `math/rand` source).

## Summary by CodeRabbit

* **Refactor**
  * Simplified random value generation by using the shared random source directly.
  * Removed repeated creation and seeding of temporary random generators.
  * Unique ID generation now uses the same streamlined approach within its loop.

### DIFF
--- a/modules/core/random/random.go
+++ b/modules/core/random/random.go
@@ -4,12 +4,11 @@ package random
 import (
 	"bytes"
 	"math/rand"
-	"time"
 )
 
 // Random generates a random int between min and max, inclusive.
 func Random(min int, max int) int {
-	return newRand().Intn(max-min+1) + min
+	return rand.Intn(max-min+1) + min
 }
 
 // RandomInt picks a random element in the slice of ints.
@@ -33,15 +32,9 @@ const uniqueIDLength = 6 // Should be good for 62^6 = 56+ billion combinations
 func UniqueID() string {
 	var out bytes.Buffer
 
-	generator := newRand()
 	for i := 0; i < uniqueIDLength; i++ {
-		out.WriteByte(base62chars[generator.Intn(len(base62chars))])
+		out.WriteByte(base62chars[rand.Intn(len(base62chars))])
 	}
 
 	return out.String()
-}
-
-// newRand creates a new random number generator, seeding it with the current system time.
-func newRand() *rand.Rand {
-	return rand.New(rand.NewSource(time.Now().UnixNano()))
 }


james00012 (approved):
Thanks, this is a real one. Per-call time seeding collides on coarse-resolution timers (Windows especially), and the top-level math/rand functions are the right fix: auto-seeded since Go 1.20 and concurrency-safe, so the shared-seed collision goes away with no API or behavior change.

Verified locally: `go test -count=5 ./modules/random/...` passes and `go vet` is clean.

Heads up that we are splitting this file into a new core module as part of the v2 work, so we will carry this fix over there too.

LGTM.

zuwasi:
Thanks for the review @james00012! Is there anything else needed on my end before this can be merged?

james00012 (approved):
Rebased onto main and ported the fix to its new home modules/core/random/random.go (the core split moved it), dropping newRand and not reintroducing the deprecated UniqueId that #1870 removed. Collision fix unchanged.

PR #1866, opened 2026-07-10
Extract tooling (docker, packer, database, opa) into /v2 modules

james00012:
Splits the tooling tier into per-domain /v2 submodules, continuing the modularization after core (#1858) and helpers (#1864).

Stacked on #1864 (the base is the helpers branch) so this diff is tooling-only. Once #1864 merges I will retarget this to main.

## What changed
- Adds go.mod/go.sum for docker, packer, database, opa
- Dependency graph: docker depends on core/v2 and httphelper/v2; packer, database, opa depend on core/v2 only. No intra-tier deps.
- Rewrites all consumers (terraform, teststructure, test/) to the /v2 import paths
- Registers the four modules in go.work and adds root replace directives so GOWORK=off builds and the tidy check resolve them locally

Cross-module requires sit at the local zero placeholder that go.work resolves; they are pinned to real versions only on a release-prep branch.

## Validation
- All four modules build and test-compile under go.work; root and its consumers compile
- Gates pass locally: acyclic-deps, single-source (9 module paths), SIV placement, release-mode (8 modules pin at root versions, external GOWORK=off consumer builds)
- Runtime: opa/v2 (real opa + terraform eval) and docker/v2 (real docker build/run, exercises core/v2 and httphelper/v2) pass

## Summary by CodeRabbit

* **New Features**
  * Added version 2 module packaging for database, Docker, OPA, and Packer testing utilities.
  * Added database testing support covering MySQL, PostgreSQL, and Microsoft SQL Server.
  * Updated related examples and integrations to use the versioned modules.

* **Chores**
  * Refreshed supporting libraries and aligned module dependencies.
  * Updated tests and examples for compatibility with the new module versions.


go.mod
```diff
@@ -255,3 +259,11 @@ replace github.com/gruntwork-io/terratest/modules/core/v2 => ./modules/core
 replace github.com/gruntwork-io/terratest/modules/ssh/v2 => ./modules/ssh
 
 replace github.com/gruntwork-io/terratest/modules/httphelper/v2 => ./modules/httphelper
+
+replace github.com/gruntwork-io/terratest/modules/database/v2 => ./modules/database
```
yhakbar, line 263: I thought we didn't need these replace statements because of the `go.work`.

james00012: Good instinct, but go.work isn't enough on its own here. It only applies in workspace mode, and two things we depend on turn that off:

1. The CI tidy check runs plain `go mod tidy`, which works on a single go.mod and ignores go.work. When it sees `require .../core/v2` it tries to download that version, but we haven't published it yet, so it fails. The replace points it at the local copy instead.
2. The release-mode check builds with `GOWORK=off` to simulate an external consumer with no workspace. Same problem, same fix.

These are dev-only scaffolding, not a permanent design choice. The root module is never published in v2 (consumers import the submodules directly), so they never end up in a release tag. And once we cut the first v2 tags, the require lines can point at real published versions and these replaces come out entirely. go.work stays for local cross-module development. Added a comment above the block explaining it.

yhakbar (approved):

PR #1873, opened 2026-07-18
fix(logger): attribute parallel logs to the correct test under go test -json

### LINKED ISSUE #1871: Parallel logging attributes output to the wrong test
**Describe the bug**
This seems like a regression of #1457 

When I run `go test -v -json` against one of my test files, I can see that the `Test` field in the JSON log does not match the start of the `Output` line. See this example, showing `Test: TestApply2` has the output from `TestApply1` and vice versa. A small snippet is included below, but the GitHub repository contains the test code behind this which reproduces the issue reliably

I've had this problem with Terratest for years with non-JSON output where it interleaves the tests together which is its own problem, but somewhat unavoidable if you're using `-v` to stream output as it happens to keep a CI run etc alive. But many people suggest that `-json` is the solution because the log lines are attributed to the test making it easier to "un-interleave" them afterwards. That doesn't seem to be the case here, and I am not getting useful JSON output that I can do any meaningful processing with.

```json
{"Time":"2026-07-15T19:05:47.003810592+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply2","Output":"TestApply1 2026-07-15T19:05:47+01:00 logger.go:79: \u001b[0m\u001b[1mrandom_integer.time_to_sleep: Destruction complete after 0s\u001b[0m\n"}
{"Time":"2026-07-15T19:05:47.008429106+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply2","Output":"TestApply1 2026-07-15T19:05:47+01:00 logger.go:79: \u001b[0m\u001b[1m\u001b[32m\n"}
{"Time":"2026-07-15T19:05:47.008469789+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply2","Output":"TestApply1 2026-07-15T19:05:47+01:00 logger.go:79: Destroy complete! Resources: 2 destroyed.\n"}
{"Time":"2026-07-15T19:05:47.011131795+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply2","Output":"TestApply1 2026-07-15T19:05:47+01:00 logger.go:79: \u001b[0m\n"}
{"Time":"2026-07-15T19:05:47.011182727+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply1","Output":"--- PASS: TestApply1 (14.42s)\n"}
{"Time":"2026-07-15T19:05:47.655959433+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply1","Output":"TestApply2 2026-07-15T19:05:47+01:00 logger.go:79: \u001b[0m\u001b[1mtime_sleep.time_to_sleep: Still creating... [10s elapsed]\u001b[0m\u001b[0m\n"}
{"Time":"2026-07-15T19:05:48.651056005+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply1","Output":"TestApply2 2026-07-15T19:05:48+01:00 logger.go:79: \u001b[0m\u001b[1mtime_sleep.time_to_sleep: Creation complete after 11s [id=2026-07-15T18:05:48Z]\u001b[0m\n"}
{"Time":"2026-07-15T19:05:48.659808962+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply1","Output":"TestApply2 2026-07-15T19:05:48+01:00 logger.go:79: \u001b[0m\u001b[1m\u001b[32m\n"}
{"Time":"2026-07-15T19:05:48.660014665+01:00","Action":"output","Package":"github.com/gchappel/terratestbug/terraform","Test":"TestApply1","Output":"TestApply2 2026-07-15T19:05:48+01:00 logger.go:79: Apply complete! Resources: 2 added, 0 changed, 0 destroyed.\n"}
```
|
**To Reproduce**
- clone https://github.com/gchappel/terratestbug
- run `go test -v -json ./terraform/`
- see the raw JSON output sometimes mixes up test names and outputs between `TestApply1` and `TestApply2`
  - each test has a `terraform_data` resource which has a unique value, supporting showing which test is which
```go
// paste code snippets here
```

**Expected behavior**
- The `Test` and `Output` fields are matched up, so that the output from `Apply1` is **always** tagged as `"Test": "Apply1"`
- 
**Nice to have**
- [x] Terminal output
- [ ] Screenshots

**Versions**
- Terratest version: `v1.0.1`
- Environment details (Ubuntu 20.04, Windows 10, etc.): Windows 11, WSL2, Ubuntu `24.04`, Go `1.26.5`

**Additional context**

james00012:
Fixes #1871.

## What

The default `Terratest` logger wrote formatted lines directly to `os.Stdout`. Under `go test -json`, the JSON runner cannot attribute a raw stdout write to a specific parallel test, so it tags each line with whichever test was active. The result is output where the `Test` field does not match the line it carries, which makes the JSON stream unusable for de-interleaving parallel tests. PR #1467 only added a mutex to stop garbled lines; attribution was never addressed.

## Change

Two commits:

1. `logger`: `DoLog` now routes through `t.Log` when a `*testing.T` is available and the destination is stdout. The framework then attributes each line to the correct test, including under `t.Parallel()`. Behavior is otherwise preserved:
   - `t.Log` streams immediately under `-v` on Go 1.14+, so the always on streaming rationale in the doc comment still holds.
   - An explicit non stdout writer (for example a `bytes.Buffer`) is always honored.
   - If the test has already completed, it falls back to a direct stdout write instead of panicking.

2. `logger/parser`: routing through `t.Log` indents each line and prepends the framework `file.go:NN:` decoration, which broke `terratest_log_parser` de-interleaving of parallel non JSON output (every line rolled up to the last test resumed via `=== CONT`). The indented lines still embed their owning test name, so the parser now detects that format and extracts the name directly. The un-indented path (stdout fallback) is unchanged.

## Verification

- Real terraform e2e: two `t.Parallel()` tests each running `terraform.InitAndApplyContext` against a `random_integer` + `time_sleep` config (mirrors the reporter repro, no cloud needed). Before: 134 of 188 terratest log lines misattributed. After: 0 of 190.
- Parallel subtests under `-json`: 0 of 24 misattributed.
- Completed-test path falls back to stdout without panicking; non `*testing.T` loggers still stream to stdout at column 0; `*testing.B` routes through `b.Log`.
- Parser de-interleaving of real post-fix `-v` output verified end to end via `SpawnParsers`; existing fixture based parser tests still pass.
- Full `modules/core` suite green under `-race`; `go build` across the workspace, `go vet`, and `gofmt` clean.

## Notes for review

- Routing through `t.Log` means Go prepends its own `file:line:` decoration, so a line now carries both Go and Terratest prefixes. I kept the Terratest prefix intact to avoid dropping the test name and timestamp. Happy to trim the redundancy if you prefer.
- The switch cannot be gated to JSON only, since `go test -v -json` is indistinguishable from `-v` at runtime, so it applies whenever a real `*testing.T` is present.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved routing of Terratest stdout output through `t.Log` when available to keep parallel `go test -json` attribution correct.
  * Enhanced log parsing to reliably associate indented Terratest “applying” lines with the owning test and de-interleave mixed parallel output.
* **Documentation**
  * Clarified how indented Terratest “applying” output is attributed to the correct test.
* **Tests**
  * Added new coverage for indented Terratest log detection, test-name extraction (including subtests), and correct splitting of `t.Log`-routed output.


modules/core/logger/parser/parser.go
```diff
@@ -56,6 +56,11 @@ var (
 	regexStatus  = regexp.MustCompile(`=== (RUN|PAUSE|CONT)\s+(.+)`)
 	regexSummary = regexp.MustCompile(`(^FAIL$)|(^(ok|FAIL)\s+([^ ]+)\s+(?:(\d+\.\d+)s|\(cached\)|(\[\w+ failed]))(?:\s+coverage:\s+(\d+\.\d+)%\sof\sstatements(?:\sin\s.+)?)?$)`)
 	regexPanic   = regexp.MustCompile(`^panic:`)
+	// regexIndentedTerratestLog matches a terratest log line emitted through t.Log: the testing framework indents it
+	// and prepends its own "file.go:NN: " decoration, e.g. "    apply_test.go:42: TestFoo 2006-01-02T15:04:05Z07:00
+	// caller.go:7: message". It captures the embedded test name so parallel output can still be de-interleaved. The
+	// "TestName <RFC3339 timestamp>" signature is what distinguishes a terratest line from an ordinary t.Log line.
+	regexIndentedTerratestLog = regexp.MustCompile(`^\s+\S+:\d+:\s+(Test[^\s]*)\s+\d{4}-\d{2}-\d{2}T`)
```
gcagle3: Minor nit: We could use 'Test\S*' here instead of 'Test[^\s]*' to match the style used in other areas. Not a blocker though.

gcagle3 (approved):

modules/core/logger/parser/parser.go
james00012: Updated.

gcagle3 (approved):

PR #1874, opened 2026-07-21
Automate v2 release into a single create-release workflow

james00012:
Folds the manual release-prep pin step into the release workflow, so cutting a v2 release is one dispatch.

Before: cut a release-prep branch, run scripts/release-prep-pin.sh by hand, push that branch (whose build is intentionally red), then dispatch a tag-only workflow with the commit SHA.

After: dispatch `Create v2 Release` (.github/workflows/create-release.yml) with a version (e.g. v2.0.0-beta.1) and a ref (default main). It:
1. Runs the pre-flight gates + the GOWORK=off release-mode build.
2. Pins every submodule's cross-module require to the version and commits that in CI (scripts/release-prep-pin.sh).
3. Pushes the 16 per-module tags in dependency order (core -> helpers -> tooling -> platforms -> IaC), then verifies proxy.golang.org.

Run once with dry_run=true to preview the exact tags, then dry_run=false to push.

Why:
- Removes the only manual, error-prone release step.
- The intentionally-broken pinned state now lives on an ephemeral in-CI commit that only the tags point at, instead of a pushed branch with red CI.
- The pin logic is now a shared, shellcheck-clean script (scripts/release-prep-pin.sh) rather than an out-of-repo helper.

Changes: add scripts/release-prep-pin.sh, replace v2-release.yml with create-release.yml, update docs/v2-release-runbook.md. YAML validated (actionlint clean).

## Summary by CodeRabbit

* **New Features**
  * Added a single manual v2 release workflow with configurable source ref and default dry-run.
  * Added automated cross-module v2 version pinning and dependency-ordered lockstep tagging.
  * Added release-mode build validation and optional Go proxy checks before publishing; generates/attaches release notes with prerelease handling for `-suffix` versions.
* **Documentation**
  * Updated the v2 release runbook with workflow inputs/behavior, the required pinning timing, and corrected tag naming format (`modules/<name>/<version>`).

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,11 +1,15 @@
-name: v2 Release (lockstep tag push)
+name: Create v2 Release (pin + lockstep tags)
 
-# Manual-trigger workflow that pushes the full v2 tag set from a release-prep commit.
-# Procedure documented in docs/v2-release-runbook.md.
+# One manual workflow that cuts a full v2 release from main:
+#   1. Pre-flight: the modularization gates + the GOWORK=off release-mode build.
+#   2. Pin every submodule's cross-module require to the release version and
+#      commit that in-CI (the commit the tags point at). Deterministic; runs
+#      scripts/release-prep-pin.sh.
+#   3. Push the 16 per-module tags in dependency order, then verify the proxy.
 #
-# Use this AFTER:
-#   - scripts/release-prep.sh ran on the release-prep branch
-#   - The release-prep commit was reviewed and approved
+# Supersedes the old two-step flow (manual release-prep-pin.sh on a throwaway
+# branch + a tag-only workflow). Procedure documented in
+# docs/v2-release-runbook.md. Run with dry_run=true first to preview the tags.
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


scripts/release-prep-pin.sh
```diff
@@ -0,0 +1,31 @@
...
+  exit 1
+fi
+
+PLACEHOLDER='v2.0.0-00010101000000-000000000000'
+count=0
+for d in modules/*/; do
+  [ -f "${d}go.mod" ] || continue
+  mods=$(grep -oE "github.com/gruntwork-io/terratest/modules/[a-z0-9]+/v2 ${PLACEHOLDER}" "${d}go.mod" 2>/dev/null | awk '{print $1}' || true)
+  for mod in $mods; do
+    go -C "$d" mod edit -require="${mod}@${VERSION}"
```
denis256, line 26: Missing "set -e" may lead the script to pass on a failed go mod edit

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,178 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: write
+    env:
+      VERSION: ${{ inputs.version }}
+      # Single source of truth for the module list and tag order (mirrors ORDER
+      # in scripts/check-release-mode.sh and the runbook tag sequence).
+      MODULE_PATHS: >-
+        modules/core
```
denis256, line 72: Hardcoded 16-module list can drift, will be good to have automation for automated collection/usage

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,178 @@
...
+        modules/terraform modules/terragrunt modules/teststructure
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ inputs.ref }}
+          fetch-depth: 0
+
+      - name: Validate version
+        run: |
+          if [[ ! "$VERSION" =~ ^v2\.[0-9]+\.[0-9]+(-[0-9A-Za-z.]+)?$ ]]; then
```
denis256, line 85: Most probably, version check should be one of the first steps to fail execution at the beginning

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,178 @@
...
+
+      - name: Verify proxy serves each tag
+        if: ${{ inputs.dry_run == false }}
+        run: |
+          # Allow the proxy a moment to populate after push.
+          sleep 30
+          fail=0
+          for path in $MODULE_PATHS; do
+            url="https://proxy.golang.org/github.com/gruntwork-io/terratest/${path}/v2/@v/${VERSION}.info"
+            code=$(curl -s -o /dev/null -w '%{http_code}' "$url")
```
denis256, line 137: Make sense to set a timeout for curl

denis256 (approved):

Issue #1875, opened 2026-07-24
'Go Report' image in the README.md reads 'retired'


wjam:
**Describe the bug**
The link to https://goreportcard.com in the README.md now links to a page that says that Go Report Card has been sunset.

**To Reproduce**
Link the `go report` image in the README.md and read the page

**Expected behavior**
Not to link to services that have been sunset.

**Versions**
- Terratest version: commit 74e48a2dcdb11f145021fc13473bba594e860ab2

**Additional context**

I'm not sure where to leave feedback on the v2 beta, so I'll leave it here as I noticed this bug while trying out this new version. I'm also not sure if this feedback is too late for the v2 migration.

It's good that Terratest gets split into more manageable modules rather than single monolithic one, as will reduce the potential for CVEs. Unfortunately, depending on the `teststructure` module negates a large amount of this work, as it pulls in `k8s`, `packer`, and `terraform` modules. A similar thing appears to be present in the `k8s` module which depends on `aws` for a single call. Without a greater split between the modules, I'm not sure whether there'll be a real difference in the number of dependencies a user of this library will have.

PR #1878, opened 2026-07-26
refactor(k8s): resolve node public IPs through an injectable lookup

james00012:
Third of three PRs for #1875. The reporter noted that the `k8s` module depends on `aws` for a single call. That is exactly right.

## The problem

One call site, `findAwsNodeHostnameContextE` in `service.go`, looked up one EC2 instance's public IP when resolving the endpoint of a NodePort service on EKS:

```go
ipMap, err := aws.GetPublicIpsOfEc2InstancesContextE(t, ctx, []string{instanceID}, region)
```

Requiring the whole `aws` module for that pulled in 23 AWS service SDKs: acm, autoscaling, cloudwatchlogs, dynamodb, ec2, ecr, ecrpublic, ecs, iam, kms, lambda, rds, route53, s3, secretsmanager, signin, sns, sqs, ssm, sso, ssooidc, sts and internal. That accounted for 37 of the 118 modules in the k8s dependency graph, or 31 percent, for one `DescribeInstances`.

Note this edge does not violate `check-acyclic-deps.sh`, since `aws` and `k8s` are both tier 3 and the script only rejects imports from a strictly higher tier.

## The change

A `NodePublicIPLookup` function type on `KubectlOptions`. Its signature matches `aws.GetPublicIpsOfEc2InstancesContextE` exactly, so EKS users wire it up in one line:

```go
options := k8s.NewKubectlOptions("", "", "default")
options.NodePublicIPLookup = aws.GetPublicIpsOfEc2InstancesContextE
```

No k8s types leak into `aws` and no aws types leak into `k8s`, so neither module gets heavier and no new module is needed.

The field is tagged `json:"-"`, since a func cannot be marshalled and `KubectlOptions` is serialized as test data.

## Resolving the public IP without an API call

Cloud controller managers record an instance's public IP as an `ExternalIP` address on the Node object. `findDefaultNodeHostnameE` only ever matched `NodeHostName`, so that address was never used, even though both `FindNodeHostname` doc comments have promised to prefer "the external IP when available" since the function was written in 2018.

Preferring `ExternalIP` covers the common case with no cloud API call and no configuration:

```go
// resolution order
node ExternalIP  ->  options.NodePublicIPLookup (if set)  ->  node internal hostname
```

## No breaking change

Because `ExternalIP` covers the common case, there was no longer a reason to change the public signatures.

`FindNodeHostnameContext` and `FindNodeHostnameContextE` keep their original signatures, byte identical to main. A caller written against v2.0.0-beta.1 was compiled unmodified against this branch: it builds and resolves the node's public IP.

The lookup is reached through new `FindNodeHostnameWithOptionsContext` and `FindNodeHostnameWithOptionsContextE` functions. `GetServiceEndpoint` and the rest of the service API are unchanged and route through those, so a configured lookup still applies. Nothing is deprecated, because nothing was removed or changed shape.

`ec2:DescribeInstances` is no longer required for the common path.

## Verified on a live EKS cluster

EKS v1.31, managed node group in public subnets. Both nodes carried an `ExternalIP`:

```
providerID=aws:///us-east-1d/i-015db3381c397594a  ExternalIP=32.197.58.143
providerID=aws:///us-east-1f/i-0fd7752d9dc57d23d  ExternalIP=100.56.17.181
```

With no lookup configured, `FindNodeHostnameContextE` returned `32.197.58.143`, and `GetServiceEndpoint` on a real NodePort service returned `32.197.58.143:31234`, which served traffic from outside the VPC. Cluster destroyed after the run.

One deliberate behaviour change remains: clusters on any provider that advertise an `ExternalIP` now resolve to it rather than to the internal hostname. That is what the documented contract has always claimed.

## Result

`k8s` go.sum drops from 259 to 160 lines, with zero AWS SDK modules remaining. `helm`, which required `aws` only transitively through `k8s`, was re-tidied: 316 to 219 lines, 72 fewer aws-sdk-go-v2 entries.

## Test plan

- Unit tests in `node_hostname_test.go` covering: `ExternalIP` preferred with no options set, `ExternalIP` short circuits a configured lookup, `ExternalIP` honoured for non-AWS providers, blank `ExternalIP` ignored, lookup used when no `ExternalIP` is present and receives the right instance ID and region, fallback with no lookup configured, fallback when the instance has no public IP, lookup skipped for a non-AWS provider ID, errors propagate, and malformed provider IDs rejected without panicking
- A separate module compiling a pre-existing caller against the branch, verifying the signatures did not change
- `go test` clean for k8s and helm
- `check-acyclic-deps.sh`, `check-single-source.sh` and `check-siv-placement.sh` OK
- golangci-lint 2.11.3 with the CI build tags reports 0 issues repo wide

## Summary by CodeRabbit

- **New Features**
  - Kubernetes service endpoint discovery can now use an optional callback to resolve externally reachable AWS node public IPs.
  - Hostname resolution can prefer a node’s `ExternalIP`, otherwise use the lookup (when configured), with automatic fallback to internal hostnames.
- **Bug Fixes**
  - Rejects malformed AWS `providerID` values safely (no panics) and propagates lookup errors.
  - Skips lookup for non-AWS providers and when `ExternalIP` is present (including empty values).
  - Ensures IPv6 service endpoints are formatted with brackets for URL safety.
- **Tests**
  - Expanded unit coverage for lookup behavior, fallbacks, JSON omission of the callback, and IPv6 endpoint formatting.


modules/k8s/service.go
```diff
@@ -256,10 +256,12 @@ func pickRandomNodeE(t testing.TestingT, options *KubectlOptions) (corev1.Node,
 // FindNodeHostnameContext returns the hostname or IP address of the given node using the provided context, preferring
 // the external IP when available. This will fail the test if there is an error.
 //
+// Resolving the external IP of a cloud instance requires options.NodePublicIPLookup to be set. See NodePublicIPLookup.
+//
 //nolint:gocritic // hugeParam: cannot change public function signature
-func FindNodeHostnameContext(t testing.TestingT, ctx context.Context, node corev1.Node) string {
+func FindNodeHostnameContext(t testing.TestingT, ctx context.Context, options *KubectlOptions, node corev1.Node) string {
```
yhakbar: This is a breaking change. How do we plan on making sure consumers can migrate cleanly?

james00012: Dropped the breaking change. Node objects already carry the public IP as `ExternalIP`, which we were ignoring, so no cloud API call is needed and the signatures stay as they are. `FindNodeHostnameContext[E]` are byte identical to main. `NodePublicIPLookup` is now just an escape hatch on `FindNodeHostnameWithOptionsContext[E]`. Consumers do nothing.

Not confirmed on a live EKS cluster that the CCM sets `ExternalIP`; if it is absent this falls through to today's behaviour.

modules/k8s/service.go
```diff
@@ -268,40 +268,113 @@ func FindNodeHostnameContext(t testing.TestingT, ctx context.Context, node corev
...
+func FindNodeHostnameWithOptionsContextE(
+	t testing.TestingT,
+	ctx context.Context,
+	options *KubectlOptions,
+	node corev1.Node,
+) (string, error) {
+	// An external IP recorded on the node is authoritative and costs no API call, so prefer it for every provider.
+	if externalIP, ok := findNodeAddress(&node, corev1.NodeExternalIP); ok {
+		return externalIP, nil
+	}
```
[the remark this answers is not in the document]
james00012: Fixed. All four endpoint sites use `net.JoinHostPort` now. This got more likely with the switch to preferring the node's `ExternalIP`, which is IPv6 on a dual-stack cluster.

modules/k8s/service.go
james00012: Confirmed on a live EKS cluster, so the caveat above is resolved. v1.31, managed node group in public subnets:

```
providerID=aws:///us-east-1d/i-015db3381c397594a  ExternalIP=32.197.58.143
providerID=aws:///us-east-1f/i-0fd7752d9dc57d23d  ExternalIP=100.56.17.181
```

With no lookup configured, `FindNodeHostnameContextE` returned `32.197.58.143`, and `GetServiceEndpoint` on a real NodePort service returned `32.197.58.143:31234`, which served traffic from outside the VPC. Cluster destroyed.

modules/k8s/service.go
```diff
@@ -268,40 +270,113 @@ func FindNodeHostnameContext(t testing.TestingT, ctx context.Context, node corev
...
+	// An AZ is a region plus a single trailing letter, so dropping that letter yields the region.
 	region := availabilityZone[:len(availabilityZone)-1]
 
-	ipMap, err := aws.GetPublicIpsOfEc2InstancesContextE(t, ctx, []string{instanceID}, region)
+	if options == nil || options.NodePublicIPLookup == nil {
+		logger.Default.Logf(t, "[WARNING] Node %s is backed by an AWS EC2 instance, but KubectlOptions.NodePublicIPLookup "+
+			"is not set, so its public IP cannot be resolved. Falling back to the internal hostname recorded on the node. "+
+			"Set options.NodePublicIPLookup = aws.GetPublicIpsOfEc2InstancesContextE to resolve public IPs.", node.Name)
+
+		return findDefaultNodeHostnameE(node)
```
denis256, line 376: Not sure if this will work for NodePort services

denis256 (approved):

PR #1880, opened 2026-08-05
fix(teststate): stop after reporting a fatal error

james00012:
Found while reviewing #1879. Pre-existing.

## Bug

Every `t.Fatalf` in `core/teststate` falls through to the code after it. With `*testing.T` that is invisible, because `FailNow` calls `runtime.Goexit`. But `TestingT` exists so other harnesses can be plugged in, and an implementation whose `FailNow` returns keeps going.

The harmful case is `save`:

```go
bytes, err := json.Marshal(value)
if err != nil {
    t.Fatalf("Failed to convert value %s to JSON: %v", path, err)
}
...
os.WriteFile(path, bytes, 0o600)   // bytes is nil, writes a zero byte file
```

A later stage then loads that empty file. `IsPresent` and `IsEmptyJSON` were quieter but wrong the same way, answering "absent" or "empty" when the real answer was an error they had just announced.

## Fix

An explicit `return` after each `Fatalf`, plus a package comment explaining why they are not dead code.

## Tests

Three regression tests driven by a `TestingT` whose `FailNow` returns. The marshal one was confirmed to fail without the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during test state saving, loading, and validation.
  * Prevented invalid or unreadable test state data from being incorrectly treated as missing or empty.
  * Ensured failed operations stop processing immediately and report errors accurately.

* **Tests**
  * Added regression coverage for save failures, inaccessible paths, and invalid JSON.

### DIFF
--- a/modules/core/teststate/teststate.go
+++ b/modules/core/teststate/teststate.go
@@ -7,6 +7,11 @@
 //
 // This package lives in core rather than teststructure so that modules such as aws, k8s, packer, and ssh can provide
 // their own helpers without teststructure having to import every one of them.
+// Every t.Fatalf in this package is followed by an explicit return. testing.TestingT documents FailNow as stopping
+// execution via runtime.Goexit, and *testing.T honours that, so those returns are unreachable in ordinary use. They
+// are not decorative: TestingT exists so other harnesses can be plugged in, and an implementation whose FailNow
+// returns would otherwise carry on past the failure. In save that meant writing a zero byte file after a marshal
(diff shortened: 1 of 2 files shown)


denis256 (approved):
Hm, the package comment says every Fatalf has a return, but three calls at the end of functions are missing them:

* Line 97 in teststate.go: save() (on os.WriteFile failure)
* Line 115 in teststate.go: Load() (on json.Unmarshal failure)
* Line 199 in teststate.go: Cleanup() (on os.Remove failure)

PR #1898, opened 2026-09-08
feat(gcp): read a Pub/Sub topic's settings back so a test can assert them (LIB-5570)

james00012:
**TL;DR:** A test holding a Pub/Sub topic name can now read that topic's labels, message retention and storage policy, instead of only asking whether the topic exists.

Closes [LIB-5570](https://linear.app/gruntwork/issue/LIB-5570), part of [LIB-5569](https://linear.app/gruntwork/issue/LIB-5569).

**Why:** the GCP module is write-only about topic settings, so a test can watch a Terraform module set the wrong message retention and still pass. This is the same gap `GetStorageBucketAttrs` closed for buckets in #1890.

**What changed**

* **modules/gcp · pubsub.go**: `GetTopicAttrs`, `GetTopicAttrsE` and `GetTopicAttrsWithClient`, returning the topic as `*pubsubpb.Topic` so a caller asserts on the API's own fields and a field the API adds later needs no change here. It goes through the same `TopicAdminClient` the existence check uses, and the error variant uses a named return so the deferred `Close` is joined onto the error rather than dropped.
* **modules/gcp · pubsub_test.go**: a case against the pstest in-memory fake, creating a topic with the labels, retention and storage policy a real module sets and reading all three back. The missing-topic case asserts the topic name and the project as well as the phrase, so an error that dropped either would fail.
* **modules/gcp · storage.go**: `GetStorageBucketAttrsContext` and its `E` variant lose the `Context` suffix and gain the same joined `Close`. That suffix is for backwards compatibility with v1 names, and both are new in v2 with no v1 name to preserve. Neither is in a release, so nothing outside this repository can be holding the old name.
* **Deliberately not handled:** subscriptions and schemas, where the same read gap exists and each is its own change. Also the twelve other places in `modules/gcp` that discard a `Close` error: those are pre-existing rather than new, so sweeping them belongs in its own change.

**Landing it**
* **Rollback:** revert is enough. Nothing existing changes behaviour; this only adds functions.
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including the new case. It fails on the base branch, where `GetTopicAttrsWithClient` does not exist.
* **Rename**: no reference to `GetStorageBucketAttrsContext` is left anywhere in the repository, and the only caller outside it is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), which pins terratest by commit and so keeps building until that pin moves.
* **Not run**: anything against a real project. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the topic module and reads it back through this function.

## Summary by CodeRabbit

- **New Features**
  - Added support for retrieving Google Cloud Pub/Sub topic settings and attributes.
  - Topic details include labels, message-retention duration, and allowed persistence regions.
  - Added clear error reporting when topics are unavailable or retrieval fails, including the affected topic and project.

- **Changes**
  - Simplified the naming of Google Cloud Storage bucket attribute retrieval helpers.

- **Tests**
  - Added coverage for successful topic retrieval and validation of topic attributes and missing-topic errors.

### DIFF
--- a/modules/gcp/pubsub.go
+++ b/modules/gcp/pubsub.go
@@ -2,6 +2,7 @@ package gcp
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"cloud.google.com/go/pubsub/v2"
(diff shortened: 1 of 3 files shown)


modules/gcp/pubsub.go
```diff
@@ -56,6 +56,53 @@ func AssertTopicExistsWithClient(ctx context.Context, client *pubsub.Client, top
 	return nil
 }
 
+// GetTopicAttrsContext returns the settings Google Cloud holds for the given Pub/Sub topic, so a
+// test can assert on what was actually created rather than only that it exists.
+// This will fail the test if there is an error.
+// The ctx parameter supports cancellation and timeouts.
+func GetTopicAttrsContext(t testing.TestingT, ctx context.Context, projectID string, topicName string) *pubsubpb.Topic {
```
yhakbar: Is there isn't already a `GetTopicAttrs`, I recommend not creating a function named `GetTopicAttrsContext`. We only added the `Context` suffixed stuff for backwards compatibility.

modules/gcp/pubsub.go
```diff
@@ -56,6 +56,53 @@ func AssertTopicExistsWithClient(ctx context.Context, client *pubsub.Client, top
...
+// The ctx parameter supports cancellation and timeouts.
+func GetTopicAttrsContextE(t testing.TestingT, ctx context.Context, projectID string, topicName string) (*pubsubpb.Topic, error) {
+	logger.Default.Logf(t, "Getting settings for Pub/Sub topic %s in project %s", topicName, projectID)
+
+	client, err := newPubSubClient(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() { _ = client.Close() }()
```
yhakbar: We should use a named return so that we can errors.Join this error on execution of the defer.

modules/gcp/pubsub_test.go
```diff
@@ -54,6 +58,38 @@ func TestPubSubTopicLifecycleWithClient(t *testing.T) {
...
+func TestGetTopicAttrsWithClient(t *testing.T) {
+	t.Parallel()
+
+	client := newFakePubSubClient(t)
+	ctx := context.Background()
+
+	require.ErrorContains(t, func() error {
+		_, err := gcp.GetTopicAttrsWithClient(ctx, client, "missing")
+		return err
+	}(), "does not exist")
```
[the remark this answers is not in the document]
james00012: Applied. The missing-topic case now asserts the topic name and the project as well as the phrase, so an error that dropped either would fail.

odgrim (approved):

PR #1913, opened 2026-09-11
feat(gcp): read a Document AI processor's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a Document AI processor id can now read that processor's type, display name and state. The GCP module had nothing for Document AI at all before this.

**Why:** a Terraform module that creates a processor has no way to prove it created the processor it was asked for, and a processor whose type came out wrong is exactly the case worth catching. It is the same gap `GetStorageBucketAttrs` closed for buckets in #1890.

**What changed**

* **modules/gcp · documentai.go**, new: `GetDocumentAIProcessorAttrs`, `GetDocumentAIProcessorAttrsE` and `GetDocumentAIProcessorAttrsWithClient`, returning the processor as `*documentai.GoogleCloudDocumentaiV1Processor` so a caller asserts on the API's own fields. A missing processor returns an error naming the processor and the project, in the wording the other read functions here use.
* **`NewDocumentAIServiceE` takes a location, unlike every other constructor here.** Document AI answers only on a per-location host such as `us-documentai.googleapis.com`, so the location decides the endpoint rather than only the resource path. A service built against the global host returns 404 for a processor that exists.
* **modules/gcp · documentai_test.go**, new: a configured processor and a missing one, both against a local test server the Google transport actually talks to, in the shape `storage_test.go` already uses.
* **No new dependency.** `google.golang.org/api` is already required for the compute client and the Document AI service lives in the same module, so `go.mod` is unchanged.
* **Deliberately not handled:** processor versions and the processing calls themselves. Processing a document costs money per page and is a different thing from proving a processor was created as asked.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including both new cases. They fail on `main`, where `documentai.go` does not exist.
* **Not run**: anything against a real project. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the processor module and reads it back through this function.

## Summary by CodeRabbit

- **New Features**
  - Added support for retrieving Google Cloud Document AI processor settings by project, location, and processor ID.
  - Added authenticated, location-aware Document AI service access.
  - Added context cancellation and custom service client support.
  - Added clear errors for missing processors, API failures, and invalid locations.

- **Tests**
  - Added coverage for successful processor retrieval, missing processors, and location validation.

### DIFF
--- a/modules/gcp/documentai.go
+++ b/modules/gcp/documentai.go
@@ -1,0 +1,72 @@
+package gcp
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/gruntwork-io/terratest/modules/core/v2/logger"
+	"github.com/gruntwork-io/terratest/modules/core/v2/testing"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/api/documentai/v1"
+	"google.golang.org/api/googleapi"
+	"google.golang.org/api/option"
+)
+
+// GetDocumentAIProcessorAttrs returns the settings Google Cloud holds for the given Document AI
+// processor, so a test can assert on what was actually created rather than only that it exists. A
+// processor lives in a multi-region such as `us` or `eu`, which has to be given, and the id is the
+// one Google assigns.
+// This will fail the test if there is an error.
+// The ctx parameter supports cancellation and timeouts.
+func GetDocumentAIProcessorAttrs(t testing.TestingT, ctx context.Context, projectID string, location string, processorID string) *documentai.GoogleCloudDocumentaiV1Processor {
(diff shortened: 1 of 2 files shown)


denis256 (approved):
Can we put location in this error? Document AI is location-scoped (host + path), so a wrong region looks like a missing processor.

Example: processor lives in eu, call GetDocumentAIProcessorAttrs(t, ctx, "my-proj", "us", "proc-1") -> today: Document AI processor proc-1 does not exist in project my-proj. Prefer something like ... in project my-proj (location us).

PR #1914, opened 2026-09-11
feat(gcp): read a Dialogflow CX agent's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a Dialogflow CX agent id can now read that agent's display name, description, default language, time zone and spell-correction flag. The GCP module had nothing for Dialogflow at all before this.

**Why:** a Terraform module that creates an agent has no way to prove it created the agent it was asked for. It is the same gap `GetStorageBucketAttrs` closed for buckets in #1890.

**What changed**

* **modules/gcp · dialogflow.go**, new: `GetDialogflowAgentAttrs`, `GetDialogflowAgentAttrsE` and `GetDialogflowAgentAttrsWithClient`, returning the agent as `*dialogflow.GoogleCloudDialogflowCxV3Agent` so a caller asserts on the API's own fields. A missing agent returns an error naming the agent and the project, in the wording the other read functions here use.
* **`NewDialogflowServiceE` takes a location.** Dialogflow answers on a per-location host for every location but `global`, so the location decides the endpoint. A service built against the global host returns 404 for a regional agent that exists.
* **modules/gcp · dialogflow_test.go**, new: a configured agent and a missing one, both against a local test server the Google transport actually talks to, in the shape `storage_test.go` already uses.
* **No new dependency.** `google.golang.org/api` is already required for the compute client and the Dialogflow service lives in the same module, so `go.mod` is unchanged.
* **Deliberately not handled:** flows, intents, entity types and sessions. Each is its own resource with its own read, and none is needed to prove an agent.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including both new cases. They fail on `main`, where `dialogflow.go` does not exist.
* **Not run**: anything against a real project. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the agent module and reads it back through this function.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving Dialogflow CX agent settings from Google Cloud.
  * Supports agents in regional and global locations.
  * Provides clear errors when an agent cannot be found.
  * Validates location values to help prevent invalid service requests.

* **Tests**
  * Added coverage for successful agent retrieval, missing-agent errors, and valid or invalid location handling.

### DIFF
--- a/modules/gcp/dialogflow.go
+++ b/modules/gcp/dialogflow.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 
 	"github.com/gruntwork-io/terratest/modules/core/v2/logger"
 	"github.com/gruntwork-io/terratest/modules/core/v2/testing"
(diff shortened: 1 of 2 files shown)


denis256 (approved):
Can we put location in this error? Dialogflow routes by location (host + path), so a wrong region looks like a missing agent. 


Example: agent in us-central1, call with global -> Dialogflow agent agent-1 does not exist in project my-proj. Prefer ... in project my-proj (location global)

PR #1916, opened 2026-09-15
feat(gcp): read a Contact Center Insights view's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a Contact Center Insights view id can now read that view's display name, filter and resource name back from Google. The GCP module had nothing for Contact Center Insights before this.

**Why:** a Terraform module that creates a view has no way to prove it created the view it was asked for. It is the same gap `GetStorageBucketAttrs` closed for buckets in #1890.

**What changed**

* **modules/gcp · contactcenterinsights.go**, new: `GetInsightsViewAttrs`, `GetInsightsViewAttrsE` and `GetInsightsViewAttrsWithClient`, returning the view as `*contactcenterinsights.GoogleCloudContactcenterinsightsV1View` so a caller asserts on the API's own fields. A missing view returns an error naming the view and the project, in the wording the other read functions here use.
* **`NewContactCenterInsightsServiceE` takes no location.** Every location answers on the one host, so the location goes in the resource name rather than the endpoint, unlike Dialogflow.
* **modules/gcp · contactcenterinsights_test.go**, new: a configured view and a missing one, both against a local test server the Google transport actually talks to, in the shape `storage_test.go` already uses.
* **No new dependency.** `google.golang.org/api` is already required for the compute client and the Contact Center Insights service lives in the same module, so `go.mod` is unchanged.
* **Deliberately not handled:** conversations, analyses and the other rule resources. Each is its own resource with its own read, and none is needed to prove a view.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including both new cases. They fail on `main`, where `contactcenterinsights.go` does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the view module and reads it back through this function; that run passed against a real project today.

## Summary by CodeRabbit

* **New Features**
  * Added helpers for retrieving Google Cloud Contact Center Insights view settings.
  * Added support for clear errors when a requested view does not exist.
  * Added context support for request cancellation and timeouts.

* **Tests**
  * Added coverage for successful view retrieval, response parsing, request paths, and missing-view errors.


modules/gcp/contactcenterinsights.go
```diff
@@ -0,0 +1,69 @@
...
+// contactcenterinsights_test.go for the pattern).
+// The ctx parameter supports cancellation and timeouts.
+func GetInsightsViewAttrsWithClient(ctx context.Context, service *contactcenterinsights.Service, projectID string, location string, viewID string) (*contactcenterinsights.GoogleCloudContactcenterinsightsV1View, error) {
+	name := fmt.Sprintf("projects/%s/locations/%s/views/%s", projectID, location, viewID)
+
+	view, err := service.Projects.Locations.Views.Get(name).Context(ctx).Do()
+	if err != nil {
+		var apiErr *googleapi.Error
+		if errors.As(err, &apiErr) && apiErr.Code == 404 {
+			return nil, fmt.Errorf("Contact Center Insights view %s does not exist in project %s", viewID, projectID)
```
denis256, line 54: We may want to add the region in error to know which region was used to make debugging easier

denis256 (approved):

PR #1930, opened 2026-09-21
feat(gcp): read a DNS record set's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a managed zone, a record name and a type can now read that record's TTL and values back from Cloud DNS. The DNS helpers read zones and policies before this, and not the records inside them.

**Why:** A module that writes a record has no way to prove it wrote the one it was asked for, and a record whose values or TTL came out wrong answers every lookup wrongly without failing anything.

**What changed**

* **modules/gcp · dns.go**: `GetDNSRecordSetAttrs`, `GetDNSRecordSetAttrsE` and `GetDNSRecordSetAttrsWithClient`, beside the zone and policy reads, returning `*dns.ResourceRecordSet`. A record is named by its zone, its fully qualified name and its type together, so the read takes all three and a missing record's error names each.
* **modules/gcp · dns_test.go**: a configured record and a missing one, against the local test server the other DNS cases use.
* **No new dependency.** It is the DNS service this file already talks to.
* **Deliberately not handled:** listing records, and routing policies beyond what the response already carries.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where the record set read does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the record set module and reads the record back through this function; that run passed against a real project today.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving Google Cloud DNS record-set details by project, managed zone, record name, and record type.
  * Added clear error reporting when a requested record set is missing or cannot be retrieved.


james00012:

PR #1931, opened 2026-09-21
feat(gcp): read a log bucket's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a project, a location and a bucket id can now read that log bucket's retention, lock and state back from Cloud Logging. The logging helpers read log-based metrics before this, and not buckets.

**Why:** A bucket whose retention came out wrong keeps logs for the wrong length of time, which is either a compliance gap or a bill, and nothing fails when it happens.

**What changed**

* **modules/gcp · logging.go**: `GetLogBucketAttrs`, `GetLogBucketAttrsE` and `GetLogBucketAttrsWithClient`, beside the metric reads, returning `*logging.LogBucket`. A bucket is named by its location as well as its id, and a deleted one is still returned for seven days with a `lifecycleState` of `DELETE_REQUESTED`, which the doc comment says.
* **modules/gcp · logging_test.go**: a configured bucket and a missing one.
* **No new dependency.**
* **Deliberately not handled:** log views and bucket links, which are their own resources.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where the bucket read does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the project bucket module and reads the bucket back through this function; that run passed against a real project today.

## Summary by CodeRabbit

* **New Features**
  * Added helpers to retrieve Google Cloud log bucket settings using project, location, and bucket identifiers.
  * Added support for custom clients and both error-returning and test-failing usage patterns.
  * Added clear handling for missing buckets and other retrieval failures.

* **Tests**
  * Added coverage for successful bucket retrieval and not-found error details.


james00012:

PR #1932, opened 2026-09-21
feat(gcp): read a workload identity pool provider's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a pool and a provider id can now read that provider's attribute mapping, condition and OIDC or other issuer settings back from IAM. The IAM helpers read pools before this, and not the providers inside them.

**Why:** The attribute condition is what keeps every other repository or account out of a project, and a provider whose condition came out wrong lets them federate in without failing anything.

**What changed**

* **modules/gcp · iam.go**: `GetWorkloadIdentityPoolProviderAttrs`, `GetWorkloadIdentityPoolProviderAttrsE` and `GetWorkloadIdentityPoolProviderAttrsWithClient`, beside the pool reads, returning `*iam.WorkloadIdentityPoolProvider`. A provider is named by its pool as well as its own id.
* **modules/gcp · iam_test.go**: a configured OIDC provider and a missing one.
* **No new dependency.**
* **Deliberately not handled:** managed identities and workforce pools, which are different resources.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where the provider read does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the pool provider module and reads the provider back through this function; that run passed against a real project today.

## Summary by CodeRabbit

- **New Features**
  - Added support for retrieving Google Cloud Workload Identity Pool provider settings.
  - Provider details now include configuration such as display name, description, status, attribute mappings, conditions, and OIDC settings.
  - Improved error reporting when a requested provider does not exist.

- **Tests**
  - Added coverage for successful provider retrieval and missing-provider responses.


james00012:

PR #1933, opened 2026-09-21
feat(gcp): read a Cloud Deploy target's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a location and a target name can now read that Cloud Deploy target's approval gate, labels and where it deploys back from Google. The Cloud Deploy helpers read delivery pipelines before this, and not their targets.

**Why:** A target that lost its approval gate rolls out to production unattended, and nothing fails when it does.

**What changed**

* **modules/gcp · clouddeploy.go**: `GetDeployTargetAttrs`, `GetDeployTargetAttrsE` and `GetDeployTargetAttrsWithClient`, beside the pipeline reads, returning `*clouddeploy.Target`. The name carries `Deploy` because a bare `Target` would be ambiguous in a package that also covers Compute Engine.
* **modules/gcp · clouddeploy_test.go**: a configured target and a missing one.
* **No new dependency.**
* **Deliberately not handled:** automations, custom target types and deploy policies.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where the target read does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the target module and reads the target back through this function; that run passed against a real project today.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving Google Cloud Deploy target settings by project, location, and target name.
  * Added clear error handling when the requested target does not exist.

* **Tests**
  * Added coverage for successful target retrieval and validation of returned settings.
  * Added coverage for missing-target errors and contextual error details.


james00012:

PR #1934, opened 2026-09-21
feat(gcp): read a Cloud Build trigger's settings back so a test can assert them

james00012:
**TL;DR:** A test holding a location and a trigger name can now read that trigger's event source, identity, substitutions and inline steps back from Cloud Build.

**Why:** A trigger whose event source or steps came out wrong builds the wrong thing on the wrong event, and nothing fails when it does.

**What changed**

* **modules/gcp · buildtrigger.go**, new: `GetBuildTriggerAttrs`, `GetBuildTriggerAttrsE` and `GetBuildTriggerAttrsWithClient`, returning `*cloudbuild.BuildTrigger`. It uses the REST client `NewCloudBuildRESTServiceE` already builds for the worker pool read, since the gRPC client in `cloudbuild.go` has nothing a local test server can stand in for.
* **modules/gcp · workerpool.go**: the constructor's comment now says the trigger read uses it too, which it did not before.
* **modules/gcp · buildtrigger_test.go**, new: a configured trigger and a missing one, through the fake `workerpool_test.go` already has.
* **No new dependency.**
* **Deliberately not handled:** reading a trigger by its id rather than its name, which nothing here proves.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where `buildtrigger.go` does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the trigger module and reads the trigger back through this function; that run passed against a real project today.

## Summary by CodeRabbit

- **New Features**
  - Added utilities for retrieving Google Cloud Build trigger details by project, location, and trigger name.
  - Added support for custom clients and contexts when querying triggers.
  - Improved error reporting for missing triggers and other lookup failures.

- **Tests**
  - Added coverage for successful trigger retrieval, request validation, decoded results, and missing-trigger errors.

- **Documentation**
  - Clarified guidance for using REST-based Cloud Build access in test environments.


james00012:

PR #1935, opened 2026-09-21
feat(gcp): read an alert policy's and an SLO's settings back so a test can assert them

james00012:
**TL;DR:** A test can now read an alert policy's conditions, severity and whether it is enabled, and an SLO's goal, window and indicator, back from Cloud Monitoring. The monitoring helpers read notification channels and uptime checks before this.

**Why:** A policy whose threshold or duration came out wrong pages at the wrong time or never, and an SLO whose goal came out wrong burns its budget at the wrong rate. Neither fails anything when it happens.

**What changed**

* **modules/gcp · monitoring.go · alert policies**: `GetAlertPolicyAttrs` and its two variants, returning `*monitoring.AlertPolicy`. Google assigns a policy's id, so the caller passes the id it got back, the last segment of the policy's name.
* **modules/gcp · monitoring.go · SLOs**: `GetServiceLevelObjectiveAttrs` and its two variants, returning `*monitoring.ServiceLevelObjective`. An SLO is named by the service it measures as well as its own id.
* **modules/gcp · monitoring_test.go**: a configured and a missing case for each.
* **No new dependency.** Both use the Monitoring v3 service this file already talks to.
* **Deliberately not handled:** dashboards, which are a different API.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where either read does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the alert policy and SLO modules and reads each resource back through these functions; that run passed against a real project today.

## Summary by CodeRabbit

- **New Features**
  - Added helper support for retrieving Google Cloud Monitoring alert policy details.
  - Added helper support for retrieving service level objective details.
  - Added options for standard error handling and reuse of existing Monitoring service clients.
  - Missing resources now return clear not-found errors, while other retrieval failures retain their underlying error details.

- **Tests**
  - Added coverage for successful and unsuccessful alert policy and service level objective retrievals.


james00012:

PR #1936, opened 2026-09-21
feat(gcp): read a subnetwork's, a Cloud Router's and an IP address's settings back

james00012:
**TL;DR:** A test can now read a subnetwork, a Cloud Router, a regional IP address and a global IP address back from Compute Engine. A Cloud NAT gateway comes back inside its router, so the router read is also how a test reads a NAT.

**Why:** These are what most network modules build next to the VPC network, and a range, a port allocation or an ASN that came out wrong breaks something far from the change that caused it. The compute helpers read networks and firewall rules before this, since [#1915](https://github.com/gruntwork-io/terratest/pull/1915).

**What changed**

* **modules/gcp · vpc.go**, new: `FetchSubnetwork`, `FetchRouter`, `FetchAddress` and `FetchGlobalAddress`, each as `FetchX`, `FetchXE` and `FetchXWithClient`, with the `X.Get(...) got error` shape `disk.go` uses. They take a context and carry no `Context` suffix, since that suffix exists only to keep older names working.
* **modules/gcp · vpc_test.go**, new: a configured and a missing case for each, built with the `respond` helper `compute_test.go` already has.
* **VPC peering needs nothing new.** A peering comes back inside its network, which `FetchNetworkContext` already reads.
* **No new dependency.**
* **Deliberately not handled:** Shared VPC host and service projects, which need an organization-level permission the test project does not have, so nothing could prove a read of them.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where `vpc.go` does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the subnetwork, router, NAT, address and peering modules and reads each resource back through these functions; that run passed against a real project today.


james00012:

PR #1937, opened 2026-09-21
feat(gcp): read the pieces of a global HTTPS load balancer back so a test can assert them

james00012:
**TL;DR:** A test can now read a health check, a backend service, a URL map, an HTTPS proxy, an SSL certificate and a global forwarding rule back from Compute Engine, which are the six resources a global HTTPS load balancer is made of.

**Why:** A load balancer whose port, timeout or routing came out wrong serves the wrong thing or nothing, and none of it fails until traffic arrives.

**What changed**

* **modules/gcp · loadbalancing.go**, new: `FetchHealthCheck`, `FetchBackendService`, `FetchURLMap`, `FetchTargetHTTPSProxy`, `FetchSSLCertificate` and `FetchGlobalForwardingRule`, each as `FetchX`, `FetchXE` and `FetchXWithClient`, taking a context and carrying no `Context` suffix, since that suffix exists only to keep older names working. A managed certificate is readable while it is still provisioning, which is the state a test usually sees.
* **modules/gcp · loadbalancing_test.go**, new: a configured and a missing case for each, built with `respond`.
* **No new dependency.**
* **Deliberately not handled:** the regional variants of each, which are separate endpoints with the same shape.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where `loadbalancing.go` does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the six load balancing modules and reads each resource back through these functions; that run passed against a real project today.


james00012:

PR #1938, opened 2026-09-21
feat(gcp): read an instance template's and a regional managed instance group's settings back

james00012:
**TL;DR:** A test can now read an instance template's machine type, disks and scheduling, and a regional managed instance group's template, size and update policy, back from Compute Engine.

**Why:** A template or an update policy that came out wrong launches every instance of a group wrongly, or replaces them all at once on the next change, and nothing fails until the group scales.

**What changed**

* **modules/gcp · instancetemplate.go**, new: `FetchInstanceTemplate` and `FetchRegionalInstanceGroupManager`, each as `FetchX`, `FetchXE` and `FetchXWithClient`, taking a context and carrying no `Context` suffix, since that suffix exists only to keep older names working. The group read returns the manager's own settings; the instances it runs are already read by `FetchRegionalInstanceGroupContext`, whose naming the new read follows.
* **modules/gcp · instancetemplate_test.go**, new: a configured and a missing case for each, built with `respond`.
* **No new dependency.**
* **Deliberately not handled:** regional instance templates and zonal managed instance groups.

**Landing it**
* **Rule bent:** none.

**Validation**
* **Unit**: the whole `modules/gcp` package passes, including every new case. They fail on `main`, where `instancetemplate.go` does not exist.
* **Not run**: anything against a real project from this repository. The consumer that does is [terraform-google-test-library](https://github.com/gruntwork-io-team/terraform-google-test-library), whose suite applies the instance template and regional managed instance group modules and reads each resource back through these functions; that run passed against a real project today.


james00012:

